PR #57, opened 2026-05-09
Add per-finding agent_action and top-level agent_summary (v0.12)

pengfei-threemoonslab:
## Summary

- Ships W3-1 / W3-2 of the agent-adoption strategy: schema bumps to v0.12 with two additive, deterministic-projection fields aimed at reducing the work a coding agent has to do when reading `report.json`.
- `findings[].agent_action` projects the routing decision into a single enum (`auto_apply`, `propose_patch_for_review`, `escalate_to_human`, `suppress_with_reason` (reserved), `informational`) so consumers don't synthesize it from `patches`/`autofix_safe`/`requires_human_review`/`suggested_patch_kind` themselves.
- `report.agent_summary` exposes the verdict, headline, blocker/review counts, auto-applicable patch count, human-review count, and a deterministic `first_recommended_action` next step — one fetch instead of array traversal.

## Type

- [ ] Input adapter change
- [x] Report, schema, or SARIF output _(additive v0.11 → v0.12 schema bump; new top-level field plus a new per-finding field)_
- [ ] Documentation only

## Verification

Additional local checks run:

- `python -m pytest` → 878 passed (was 860 on main; +18 new), 3 skipped
- `python -m ruff check .` → clean
- `python scripts/build-llms-full.py` regenerates byte-identical with the post-bump sources
- `python scripts/generate_schemas.py` produces `docs/report-schema.v0.12.json` from the model defaults; the contract test asserts the schema and the runtime model agree

## Release-readiness notes

- [x] No user-code import added to default scan paths _(both new fields are pure derivations from already-available data)_
- [x] New or changed check IDs are documented in `docs/checks.md` _(no new check IDs)_
- [x] Report/schema changes are additive or documented in `STABILITY.md` _(both fields documented; v0.11 added to the frozen-reference list; legacy schema-version pattern extended in the contract test)_

## Notable design choice

`derive_agent_action` deviates from the strategy doc's pseudocode on one branch. The strategy ordered `requires_human_review` BEFORE the medium/low confidence check, which would have mapped a non-manual medium-confidence patch to `escalate_to_human` even though the value's defined semantic ("no machine-applicable patch; needs human judgment") excludes that case. The implementation reorders so non-manual medium/low patches → `propose_patch_for_review`. Documented in `derive_agent_action`'s docstring; pinned by `tests/test_agent_action_summary.py::test_derive_agent_action_medium_confidence_non_manual` and the low-confidence twin.

## What's next per the plan

- W3-3 (prompts) is in this PR alongside the model/schema changes — `prompts/recommend-fixes.md` and `prompts/fix-top-finding.md` plus their skill mirrors now read `agent_action` first with a legacy fallback for v0.11 consumers, and surface `agent_summary.first_recommended_action` as the default next step.
- W3-4 (upstream-framework PR) and W3-5 (re-run benchmark) remain external/manual tasks per the plan.


pengfei-threemoonslab:
Three review fixes pushed in [74daa68](https://github.com/ThreeMoonsLab/agents-shipgate/pull/57/commits/74daa68).

**P1.1 — Path threaded through, no more hardcoded default.** `build_agent_summary` now takes a `json_report_path` kwarg, sourced from `ReadinessReport.generated_reports["json"]` in `build_report`. When the path is known (the common case), the apply-patches command names the real on-disk path. When JSON output is disabled, the action falls back to `kind: "info"` with a parameterised hint — we never emit a command pointing at a phantom file or a stale default-path report. Verified end-to-end against `scan --out custom-reports`. Pinned by two new tests.

**P1.2 — Schema now requires both fields and rejects null.** `scripts/generate_schemas.py`:
- Added `agent_summary` to the top-level required list and forced its property schema to a direct `$ref: "#/$defs/AgentSummary"` (same pattern as `release_decision` — no `anyOf` with null).
- Added `agent_action` to Finding's required list and inlined the enum directly (AgentAction is a Literal, not a BaseModel, so the inline form drops the null branch).

The new test `test_v12_schema_requires_agent_summary_and_agent_action_non_nullable` pins the required-list shape AND end-to-end verifies that stripping either field — or setting `agent_summary` to null — fails validation.

**P2 — Stale docs bumped, test now self-derives.** `docs/INDEX.md`, `docs/examples.md`, `docs/baseline.md`, and `docs/autofix-policy.md` now identify v0.12 as current with v0.11 in the frozen-reference list. Reworked `tests/test_docs_links.py::test_index_lists_current_schema` to read the expected version from `ReadinessReport.report_schema_version` (no more hardcoded literal that codifies stale expectations). Added a second guard, `test_no_other_doc_falsely_advertises_an_older_schema_as_current`, that scans the four secondary docs and fails when an older-minor schema appears near the word "current" without a frozen/legacy marker — catches drift earlier than downstream test fixtures would.

882 tests pass (was 878; +4 new). Ruff clean.

pengfei-threemoonslab:
Three more review fixes pushed in [b18b9c3](https://github.com/ThreeMoonsLab/agents-shipgate/pull/57/commits/b18b9c3).

**P1 — `needs_human_review` no longer undercounts.** Now includes both `escalate_to_human` AND `propose_patch_for_review` (the latter is "patch attached, but needs explicit user confirmation" — that's a human-review case too). `release_decision.review_items` already routes both, so `agent_summary` now agrees. Reproduced the reviewer's exact case: a single `propose_patch_for_review` finding now produces `needs_human_review=1` with action text "Walk the 1 review item starting with ..." — pinned by `test_needs_human_review_only_finding_is_propose_patch_for_review`.

**P2 — `shlex.quote(json_report_path)` so paths with spaces round-trip.** Verified against `/tmp/shipgate review/custom reports/report.json`: `shlex.split(action.command)` now puts the full path back together as a single `--from` argument. Pinned by `test_first_recommended_action_command_is_shell_safe`.

**P3 — Two more stale docs bumped, drift-guard widened.** `docs/overview.md` and `docs/ai-search-summary.md` now point at v0.12 with v0.11 frozen. Renamed and broadened `test_no_doc_falsely_advertises_an_older_schema_as_current` to walk `docs/**/*.md` recursively instead of a hand-curated 4-element list — adding a new doc that references the schema can no longer slip past the guard.

885 pytest passes (was 882; +3 new). Ruff clean.

pengfei-threemoonslab:
Two more review fixes pushed in [49477ad](https://github.com/ThreeMoonsLab/agents-shipgate/pull/57/commits/49477ad).

**P1 — Contract docs and prompt now match the runtime semantics.**
- `docs/agent-contract-current.md` now defines `needs_human_review` as `agent_action ∈ {escalate_to_human, propose_patch_for_review}` with a callout that `release_decision.review_items` tracks the same population.
- `prompts/recommend-fixes.md` (and the skill mirror) now reads `needs_human_review = buckets B + C + D combined`, with explicit guidance that to split B from C+D you walk `findings[].agent_action`. The previous formula ("bucket-B = total − A − C+D − informational") is gone — it produced wrong counts because it relied on the old `needs_human_review = C+D` definition.

**P2 — Schema now requires the full `agent_summary` and `AgentSummaryAction` shapes.** The schema generator post-processes:
- `AgentSummary.required` = `[verdict, headline, blocker_count, review_item_count, auto_appliable_patches, needs_human_review, first_recommended_action]`
- `AgentSummaryAction.required` = `[kind, command, why]` (`command` nullable as a value, required as a key)

The new test `test_v12_schema_requires_full_agent_summary_shape` pins both required lists AND end-to-end verifies that stripping any expected key from a real scan payload fails validation. Verified all 7 summary fields and all 3 action fields fail individually when removed.

886 tests pass (was 885; +1 new). Ruff clean.

pengfei-threemoonslab:
Fixed in [f2c0a9a](https://github.com/ThreeMoonsLab/agents-shipgate/pull/57/commits/f2c0a9a). You're right that `review_item_count` (severity-driven, includes medium auto_apply findings) and `needs_human_review` (action-driven, only escalate/propose) track different populations and can legitimately diverge. The bug was the headline using the wrong one.

**Fix.** Headline now uses `needs_human_review` for the "require human review" wording, with explicit branches for the divergence cases:

| Case | Headline |
|---|---|
| `needs_review > 0` (with or without auto-apply) | "N finding(s) require human review[; M also auto-applicable]." |
| `needs_review == 0` and `auto_appliable > 0` | "M auto-applicable finding(s) flagged for release review; none require human input beyond apply-patches." |
| Both 0 (rare — review_items has only informational entries) | "K review item(s) flagged for release review." |

Verified the reviewer's exact case (medium auto_apply, `verdict: review_required`):
- Before: `1 finding(s) require human review. 1 review item.`
- After: `1 auto-applicable finding(s) flagged for release review; none require human input beyond apply-patches. 1 review item.`

Also fixed the action text's "Walk the N review item(s)" wording to fall back to `review_item_count` when `needs_review==0`, so the prose can never read "Walk the 0 review items".

**Contract doc.** `docs/agent-contract-current.md` now explicitly distinguishes the two populations: `review_item_count` is severity-driven (release-review debt), `needs_human_review` is action-driven (agent work). The "track the same population" line that I added in the previous round is removed; a dedicated bullet calls out that they can diverge and gives the medium-severity-stale-suppression case as the canonical example.

Pinned by `test_review_required_with_only_auto_apply_does_not_claim_human_review` — it reproduces the reviewer's case and asserts the headline neither claims "require human review" nor omits the auto-applicable callout.

887 pytest passes (was 886; +1 new). Ruff clean.

pengfei-threemoonslab:
Three more review fixes pushed in [c612720](https://github.com/ThreeMoonsLab/agents-shipgate/pull/57/commits/c612720).

**P1 — Reason no longer reintroduces the contradiction.** Added an `append_reason` switch that's set False in the two `review_required` branches where the headline already explains the agent-level situation in agent-driven terms (auto-applicable-only, and the rare zero-needs-zero-auto fallback). The test now uses the realistic `release_decision.reason` text the runtime actually emits in this scenario ("1 finding requires human review before shipping.") and asserts that string does NOT appear in the final headline. The previous test stubbed a benign reason and missed the regression — that's exactly the kind of test trap you flagged.

**P2 — `derive_agent_action` is order-invariant.** Now uses the first NON-manual patch — same rule `_derive_from_patches` already uses for `suggested_patch_kind`. All-manual lists still escalate. Verified both orderings produce `propose_patch_for_review` for `[ManualPatch, medium SetPointer]`. Replaced the old `test_derive_agent_action_mixed_manual_and_non_manual` test (which was encoding the buggy behavior) with a parametrized order-invariance test plus a dedicated all-manual escalate test.

**P3 — Sample expected JSON reports regenerated against v0.12.** All four (`support_refund_agent`, `simple_openai_api_agent`, `simple_langchain_agent`, `simple_crewai_agent`) now carry `report_schema_version: "0.12"`, a populated `agent_summary`, and `findings[].agent_action`. Cwd-relative paths normalized to `<REPO>` for portability. Existing markdown goldens still pass (they used dynamic fields, so they didn't drift).

Tests: 889 pass (was 887; +2 new). Ruff clean.

pengfei-threemoonslab:
Fixed in [0312f31](https://github.com/ThreeMoonsLab/agents-shipgate/pull/57/commits/0312f31). You're right that the previous fix improved the contradictory case but introduced a worse one — the rare branch was actively dropping the most useful piece of context.

**Detection.** `build_agent_summary` now reads `release_decision.evidence_coverage.human_review_recommended` and threads an `evidence_recommended` flag through both the headline-builder and the action-builder.

**Headline.** When the verdict is `review_required` AND no findings need fixing AND evidence-coverage drove the decision, the headline IS the reason text:

| Sample | Old headline | New headline |
|---|---|---|
| `simple_langchain_agent` | `0 review item(s) flagged for release review.` | `Static-only scan with low-confidence evidence; human review recommended.` |
| `simple_crewai_agent` | (same) | (same) |

**Action.** `first_recommended_action` is no longer null in this case. It carries an info action that names the situation and offers a concrete remediation path: "gather better evidence (e.g. add MCP/OpenAPI inputs, eval traces) or accept the static-only review posture; no machine-applicable fix is available."

**Test.** `test_evidence_only_review_surfaces_reason_and_info_action` constructs the langchain/crewai shape (review_required, evidence-coverage-driven, no findings) and pins (a) headline equals the reason text verbatim, (b) it does NOT contain the unhelpful "0 review item" placeholder, and (c) the action's why-text points at concrete remediation paths.

**Goldens.** All four sample `expected/report.json` files regenerated — langchain and crewai now show useful headlines and non-null actions; the other two are unchanged in shape but reflect the current emitter.

890 tests pass (was 889; +1 new). Ruff clean.

pengfei-threemoonslab:
Two more review fixes pushed in [c2fab8a](https://github.com/ThreeMoonsLab/agents-shipgate/pull/57/commits/c2fab8a).

**P3 — Mixed manual + high-confidence patches now match the enum prose.** Collapsed the high/medium/low non-manual branches: ANY non-manual patch with declared confidence routes to `propose_patch_for_review` when `autofix_safe` is False. The high-confidence non-manual patch is genuinely safe (`apply-patches --confidence high` will run it and skip the manual sibling), but the agent still has to surface the manual instructions — that's the propose-for-review semantic. Verified order-invariant for `[Manual, high]` and `[high, Manual]`. Pinned by `test_derive_agent_action_mixed_high_confidence_with_manual_proposes_review`.

| Patches | Old verdict | New verdict |
|---|---|---|
| `[high]` only | `auto_apply` | `auto_apply` |
| `[medium]` only | `propose_patch_for_review` | `propose_patch_for_review` |
| `[Manual]` only | `escalate_to_human` | `escalate_to_human` |
| `[Manual, high]` | `escalate_to_human` ❌ | `propose_patch_for_review` ✅ |
| `[high, Manual]` | `escalate_to_human` ❌ | `propose_patch_for_review` ✅ |

**P2 — fix-top-finding prompt now lists the exception cases explicitly.** When the verdict is blocked/review_required, `agent_summary.first_recommended_action.why` no longer always names a specific finding. The prompt now flags two exceptions:
- **Evidence-coverage-driven `review_required`**: the why-text describes the evidence situation (low-confidence/static), no `check_id` to parse — follow the why-text guidance directly.
- **`auto_appliable_patches > 0`**: the why-text names `apply-patches`, not a specific finding — fall back to walking `findings[]` for the top entry.

Mirrored to `skills/agents-shipgate/prompts/`. Goldens regenerated.

891 tests pass (was 890; +1 new). Ruff clean.

pengfei-threemoonslab:
Fixed in [3bf12ed](https://github.com/ThreeMoonsLab/agents-shipgate/pull/57/commits/3bf12ed). Both prose surfaces (`src/agents_shipgate/core/models.py` enum comment and `docs/agent-contract-current.md`) now describe `propose_patch_for_review` as:

> at least one non-manual patch is attached and machine-applicable, but the full patch set is not auto-safe. Two shapes land here: (a) every non-manual patch is medium- or low-confidence, and (b) a high-confidence non-manual patch sits alongside one or more `ManualPatch` siblings (the non-manual is safe to apply, but the manual instructions still need a human). In both cases the agent should ask the user before `--apply` and surface any manual instructions verbatim.

`escalate_to_human` stays "no machine-applicable patch" — that's now exclusive of the mixed-manual-with-high case that previously contradicted it. `llms-full.txt` regenerated to pick up the contract doc change.

891 tests pass (unchanged); ruff clean.

PR #77, opened 2026-05-15
Add manual release dispatch workflow

pengfei-threemoonslab:
## Summary
- add `workflow_dispatch` support to the Release workflow with a required version input
- validate requested releases against `pyproject.toml` and `src/agents_shipgate/__init__.py`, then create an annotated release tag from the default branch in the same approved run
- verify GitHub Release, PyPI, and GitHub Marketplace visibility after publishing
- document the manual release path and pin it with action metadata tests

## Validation
- `python -m ruff check .`
- YAML parse check for `.github/workflows/release.yml`
- `python -m pytest tests/test_action_metadata.py tests/test_public_surface_contract.py tests/test_docs_links.py` -> `254 passed`

## Notes
- The job still supports the existing `push.tags: v*` release path.
- Marketplace visibility is checked with retry/backoff and emits a warning if GitHub has not refreshed it within the retry window.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,6 +4,12 @@ on:
   push:
     tags:
       - "v*"
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Version to release without leading v, for example 0.10.0. Must match pyproject.toml and agents_shipgate.__version__."
+        required: true
+        type: string
 
 permissions:
   contents: write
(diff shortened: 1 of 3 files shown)


pengfei-threemoonslab:
Closing as superseded. Since this was opened (2026-05), the release path was rebuilt under the release-integrity milestone (#359): tag-driven publication with source-to-wheel binding, the qualification gate, and a separate non-publishing **Release Rehearsal** workflow that already carries `workflow_dispatch` with a ref selector. A manual dispatch that creates an annotated tag from a typed version input would bypass the protected release input that model exists to enforce (`docs/release-evidence-policy-decision.md`, `docs/release-runbook.md`). Nothing here is needed for the imminent v0.16.0 tag.

PR #90, opened 2026-05-18
Add dynamic_default contract on CheckMetadata (v0.18 PR #1)

pengfei-threemoonslab:
## Summary

Formalizes the M1 dynamic-severity contract closed in v0.17 (silent severity-override bypass via `SHIP-ACTION-POLICY-VIOLATION`). Makes the contract **explicit in the schema** so a future contributor cannot silently miss the wiring and re-introduce the bypass class.

- New `CheckMetadata.dynamic_default: bool = False` field + model validator (requires `floor_severity` when `True`).
- `SHIP-ACTION-POLICY-VIOLATION` declares `dynamic_default=True` and `floor_severity="medium"`. **Breaking** for manifests downgrading this check below `medium` without a tier-crossing acknowledgement — failure mode is loud (exit 2) per the M1 contract.
- New `cli/scan.py:_dynamic_check_defaults` centralizes the aggregator with a 3-step seed/overlay/policy-pack pattern.
- Resolver internal-consistency guard (`RuntimeError`) catches catalog/aggregator drift pre-merge.
- Plugin gate `dynamic_default_not_supported` rejects plugins declaring `dynamic_default=True` (no path to wire into the aggregator). Gate placement runs **before** `_coerce_metadata` so a plugin declaring `dynamic_default=True` without `floor_severity` lands under a precise status instead of being mis-classified as `bad_floor`.

## Contract for future contributors

To add a new built-in dynamic-severity check:
1. Set `dynamic_default=True` in `CHECK_METADATA` (forces `floor_severity` via the model validator).
2. Add an overlay branch in `cli/scan.py:_dynamic_check_defaults`. The seed loop auto-includes the new check ID; the overlay supplies the manifest-effective default.

Contract test `test_dynamic_default_aggregator_completeness` fails the moment (1) is done without (2).

## Schema impact

No report schema bump. `CheckMetadata` is catalog-internal — not in emitted `report.json`. Only `docs/checks.json` is regenerated (additive: 5 new fields per entry).

## Test plan

- [x] All 1192 tests pass (62 → 69 in `test_severity_override_floor.py`; 18 → 20 in `test_plugin_validation.py`).
- [x] `python scripts/generate_schemas.py --check` exits 0.
- [x] `ruff check .` clean.
- [x] Coverage 87.24% (well above 75% floor).
- [x] Sample-report diff: byte-identical except path-dependent fields (catalog field doesn't leak into emitted reports).
- [x] STABILITY.md "Dynamic-severity check classes" rewritten around the schema field. Plugin-validation enum documents `dynamic_default_not_supported`.


pengfei-threemoonslab:
All three review findings addressed in commit 8852e9d. Summary:

**P1 — Coerced-truthy bypass closed** ([`src/agents_shipgate/checks/plugin_validation.py`](https://github.com/ThreeMoonsLab/agents-shipgate/blob/feat/v018-dynamic-default-contract/src/agents_shipgate/checks/plugin_validation.py))
- Replaced `value.get("dynamic_default") is True` with `TypeAdapter(bool).validate_python(...)` so the raw gate mirrors Pydantic's coercion exactly. Anything Pydantic stores as `True` (`"true"`, `"yes"`, `"on"`, `1`, etc.) is now rejected by the gate.
- 12 new parametrized regression cases in `tests/test_plugin_validation.py`: 6 truthy variants must land in `dynamic_default_not_supported`; 7 falsey controls (`False`, `"false"`, `0`, absent key, etc.) must NOT trip the gate.

**P2 — Aggregator-completeness test was vacuous** ([`tests/test_severity_override_floor.py`](https://github.com/ThreeMoonsLab/agents-shipgate/blob/feat/v018-dynamic-default-contract/tests/test_severity_override_floor.py))
You were right — the old test only checked key presence, which the seed loop guarantees regardless of overlay wiring. Replaced with a per-check overlay-verification registry `_DYNAMIC_DEFAULT_OVERLAY_CASES` mapping each `dynamic_default=True` catalog ID to `(manifest_overlay, expected_dynamic_severity)` where the expected severity is provably STRONGER than the catalog static. Two new tests:
- `test_dynamic_default_overlay_registry_is_complete` — every catalog `dynamic_default=True` ID must have a registry entry. Fails when a future contributor adds a dynamic-default check without thinking through (and documenting) the manifest path.
- `test_dynamic_default_aggregator_overlay_fires` (parametrized) — builds the manifest with the strengthened section, asserts `_dynamic_check_defaults` returns the manifest-effective severity (not the catalog static). Fails when an overlay branch is missing.

The seed loop stays (it prevents the resolver guard from false-positiving on legitimate user input that overrides a swing check without populating the manifest section). The new tests catch the regression class the seed loop alone cannot.

**P3 — `docs/checks.md` "five gates" → "six gates"**
Already addressed in the earlier follow-up commit [3fbb1e3](https://github.com/ThreeMoonsLab/agents-shipgate/commit/3fbb1e3) (`Expand dynamic_default coverage: tests + docs (PR #1 follow-up)`). Your checkout likely predated that commit — the current branch shows:

> "**Plugin validation (v0.17+; six gates v0.18+).** Shipgate runs six load-time gates against every entry point — load, signature, metadata, dynamic-default-not-supported (v0.18+), ID-collision, and floor-consistency…"

with cross-links to the relevant STABILITY.md sections.

**Verification:** 1198 tests pass (1192 + 12 new + −6 net from P2 restructuring); ruff clean; `python scripts/generate_schemas.py --check` exits 0; coverage 87.23%.

pengfei-threemoonslab:
Right — that bullet conflated the hard floor with the tier-crossing ack. A user reading it could reasonably try to fix a `low`/`info` override by adding an `acknowledge_overrides` entry, and the manifest would still fail to load. Fixed in [d3a3720](https://github.com/ThreeMoonsLab/agents-shipgate/commit/d3a3720): the breaking-change bullet now spells out both contracts separately so the right remedy is obvious.

- **Hard floor (no bypass)**: below `medium` (i.e., `low` / `info`) is rejected by the `floor_severity` validator regardless of `acknowledge_overrides`. Remedy: raise to `>= medium` or remove the override.
- **Tier-crossing (ack required)**: `high` (catalog default) → `medium` (floor) crosses the high → normal tier boundary. Allowed only with a matching `acknowledge_overrides` entry that supplies a reason.

Closing the loop on the failure modes: a `low` override + ack still fails with the floor-error message; a `medium` override without an ack fails with the tier-boundary message. The two error paths are distinct, and the new bullet directs users to the appropriate fix.

PR #91, opened 2026-05-18
Widen trust-invariant lint to whole scanner package (v0.18 PR #2)

pengfei-threemoonslab:
## Summary

Brings the trust-invariant AST lint up to the public claim in STABILITY.md and README: "the scanner does not execute or import user code." Pre-v0.18 the lint covered only `src/agents_shipgate/inputs/`; this PR widens it to every `.py` under `src/agents_shipgate/` with a small allowlist of four legitimate first-party meta-CLI surfaces.

**No production code changes.** All changes live in `tests/test_adapter_static_only.py` plus STABILITY.md / CHANGELOG.md.

## Allowlisted surfaces (7 entries across 4 files)

Each `ALLOWED_EXCEPTIONS` entry carries prose rationale and is pinned by a contract test:

- **`cli/bootstrap.py`** — `subprocess.run([sys.executable, "-m", "agents_shipgate", ...])` chains `detect → init → scan → apply-patches` against Shipgate's own CLI.
- **`cli/discovery/artifacts.py`** — `subprocess.run(["git", ...])` probes the user repo for file inventory. Reads git metadata only.
- **`triggers.py`** — `subprocess.run(["git", "diff", ...])` for trigger evaluation. Reads diff content only.
- **`cli/self_check.py`** — `__import__(name)` validates that supplied modules are installed. Runs only under `agents-shipgate self-check`, never during scan.

## Contract tests prevent allowlist rot

- `test_allowlist_entry_matches_real_surface` (parametrized) — every entry MUST correspond to a real surface; entries become stale if the file is later refactored.
- `test_no_unallowlisted_forbidden_surface_in_scanner` — definitive PASS/FAIL signal over the whole sweep, with a failure message listing every `(relative_path, surface)` pair to address.

## Other changes

- `_scan_source` refactored to return structured `Violation` objects (`line`, `surface`, `message`) instead of preformatted strings, so callers can route by `surface` against `ALLOWED_EXCEPTIONS`.
- `importlib.resources` added to `ALLOWED_FORBIDDEN_MODULE_IMPORTS` for bundled-package files inside the wheel.
- Legacy `test_invariant_lint_covers_every_adapter_module` deleted (paranoid for the 18-file `inputs/` case; doesn't scale to ~80 files). Replaced by the new contract tests.

## Test plan

- [x] 170 trust-lint tests pass (62 → 170; 41 negative-control snippet shapes + 41 safe-control + 7 allowlist contract + 79 whole-scanner per-file + 2 contract aggregate).
- [x] Full suite 1192 passes; ruff clean; coverage 87.23%.
- [x] **Negative control:** appending `import subprocess; subprocess.run(['echo', 'hi'])` to `core/findings.py` causes `test_no_unallowlisted_forbidden_surface_in_scanner` to fail with `core/findings.py::import:subprocess, core/findings.py::attr_call:subprocess.run` — clear, actionable message. Revert returns to green.
- [x] STABILITY.md "Trust-model invariants" widened to cite the entire scanner package; new "Meta-CLI surfaces (allowlisted, audited)" subsection documents each entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,47 @@
     `dynamic_default=True` without `floor_severity` lands in
     `dynamic_default_not_supported` rather than being mis-classified
     as `bad_floor` by the new `CheckMetadata` model validator.
+- **v0.18 / PR #2 review follow-up: per-call-site allowlist pinning.**
+  PR #91 review caught two structural holes in the v0.18 trust lint
+  extension:
+  - **P1**: the allowlist matched on `(relative_path, surface)` only,
+    so one entry blanket-permitted every occurrence of a surface in
+    a file. A future unreviewed `subprocess.run(...)` added to an
+    already-allowlisted file would slip past silently.
+  - **P2**: `importlib.resources` was globally exempted, so
+    `files(name)` calls produced no violation. The current uses
+    pass a literal `'agents_shipgate'` anchor, but a future
+    user-controlled anchor would bypass the dynamic-import lint.
+
+  Both are closed by tightening the allowlist contract:
+  - `AllowedException` now carries `line: int` and `snippet: str`
+    (canonical `ast.unparse` of the offending node) in addition to
+    `relative_path` and `surface`. `_violation_allowed` matches on
+    all four fields. Adding a new `subprocess.run` call to an
+    already-allowlisted file now requires a new entry; changing an
+    existing call's argv shape changes the `snippet` and fails the
+    contract test.
+  - `importlib.resources.files` joins `FORBIDDEN_ATTR_CALLS_EXACT`,
+    and `importlib.resources` joins `TRACKED_NON_FORBIDDEN_MODULES`
+    so `from importlib.resources import files; files(...)` is
+    resolved through name-aliases. Both call sites in `triggers.py`
+    and `fixtures.py` are individually pinned with the literal
+    `'agents_shipgate'` anchor in the snippet — a future
+    `files(some_user_anchor)` call would change the snippet and
+    fail the test.
+  - `Violation` gains `snippet: str` captured via `ast.unparse(node)`.
+  - New regression test
+    `test_allowed_exceptions_pin_subprocess_run_per_call_site`
+    asserts that multi-call files (triggers.py, artifacts.py) have
+    distinct entries per call site, so the P1 bypass cannot
+    reappear via consolidation.
+  - New regression test `test_allowed_exceptions_have_no_duplicates`
+    asserts no two entries cover the same call site.
+  - Negative-control: injecting a 4th `subprocess.run` into
+    `triggers.py` now fails the contract test with the precise
+    `(line, surface, snippet)` triple. Injecting
+    `files(user_var)` in place of `files('agents_shipgate')` fails
+    similarly.
 
 - **v0.18 / PR #2 trust-hardening: static AST lint widened to entire scanner.**
   Previously `tests/test_adapter_static_only.py` AST-scanned only
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,14 +78,22 @@
     already-allowlisted file now requires a new entry; changing an
     existing call's argv shape changes the `snippet` and fails the
     contract test.
-  - `importlib.resources.files` joins `FORBIDDEN_ATTR_CALLS_EXACT`,
-    and `importlib.resources` joins `TRACKED_NON_FORBIDDEN_MODULES`
-    so `from importlib.resources import files; files(...)` is
-    resolved through name-aliases. Both call sites in `triggers.py`
-    and `fixtures.py` are individually pinned with the literal
-    `'agents_shipgate'` anchor in the snippet — a future
-    `files(some_user_anchor)` call would change the snippet and
-    fail the test.
+  - `importlib.resources.` joins `FORBIDDEN_ATTR_CALL_PREFIXES`, and
+    `importlib.resources` joins `TRACKED_NON_FORBIDDEN_MODULES`. The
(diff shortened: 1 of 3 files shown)


pengfei-threemoonslab:
Both findings addressed in commit 4e122b2. The structural changes are non-trivial — summary:

**P1 — Per-call-site allowlist (was: blanket per-file/surface)**

You were right: one `AllowedException("triggers.py", "attr_call:subprocess.run")` covered all three current call sites AND would have covered a future fourth. I verified the bypass — injecting a synthetic 4th `subprocess.run` in-memory left `unallowed == []`.

Fix: `AllowedException` now carries `line: int` and `snippet: str` (canonical `ast.unparse` of the offending AST node). `Violation` also captures `snippet`. The match in `_violation_allowed` requires all four fields — adding a new call site requires a new entry; changing an existing call's argv shape flips the `snippet` and forces re-review.

The allowlist grew from 7 entries to 13 (3 distinct entries for `triggers.py` subprocess.run + 2 for `artifacts.py` subprocess.run + others). A new regression test `test_allowed_exceptions_pin_subprocess_run_per_call_site` asserts the multi-call files have distinct per-site entries so this hole cannot reappear via consolidation. A second new test `test_allowed_exceptions_have_no_duplicates` rules out two entries covering the same site.

**P2 — `importlib.resources.files` is now a tracked forbidden surface**

You were right that `files()` was unrestricted. Fix:
- `importlib.resources.files` added to `FORBIDDEN_ATTR_CALLS_EXACT`.
- `importlib.resources` added to `TRACKED_NON_FORBIDDEN_MODULES` so `from importlib.resources import files; files(...)` is resolved through name-aliases.
- Both first-party call sites (`triggers.py:61` and `fixtures.py:40`) are individually pinned with snippet `files('agents_shipgate')`. A future `files(user_var)` call would change the snippet to `files(user_var)` and fail the contract test.

**Negative controls (both reverted before commit):**
1. Injecting `subprocess.run(['git', 'status'], capture_output=True, text=True, check=False)` into `triggers.py:364` fails `test_no_unallowlisted_forbidden_surface_in_scanner` with:
   `triggers.py:364 attr_call:subprocess.run snippet="subprocess.run(['git', 'status'], capture_output=True, text=True, check=False)"`
2. Replacing `files('agents_shipgate')` with `files(some_user_anchor)` at `triggers.py:61` fails with snippet mismatch.

**Failure-mode diagnostics**: the per-call-site contract test distinguishes three failure modes — removed surface (entry stale), line drift (entry's `line` no longer matches), and snippet drift (entry's `snippet` no longer matches the actual call argv). Each gives a specific error message naming the field that drifted.

**Verification:** 179 lint tests pass (170 → 179); 1192 full suite passes; ruff clean; coverage 87.23%. STABILITY.md "Meta-CLI surfaces (allowlisted, audited)" rewritten around per-call-site pinning.

pengfei-threemoonslab:
Both items addressed in commit [`acb3965`](https://github.com/ThreeMoonsLab/agents-shipgate/commit/acb3965), force-pushed after rebasing onto current `origin/main` (which now includes PR #90's squash-merge `95f356d`).

## Merge status
Rebased — PR is no longer conflicting. The branch is now 3 commits ahead of main:
- `f6e670c` Widen trust-invariant lint to whole scanner package (v0.18 PR #2)
- `40bd127` Address PR #91 P1 + P2 review: per-call-site allowlist + tighten importlib.resources
- `acb3965` Broaden importlib.resources surface to whole-module prefix (PR #91 P2 follow-up #2)

CHANGELOG conflicts resolved by keeping both PR #1 and PR #2 entries side-by-side under `## Unreleased` (PR #1 entries first since they shipped first via #90, then PR #2 review follow-up, then PR #2 original).

## P2 — broadened to the whole module
You were right: I added only `files` to `FORBIDDEN_ATTR_CALLS_EXACT`, leaving `read_text`, `read_binary`, `path`, `open_text`, `open_binary`, `is_resource`, `contents`, `as_file` (and any future Python-stdlib addition) as a parallel hole. I verified your exact reproductions:

```python
# both produced [] from _scan_source before this commit
from importlib.resources import read_text; read_text(package_name, "x")
import importlib.resources as res;        res.read_text(package_name, "x")
```

**Fix:** replaced the per-attr entry with the module-wide prefix `"importlib.resources."` in `FORBIDDEN_ATTR_CALL_PREFIXES` (trailing dot scopes to sub-attributes). The prefix catches the entire module surface in one entry, present and future. Manually reproduced your two cases after the fix:
```
from-import-then-call: ['import:importlib.resources.read_text',
                        'attr_call:importlib.resources.read_text']
aliased-attr-chain:    ['attr_call:importlib.resources.read_text']
```

10 new negative-control snippet cases in `test_lint_scanner_catches_known_violation_shapes` pin the broader family (files, read_text, read_binary, open_text, path, is_resource, as_file, plus attribute-chain / aliased / from-import-alias variants). A regression that narrows the prefix back to `files` only fails immediately.

The existing first-party allowlist entries for `files('agents_shipgate')` at `triggers.py:61` and `fixtures.py:40` keep working because they pin `surface="attr_call:importlib.resources.files"` plus the literal-anchor snippet. Any future `read_text(some_user_anchor, ...)` would emit `surface="attr_call:importlib.resources.read_text"` with a snippet containing the non-literal — neither matches any existing entry, so the contract test fails and forces re-review.

STABILITY.md updated: `importlib.resources` is allowed at the import line **only** so name-aliases get built; every call site is forbidden via the prefix and needs per-call-site allowlisting with snippet pinning. CHANGELOG bullet for `importlib.resources` rewritten around the prefix and explicitly enumerates the bypass shapes the per-attr design missed.

**Verification:** 189 lint tests pass (was 179, +10 negative-control); 1192 full suite passes; ruff clean; coverage 87.23%.

PR #111, opened 2026-05-22
Add third-party adapter entry-point discovery (v0.20 E4)

pengfei-threemoonslab:
## Summary

Closes **E4** from the round-3 architecture review (Nov/Dec 2025). Opens the same extension surface for adapters (input loaders) that M5 already opened for plugin checks. Adapters now register through the `agents_shipgate.adapters` entry-point group, gated by the existing `AGENTS_SHIPGATE_ENABLE_PLUGINS=1` env var and `--no-plugins` CLI flag.

## Why

Third parties that wanted to ship a new input adapter (AutoGen, LlamaIndex, Semantic Kernel, additional ADK languages) had no path: the dispatcher's `REGISTRY` was builtin-only. The round-3 review noted M5 plugin validation made adapter discovery safe to land — this PR is that follow-up.

## How

Four load-time gates in a new `inputs/adapter_validation.py` (parallel to `checks/plugin_validation.py`):

1. **`load_failed`** — `entry_point.load()` raised.
2. **`bad_protocol`** — missing ClassVars, empty `source_type`, or wrong `load()` arity.
3. **`bad_scope`** — `scope` not in `{per_source, per_scan}`.
4. **`source_type_collision`** — load-bearing trust rule. Adapter's `source_type` cannot shadow a built-in (`mcp`, `openapi`, `langchain`, …) or another third-party adapter registered earlier in the same scan. Without this rule, a malicious plugin could displace `mcp` and intercept every scan that loads an MCP source.

`discover_third_party_adapters(registry, *, plugins_enabled, loaded_adapters)` in `inputs/protocol.py` walks `entry_points("agents_shipgate.adapters")`, validates each, registers the valid ones, and appends per-entry info dicts to `loaded_adapters[]`. Discovery runs before `_load_sources` so the dispatcher resolves third-party source types alongside built-ins.

## Report surface

New `ReadinessReport.loaded_adapters: list[dict[str, Any]]` field — parallel shape to `loaded_plugins[]`, additive within the existing v0.20 schema. Required + present on every emitted scan; empty list when `--no-plugins` or no third-party adapters installed. Schema generator pins per-item required fields (`name`, `value`, `distribution`, `version`, `source_type`, `validation_status`, `validation_errors`, `runtime_errors`).

## `--strict-plugins` (extended)

`--strict-plugins` now elevates exit code 4 on EITHER plugin failures (M5) OR adapter failures (this PR). The human-readable lines tag each failure as `plugin` or `adapter`. `--no-plugins` help text updated to mention adapter discovery is also disabled.

## Runtime safety net

`run_validated_adapter` in `inputs/adapter_validation.py` wraps `adapter.load()` and captures (a) exceptions, (b) wrong return type, (c) **artifact-class smuggling** — adapter declares `artifact_class=X` but returns an artifact of another type. Mirrors the `Finding.check_id` smuggling rule from `plugin_validation`. The dispatcher's existing `_absorb` artifact-class check fires `TypeError` for built-ins; this wrapper is opt-in for future adapter-execution paths.

## Test plan

- [x] 21 new tests in `tests/test_adapter_entry_point_discovery.py`:
  - valid third-party adapter end-to-end (class form, instance form)
  - each of the four gates exercised in isolation
  - `source_type_collision` parametrized across 4 built-in source types
  - `source_type_collision` between two third parties
  - gating by `AGENTS_SHIPGATE_ENABLE_PLUGINS` env (default off)
  - gating by `--no-plugins` overriding env (forces off)
  - `--strict-plugins` exits 4 end-to-end via CliRunner
  - `run_validated_adapter` catches exceptions, rejects wrong return type, rejects smuggled artifact
- [x] Full suite 1257 passing; ruff clean; coverage 88.00%
- [x] `scripts/generate_schemas.py --check` exits 0
- [x] STABILITY.md "Third-party adapter discovery (v0.20+)" subsection added under Trust-model invariants
- [x] `tests/test_schema_boundaries.py` wire-field-order pin updated for the new field
- [x] `docs/report-sensitive-fields.json` inventory extended for the privacy redaction layer

## Out of scope (intentional)

- **Runtime wrapping in the dispatcher.** The current `_load_sources` invokes `adapter.load()` directly. Wrapping every adapter with `run_validated_adapter` would change the contract for built-ins too; the wrapper is shipped as opt-in infrastructure for future adapter-execution work. The four load-time gates + the existing `_absorb` artifact-class check already cover the trust-bypass cases.
- **Sample-fixture regeneration.** The 4 `samples/*/expected/report.json` files don't yet carry `loaded_adapters: []`. The sample-report tests only pin `report_schema_version` + `release_decision.decision`, so they pass; the next fixture regeneration will pick up the new field naturally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,57 @@
 
 ## Unreleased
 
+- **v0.20 — third-party adapter entry-point discovery (E4 from round-3 review).**
+  Opens the same extension surface for adapters (input loaders) that M5
+  already opened for check plugins. Discovery is gated by the existing
+  `AGENTS_SHIPGATE_ENABLE_PLUGINS=1` env var and `--no-plugins` CLI flag.
+  - New entry-point group: `agents_shipgate.adapters`. A third-party
+    package declares an adapter class (or instance) in its
+    `pyproject.toml` under
+    `[project.entry-points."agents_shipgate.adapters"]`; the class must
+    satisfy the `ToolSourceAdapter` Protocol — `source_type` ClassVar,
+    `scope` ClassVar (`per_source` or `per_scan`), `artifact_class`
+    ClassVar, and a `load(source, base_dir, manifest)` method.
+  - New module `src/agents_shipgate/inputs/adapter_validation.py` with
+    four load-time gates: `load_failed`, `bad_protocol`, `bad_scope`,
+    and **`source_type_collision`** — the load-bearing trust rule
+    rejecting any third-party adapter whose `source_type` shadows a
+    built-in or another already-registered third-party adapter.
+  - New top-level `discover_third_party_adapters(registry, *,
+    plugins_enabled, loaded_adapters)` in `inputs/protocol.py` walks
+    `entry_points("agents_shipgate.adapters")`, validates each entry,
+    and registers the valid ones onto the supplied registry. Both
+    valid and invalid records surface in
+    `report.loaded_adapters[]` so reviewers can see what was skipped.
+  - New report field `loaded_adapters: list[dict[str, Any]]` parallel
+    to `loaded_plugins[]`. Items carry `name`, `value`, `distribution`,
+    `version`, `source_type`, `validation_status`,
+    `validation_errors[]`, `runtime_errors[]`. Required + present on
+    every emitted scan (empty list when `--no-plugins` or no
+    third-party adapters are installed). The schema generator marks
+    each item's eight fields as required.
+  - `--strict-plugins` (v0.17+) extended to cover adapter failures.
+    Any non-`valid` `loaded_adapters[]` row OR non-empty
+    `loaded_adapters[].runtime_errors` now elevates the scan to exit
+    code 4 alongside the existing plugin failures.
+  - `--no-plugins` flag help text updated to mention third-party
+    adapter discovery is also disabled.
+  - `run_validated_adapter` (in `adapter_validation.py`) provides a
+    runtime safety wrapper for callers that want to capture
+    exceptions into `loaded_adapters[].runtime_errors` instead of
+    propagating them. The dispatcher's existing `_absorb` artifact-
+    class check already fires `TypeError` for artifact smuggling;
+    runtime wrapping is opt-in for future adapter-execution paths.
+  - 21 new tests in `tests/test_adapter_entry_point_discovery.py`:
+    each of the four gates + valid-class + valid-instance + env-var
+    gating + `--no-plugins` overrides + collision-with-each-builtin
+    parametrize + collision-between-third-parties + `--strict-plugins`
+    end-to-end + runtime safety net (exception capture, wrong return
+    type, artifact smuggling).
+  - STABILITY.md gains a new "Third-party adapter discovery (v0.20+)"
+    subsection under "Trust-model invariants" documenting the four
+    gates + the `source_type_collision` load-bearing rule.
+
 - **v0.20 — top-level `reviewer_summary` block.** Adds a deterministic
   projection of the reviewer lens surfaces (`tool_surface_diff`,
   capability/intent diff, `action_surface_diff`, evidence matrix) and
--- a/STABILITY.md
+++ b/STABILITY.md
@@ -361,7 +361,7 @@ If a contributor introduces a real need for one of the forbidden surfaces,
 update this section in the same PR. The intent is not "we tried to forbid X"
 — it is that X is *structurally absent* from the scanner's parsing path.
 
-Plugins are off by default. `AGENTS_SHIPGATE_ENABLE_PLUGINS=1` enables loading; `--no-plugins` overrides at the CLI level. When loaded, every plugin is enumerated in `report.loaded_plugins`.
+Plugins are off by default. `AGENTS_SHIPGATE_ENABLE_PLUGINS=1` enables loading; `--no-plugins` overrides at the CLI level. When loaded, every plugin is enumerated in `report.loaded_plugins`, and every third-party adapter (v0.20+) is enumerated in `report.loaded_adapters`.
 
 Plugin validation (v0.17+ / M5). Every entry point is checked against five load-time gates before it can run:
 
@@ -372,6 +372,21 @@ Plugin validation (v0.17+ / M5). Every entry point is checked against five load-
 5. **bad_floor** — `floor_severity` must not exceed `default_severity` on the same metadata block.
 
 Plugins that pass every gate run with the same trust as built-ins. Runtime validation additionally drops findings whose `Finding.check_id` does not match the plugin's declared `id`/`check_id`, drops non-`Finding` items, and captures any exception raised during the plugin call into `loaded_plugins[].runtime_errors`. The scan continues regardless; `--strict-plugins` elevates any non-`valid` plugin or non-empty `runtime_errors` to exit code 4.
+
+#### Third-party adapter discovery (v0.20+)
+
+Third-party adapters register through the `agents_shipgate.adapters` Python entry-point group and provide a class (or instance) satisfying the `ToolSourceAdapter` Protocol — a `source_type: str` ClassVar, a `scope: Literal["per_source", "per_scan"]` ClassVar, an `artifact_class: type | None` ClassVar, and a `load(source, base_dir, manifest)` method returning `LoadedAdapterResult`. Discovery is gated by the same `AGENTS_SHIPGATE_ENABLE_PLUGINS=1` env var as plugin checks; `--no-plugins` forces it off.
+
+Every discovered entry point is checked against four load-time gates before it can register on the scan's `REGISTRY`:
+
+1. **load** — `entry_point.load()` must not raise. Captured as `validation_status="load_failed"`.
+2. **bad_protocol** — the loaded value (a class is instantiated with no args; an instance is used directly) must have all three ClassVars (`source_type` non-empty string, `scope`, `artifact_class`) and a callable `load` method accepting ≤ 3 required positional parameters. Captured as `validation_status="bad_protocol"`.
+3. **bad_scope** — `scope` must be exactly `"per_source"` or `"per_scan"`. Out-of-range values would be silently skipped by the dispatcher. Captured as `validation_status="bad_scope"`.
+4. **source_type_collision** — the adapter's `source_type` must not shadow a built-in (`mcp`, `openapi`, `langchain`, etc.) or another third-party adapter discovered earlier in the same scan. **This is the load-bearing trust rule** — without it, a malicious plugin could displace a built-in adapter and intercept every scan targeting that source type. Captured as `validation_status="source_type_collision"`.
+
+Adapters that pass every gate are registered on the live `REGISTRY` and dispatched by the same pass-1/pass-2 loop as built-ins. The dispatcher's existing `_absorb` artifact-class check fires `TypeError` if a third-party adapter declares one `artifact_class` but returns an artifact of another (the same artifact-smuggling-prevention rule built-ins are subject to). `run_validated_adapter` (in `inputs/adapter_validation.py`) additionally captures runtime exceptions and wrong-return-type failures into `loaded_adapters[].runtime_errors` for callers that wrap the invocation themselves.
+
+`--strict-plugins` (v0.17+) covers BOTH plugin and adapter failures from v0.20+ — any non-`valid` `loaded_plugins[]` row, any non-empty `loaded_plugins[].runtime_errors`, any non-`valid` `loaded_adapters[]` row, OR any non-empty `loaded_adapters[].runtime_errors` elevates the scan to exit code 4. Default behavior remains lenient — failures are recorded in the respective provenance arrays and the scan proceeds.
 
 ### Manifest Schema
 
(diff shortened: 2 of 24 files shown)


pengfei-threemoonslab:
All four findings addressed in commit [`14b943a`](https://github.com/ThreeMoonsLab/agents-shipgate/commit/14b943a). Three P1s required real structural changes; the P2 was a tightening. 7 new regression tests pin each fix.

## P1 #1 + P1 #2 — Per-scan registry (was: global mutation)

You're right — mutating the module-global `REGISTRY` broke both the `--no-plugins` trust boundary AND collision detection across consecutive scans. Reproduced both your test cases.

**Fix**: new `AdapterRegistry.clone()` returns a shallow-copy registry with a fresh `_adapters` dict. `_load_inputs` now builds `scan_registry = REGISTRY.clone()` and threads it through `_load_sources(..., registry=scan_registry)`. Discovery registers ONLY onto the clone; the global stays builtin-only across the process. Existing tests that monkeypatch `REGISTRY._adapters` keep working because `clone()` captures the global's state at the moment `_load_inputs` runs (after monkeypatch).

Two regression tests pin the fix:
- `test_no_plugins_disables_third_party_after_prior_enabled_scan` — scan 1 enables the adapter (runs once); scan 2 in the same process with `plugins_enabled=False` must NOT increment the invocation counter. Asserts both `report.loaded_adapters == []` AND `_RecordingAdapter.invocations == 1`.
- `test_second_scan_does_not_misclassify_third_party_as_collision` — two consecutive scans of the same adapter both report `validation_status == "valid"`. Pre-fix scan 2 would have reported `source_type_collision` while still executing the stale registration.

The autouse `_reset_registry_after_test` fixture I'd added with the original PR is gone — no cleanup needed since discovery doesn't mutate the global.

## P1 #3 — `ToolSourceConfig.type` opened from `Literal` to `str`

You're right — the closed `Literal` made the v0.20 third-party adapter surface unusable for its main advertised use case. A manifest declaring `type: my_third_party_source` failed Pydantic validation before discovery could register the loader.

**Fix**:
- `ToolSourceConfig.type: Literal[...]` → `type: str`.
- New module-level `BUILTIN_TOOL_SOURCE_TYPES` constant for documentation + tests that iterate the curated set.
- New `BUILTIN_PER_SCAN_ONLY_TOOL_SOURCE_TYPES = {"n8n", "openai_api", "anthropic_api", "validation"}` + a `reject_per_scan_only_builtins` model validator preserves the safety rejection (your `test_n8n_top_level_config_is_accepted_and_tool_source_type_is_rejected` still passes for the exact `type: n8n` case at manifest-load time).
- Unknown source types (typos, third-party without a registered adapter) still fail with `ConfigError` (exit 2) — now at dispatcher time via `AdapterRegistry.require`, not at parse time. The exit-code contract is unchanged. `test_n8n*::n8N` case updated to expect the failure at `run_scan` instead of `load_manifest` (same exit code, different layer).
- New regression test `test_per_source_third_party_adapter_referenced_from_manifest`: builds a manifest with `type: demo_source`, registers a matching third-party per_source adapter via entry-points, asserts `run_scan` succeeds, the adapter actually executes, and `report.loaded_adapters[0].validation_status == "valid"`.

The JSON manifest schema (`docs/manifest-v0.1.json`) loses the `enum` constraint on `tool_sources[].type` as a side effect. IDE autocomplete for built-in names is a follow-up; the trust gate is the dispatcher, not the schema.

## P2 #4 — Tighter signature validation

You're right — `_protocol_error` only caught >3 required positional and zero positional. Two failure modes slipped through and would crash the dispatcher at runtime instead of failing at the `bad_protocol` gate:

- `def load(self, source)` — only 1 positional slot; dispatcher's `adapter.load(source, base_dir, manifest)` would fail with `TypeError: too many positional arguments`.
- `def load(self, source, base_dir, manifest, *, must_set)` — required keyword-only; dispatcher passes no kwargs.

**Fix**: `_protocol_error` now requires `accepting_slots >= 3 OR has_var_positional`, AND zero required keyword-only parameters. Four new regression tests:

- `test_gate_bad_protocol_load_too_few_positional` — your example, expects `bad_protocol` with "at least 3 positional" in the error.
- `test_gate_bad_protocol_load_required_keyword_only` — required kw-only is rejected; error mentions both "keyword-only" and the offending param name.
- `test_gate_bad_protocol_accepts_var_positional` — `def load(self, *args)` is valid (can bind 3 args).
- `test_gate_bad_protocol_accepts_optional_keyword_only` — optional kw-only with default is valid.

## Verification

- 1264 tests pass (was 1257; +7 regression tests covering each finding).
- ruff clean.
- `scripts/generate_schemas.py --check` exits 0.
- Coverage 88.02%.
- The four findings reproduce against the previous commit `5400fb2`; each is closed against `14b943a`.

pengfei-threemoonslab:
Round-3 review findings addressed in commit [`943afda`](https://github.com/ThreeMoonsLab/agents-shipgate/commit/943afda). Two P1s and one P2; 4 new regression tests pin each.

## P1 #1 — Runtime failures bypassed the provenance contract

You're right — `run_validated_adapter` existed in `inputs/adapter_validation.py` but was never invoked. The dispatcher in `_load_sources` called `adapter.load(...)` directly for everything, so a third-party adapter that raised at runtime aborted `run_scan` with the raw exception, no report was emitted, and `--strict-plugins` had no chance to see `loaded_adapters[].runtime_errors`.

**Fix**:
- `_load_inputs` now builds a `third_party_records: dict[str, LoadedAdapter]` map (source_type → record) from discovery and threads it to `_load_sources`.
- `_load_sources` branches per adapter: built-ins keep the direct call (their failures should abort loudly — a built-in raising is a scanner bug). Third-party adapters route through `run_validated_adapter`, which captures every exception into `LoadedAdapter.runtime_errors` (mirrored into the matching `loaded_adapters[]` dict) and returns `None`. The dispatcher skips `_absorb` on `None` and the scan continues.
- `_invoke_per_source_adapter` learned a `third_party_record` kwarg that switches to the wrapper.

Two regression tests pin the fix:
- `test_runtime_error_in_third_party_adapter_captured_not_propagated` — a `RuntimeError` raised in `load()` no longer crashes `run_scan`; lands in `runtime_errors`; exit 0 in lenient mode.
- `test_strict_plugins_exits_on_third_party_adapter_runtime_error` — end-to-end via CliRunner, `--strict-plugins` exits 4 specifically on the runtime-error path.

## P1 #2 — `doctor` couldn't introspect third-party source types

You're right — `inspect_sources` called `_load_sources` against the global `REGISTRY` (now builtin-only after the per-scan-clone refactor) with no discovery. A manifest with `tool_sources[].type: demo_source` scanned fine but crashed `doctor` with `ConfigError: No adapter registered`.

**Fix**:
- `inspect_sources` now mirrors `_load_inputs`: `REGISTRY.clone()` → `discover_third_party_adapters(...)` → pass the per-scan registry and `third_party_records` to `_load_sources`.
- New `plugins_enabled: bool | None = None` kwarg defaults to `None` so the existing `AGENTS_SHIPGATE_ENABLE_PLUGINS` env var governs (no doctor CLI flag change needed yet).
- `loaded_adapters` surfaced in the doctor payload alongside `policy_packs` so operators can confirm discovery results without running a full scan.

Regression test `test_doctor_resolves_third_party_source_types` builds the exact manifest you described, asserts `inspect_sources` returns a clean payload, the adapter actually ran, and `loaded_adapters[]` is surfaced.

## P2 #3 — Markdown reports hid adapter validation failures

You're right — `report/markdown.py` rendered `loaded_plugins` but had no `loaded_adapters` section. `load_failed` adapters appeared only in `report.json`.

**Fix**: new `_append_loaded_adapters(lines, report)` renders a `## Loaded Adapters` section. Each entry:

```
- <distribution> <version>: <source_type> — `<validation_status>`
  - validation_error: <message>
  - runtime_error: <message>
```

Hidden when `loaded_adapters` is empty so clean repos see no extra noise. Wired right after `_append_loaded_plugins`.

Regression test `test_markdown_report_renders_loaded_adapters_section` reproduces a `load_failed` adapter and asserts the section appears in `report.md` with both the status and the validation_error visible (accepting markdown's underscore-escape form so `load_failed` / `load\_failed` both satisfy the assertion).

## Verification

- 1268 tests pass (was 1264; +4 regression tests covering the three new findings).
- ruff clean.
- `scripts/generate_schemas.py --check` exits 0.
- Coverage 88.06%.
- All three findings reproduce against commit `14b943a`; each is closed against `943afda`.

pengfei-threemoonslab:
Doc fix landed in commit [`bdbcae1`](https://github.com/ThreeMoonsLab/agents-shipgate/commit/bdbcae1) — STABILITY.md "Third-party adapter discovery (v0.20+)" subsection rewritten to match the code as merged on this branch.

You're right — two prose claims were stale relative to the review fixes:

1. **"register on the live `REGISTRY`"** was wrong since commit `14b943a` (the per-scan-clone fix for P1 #1 + P1 #2). Adapters now register on a per-scan `AdapterRegistry.clone()` built at the start of each `run_scan` / `inspect_sources`. The global `REGISTRY` stays builtin-only across the process lifetime.
2. **"captures runtime exceptions … for callers that wrap the invocation themselves"** was wrong since commit `943afda` (the dispatcher-managed wrapper for P1 #1 review-round-3). The dispatcher in `_load_sources` now routes EVERY third-party adapter `load()` call through `run_validated_adapter` (both pass-1 per-source and pass-2 per-scan loops). Callers do not opt in.

Rewritten section now documents:

- **Per-scan registry contract** with the two invariants it guarantees (`--no-plugins` honest across scans; collision detection honest across scans).
- **Dispatcher trust mechanisms**, separated: artifact-class smuggling prevention (existing `_absorb` `TypeError`, applies to all adapters) and runtime-error capture (new third-party-only path through `run_validated_adapter`). Built-ins keep the direct call shape — a built-in raising means the scanner is broken and must abort loudly; this is now stated explicitly.
- **`doctor` parity**: `inspect_sources` uses the same per-scan registry + discovery + dispatcher path and surfaces `loaded_adapters[]` in its payload.
- **`bad_protocol` gate description** tightened to match the P2 #4 fix: "at least 3 positional slots (or `*args`)" and "no required keyword-only parameters" are now both gated, in addition to the "no more than 3 required positional" rule.
- **New "Manifest `tool_sources[].type`" paragraph**: documents the `Literal → str` relaxation, the `BUILTIN_TOOL_SOURCE_TYPES` constant for docs/tools, the `BUILTIN_PER_SCAN_ONLY_TOOL_SOURCE_TYPES` rejection at manifest-load (preserving the n8n safety), and the `ConfigError` (exit 2) dispatch-time fallback for unknown types.

Pure docs change. Ruff clean; public-surface contract tests pass (the existing drift guards already covered cross-doc consistency for the wedge positioning, gating signal, and schema version — none of which this edit touches).

pengfei-threemoonslab:
## Round-4 review fix: user-facing unknown-adapter remediation

> [P2] The unknown adapter error now gives the wrong recovery path for the new third-party adapter flow. Since `tool_sources[].type` is now open `str` for custom adapter names at [manifest.py:101], a valid third-party manifest can reach `AdapterRegistry.require()` when the adapter package is missing or plugins are not enabled. The error at [protocol.py:197] still says to add the adapter to `_register_builtin_adapters()`, and agent mode routes to "edit shipgate.yaml." That is wrong for the new public extension path; it should mention installing/enabling the third-party adapter, `AGENTS_SHIPGATE_ENABLE_PLUGINS=1`, or fixing a typo.

Fixed in 92c4488. The fix is two-layer:

**1. `AdapterRegistry.require()` now emits a user-facing message.** It branches on `_adapter_plugins_enabled(None)`:
- plugins enabled → "install the third-party adapter package" / "fix a typo"
- plugins disabled → "set `AGENTS_SHIPGATE_ENABLE_PLUGINS=1`" / "install the package" / "fix a typo"

No more "Add the adapter to `_register_builtin_adapters()`".

**2. Agent mode no longer routes to "edit shipgate.yaml".** `_diagnose_config_error` in `cli/_helpers.py` now pattern-matches the `require()` error prefix and routes to a new diagnostic, `SHIP-DIAG-UNKNOWN-ADAPTER-SOURCE-TYPE`, instead of falling through to `SHIP-DIAG-INVALID-MANIFEST`. The new diagnostic emits 2 (plugins enabled) or 3 (plugins disabled) ranked `next_actions`, with the rank-1 being:
- plugins enabled: `kind=command`, `command="pip install <third-party-adapter-package>"`
- plugins disabled: `kind=command`, `command="AGENTS_SHIPGATE_ENABLE_PLUGINS=1 agents-shipgate scan -c <path>"`

The manifest is valid; `kind=edit` is rank-N (only for the typo case), never rank-1.

**Regression tests** in `tests/test_adapter_entry_point_discovery.py`:
- `test_unknown_adapter_source_type_routes_to_install_enable_diagnostic` — drives a scan with `type: not_a_real_adapter`, captures the agent-mode error JSON on stderr, asserts `next_actions[0].kind == "command"`, the command contains `AGENTS_SHIPGATE_ENABLE_PLUGINS=1`, and the diagnostic id is `SHIP-DIAG-UNKNOWN-ADAPTER-SOURCE-TYPE` (not `SHIP-DIAG-INVALID-MANIFEST`).
- `test_require_error_message_is_user_facing` — pins `str(exc)` to not contain `_register_builtin_adapters` and to contain `AGENTS_SHIPGATE_ENABLE_PLUGINS` and "typo".

Docs: `docs/diagnostics.md` gets a catalog row for the new ID, marked `v0.20+`.

**Verification**: 1772 passed, 4 skipped; ruff clean.

PR #201, opened 2026-06-11
Add stale-install agent-protocol handshake (Phase 1 close-out)

pengfei-threemoonslab:
## Summary

- Adds the **stale-install** fail-safe to the agent-native protocol. The protocol already covered an *absent* binary with a canonical schema-valid `agent_result_v1` object (`missing-install.json`), but a binary that is *present yet older than the required `>=0.13.0`* had only prose version gates in the skill prompts — no canonical object. A stale copy lingering on `PATH` can emit an outdated schema or miss commands (a plain `pipx install` is a no-op over an older build), which is the exact stale-binary footgun seen in the design-partner pilot.
- Closes the gap symmetrically with the existing missing-install pattern:
  - `stale-install.json` fixture, mirrored byte-identically in `tests/golden/agent_protocol/` and `examples/agent-protocol/expected/`
  - a "Stale Install" section in `docs/agents/protocol.md`
  - an `examples/agent-protocol/README.md` bullet
  - two tests in `tests/test_agent_protocol.py`
- **No schema churn:** reuses the existing `install` action kind (`command: pipx upgrade agents-shipgate`) rather than adding an `upgrade` value to the closed `AgentResultActionKind` Literal. The `install`-kind validator branch short-circuits the block-route checks, so `block`+`install`+`must_stop` validates exactly like missing-install. Stale binaries now fail closed with a routable object instead of agent-authored prose.

## Type

- [ ] Input adapter change
- [x] Documentation only (protocol doc + fixtures + tests; no Pydantic model or runtime behavior changed)

## Verification

Additional local checks run:

- `pytest tests/test_agent_protocol.py` — green (new fixtures validate against `docs/agent-result-schema.v1.json` and `AgentResultV1`)
- `ruff check tests/test_agent_protocol.py` — clean
- `scripts/generate_schemas.py --check` — clean (no model changed)
- Live `shipgate check --agent claude-code --format agent-json` smoke — healthy

## Release-readiness notes

- [x] New or changed check IDs are documented in `docs/checks.md` (n/a — no check IDs added)
- [x] Report/schema changes are additive or documented in `STABILITY.md` (n/a — no schema change; reuses existing `install` action kind)


pengfei-threemoonslab:
Thanks — both addressed in `b211618`.

**P2 (state machine lacks an `install` branch): fixed.** Good catch; this actually pre-dated the PR (`missing-install` already used `block`+`actor=coding_agent`+`kind=install`). Added an explicit state-machine row — run the install/upgrade command, rerun `shipgate check`, do not report completion until `allow`/`warn` — plus a note distinguishing it from the repair loop, and a content test (`test_protocol_state_machine_documents_install_branch`).

**P1 (stale-install not wired into shipped prompts): addressed, with one correction.** The premise that the stale path is uniquely prose-only isn't quite right — `missing-install` is *also* not referenced in any agent-ingested prompt/skill/renderer (grep across `adoption-kits/`, `prompts/`, `skills/`, `.agents/`, `src/.../discovery` returns zero hits for both), and the Codex preflight is already fail-closed (`recipes.md`: "Do not continue … until the CLI exists and is `>=0.13.0`"). So this PR added the fixture at parity with the existing pattern, not a regression.

That said, the underlying point is fair: the preflight stopped the agent but never pointed at the canonical object. The Codex skill `references/recipes.md` CLI Preflight now cross-references the protocol's Missing/Stale Install sections and both fixtures, and reinforces "do not report the task complete" until the version is current — synced across all three byte-identical copies (adoption-kit source, `.agents` dogfood, plugin), with the rendered `EXPECTED_CODEX_SKILL_RENDER_SHA256` bumped, the outgoing hash appended to the kit sidecar `prior_render_sha256`, and a renderer parity assertion that the preflight references both `missing-install.json` and `stale-install.json`.

I deliberately kept the existing **ask-the-user / run-the-pinned-command** stance rather than having the agent autonomously `pipx upgrade` the user's environment — the protocol's `actor="coding_agent"` vs. the skill's "ask the user" is a real product decision worth settling on its own rather than slipping in here. Happy to follow up with the symmetric tightening of the Claude Code skill `SKILL.md` and that autonomous-install question if you'd like.

No `agent_result_v1` schema change. Full suite, ruff, and schema-drift all green.

PR #203, opened 2026-06-11
Close the check/verify coverage gap: defer tool-surface changes to verify

pengfei-threemoonslab:
## Summary

- Closes the **`check`/`verify` semantic gap** — the Phase 2 correctness item where the agent-native local loop could disagree with the one decision engine.
- `shipgate check` routes only through the boundary evaluator (host/trust-root surfaces: Codex/host config, MCP approvals, CI gate, agent instructions, policy, skills). It never inspects the **tool-use capability surface** — only `verify` computes the capability delta. So a diff that adds a capability to a declared tool source returned `decision="allow"`, `completion_allowed=true` from `check`, while `verify` would block it. That is a second decision engine that can contradict `release_decision.decision`.
- **Fix:** when `check` produces a clean boundary `allow`, has no full-scan `release_decision`, **and** the diff changes a file the manifest declares as a `tool_sources[].path`, escalate `allow → warn` and route `first_next_action` to `verify`, with a `diagnostics[].code="capability_change_requires_verify"` marker and a `trace[].step="coverage"` event. `check` defers capability gating to the one decision engine instead of green-lighting a change it never evaluated.

## Why this design

- **Manifest-declared `tool_sources[].path` is the authoritative signal**, not a heuristic. In an opted-in repo the trigger force-runs on *every* PR (even docs-only), so `should_run` alone would warn on docs and regress the prized "0 docs-only noise" property. Using the manifest's declared tool sources means the warn fires only on genuine capability surfaces — docs/test/boundary-only diffs still return `allow`.
- **`warn`, not `block`.** Completion stays allowed (check isn't the merge gate; verify/CI is), but the agent is told to run the real gate. Blocking every tool-surface edit would be a false positive and make `check` unusable.
- **No second engine.** `check` doesn't try to reproduce the capability scan; it defers to `verify` / `release_decision.decision`.
- **No `agent_result_v1` schema change** — reuses existing fields (`warn` decision, `diagnostics`, `trace`). Best-effort manifest load; an absent/invalid manifest degrades to prior behavior. Covers both the `check` CLI and the MCP server (both route through `build_codex_agent_result`).

## Behavior

| Diff in an opted-in repo | Before | After |
|---|---|---|
| changes a declared `tool_sources[].path` | `allow` | `warn` → `agents-shipgate verify --json` |
| docs/test-only | `allow` | `allow` (unchanged) |
| boundary file (`.codex/config.toml`, policy, CI gate) | `allow`/`block` per rules | unchanged |
| boundary block + tool-surface change | `block` | `block` (coverage only escalates from `allow`) |

## Type

- [x] CLI or GitHub Action behavior (`shipgate check` routing; no schema change)

## Verification

Additional local checks run:

- New tests in `tests/test_codex_boundary_check.py`: declared tool-source change → warn+verify+diagnostic+`trace.step="coverage"`; no manifest / docs change → allow; boundary block + tool-surface change stays blocked.
- `ruff check .`, `compileall -q src tests`, `scripts/generate_schemas.py --check` — all clean
- Manual repro confirmed: declared tool-source change → warn; README → allow (no noise); `.codex/config.toml` → allow.

## Release-readiness notes

- [x] No user-code import added to default scan paths (best-effort YAML manifest load only)
- [x] New or changed check IDs documented in `docs/checks.md` (n/a — no check IDs added)
- [x] Report/schema changes are additive or documented (n/a — no schema change; reuses existing `agent_result_v1` fields; protocol documented in `docs/agents/protocol.md` → Coverage)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## 0.13.0 - Unreleased
 
+- **`check` defers tool-surface changes to `verify` (coverage boundary).**
+  `shipgate check` is boundary-scoped and does not compute the capability
+  delta, so a clean boundary result over a diff that changes a
+  manifest-declared `tool_sources[].path` no longer returns `allow` — it
+  returns `decision="warn"` routing `first_next_action` to `verify`, with a
+  `diagnostics[].code="capability_change_requires_verify"` marker and a
+  `trace[].step="coverage"` event. Completion is still allowed, but `check`
+  no longer green-lights a capability change only `verify` gates, so the local
+  loop cannot disagree with `release_decision.decision`. Docs/test/boundary-only
+  diffs are unaffected (still `allow`); no `agent_result_v1` schema change.
 - **Agent-mode auto-detection.** Agent mode now auto-enables when a known
   coding-agent harness environment is detected (Claude Code exports
   `CLAUDECODE=1`, Cursor `CURSOR_TRACE_ID`), so structured `next_action`
--- a/docs/agents/protocol.md
+++ b/docs/agents/protocol.md
@@ -113,6 +113,25 @@ must set `human_review.required=true` and stop the agent.
 `repair.forbidden_shortcuts` is present on every result, including `allow`, so
 agents have the same trust-root boundary even when no finding fires.
 
+## Coverage
+
+`shipgate check` is boundary-scoped: it evaluates host and trust-root surfaces
+(Codex/host config, MCP approvals, the Shipgate CI gate, agent instructions,
+policy, and skills) from the diff. It does **not** compute the tool-use
+capability delta — that is `verify`'s job, and `release_decision.decision`
+remains the one authoritative capability gate.
+
+So that `check` never disagrees with that gate, a clean boundary result over a
+diff that changes a **manifest-declared tool source** (a `tool_sources[].path`
+entry) does not return `allow`. It returns `decision="warn"` with
+`first_next_action.kind="warn"` routing to `verify`, plus a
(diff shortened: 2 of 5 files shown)


pengfei-threemoonslab:
Good catch — fixed in `fb1a1c0`. Confirmed your repro: `openai_agents_sdk` with `path: agents` + a change to `agents/refund_agent.py` returned `allow/continue` because the matcher only exact-matched the declared path.

**Fix:** match a changed file when it equals the declared path **or sits under it** (`changed == decl or changed.startswith(decl + "/")`), which covers both file sources (`mcp`, `openapi`) and the directory-scanning types (`openai_agents_sdk`, `google_adk`, `langchain`, `crewai`, `codex_plugin`, `codex_config`) without needing to stat. The trailing `/` prevents prefix collisions (`agent` vs `agents/...`).

**On the broad-root caution you raised** — handled with two guards so directory matching doesn't become docs-only noise:
- A declared path that resolves to the workspace root (`codex_config path: .`) is dropped — it would otherwise match every file.
- Changed files the boundary evaluator already inspects are dropped via a new `is_boundary_path()` helper, so a `codex_config` source overlapping `.codex/config.toml` doesn't produce a redundant warn for a file `check` did evaluate.

Verified end-to-end:

| Case | Result |
|---|---|
| `openai_agents_sdk path: agents` + `agents/refund_agent.py` | `warn` → verify |
| `codex_config path: .` + `README.md` | `allow` (no noise) |
| `codex_config path: .` + `.codex/config.toml` | `allow` (boundary-evaluated) |

Added tests for the directory-source match and the broad-root exclusion. Full suite, ruff, compileall, and schema-drift green.

PR #204, opened 2026-06-11
Add host-grant drift detection (Phase 3: proactive governance)

pengfei-threemoonslab:
## Summary

- **Phase 3 ("make Shipgate proactive") construction item: policy drift detection.** PR-time `SHIP-HOST-BOUNDARY-*` checks only see grants that change inside a reviewed diff. Host grants also change *outside* that loop — a coding agent (or a hurried human) edits `.mcp.json`, `.claude/settings.json`, or a workflow directly, and nothing re-reviews the new authority. This PR closes that loop on top of the existing deterministic `audit --host` inventory.
- **`audit --host --save-baseline`** records the current grants as the acknowledged state in `.agents-shipgate/host-grants.json`. The snapshot is content-only (no timestamps, machine paths, or CLI version), so re-saving an unchanged state is **byte-identical** (no commit noise). It is meant to be committed — and `.agents-shipgate/` is already a verify trust-root surface, so PR edits to the snapshot stay release-visible with zero new machinery.
- **`audit --host --drift`** deterministically diffs current grants against the baseline: per-category added/removed/changed buckets plus **`expansion_signals`** naming the authority-broadening shapes — new MCP server, allow rule added (wildcards flagged), **`deny`/`ask` rule removed** (removals broaden here — which is exactly why the gate fires on *any* drift instead of trying to classify "safe" directions), hook added, workflow gaining write scopes / `write-all` / `pull_request_target`, Codex config appearing.
- **Advisory by default; `--fail-on-drift` exits 20** for scheduled CI gates. Missing / corrupt / unknown-schema-version baselines exit 2 with the exact recovery command in the message.

## Why this design

- **Any-drift gate, expansion signals as presentation.** Direction is not a safe/unsafe oracle (a removed `deny` rule is a *broadening*; an expansion-only gate could also be gamed by remove-then-re-add). The gate is snapshot equality; the signals tell the reviewer where to look first.
- **No new decision engine, no new check IDs.** This is a read-only command over the existing inventory — it avoids the three-place check registration and never competes with `release_decision.decision`.
- **Reuses the existing trust model.** Snapshot tampering in a PR is already caught by the verify trust-root surface for `.agents-shipgate/**`; the report's footer explicitly tells agents/humans "do not re-save to silence drift you have not reviewed."
- **Idempotent + deterministic** by construction (canonically sorted categories, content hashing, stable output ordering) — consistent with the repo's determinism discipline.

## Usage

```bash
# One-time, after a human reviews the current grants:
agents-shipgate audit --host --save-baseline

# Scheduled (or pre-commit):
agents-shipgate audit --host --drift --fail-on-drift   # exit 20 on drift
```

Scheduled-CI recipe: `examples/github-actions/11-host-grant-drift.yml`. Docs: new "Host-grant drift detection" section in `docs/mcp-governance.md`.

## Type

- [x] CLI or GitHub Action behavior (new `audit` flags; no scan/verify behavior change)
- [ ] Report, schema, or SARIF output (no report-schema change; new standalone `host_grants_schema_version: 0.1` snapshot/diff payloads)

## Verification

Additional local checks run:

- 13 new tests in `tests/test_host_audit.py`: snapshot normalization + secrets exclusion, idempotent re-save, clean no-drift, expansion signals (incl. removed-deny and the already-baselined-wildcard non-signal), workflow write-scope change → `changed` + `workflow_write_expanded`, exit 20 / exit 0 gating, missing/corrupt/unknown-version baselines → exit 2, mutually-exclusive flags, payload determinism
- `ruff check .`, `compileall -q src tests`, `scripts/generate_schemas.py --check` — all clean
- Manual end-to-end: save → no-drift → expand (server + wildcard + deny removal) → drift report with 4 expansion signals → exit 20 with the flag

## Release-readiness notes

- [x] No user-code import added to default scan paths (JSON/YAML host files only, same as the existing inventory)
- [x] New or changed check IDs are documented in `docs/checks.md` (n/a — no check IDs; command surface only)
- [x] Report/schema changes are additive or documented (no report-schema change; the snapshot carries its own `host_grants_schema_version` and unknown versions fail closed)


pengfei-threemoonslab:
Both findings were valid — I reproduced all four cases exactly as described before fixing. Addressed in `441981c`.

**P1 (existing-grant edits invisible to drift): fixed with redacted config hashing.** The inventory stored only display projections, so an args edit on an existing server and a hook-command edit under an existing event both produced `has_drift=false`. MCP server and hook entries now carry a `config_sha256` over their **full** configuration, with secret-bearing `env`/`headers` *values* redacted before hashing (keys kept). Consequences, all test-locked:

| Edit | Before | After |
|---|---|---|
| existing server's `args` changed | no drift | drift + `mcp_server_changed` signal |
| hook command changed under existing event | no drift | drift + `hook_changed` signal |
| env/header token **value** rotated | no drift | no drift (by design — rotation ≠ authority change; key-set changes still fire) |

The hash-not-raw-config choice keeps the committed baseline free of anything secret-shaped while making every grant-shaping field (args, commands, matchers, URL, env/header keys) drift-visible.

**P2 (baselines not failing closed): fixed at load time.** `load_host_grants_baseline` now (1) validates per-category shapes — lists, with dict or string entries as appropriate — and (2) verifies the stored `inventory_sha256` against the inventory content. Your two repro cases now both take the routable exit-2 path: bogus stored hash → "failed its integrity check … re-record" (was silent exit 0), malformed-shape JSON → exit 2 with the re-record command (was a `TypeError`/exit 1). This also makes the baseline self-tamper-evident, complementing the PR-time trust-root visibility of `.agents-shipgate/`.

Six new tests cover the four reproduced cases plus the env-rotation non-drift property and a three-shape malformed-baseline sweep. Docs (`mcp-governance.md`) and the CHANGELOG bullet updated to state the exact drift semantics. Full suite, ruff, compileall, and schema-drift all green.

One known boundary left as-is (documented, happy to discuss): `.codex/config.toml` content edits are tracked by presence only here — diff-time semantics for those files belong to the `SHIP-CODEX-BOUNDARY-*` checks and `shipgate check`, and a raw-bytes hash would make token rotation inside the TOML count as drift (the noise this PR deliberately avoids for MCP/hooks). A redacted TOML projection would be the right follow-up if we want scheduled drift parity there too.

pengfei-threemoonslab:
Valid again — the blanket redaction traded one blind spot for another. Reproduced your exact case (`READ_ONLY=true→false` on an existing server → `has_drift=false`, `--fail-on-drift` exit 0) before fixing. Addressed in `e300c7c`.

**Fix: per-key redaction instead of blanket redaction.** Inside `env`/`headers`, only values under **secret-looking keys** are redacted before config hashing. Classification reuses the existing report-redaction vocabulary (`core/privacy.SENSITIVE_VALUE_KEYS` + `cookie`/`passphrase`), matched as substrings of the normalized key — so `GITHUB_TOKEN`, `OPENAI_API_KEY`, `AWS_SECRET_ACCESS_KEY`, `Authorization`, `Proxy-Authorization` classify as secret, while `READ_ONLY`, `ALLOWED_PATHS`, `GITHUB_TOOLSETS`, `X-Scope` hash normally. No new vocabulary to maintain; one shared notion of "secret-looking" across the product.

Resulting semantics, all test-locked:

| Edit | Drift? |
|---|---|
| `READ_ONLY=true → false` on existing server | **yes** + `mcp_server_changed` (your repro) |
| `GITHUB_TOKEN` / `Authorization` value rotation | no |
| new non-secret header key (`X-Scope: admin`) | **yes** |
| env/header key added or removed | yes (unchanged) |

**On the failure direction:** misclassification now fails safe in the direction that matters. A secret stored under an unconventional key name (`MY_CREDS=...`) causes drift *noise* on rotation — the gate fires and a human looks — never a silent blind spot. And the privacy posture is unchanged in all cases: raw values never enter the committed baseline, only the final `config_sha256`.

**On the "let projects mark secret keys" half of your suggestion:** deliberately deferred. `audit` is a zero-config surface, and adding a manifest knob for key classification before any real-world false-positive shows up would be premature config surface. The shared-vocabulary heuristic plus the fail-safe direction covers the conventional cases; if a project hits rotation noise from an unconventional key name, that's the signal to add the (small, additive) override — happy to do it then.

Docs and CHANGELOG updated to state the exact per-key semantics. Full suite, ruff, compileall, schema-drift all green.

PR #205, opened 2026-06-12
Add accepted-debt exception workflow (Phase 4: organizational governance)

pengfei-threemoonslab:
## Summary

- **Phase 4 ("make Shipgate organizational"), built selectively.** The roadmap's own discipline says org features are earned, not presumed — so this PR ships the one Phase 4 item that strengthens governance even for a single repo: the **exception workflow with approval requirements**, on the baseline (accepted-debt) mechanism.
- **The gap was real and sharp:** the v0.5 provenance contract *documented* reviewer-set `reason`/`expires`, but `baseline save` offered **no flags to set them**, and hand-editing the JSON trips the integrity hash (`SHIP-BASELINE-INTEGRITY-MISMATCH`) — the documented exception workflow was effectively unusable. And unlike the sibling exception mechanisms (`human_ack`, severity-override acknowledgements), baseline entries carried **no approver identity at all**.
- **Baseline schema 0.6** adds optional `provenance.owner`; legacy 0.2–0.5 files still load, and re-saving upgrades them.

## What's new

**Approval metadata, settable through the CLI (never inferred):**
```bash
agents-shipgate baseline save --owner alice --reason "accepted pending Q3 fix" --expires 2026-09-30
```
Stamped on newly-accepted entries; flagless re-saves preserve prior metadata verbatim. `--apply-to-existing` fills the fields into existing entries that lack them — **never overwrites a previously-set value and preserves each entry's original `recorded_at`/`run_id`**, so acknowledging legacy debt doesn't rewrite its history. Approval is declared, never inferred — the `human_ack` contract.

**Debt aging report + org governance gate:**
```bash
agents-shipgate baseline status --json                                  # advisory
agents-shipgate baseline status --require-owner --require-expiry --max-age-days 180   # exit 20 on violations
```
Per-entry owner/reason/age/expiry with expired/unowned/no-expiry flags and a summary (expiring-soon, oldest entry). `--as-of YYYY-MM-DD` pins the date for reproducible CI output.

## Design decisions a reviewer should weigh

- **Gate-by-CLI-flags, not manifest knobs** — same pattern as `audit --host --drift --fail-on-drift` (#204): org enforcement composes in a scheduled workflow without new manifest surface, new check IDs (no three-place registration), or a second decision engine. Advisory by default; exit 20 only when a gate flag is passed.
- **Expired entries violate `--require-expiry`** — an expired exception is no longer a valid exception.
- **Entries without provenance fail every active gate** — unknown history is *ungoverned* debt, not exempt debt (fail closed, consistent with the trust model).
- **Fill-only `--apply-to-existing`** — bulk acknowledgement cannot silently re-attribute debt someone else approved.

## Deliberately NOT built (org plane stays earned-only)

Hosted/cross-repo aggregation, GitHub App, signing, a per-entry `baseline annotate <fingerprint>` command (the per-save batch + fill-only covers the real workflow; granular ownership is a follow-up if a team actually needs split ownership), and manifest governance knobs.

## Type

- [x] CLI or GitHub Action behavior (`baseline save` flags, new `baseline status`)
- [x] Report, schema, or SARIF output (baseline schema 0.5 → 0.6, additive `provenance.owner`; documented in `docs/baseline.md`; no report-schema change)

## Verification

Additional local checks run:

- New `tests/test_baseline_status.py` (12 tests): stamping on new entries, flagless re-save preservation, fill-only/never-overwrite semantics with history preservation, legacy-file load, status math (ages, expiry windows, summary counts), fail-closed legacy gating, CLI advisory/gate exit codes (0/20), `--as-of` validation (2), missing baseline (3), determinism
- Updated the three version-literal assertions (0.5 → 0.6); the stale `docs/baseline.md` schema-version paragraph (said 0.4) corrected
- Full suite `pytest -n auto -m "not perf"` — green (exit 0); `ruff check .`, `compileall`, `scripts/generate_schemas.py --check` — clean
- Manual end-to-end on `samples/support_refund_agent`: save with metadata → status advisory → gates exit 0 → time-travel `--as-of` past expiry → exit 20 → flagless re-save preserves owner → gate exit 0

## Release-readiness notes

- [x] New or changed check IDs are documented in `docs/checks.md` (n/a — no check IDs added)
- [x] Report/schema changes are additive or documented (baseline 0.6 additive `owner`; documented in `docs/baseline.md` + CHANGELOG; legacy versions load)

### DIFF
--- a/.well-known/agents-shipgate.json
+++ b/.well-known/agents-shipgate.json
@@ -206,7 +206,7 @@
     "3": "input_parse_error",
     "4": "other_error",
     "6": "baseline_integrity_failure",
-    "20": "strict_gate_failure"
+    "20": "gate_failure (strict-mode scan/verify; or opt-in governance gates: baseline status --require-owner/--require-expiry/--max-age-days, audit --host --drift --fail-on-drift)"
   },
   "agent_mode_env_var": "AGENTS_SHIPGATE_AGENT_MODE",
   "plugin_opt_in_env_var": "AGENTS_SHIPGATE_ENABLE_PLUGINS"
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -331,8 +331,15 @@ checks:
 ### Task 4 · Save a baseline before enabling strict CI
 
 ```bash
-agents-shipgate baseline save -c shipgate.yaml --out .agents-shipgate/baseline.json
+agents-shipgate baseline save -c shipgate.yaml --out .agents-shipgate/baseline.json \
+  --owner <human> --reason "<why accepted>" --expires <YYYY-MM-DD>
 ```
+
+`--owner`/`--reason`/`--expires` (v0.13+) record who accepted the debt, why,
+and the review-by date on newly-accepted entries. They are human-declared
+values: an agent must ask the user, never invent them, and blank values are
+rejected. `--apply-to-existing` fills the fields into existing entries that
+lack them without overwriting previously-set values.
 
 Then in CI:
 
@@ -343,6 +350,11 @@ agents-shipgate scan -c shipgate.yaml \
 ```
 
 Strict mode fails CI only on **new** findings (those not in the baseline).
+`agents-shipgate baseline status --json` reports accepted-debt aging
+(owner, age, expiry); with `--require-owner` / `--require-expiry` /
+`--max-age-days N` it exits `20` on violations (advisory exit `0` without
+gate flags) — parse `violations[]` from the JSON, then route to a human:
+acknowledging debt is a human decision.
 
 ### Task 5 · Explain a check or a specific finding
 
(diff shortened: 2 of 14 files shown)


pengfei-threemoonslab:
Both addressed in `e351c44`.

**P2 (blank owner satisfies `--require-owner`): fixed in two layers.** Confirmed the mechanism — the save flag accepted `""`/whitespace, `unowned` tested only `is None`, and the human output rendered the same entry as `owner=—` while the gate passed.

1. **Save-side rejection:** `baseline save` now exits 2 on blank `--owner`/`--reason` ("declared human authority requires a non-empty value" — the `human_ack` non-empty contract), and strips surrounding whitespace before stamping.
2. **Status-side normalization (defense in depth):** `baseline status` treats blank/whitespace owner/reason as missing — the JSON payload normalizes them to `null`, `unowned` flips true, and `--require-owner` exits 20 — so files produced by any means other than the CLI still gate honestly.

Bonus consistency already in place and now test-locked: the `--apply-to-existing` fill-only pass treats a blank owner as missing, so a later acknowledged save repairs it without overwriting anything real. New tests: blank/whitespace flags → exit 2; hand-built blank-owner baseline → `unowned` + gate exit 20; blank-owner fill.

**P3 (contract docs omit the gate surface): updated everywhere automation authors look.**
- `STABILITY.md`: `baseline save` row gains the v0.13 flags; new `baseline status` row with gate semantics; exit-code `20` extended from "strict-mode gate failure" to cover the opt-in governance gates (`baseline status --require-owner/--require-expiry/--max-age-days`, and `audit --host --drift --fail-on-drift` from #204, which had the same doc gap).
- `AGENTS.md`: Task 4 now shows the metadata flags with the agent boundary spelled out ("human-declared values: an agent must ask the user, never invent them"), plus the command-table rows and the parse-`violations[]`-then-route-to-human loop for `baseline status`.
- Machine-readable catalogs: `docs/errors.json` and `.well-known/agents-shipgate.json` exit-code `20` entries updated to the same contract; `llms-full.txt` regenerated (its up-to-date contract test caught the doc edits — nice safety net).

Full suite, ruff, and JSON validity green; CI running on the pushed commit.

pengfei-threemoonslab:
Confirmed and fixed in the latest commit. Classic falsiness gap: my previous round's "blank counts as missing for fill" claim was only true for `""` — whitespace-only is truthy, so `provenance.owner or owner` preserved `"   "` while `status` (correctly) kept treating it as unowned. The recommended repair path could never clear the gate it was recommended for.

The fill-only merge now uses the exact same blank-as-missing rule as `baseline_status_payload`, via a shared-semantics `_filled_text()` helper: an existing value counts only when it has non-whitespace content; a real value with incidental padding (`"  alice  "`) is still never overwritten.

Tests now assert the full repair loop for both `""` and `"   "` — not just that the owner gets filled, but that `baseline_status_violations(..., require_owner=True)` is empty afterwards — plus the padded-real-value preservation case. Reproduced your exact scenario before the fix (`owner` stayed `'   '`) and after (`'alice'`, gate clears). Full suite, ruff green; CI running.

PR #255, opened 2026-07-06
Publish the first adjudicated real-history accuracy row (2026-W26)

pengfei-threemoonslab:
**Stacked on #254** (merge that first; GitHub retargeting note: re-base this to main after #254 lands if the branch isn't auto-deleted).

Roadmap item 1 (阻塞级): the W26 mined corpus (120 real merged PRs: stripe/agent-toolkit, block/goose, pydantic/pydantic-ai) is now labeled, scored, and **published — ugly numbers included**, per the "难看也发" instruction.

## What the numbers say

- Funnel: 110/120 trigger-skip (consistent with the 93% base rate across W24–W26's 361 PRs), 6 evaluated (all stripe), 4 scan_failed (the chipped goose duplicate-action_id crash).
- Adjudicated ground truth (10 rows): 8 `safe_to_merge`, 2 `needs_human`, 0 `must_block`.
- Score: `needs_human_caught` **1.0** — both authority-bearing changes (a new auto-synced skill directing CLI installs; a skill-sync supply-chain rewiring with a dropped API-key requirement) were held for a human. `benign_escalation_rate` **0.0**. `ie_rate_on_safe` **0.5**; the other 4 safe rows crashed. **The gate never wrongly passes and never cleanly passes** on this corpus — abstention (config-bound gap) or crash, which is precisely what the in-flight dynamic-toolkit work must move.

## Method disclosure (in the run notes, so the caveat travels with the numbers)

Two **independent AI labelers** (separate contexts, no coordination, real diffs fetched) per `LABELING.md`; disagreement **0/10** — both independently flagged the same two rows as `needs_human` with the same reasoning; third-pass adjudication. Labels are AI-generated pending human spot-check.

## Where it's published

- `benchmark/miner/results/2026-W26-mined.labels.csv` (adjudicated, per protocol: only the adjudicated file is committed).
- `benchmark/miner/README.md` § "2026-W26 labels + score" (full confusion matrix, method, reading).
- README status banner now states the real numbers instead of "no precision/recall numbers are published yet" — small n, framed exactly as small n.

`test_public_surface_contract` + `test_miner_corpus` green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #256, opened 2026-07-06
Fix the two real-history failure modes: action_id crash degrade + config-bound capability detection

pengfei-threemoonslab:
Roadmap item 2 (the W26 corpus's active fix), implemented per [`docs/engineering/config-bound-capability-detection.md`](docs/engineering/config-bound-capability-detection.md) and **calibrated against the real repos** as the design doc demands.

## Fix 1 — duplicate `action_id` degrades instead of crashing (`0e5e282`)

The goose crash (`Duplicate action_surface action_id`, 4/120 W26 rows dead at exit 2) survived #226 via a second path: `compute_action_surface_diff` on a base reference serialized by a pre-#226 engine. Reproduced first, then fixed with #226's own pattern — an optional warnings sink, degrade via the existing tool-name-suffix disambiguator (reconstructs the same ids head-side dedup produces, so base pairs cleanly), warning lands in `source_warnings` → review_required. Sinkless callers keep the hard error.

## Fix 2 — config-bound capability detection (`1e1e488`)

All three detections from the design doc, hung on the existing `toolkit_scope_bound` fact channel (the one that already round-trips base↔head):
- `SHIP-CAP-CONFIG-BINDING-REMOVED` (high, suppression-immune): a factory's config binding deleted between base and head — the dangerous "cleanup that expands authority" case.
- `SHIP-CAP-CONFIG-BINDING-CHANGED` (medium review item): both sides bound + the bound config file in `changed_files`.
- `SHIP-SCOPE-TOOLKIT-UNBOUNDED` remedy upgraded with the explicit-inventory alternative + factory site; broadening checks now skip `config`/`unknown` states so they can't claim "full surface mounted" on surfaces they can't see.
- New conservative tracer (`inputs/config_trace.py`): json/yaml/toml loads, `os.environ`, in-file pydantic settings; same-file, enclosing-scope only; unresolvable → `unknown`, and **`config → unknown` never fires** — the design doc's false-positive guard.

## Real-repo calibration (fresh clones, all 10 W26 engaged rows re-evaluated on this branch)

| Row | before | after |
|---|---|---|
| stripe/agent-toolkit ×6 (skill-sync PRs) | `insufficient_evidence` | `insufficient_evidence` — **correct**: these diffs don't touch the binding; the blind spot this PR closes (silent binding removal) is proven by the fixture e2e (`IE parity → SHIP-CAP-CONFIG-BINDING-REMOVED + review_required`) |
| block/goose#9798 / #9684 | `scan_failed` (crash) | `insufficient_evidence` |
| block/goose#9637 | `scan_failed` (crash) | `insufficient_evidence` |
| block/goose#9717 | `scan_failed` (crash) | `human_review_required` |

Post-fix score on the W26 labels (#255): `unscored` 4→0; the crash class is eliminated from real history. One safe PR (goose#9717, an endpoint-deletion capability *reduction*) now review-routes — conservative, surfaced, and honest, vs silently dead before.

## Acceptance (design doc, all three)

1. Fixture repo reproduces the pilot scenario: binding removal → `SHIP-CAP-CONFIG-BINDING-REMOVED`, `review_required`, not IE parity. ✅
2. **Zero golden churn** — full suite green, `samples/` untouched (no false positives on static repos). ✅ (independently re-run: exit 0)
3. `fix_task` carries the inventory remedy with `source path:line`. ✅ (asserted in-test)

Scope cuts documented in-code: LangChain/CrewAI surfaces get the `config_binding` annotation but no base↔head carriage channel yet; severity calibration removal=high / changed=medium per spec.

Merge-order note: touches generated `llms-full.txt` + `docs/checks.json`, as does #254 — whichever merges second re-runs the generators.


pengfei-threemoonslab:
P1 fixed — good catch, and thank you for reproducing it. The retarget (`config/safe.json → config/broad.json` with only the `.py` in the diff) is now its own trigger: when both sides carry a literal `config_path` and they differ, `SHIP-CAP-CONFIG-BINDING-CHANGED` fires with `kind: config_binding_retargeted` and `previous_config_path → config_path` evidence, independent of `changed_files`. Conservative guards kept: pathless env/settings bindings never compare (existing `test_pathless_config_binding_never_guesses_a_match` still pins this) and `config → unknown` still never fires. Regression test reproduces your exact scenario. Full suite + ruff green.

PR #253, opened 2026-07-06
Add the any-coding-agent guide; route non-named harnesses explicitly

pengfei-threemoonslab:
## Why

Agent-adoption review finding: the agent-native surface is real but **gated on being one of three named harnesses**. Cline, Windsurf, Devin, Aider, OpenHands get no auto agent-mode (only `CLAUDECODE=1` / `CURSOR_TRACE_ID` auto-enable), the `AGENTS_SHIPGATE_AGENT_MODE=1` override lived only deep in AGENTS.md §errors, `llms.txt` never mentioned it, and `docs/agents/` listed exactly codex/claude-code/cursor. An agent outside the big three that fetched llms.txt had no path to structured errors or a named entry point.

## What

- **`docs/agents/any-coding-agent.md`** (new): the harness-agnostic control guide — force agent mode, the three prominent commands with the agent-handoff-first read order, exit-code table, the MCP-server path for shell-less harnesses, forbidden shortcuts, and the redacted `feedback export` loop with the `agent_feedback.yml` template link. Generic profile documented as `--agent codex`.
- **`docs/agents/README.md` + `llms.txt`**: explicit "any other agent" routing (llms.txt gains its first mention of the agent-mode env var).
- **`AGENTS.md`**: header names the wider harness set; a two-line router atop *Run (canonical)* sends in-flight capability changes to the control loop first (setup only when no `shipgate.yaml`) — closing the last first-command divergence from README's check-first ordering (the read-order half was #251).
- **`llms-full.txt`** regenerated (AGENTS.md is a build source).

## Discipline

Docs-only — no new CLI surface, schema, adapter, or check; deliberately did **not** guess env vars for auto-detecting other harnesses (that would be invented facts; the explicit override is the honest path until verified detection signals exist).

## Verification

`test_public_surface_contract` (120), `test_trigger_command`, `test_packaging` green; ruff clean; `shipgate check --agent claude-code` on this diff (touches the `agent_instructions` trust-root surface): `decision: allow`, `completion_allowed: true`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #251, opened 2026-07-02
Align read-first prose with the contract's agent_read_order

pengfei-threemoonslab:
## Why

Problem #9 from the 2026-07-01 repo review: README shipped **both orderings simultaneously** — "Read `verifier.json` first" (Verify-first quickstart, §189–195; copy-into-your-agent block, §280–285; and the "What it produces" list claimed *in read order* with `verifier.json` ahead of the handoff) vs "read `agent-handoff.json.gate.merge_verdict` first" (§How to read your first result). The `/shipgate` slash command carried the same stale sentence.

The canonical order is not ambiguous: `contract --json` → `agent_read_order[0]` is `agent-handoff.json`, and [`docs/agent-contract-current.md` § Two read entry points](docs/agent-contract-current.md) spells out the design — handoff first for the compact continue/repair/stop step, `verifier.json` as the authoritative controller substrate, `report.json.release_decision.decision` as the gate. `llms.txt`, `AGENTS.md`, and `.well-known` were already aligned; README and the slash command were the stragglers.

## What

- **README ×3**: quickstart read-order sentence, the copy-into-your-agent block, and the "What it produces" list (handoff bullet now first and marked as the first read; verifier bullet reframed from "primary PR/controller evidence artifact" to "authoritative PR/controller evidence substrate", matching the contract doc's language). No field or artifact semantics change.
- **`.claude/commands/shipgate.md`**: same alignment (this file ships in the wheel byte-identical via force-include, so no second copy to sync — covered by `test_wheel_claude_command_is_byte_identical_to_repo_file`).
- **New guard** in `tests/test_public_surface_contract.py`: `test_read_first_instructions_match_contract_agent_read_order` pins every ``Read `<artifact>` first`` instruction on README.md / AGENTS.md / llms.txt / the slash command to the **runtime** contract's `agent_read_order[0]` (via `build_contract_payload()`, the function behind `contract --json`), with a vacuity check (each surface must still contain at least one read-first instruction).

## Scoping notes

- The review item also asked for "CI validation of `.well-known` + `llms.txt` against `contract --json`" — investigation showed this **largely already exists** in `test_public_surface_contract.py` (`.well-known` is validated against `build_contract_payload()` for `agent_read_order`, `verifier_read_order`, `primary_commands`, schema versions, exit codes, and MCP tools; `llms.txt` for version literals, triggers, and llms-full references). The genuine uncovered drift was the prose read-first instructions — that's what the new test covers. Filed no redundant checks.
[...]

PR #257, opened 2026-07-07
Add the Claude Code plugin marketplace; fix silently-broken skill frontmatter

pengfei-threemoonslab:
Roadmap item 3 — the largest untapped discovery channel, and the symmetric counterpart of the existing Codex marketplace.

## What

```text
/plugin marketplace add ThreeMoonsLab/agents-shipgate
/plugin install agents-shipgate@agents-shipgate
```

- `.claude-plugin/marketplace.json` + `plugins/claude-code/` (plugin.json, the `/shipgate` command → installs namespaced as `/agents-shipgate:shipgate`, and the auto-triggering skill).
- **Skill-only by design**, matching the Codex plugin's test-pinned shape: workflows, not the scanner; no hooks (`install-hooks` stays the explicit CLI path where the binary must exist), no MCP config. Test enforces this shape.
- `tests/test_claude_code_plugin_package.py` mirrors the Codex package contract: installable manifest, `version == __version__` on both manifests (release bumps can't strand the marketplace), byte-identity of skill+command with their canonical sources, and frontmatter-parses assertions.
- Docs: README Claude Code section, `use-with-claude-code.md` marketplace subsection (incl. team pre-provisioning via `extraKnownMarketplaces`/`enabledPlugins`), CHANGELOG.

## The bug found in passing (worth the PR by itself)

`claude plugin validate` caught that the canonical `SKILL.md` **and** the `/shipgate` command ship YAML frontmatter whose `description` contains an unquoted `:` — the YAML fails to parse and Claude Code loads the file with **silently-empty metadata**. Concretely: description-based skill auto-triggering has been broken for every existing `init --claude-code` install; the skill only fired when named explicitly. Both descriptions are now quoted, all byte-identical copies synced per the kit-change protocol (render hash updated, outgoing hash appended to `prior_render_sha256`), and the new test parses the frontmatter so this cannot regress.

## Verification

- `claude plugin validate .` and `claude plugin validate ./plugins/claude-code` both pass (also surfaced that the validator rejects `displayName` — removed; docs and validator disagree there).
- Full CI-style suite + static-only lint + ruff green.
- `shipgate check` on this diff: `warn` routing to verify (agent-instructions surface — correct boundary behavior); self-`verify`: `review_required` via `SHIP-VERIFY-TRUST-ROOT-TOUCHED` on the skill/command edits — the human review it demands is this PR review.
- Codex plugin adapter confirmed blind to `.claude-plugin/` (no cross-parsing).

Merge-order note: independent of #253–#256; only CHANGELOG's Unreleased section may trivially conflict.


pengfei-threemoonslab:
Rebased onto post-#253–#256 main (one trivial CHANGELOG conflict; both Unreleased entries kept, full suite + both `claude plugin validate` levels re-green on the rebased state).

Diagnostic note for the record: this PR initially showed **zero CI runs** across push/reopen/synchronize events. Root cause was not event-delivery failure — the PR went `CONFLICTING` the moment #253–#256 merged, and GitHub silently skips `pull_request` workflow runs when it cannot construct the merge ref. The rebase fixed mergeability and CI triggered immediately. (Distinct from the 2026-07-02 incident on #251, which was a genuine dropped push event.)

pengfei-threemoonslab:
P1 fixed — and thank you for actually installing the plugin to test it; that's exactly the blind spot a repo-side review misses. The recipe lookup now starts at `${CLAUDE_PLUGIN_ROOT}/skills/agents-shipgate/prompts/<recipe>.md`, with both install shapes documented inline: running from the installed plugin, Claude Code expands the variable to the plugin cache and finds the version-matched bundled recipes; as a committed project file the unexpanded literal simply doesn't exist and the lookup continues down the original list. The raw-GitHub fallback is now explicitly labeled *last-resort unpinned; prefer any bundled copy above*.

Kept as **one canonical text** (`.claude/commands/shipgate.md`) synced to the plugin copy, so the byte-identity test at `tests/test_claude_code_plugin_package.py` stands unchanged — no divergent plugin-aware fork to maintain. Package/renderer/surface tests + both `claude plugin validate` levels green.

Issue #293, opened 2026-07-27
Host-authenticated in-session approval receipts (unblock user-approved high-risk host-grant edits)


pengfei-threemoonslab:
## Symptom

A user asks their agent to add an MCP server to `.mcp.json`. The host prompts, the user allows it, the edit lands — and `shipgate check` returns `human_review_required` with `must_stop: true`. The user *is* the human, and they already said yes. There is no sanctioned way to record that answer, so the turn ends by telling the person who just approved the change that a human must approve the change.

The PreToolUse hook even documents its `ask` as carrying "the same authority semantics as `merge_verdict: human_review_required`" — and then persists nothing.

## Scope after contract v19

v19's graded band removed this for low/medium rows, and deliberately did **not** for the rows that matter here: `HOST-MCP-SERVER-ADDED` is `high`, `HOST-PERMISSION-ALLOW-EXPANDED` is `high`, gate-governing trust roots are excluded at any risk. So the stuck state is now narrow and well-defined: user-approved high-risk host-grant edits.

## Why the obvious version is wrong

An unsigned local receipt was designed, reviewed, and **rejected** in July 2026. Recorded in the design doc so it is not re-proposed by accident:

1. Agent-forgeable — the "is an agent running me" gate is an env heuristic, so a receipt that clears a human route is self-approval with extra steps.
2. Preflight evaluates the wrong tense — receipts bind landed bytes; preflight routes proposed edits.
3. It cannot reach CI and must not reach reports — the ledger lives in one developer's `.git`, and annotating `report.json` would make the same commit produce different artifacts on different machines, breaking content-addressed reproducibility.

## Proposed direction

Design doc: [`docs/engineering/host-authenticated-approval-receipts.md`](https://github.com/ThreeMoonsLab/agents-shipgate/blob/main/docs/engineering/host-authenticated-approval-receipts.md)

A receipt is a statement **by the host** that a specific human answered a specific prompt about a specific proposed edit:

- bound to `tool_use_id` (exposed by both PreToolUse and PostToolUse), `session_id`, the host's decision value, and both the proposed and post-edit content hashes;
- authentic only when host-attested — signed like the v18 `human_authorization` overlay, or written to a host-owned location the agent's tool surface cannot reach. Absent attestation it is annotation-only;
[...]

Issue #292, opened 2026-07-27
insufficient_evidence attaches a human-review notice to every turn in first-adoption repos


pengfei-threemoonslab:
## Symptom

On a freshly adopted repo (framework detected, small tool surface), *every* turn that runs the local loop ends with "A human must review this change before it can merge" — including a docs-only or README-only edit. Reproduced 2026-07-27 on a fresh fixture built from `main` at contract v19.

Two mechanics combine:

1. `verify` passes `user_requested=True` (`cli/verify/orchestrator.py:155`), so the capability scan runs on every invocation regardless of what the diff touched — the adopted-repo force-run contract.
2. A repo whose tool surface is not statically enumerable trips `evidence_below_ie_threshold`, returning `insufficient_evidence` → `merge_verdict: human_review_required`.

The abstention is correct as a *release* judgement. Attached to a docs-only turn it reads as a non sequitur, and because `insufficient_evidence` is immune to baselines, suppressions, severity overrides, and `human_ack` by explicit contract — with every generated remedy `actor: human` — there is no action the agent can take.

## Why now

Contract v19 (#291) graded the local boundary stop, so benign edits no longer hard-stop at `check`. That moved this to the top of the remaining coding-agent UX friction: it is now the main reason a routine turn still ends with a review notice.

## Aggravating detail (disproved — see Revision)

~~`init` detects the framework with high confidence and writes a manifest that does **not** declare `agent_bindings` — so the routed onboarding path manufactures the very gap `verify` then abstains on.~~

Measured 2026-07-28: a repository with one literal `Agent(...)` resolves its root through the single-top-level-agent fallback and reports `binding_coverage.gap_count == 0` **without** any `agent_bindings` key. `init` does not manufacture the gap.

## Revision (2026-07-28, after PR #296)

Measuring the two real cold-start shapes retired most of the original direction, and found the actual root cause elsewhere. Kept open with the criteria below.

**The root cause was discoverability, not the verdict.** The decision engine already generated the exact manifest snippet each gap wants (`EvidenceGapAction.declaration_template`), but the only way to reach one was to walk `report.json` down to `release_decision.evidence_coverage.evidence_gaps[].next_action.declaration_template`. A three-line, one-time task looked like schema archaeology, so repositories stayed abstained indefinitely.

[...]

PR #296, opened 2026-07-28
Give insufficient_evidence a way out

pengfei-threemoonslab:
Refs #292 — **does not close it.** See "Scope" below.

## Summary

`insufficient_evidence` was correct and unactionable at the same time. The decision engine already generated the exact manifest snippet each evidence gap wants — but the only way to reach one was to walk `report.json` down to `release_decision.evidence_coverage.evidence_gaps[].next_action.declaration_template`, so a three-line one-time task looked like schema archaeology and repositories stayed abstained indefinitely. **That, not the verdict, is what made every turn end with an unhelpful human-review notice.**

- **`suggested-declarations.yaml`** assembles those templates next to the report, the way `suggested-inventory.json` already does for low-confidence sources. Blocks are merged per target — two gaps on one tool want *one* `action_surface.actions` row, and emitting two would hand the reviewer something invalid to paste. Every gap carrying a template names the file in its `expects`. Human-owned values stay `<REVIEW_REQUIRED>`, and the file states that a block still containing a sentinel closes nothing.
- **A defect the templates were hiding.** The authority template offered `mode` alone, but the manifest requires `auth_type` for every mode except `none` and non-empty `scopes` for `scoped` — a reviewer who filled it in exactly as written got a config error. Nobody had hit it because nobody could find it. It now names the co-required fields, and a regression test validates the shipped shape against the manifest schema.
- **Verify says whether the diff caused the gaps.** It already scans the base, so it compares the base and head gap sets: "introduces no new evidence gap; all N are pre-existing" versus "M of N are new in this diff". A docs-only turn stops reading as an accusation about the current change.
- **Framework-correct remedy.** The low-confidence advice named `tool_inventories` for every framework, but only four have that key — `openai_agents_sdk`, the quickstart framework and the most common cold-start case, has none, so readers were sent after a key the schema rejects.

## Scope: this does not satisfy #292's acceptance criteria

The issue asked for two things: a docs-only turn that ends with no human-review notice, and an agent-executable remediation. **Neither ships here, deliberately.** The notice follows the verdict, and the verdict stays for the circularity reason below. The remediation is human-owned by the `do_not_auto_assert` contract — an agent proposing a tool's effect or authority *is* the invented evidence that contract forbids — so no agent-executable path exists to build.

What this PR does instead is remove the **dead end**: the one-time human act is now concrete and cheap rather than schema archaeology, and the notice stops implicating the current change. That is a real improvement on the reported symptom but not the stated criteria, so the issue should stay open with its criteria revised (or split: the residual belongs with the host-authenticated approval work in #293).

## The verdict is deliberately unchanged

The issue proposed scoping the abstention to the evaluated diff. I rejected it rather than deferring it, and the design doc now records why: `insufficient_evidence` means *the static view of the agent surface is unreliable*, so a rule of the form "the diff touched no capability surface, therefore do not abstain" reasons from that same view and is circular. The failure is concrete, not theoretical — a config-bound or dynamically constructed toolkit (the documented design-partner finding) presents as a diff touching nothing while changing what the agent can do, so precisely the case the verdict exists for would escape it. The delta variant ("no new gaps versus base") fails identically, because an unseeable addition yields an identical gap set.

Two premises in the design doc were also wrong and are corrected in place: omitting `agent_bindings` does **not** manufacture the binding gap (a repo with one `Agent(...)` resolves its root and reports `binding_coverage.gap_count == 0` without the key), and the proposed `init` scaffold would have *caused* abstentions rather than prevented them, because each `CHANGE_ME` line becomes a source warning and four warnings trips the threshold on its own. No `init` scaffold shipped.

## Type

- [ ] Input adapter change
- [ ] Documentation only

## Verification

Additional local checks run:

- Full `pytest` suite green (exit 0); `ruff check .` clean.
- Measured both real cold-start shapes on fresh fixtures: a repo with a literal `Agent(...)` abstains on semantic gaps (effect / authority / surface); a pure-decorator repo abstains on a single `ambiguous_root_agent` binding gap. Both now get a scaffold.
- **End-to-end close-the-loop:** filled the scaffold's effect and authority blocks, merged them, re-scanned — `inferred_effect_only` and `missing_authority_evidence` cleared and the verdict moved off abstention to `blocked`, correctly, because the declared `financial_write` tool has no approval policy. A substantive verdict instead of "the scan cannot tell."
- Provenance note checked in both directions: identical base/head gap sets → "introduces no new evidence gap; all 4 are pre-existing"; wiring a new reachable tool → "4 of 8 evidence gap(s) are new in this diff". Verdict `insufficient_evidence` in both.
- Confirmed the pre-fix authority template really did fail: pasting `authority: {mode: scoped}` produced `action_surface.actions[].authority.mode 'scoped' requires auth_type`.
- New `tests/test_declaration_scaffold.py`: merge-per-target, distinct tools stay separate, nothing-owed returns `None`, sentinel present with the "closes nothing" warning, schema fillability, and provenance including the unreadable-base and no-base fallbacks.

## Release-readiness notes

- [x] New or changed check IDs are documented in `docs/checks.md` (none added or removed)
- [x] Report/schema changes are additive or documented in `STABILITY.md` — no schema or contract version change. `release_decision` branching, the IE threshold constants, and the precedence ladder are untouched; `declaration_template` content changed within the existing field, and the new artifact is advisory like `suggested-inventory.json`.


src/agents_shipgate/ci/release_decision.py
```diff
@@ -433,22 +456,9 @@ def _binding_coverage(
...
-                                    "agent": "root",
-                                    "complete": True,
-                                    "tools": [],
-                                    "handoffs": [],
-                                    "reason": "<REVIEW_REQUIRED>",
-                                }
-                            ],
-                        }
-                    },
+                    declaration_template=deepcopy(AGENT_BINDINGS_ROOT_TEMPLATE),
```
pengfei-threemoonslab: **[P1] Emit a template that matches the binding repair.** This root-only template is attached unconditionally. For `missing_binding_evidence` and `conflicting_binding_evidence`, the action/path says to edit `agent_bindings.declarations`, but the generated document contains only `agent_bindings.root`. In a decorator-only repo the agent-node list is empty, so no filled root selector can match and `ambiguous_root_agent` remains. Restrict this to resolvable root-selection cases or emit/omit an issue-specific safe template.

src/agents_shipgate/cli/scan/declarations.py
```diff
@@ -0,0 +1,112 @@
...
+    for gap in gaps:
+        action = gap.next_action
+        template = getattr(action, "declaration_template", None)
+        if not isinstance(template, dict) or not template:
+            continue
+        path = str(getattr(action, "path", "") or "shipgate.yaml")
+        # Key on the gap's own subject, not just the tool name: two sources can
+        # expose the same tool name, and merging those would fold two distinct
+        # declarations into one row.
+        target = (path, str(gap.subject or ""), str(template.get("tool") or ""))
```
pengfei-threemoonslab: **[P1] Preserve canonical identity in the rendered selector.** Keying on `subject` separates some gaps, but the templates still serialize only the display name. Two canonical tools named `lookup` therefore produce identical filled action selectors: merging both fails with `Duplicate action_surface.actions[] tool selectors`, while merging one resolves neither uniquely. Carry a stable `tool_id` or source-qualified selector in the template and grouping key.

src/agents_shipgate/ci/release_decision.py
```diff
@@ -738,9 +749,24 @@ def _semantic_gap(
...
+        # and ``ambient`` need ``reason`` and empty ``scopes``. Naming all of
+        # them keeps the template fillable for every supported answer, and the
+        # scaffold tells the reviewer to delete what their mode does not take.
         declaration_template = {
             "tool": tool.name,
-            "authority": {"mode": "<REVIEW_REQUIRED>"},
+            "scopes": [REVIEW_REQUIRED_SENTINEL],
+            "authority": {
+                "mode": REVIEW_REQUIRED_SENTINEL,
+                "auth_type": REVIEW_REQUIRED_SENTINEL,
```
pengfei-threemoonslab, line 778: **[P1] Do not allow the sentinel to become reviewed evidence.** The manifest only checks that `auth_type` is nonblank, so a scoped declaration with `auth_type: <REVIEW_REQUIRED>` is assessed as `declared` with no authority issue. `reason` sentinels are also accepted where optional, contradicting the scaffold's promise that any remaining sentinel closes nothing. Reject sentinel values before semantic assessment or render mode-specific alternatives. Also note that `ambient` requires a reason and empty scopes, not `auth_type`.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -961,6 +963,84 @@ def _self_approval_note(capability_review: VerifierCapabilityReview | None) -> s
...
+    capability change looks like, so the diff can never argue the abstention
+    away. What it can do is stop misattributing it.
+    """
+
+    if report is None or report.release_decision is None:
+        return None
+    coverage = report.release_decision.evidence_coverage
+    if coverage is None or not coverage.evidence_gaps:
+        return None
+    head = Counter(
```
pengfei-threemoonslab, line 1022: **[P2] Use a stable gap identity for provenance.** Raw `(kind, subject)` is not stable: source-warning subjects can contain generated archive roots, so an unchanged warning gets different base/head identities. A docs-only reproduction falsely reported `1 of 2 evidence gap(s) are new`. Normalize transient paths or introduce a structured stable gap fingerprint before multiset subtraction.

pengfei-threemoonslab (commented):
## Review outcome: changes recommended before merge

Second-pass review on `cda6d222`.

The update resolves the earlier CI failures and removes the unsafe prefilled empty-binding assertion. Hosted checks, Ruff, compile checks, packaging tests, the blocked-refund fixture, and Shipgate verification/receipt reproduction are all clean.

I still recommend changes before merge because the advertised remediation contract has three high-priority correctness gaps: binding issues receive a root-only template that cannot repair several gap kinds, same-name tools produce unusable duplicate selectors, and a `<REVIEW_REQUIRED>` authority value can be accepted as reviewed evidence. Additional inline comments cover unstable gap attribution, the remaining selector-path mismatch, stale-file cleanup, and mixed-gap wording.

Please add scan → fill scaffold → rescan coverage for decorator-only binding gaps, missing/conflicting binding evidence, same-name canonical tools, sentinel rejection, and mixed template/non-template gap sets.

_Submitted as a comment review because GitHub prevents a PR author from requesting changes on their own pull request._

src/agents_shipgate/ci/release_decision.py
```diff
@@ -654,11 +664,12 @@ def _semantic_gap(
         accepted_values = ["tool_id", "provider", "source_type", "source_id"]
         action_why = "A one-to-one selector must resolve before policy or evidence can apply."
         expects = "Add tool_id or source/provider qualifiers, then rerun verification."
-        declaration_template = {
-            "tool": tool.name,
-            "tool_id": tool.id,
-            "provider": tool.provider,
-        }
+        # No template: the action points at ``tool_identity``, whose schema
```
pengfei-threemoonslab, line 678: **[P2] Route selector repair to the owning declaration.** Removing the invalid flat template is correct, but `next_action.path` still routes these cases to `shipgate.yaml#tool_identity`. For an ambiguous `action_surface.actions` selector, the qualifier belongs on that action row; `tool_identity` accepts reviewed bindings rather than selector qualifiers. Preserve the issue source pointer and route the human to the actual selector.

src/agents_shipgate/cli/scan/writing.py
```diff
@@ -72,6 +80,37 @@ def _write_outputs(
...
+    stays ``<REVIEW_REQUIRED>``. Consumes the sanitized public report so no
+    redacted content reaches the scaffold.
+    """
+
+    anchor = generated_paths.get("json") or next(iter(generated_paths.values()), None)
+    if anchor is None:
+        return
+    out_path = anchor.parent / SUGGESTED_DECLARATIONS_FILENAME
+    scaffold = scaffold_for_report(report)
+    if scaffold is None:
```
pengfei-threemoonslab, line 102: **[P2] Clean this artifact on verifier reset paths too.** This handles a later successful scan with no scaffold, but early verifier failures call `_remove_scan_artifacts`, whose list omits `suggested-declarations.yaml`. A missing-manifest or missing-ref run can therefore remove `report.json` while leaving stale remediation instructions behind.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -961,6 +963,84 @@ def _self_approval_note(capability_review: VerifierCapabilityReview | None) -> s
...
+    # Multiset, not set: two gaps sharing a (kind, subject) are two gaps, and
+    # collapsing them would report a genuinely new one as inherited.
+    introduced = sum((head - base).values())
+    total = sum(head.values())
+    if introduced:
+        return f"{introduced} of {total} evidence gap(s) are new in this diff."
+    # Only name the scaffold when one will exist. Low-confidence and
+    # source-warning gaps carry no declaration template, so a repository whose
+    # gaps are all of that kind gets no file and must not be sent to one.
+    scaffolded = any(
```
pengfei-threemoonslab: **[P2] Do not claim one declaration closes a mixed gap set.** `any(declaration_template)` produces “a one-time human declaration closes them” even when other gaps are low-confidence tools or source warnings with no declaration remedy. Describe only the scaffolded subset, ideally with its count, rather than all gaps.

Issue #295, opened 2026-07-27
One verify per turn: reuse a verification the agent already ran, without trusting workspace artifacts


pengfei-threemoonslab:
## Context

`check` routes the agent to run `agents-shipgate verify`; the Stop hook then runs an identical verify again at turn end. Measured on a small fixture that is ~1.2s of pure duplicate work per turn (0.9s vs 2.1s), and it grows with repo size.

A first attempt shipped in #294 and was **reverted in the same PR** after review found two defects. Recording both so the next attempt does not repeat them:

1. **It trusted a workspace-writable artifact.** The hook read `agents-shipgate-reports/verifier.json`, which the agent can write. A forged (or interrupted) artifact with `control.state=complete` skipped verification entirely, and an `agent_action_required` artifact could inject an arbitrary "exact" command into the hook's block reason.
2. **The identity binding was insufficient.** Comparing the recorded `verification-input.diff` to the hook's own snapshot diff is not a binding to the current verification *request*: committing an uncommitted change reproduces the same diff text, which was used to reproduce an outcome-changing stale pass — an old `passed` result reused while a fresh verify returned `review_required` for `permissions.scopes: ["*"]`.

Also noted: on a feature branch with worktree edits, the hook hashes only the worktree diff while the verifier evaluates `base...HEAD` plus that diff, so the optimization would silently miss anyway.

## What a correct design needs

- Binding to the verifier's own content-addressed request identity (`request_id` / `input_set_id` / the terminal receipt), not to a re-derived diff. A hook script cannot compute those, so the producer has to hand the consumer something checkable.
- A trust story for the handoff. Any path the agent can write is not evidence; the plausible shapes are the verify CLI recording a consumable marker outside the workspace, or the hook invoking a cheap "is this result current?" CLI subcommand that re-derives identity in the engine.
- Correct behavior on a feature branch, where the hook's snapshot and the verifier's evaluated range differ.

## Non-negotiable

No design may let a reused result be more trusted than a fresh one. The Stop hook is advisory, but "advisory" must not mean "skippable by writing a file".

pengfei-threemoonslab:
Attempted in #297 and withdrawn. Recording what two review rounds established, as constraints any future attempt has to meet.

**Why it was withdrawn.** The Stop hook is a standalone rendered script with no Shipgate imports, so it cannot compute `input_set_id` — the canonical "what was actually verified". Every cheap thing it *can* compute is an approximation, and each approximation leaked:

[...]

PR #297, opened 2026-07-28
Adoption UX: honest first-adoption verdict, custom-manifest protection, agent detection

pengfei-threemoonslab:
Three debts from the coding-agent adoption reviews, in one branch, one commit each.

## 1. A first adoption is not a policy weakening

Adding the manifest to a repository that had none is the first verdict every new adopter sees. It said:

> This PR weakens the release policy that evaluates it; a coding agent cannot self-approve that change — a human must review it.

carried a finding titled *"Policy change cannot be proven safe (no base snapshot)"*, and — because a missing-manifest base counted as a `safe_recovery` — shipped **no `fix_task` at all**. The run said stop and named nothing to do.

Adoption is now proved from git rather than inferred from the diff, and proved the hard way: the comparison base must carry no manifest *under any name*. A PR that moves the manifest while loosening it also finds nothing at the configured path and would otherwise get to call itself a first adoption — reproduced, and now closed (`test_moved_manifest_cannot_pass_itself_off_as_an_adoption`, plus a live check against a rename-and-loosen diff, which still gets the "cannot be proven safe" copy).

Unchanged: check id, `medium` severity, `human_review_required`, `can_merge_without_human=false`. Only the claim about what happened changed, plus a `fix_task` that now names the one act that clears it. `check` gets the same correction locally, keyed on the diff carrying exactly one manifest record and that record being a plain addition — a composite diff (new manifest *plus* an edit to a live one) keeps the generic wording, per #282's rule that a block-level "safe" signal must never soften a path-wide fail-closed guard.

## 2. Agent detection and the agent-mode error promise

`check --agent` defaulted to `codex` and never read the harness variables Shipgate already consults to switch agent mode on, so every Claude Code and Cursor run labelled its result and audit id `codex`. One table now defines both the hints and the actor they identify.

The skills and slash command promise that with `AGENTS_SHIPGATE_AGENT_MODE=1` a failing command emits a structured `next_action` line on stderr. `check`, `audit`, and `preflight` printed prose only. Each error path now emits the line with its `exit_code`, and `preflight` no longer lets an unexpected failure escape as a bare traceback.

## 3. Withdrawn: one verify per turn (#295)

Attempted, reviewed twice, and **removed from this PR**. Its soundness rests on `input_set_id` being the complete input set, and it is not — adapter inputs such as `openai_api.prompt_files` never enter it, so an ignored one can change with neither the plan identity nor a worktree digest moving. The Stop hook is a standalone script that cannot compute the real identity, so every cheap approximation leaks somewhere. It was a latency optimization with no correctness benefit and it produced six review findings.

[#295](https://github.com/ThreeMoonsLab/agents-shipgate/issues/295) stays open with every constraint recorded, and the incomplete input set is filed as [#299](https://github.com/ThreeMoonsLab/agents-shipgate/issues/299) because it also underpins receipts and attestations.

## 3. Correctness fixes from review

- **The manifest a run actually loaded is a trust root.** The table only knew `**/shipgate.yaml`, so `--config new-gate.yml` had no manifest trust root at all — its gate could be introduced or rewritten with an empty release substrate, and a clean scan then produced `passed`/`mergeable`/`complete` beneath an adoption headline saying a human was required. One shared predicate now serves the trust-root check, the policy fail-safe, and the adoption wording.
- **`policy_weakened` is derived from the evidence kind**, so an adoption stops being recorded as a weakening by the registry, attestations, feedback's gate-bypass alarm, and reviewer routing.
- **Rerun commands carry the whole evaluated request** — config, base (or an explicit `--no-base`), policy packs, baseline, ci-mode, fail-on, plugin and heuristic flags — because a rerun that drops them evaluates a different question.
- **Preflight recovery reproduces the invocation that failed** instead of a bare `preflight --json`, which discarded the target and answered a question nobody asked.
- **Agent-mode errors follow `docs/errors.json`**: both `next_action` and `next_actions[]`, only published kinds, and filesystem failures as `other_error`/4.
- Adoption wording requires every touched policy surface to be the manifest; `ls-tree` failure is "cannot prove"; git rename/delete detection catches `old-gate.yml → new-gate.yml`.

## Verification

- Full suite green (`ruff` and `pytest` run unpiped); `fixture run ai_generated_refund_pr` still `blocked`.
- End to end on scratch repos: first adoption via `init --write` → `verify` (both the local worktree path and the CI `--base`/`--head` path) produces the introduction headline, `manifest_introduced` evidence kind, and a populated `fix_task`; after committing, the ordinary wording returns.
- `CLAUDECODE=1` / `CURSOR_TRACE_ID` / bare shell each produce the right actor; explicit `--agent` still wins.
- A custom-named manifest (`--config new-gate.yml`) is now protected in `verify`, `check`, and `preflight`, and stays out of the graded agent route; a base that keeps its own manifest under another name no longer reads as a first adoption.
- Recovery commands were replayed by hand: the `check` correction preserves actor/workspace/config/policy, the verify rerun runs clean from another directory, and stdin-backed or conflicting requests offer a review action instead of an unreplayable command.

## Found in passing, not fixed here

`ci_mode_weakened` cannot fire through the real `--base` path. `_prepare_base_report` scans the base with `ci_mode="advisory"`, `prepare.py` applies that by mutating the manifest, and `final_report.py` builds `effective_policy` from the mutated manifest — so every cached base snapshot records `advisory` and a strict→advisory downgrade compares equal. Existing coverage misses it because the tests inject the base `EffectivePolicy` directly. Filed as https://github.com/ThreeMoonsLab/agents-shipgate/issues/298; such a PR still routes to human review via `SHIP-VERIFY-TRUST-ROOT-TOUCHED`, so it is a missed specific finding, not a silent merge.


pengfei-threemoonslab (commented):
Re-review of `187757267903174cafce8d2950c9a23fa256cdbd` after the latest update.

The reuse/TOCTOU findings from the prior round are resolved: the optimization was fully removed. The rerun command also now preserves config, base/`--no-base`, policy packs, baseline, CI/fail policy, plugin flags, patch/heuristic flags, and authorization, and the structured adoption repair targets the configured manifest.

I found 9 remaining actionable issues: 4 P1 and 5 P2. I recommend addressing these before merge. Validation performed: Ruff passed; source verification and receipt reproduction passed; live GitHub checks are green; the full local suite had only sandboxed isolated-build dependency errors, and `tests/test_packaging.py` passed 8/8 when rerun with network access.

PR metadata cleanup: the title and the final Verification bullet still advertise the removed “one verify per turn” behavior.

src/agents_shipgate/checks/verify.py
```diff
@@ -70,6 +71,19 @@ def _classify(path: str) -> tuple[str, str] | None:
...
+def _configured_manifest(context: ScanContext, path: str) -> tuple[str, str] | None:
+    """Classify the manifest this run was actually pointed at.
+
+    Whatever a run loaded as its gate is a manifest trust root, even when it is
+    not called ``shipgate.yaml``.
+    """
+
+    config = getattr(context, "config_path", None)
+    if not is_configured_manifest(config, path):
+        return None
```
pengfei-threemoonslab, line 83: [P1] Keep custom-manifest evidence deterministic.

For a committed-head run, `context.config_path` points inside a freshly named `agents-shipgate-verify-head-*` archive. Returning it as `matched_glob` persists that random path in finding evidence, and finding identity hashes this evidence. Two identical runs had the same request ID but different fingerprints and report run IDs. Please emit a stable logical/workspace-relative manifest path instead; the normalized changed `path` is sufficient.

src/agents_shipgate/core/agent_boundary.py
```diff
@@ -735,11 +741,42 @@ def _boundary_summary(decision: str, violations: list[AgentResultViolatedRule])
...
+
+    Only the wording moves: the rule id, action, and risk level are unchanged,
+    so the local decision and the control state are identical either way.
+    """
+
+    records = [
+        item
+        for item in diff_files
+        for candidate in (item.new_path, item.old_path)
+        if candidate and trust_root_class_for(candidate.replace("\\", "/")) == "manifest"
```
pengfei-threemoonslab: [P1] Protect the configured manifest in `check` and `preflight`, not only in verify checks.

The configured-manifest predicate is not threaded into these earlier control surfaces. A diff for `new-gate.yml` loaded via `--config new-gate.yml` produced `check` decision `allow` with no violations, while preflight reported no protected touch and `requires_human_review=false`; nested configured manifests also miss the local boundary filter. Both routes then emit verify commands for the default `shipgate.yaml`. Please pass the resolved configured-manifest identity through boundary/preflight classification and exact command construction.

src/agents_shipgate/cli/verify/fix_task.py
```diff
@@ -496,13 +567,47 @@ def _dedupe_cap(items: list[str]) -> list[str]:
...
+    - ``options`` carries the rest of the evaluated request — policy packs, a
+      baseline, an explicit ``--no-base``, plugin and heuristic modes. Omitting
+      them produced a command that ran a *different* evaluation than the one
+      whose findings it was supposed to reproduce.
+
+    Refs and paths come from CLI / GitHub inputs and a valid git ref may
+    contain shell metacharacters, so every interpolated value is quoted.
+    """
+
+    parts = ["agents-shipgate", "verify"]
```
pengfei-threemoonslab, line 811: [P1] Preserve the evaluated workspace in the exact rerun command.

This builder still has no workspace or output-directory input. When verification is invoked from outside its target repository, the emitted command runs in the caller directory, where it can fail or evaluate another checkout; with custom `--out`, it also leaves the requested artifact directory stale. Please include and quote the resolved `--workspace` and selected `--out` (and consider other output-shaping request options such as `--pr-comment-style`).

src/agents_shipgate/cli/check.py
```diff
@@ -66,22 +87,29 @@ def check(
...
+            command="agents-shipgate check --agent codex",
+            expects="An --agent value the boundary evaluator recognizes.",
+        )
     if format_ == "agent-json":
-        typer.echo(
+        raise _flag_error(
             "--format agent-json was removed in the 0.14.0 contract cleanup. "
             "Use --format agent-boundary-json.",
-            err=True,
+            command="agents-shipgate check --format agent-boundary-json",
```
pengfei-threemoonslab: [P1] Keep structured recovery on the same check request.

These fixed commands discard all valid arguments surrounding the bad field. For example, an invalid format on a `cursor` run against another workspace emits only `agents-shipgate check --format agent-boundary-json`, switching actor, workspace, config, policy, and diff/range context. Following the documented rank-1 action can therefore answer a different boundary question. Rebuild it from the original request, changing only the invalid value and shell-quoting arguments; stdin-backed input should not claim a replayable command.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -906,24 +948,159 @@ def _map_optional_tree_path(
...
+    if base_status == "missing_manifest" and base is not None:
+        ref: str = base
+    elif base_status in {"not_requested", "skipped"}:
+        # No base was compared against, so the manifest's own history is the
+        # base: present in the workspace but absent from the head commit means
+        # this working tree is introducing it.
+        ref = head
+    else:
+        return False
+    names = _MANIFEST_FILE_NAMES | {config_relative.name}
```
pengfei-threemoonslab: [P2] Do not prove first adoption by checking only the default and new basenames.

If the base retains an operational custom manifest under another name, no YAML is removed, so this can incorrectly prove adoption. Repro: base retains valid strict `old-gate.yml`; head adds configured advisory `new-gate.yml` without deleting the old file. The result was `manifest_introduced`, `policy_weakened=false`, and a headline claiming the base carries no manifest. Please fail closed unless absence of an existing manifest can actually be established.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -939,15 +1116,38 @@ def _can_merge_without_human(
...
+    a PR that adds the manifest to a base that had none weakens nothing, and
+    saying it does is both wrong and the first thing every new adopter reads.
+    An adoption is its own case and must not depend on ``policy_weakened``:
+    that flag is now false during an adoption, because it answers "was the gate
+    weakened" for the registry, attestations, and the gate-bypass alarm. Firing
+    on ``manifest_introduced`` alone is what keeps every caller that reads this
+    as "a trust root is in play" — ``_can_merge_without_human`` above all —
+    fail-closed regardless of the other two flags.
     """
+    if manifest_introduced:
```
pengfei-threemoonslab: [P2] Do not let adoption wording override mixed existing-policy changes.

`manifest_introduced` unconditionally selects this headline even when `capability_review.policy_weakened` is true. Adding a manifest while editing an existing policy produced a `base_snapshot_unavailable` policy finding, yet the headline said there was no prior gate to weaken, the fix task said nothing existing was weakened, and `review_policy_weakening` was omitted. Pure-adoption wording should require no other policy/trust-root change needing review, or the result should surface both actions.

src/agents_shipgate/cli/check.py
```diff
@@ -66,22 +87,29 @@ def check(
 ) -> None:
     """Run the agent-native local boundary check."""
 
+    # The actor lands in the result and in the audit id, so an undetected
+    # harness mislabels every row it writes. An explicit flag always wins.
+    agent = agent or detect_actor()
```
pengfei-threemoonslab: [P2] Bind the detected actor into the audit identity.

The result’s actor field now changes, but the central audit digest omits actor and the legacy digest still hardcodes `codex`. Identical evaluations for `codex`, `claude-code`, and `cursor` therefore return the same `audit_id`, leaving the audit-attribution issue described by this PR unresolved. Please include the effective actor in the audit identity, or revise the claimed contract if caller-independent IDs are intentional.

src/agents_shipgate/cli/preflight.py
```diff
@@ -72,6 +137,18 @@ def preflight(
 ) -> None:
     """Run the proactive static preflight contract for coding agents."""
 
+    rerun = _rerun_command(
```
pengfei-threemoonslab: [P2] Do not emit a command that repeats an invalid preflight request shape.

The rerun is built before mutually exclusive plan flags are rejected and is reused for every `ConfigError`. With both `--plan CHANGELOG.md` and `--changed-files CHANGELOG.md`, the rank-1 `kind=command` action contains both flags again, so it can never satisfy its own `expects` without manual modification. Emit a review action for ambiguous request-shape conflicts or construct an explicitly normalized valid command.

src/agents_shipgate/cli/host_audit.py
```diff
@@ -144,9 +213,22 @@ def audit(
     if drift:
         try:
             baseline = load_host_grants_baseline(resolved_baseline)
+        except OSError as exc:
```
pengfei-threemoonslab: [P2] This filesystem-error handler is unreachable for baseline reads.

`load_host_grants_baseline()` converts every read `OSError` into `ValueError`, so using a directory as `--baseline-file` returns `config_error`/2 and recommends bare `agents-shipgate audit --host`, dropping drift, workspace, and baseline path and performing an unrelated inventory. Preserve/inspect the underlying I/O cause and route it through the documented filesystem-error path with a real path-repair action.

src/agents_shipgate/cli/host_audit.py
```diff
@@ -145,8 +215,36 @@ def audit(
...
+                    ),
+                ) from exc
+            raise _config_error(
+                str(exc),
+                next_action=(
+                    "Record a baseline with `agents-shipgate audit --host "
+                    "--save-baseline`, or point --baseline-file at a valid one."
+                ),
+                command=(
+                    "agents-shipgate audit --host --workspace "
```
pengfei-threemoonslab: **[P1] Never overwrite a failed baseline.** All non-I/O baseline load failures emit a rank-1 `--save-baseline` command targeting the same path. In a disposable repo, following this recovery for an existing malformed file replaced it with baseline JSON and acknowledged the current grants without human review. It also drops `--scope local-static`. Missing, malformed, or integrity-failed baselines must route to review or a non-destructive correction.

src/agents_shipgate/cli/verify/git.py
```diff
@@ -367,6 +367,100 @@ def read_file_at_ref(workspace: Path, ref: str, path: Path) -> str | None:
...
+    """
+
+    result = _run_git(
+        workspace,
+        [
+            "grep",
+            "--all-match",
+            "-l",
+            "-e",
+            "^project:",
```
pengfei-threemoonslab: **[P1] Parse retained manifests structurally.** This proof recognizes only exact unquoted `project:` and `agent:` lines. A valid retained manifest using quoted YAML keys loads successfully, but the probe returns false and `_manifest_introduced` reports a first adoption with `policy_weakened=false`. Enumerate and parse candidate YAML blobs, failing closed on unreadable results.

src/agents_shipgate/cli/agent_result.py
```diff
@@ -114,6 +115,7 @@ def _assessment_for_diff(
         ),
         input_mode=input_mode,  # type: ignore[arg-type]
         input_issues=input_issues,
+        config_path=config_path,
```
pengfei-threemoonslab, line 248: **[P1] Bind check verification to its target.** The configured manifest now reaches classification, but an ordinary force-run checked with `--config shipgate-self.yaml` still authorizes only `agents-shipgate verify --json`, dropping both workspace and config and evaluating the default manifest. Thread an invocation-specific verify command through the boundary-control projection.

src/agents_shipgate/core/preflight.py
```diff
@@ -128,6 +130,35 @@
...
+    """
+
+    if workspace is None or config is None:
+        return _VERIFY_COMMAND
+    return " ".join(
+        [
+            "agents-shipgate",
+            "verify",
+            "--workspace",
+            shlex.quote(workspace.as_posix()),
```
pengfei-threemoonslab: **[P1] Nested preflight can verify the wrong gate.** Preflight interprets config relative to its requested workspace, while verify first replaces that workspace with the Git root and then resolves config. With distinct root and `services/api` manifests, this emitted command succeeded but generated a report for the root gate. Render config relative to the Git root or make verify honor the requested workspace.

src/agents_shipgate/core/trust_roots.py
```diff
@@ -107,6 +107,33 @@ def trust_root_class_for(path: str) -> str | None:
...
+    path — while refusing a same-basename file in another directory.
+    """
+
+    if config_path is None:
+        return False
+    configured = str(config_path).replace("\\", "/").strip()
+    candidate = path.replace("\\", "/").strip()
+    if not configured or not candidate:
+        return False
+    return (
```
pengfei-threemoonslab: **[P1] Canonicalize configured-manifest identity.** Raw suffix comparison is bypassable by equivalent path spellings. Configuring `docs/engineering/../manifest-v0.1.example.full.yaml` and changing the normalized path returned `allow` with zero violations; the same logic can false-positive unrelated suffixes. Compare normalized, containment-checked workspace-relative identities.

src/agents_shipgate/core/preflight.py
```diff
@@ -265,7 +296,8 @@ def build_preflight_result(
         )
         for node in graph.nodes
     ]
-    touches = classify_protected_touches(changed)
+    verify_command = _verify_command(workspace, config)
+    touches = classify_protected_touches(changed, config_path)
```
pengfei-threemoonslab: **[P1] Include rename source paths.** The diff-derived `changed` set keeps only each destination path. Renaming configured `shipgate-self.yaml` to an unrecognized name produced no protected touch, `requires_human_review=false`, and an agent-action route. Include both `old_path` and `new_path`, deduplicated, before classification.

src/agents_shipgate/cli/check.py
```diff
@@ -15,13 +20,73 @@
...
+    if diff == "-":
+        return None
+    parts = ["agents-shipgate", "check"]
+    if agent in {"codex", "claude-code", "cursor"}:
+        parts.extend(["--agent", agent])
+    parts.extend(["--workspace", shlex.quote(str(workspace))])
+    parts.extend(["--config", shlex.quote(str(config))])
+    if policy is not None:
+        parts.extend(["--policy", shlex.quote(str(policy))])
+    if diff:
```
pengfei-threemoonslab: **[P1] Recovery is still injectable and incomplete.** This safe serializer is used only for flag errors. Diff-input failures still call the legacy builder, which joins unquoted user-controlled paths/refs, hardcodes workspace to `.`, and drops config/policy. A semicolon in `--diff` becomes a shell command boundary in the allowed action; a one-sided range plus another bad flag is silently dropped. Use one complete quoted serializer and return review when repair is ambiguous.

src/agents_shipgate/core/agent_boundary.py
```diff
@@ -536,12 +553,22 @@ def _project_legacy(
...
+
+    The actor belongs in it. The result records which agent was evaluated, so
+    two runs differing only by actor are two audit rows, not one — and without
+    it, detecting the actor changed the label while leaving every Claude Code
+    and Cursor run indistinguishable from a codex run in the audit trail.
+    """
+
     payload = {
         "schema": AGENT_BOUNDARY_RESULT_SCHEMA_VERSION,
+        "actor": actor,
```
pengfei-threemoonslab: **[P2] Existing Codex audit IDs rotate.** Adding actor unconditionally changes established Codex IDs despite the changelog saying Codex IDs remain unchanged. The docs-only golden rotates from `agent_boundary_bcae9dc247f5dae731a41e8c` to `agent_boundary_6992d494be4b9b5ad459b3e3`. Preserve the legacy Codex digest payload or explicitly version the identity contract.

src/agents_shipgate/core/agent_boundary.py
```diff
@@ -159,6 +167,7 @@ def evaluate_agent_boundary(
     input_mode: Literal["worktree", "git_range", "provided_diff"] = "provided_diff",
     input_issues: list[BoundaryInputIssue] | None = None,
     host_snapshot: HostBoundarySnapshot | None = None,
+    config_path: Path | None = None,
```
pengfei-threemoonslab, line 181: **[P2] Pass the configured manifest into the cached assessment.** The new parameter is supplied by local check but not by `assessment_for_scan_context`. Consequently, a custom-manifest diff gets the protected-surface boundary finding locally but not in full verify. Tier-A trust-root detection still routes review, but the public surfaces publish inconsistent evidence.

pengfei-threemoonslab (commented):
Re-reviewed exact head `71047f6`. The common-path fixes for deterministic custom-manifest evidence, configured-manifest classification, mixed adoption/policy wording, full verify reruns, and actor-distinct attribution are real. I still found nine reproducible issues: seven P1 and two P2, detailed inline.

Validation: GitHub CI is green; Ruff and diff checks pass; all non-packaging tests passed, and the packaging module passed 8/8 with isolated-build network access. Shipgate produced a valid reproducible receipt with `mergeable` / `passed` and no reported blockers. Its local boundary result nevertheless remains `agent_action_required` after the exact successful verify rerun, consistent with the wrong-target verification-action finding below.

PR #306, opened 2026-08-03
Make a protected-surface stop name its route (and rule out chat approval)

pengfei-threemoonslab:
## Summary

- A human-routed preflight signal said only that a coding agent must not self-approve the edit. That leaves the agent to guess *how* a human decides, and the plausible guess is to ask the operator in conversation — which preflight never reads. The agent then either stalls on an answer nothing consumes, or treats a spoken "yes" as authority and proceeds past the gate. The first wastes a human context switch on a reviewer-requested edit; the second is the gate teaching the exact behaviour it exists to prevent.
- The recommendation now states both halves: approval goes through the **pull request**, and the agent must **not** ask the operator to approve the edit in chat.

Text on an existing signal. No check id, schema, report field, or routing changed. A trust-root edit is still `critical` and still stops the turn.

## History — this PR was much larger

It opened with an authoring-safe exception that let a provable version synchronization in an agent-instruction document continue to verification instead of stopping. Review found **three** distinct fail-opens across three rounds ([1](https://github.com/ThreeMoonsLab/agents-shipgate/pull/306#discussion_r3709657637), [2](https://github.com/ThreeMoonsLab/agents-shipgate/pull/306#discussion_r3714511268), [3](https://github.com/ThreeMoonsLab/agents-shipgate/pull/306#discussion_r3715123653)), each valid, each reproduced. The exception has been **withdrawn**, not narrowed again.

All three shared one cause: whether a number in a document is a *citation* or the *operand of a rule* depends on the surrounding English, and reading English is the judgement Shipgate does not make (Principle 3). Successive fixes — positional comparison, field binding to the exact contract key, syntactic templates — each proved something true and none proved the thing that mattered.

The remaining sound container was managed blocks. Measured rather than assumed: **no instruction document in this repository carries one**, and none of the five recognized citations sits in a managed block. That path covers zero real cases here and would need a `core` → `cli` layering inversion for code that never fires, so it was not worth shipping.

Consequence, stated plainly: **a reviewer-requested version sync still stops the turn.** Only the useless question is gone. Closing that gap properly is the signed approval receipt (P2), which needs its own contract bump and is not attempted here.

## Type

- [ ] Input adapter change
- [ ] Documentation only

Plus one unrelated dependency fix, kept as its own commit — see below.

## Verification

CI is authoritative for `python -m ruff check .`, `python -m compileall -q src tests`, and `python -m pytest`. All three clean — **4331 passed, 5 skipped**.

Additional local checks run:

- All three reported inputs now stop against this repository’s real `AGENTS.md`, returning `human_review_required` / `must_stop=True` with the new recommendation.
- The #282 manifest proposal branch shares this code path and is preserved exactly; its tests are untouched and pass.

## Release-readiness notes

- [x] New or changed check IDs are documented in `docs/checks.md` — none added, renamed, or removed

## Reviewer notes

**Unrelated CI fix included.** `e474ea81` raises `cryptography` from `>=49.0.0,<50` to `>=50.0.0,<51`. The dependency audit fails on 49.0.0 (CVE-2026-69247) and the old ceiling structurally excluded the only fix version. This branch touches no dependency file; `main` last ran green on 2026-07-31, before the advisory. Any PR opened today hits it. Verified with a clean-venv dry-run resolve (`pyOpenSSL 26.4.0` supports cryptography 50), `pip check`, `pip_audit .`, and a full suite run on 50.0.0. **Worth cherry-picking to `main` independently** — `main` will fail the same way on its next run.

**Net diff is now 4 files, +87 lines** (from +1459 at its peak).


src/agents_shipgate/core/instruction_sync.py
```diff
@@ -0,0 +1,251 @@
...
+        return _unsafe("change contains no edited instruction lines")
+
+    synchronized: list[tuple[str, str]] = []
+    for hunk in diff_file.hunks:
+        removed = [text for kind, text in hunk.lines if kind == "-"]
+        added = [text for kind, text in hunk.lines if kind == "+"]
+        if len(removed) != len(added):
+            return _unsafe("a hunk adds or removes instruction lines")
+        for old_line, new_line in zip(removed, added, strict=True):
+            reason = _assess_line_pair(old_line, new_line, synchronized)
```
pengfei-threemoonslab: **[P1] Preserve line positions across context.** Collecting all removed and added lines and zipping them ignores unchanged context. Moving `Contract v14 applies to the next rule` from above an instruction to below it while changing `v14` to `v19` returns `sync_safe=True`. Reconstruct and compare the complete old/new hunk sequences, including context, after masking.

pengfei-threemoonslab (commented):
Review verdict: changes recommended. The proposed authoring-safe exception is not fail-closed for several reproducible inputs. Each inline case currently returns `sync_safe=True` and propagates through `build_preflight_result` as `agent_action_required` with `requires_human_review=False`. GitHub CI, the targeted suite, Ruff, compileall, packaging tests, and the deterministic Shipgate verifier are otherwise clean; these are risk-model soundness gaps that the existing tests and static verifier do not cover. GitHub does not permit the PR author to submit a formal “Request changes” review on their own PR.

src/agents_shipgate/core/instruction_sync.py
```diff
@@ -0,0 +1,251 @@
...
+    authoritative = _authoritative_version_literals()
+    for old_token, new_token in zip(old_tokens, new_tokens, strict=True):
+        if old_token == new_token:
+            continue
+        if new_token not in authoritative:
+            return (
+                f"{new_token!r} is not a version this CLI publishes; a coding "
+                "agent must not invent a contract, schema, or release number"
+            )
+        synchronized.append((old_token, new_token))
```
pengfei-threemoonslab: **[P1] Published versions can still weaken thresholds.** Checking only that the new token appears somewhere in the global contract payload does not preserve its meaning. `Only trust control contract version >= 19` → `>= 14` returns `sync_safe=True`, even though it lowers a load-bearing threshold. Bind replacements to the corresponding contract field or fail closed for comparison/conditional language.

src/agents_shipgate/core/instruction_sync.py
```diff
@@ -0,0 +1,251 @@
...
+    return None
+
+
+def _has_version_context(line: str) -> bool:
+    folded = line.casefold()
+    return any(term in folded for term in _VERSION_CONTEXT_TERMS)
```
pengfei-threemoonslab: **[P1] Match version context as a real term.** Substring matching makes unrelated words such as `conversion` satisfy the version-context guard. Consequently, `Require at least 2 approvals for conversion` → `19 approvals` is accepted and preflight removes human routing. Require an actual word or recognized identifier boundary; ideally associate each number with a specific version field.

src/agents_shipgate/core/instruction_sync.py
```diff
@@ -0,0 +1,251 @@
...
+    def capture(match: re.Match[str]) -> str:
+        tokens.append(match.group(0))
+        return _MASK
+
+    # A pre-existing placeholder would let two different lines mask to the same
+    # text. Control characters are not legal repository content here, so their
+    # presence is refused rather than normalized.
+    if _MASK in line:
+        return "\x00unmaskable\x00", []
+    return pattern.sub(capture, line), tokens
```
pengfei-threemoonslab: **[P2] Sentinel presence must fail explicitly.** When both sides contain the mask sentinel, `_mask` returns the same constant for two different lines instead of reporting failure. Combined with one legitimate version sync elsewhere in the hunk, arbitrary prose changes are accepted. Return an explicit unsafe result whenever the sentinel or disallowed controls occur.

src/agents_shipgate/core/instruction_sync.py
pengfei-threemoonslab: Confirmed and fixed in e4b7e56e. Reproduced exactly as described — the move returned `sync_safe=True`.

Each hunk's complete old and new sequences are now reconstructed with context included (`kind in {" ", "-"}` vs `kind in {" ", "+"}`) and compared position for position, so a line that keeps its text but changes position relative to its neighbours is refused.

Fixing this surfaced a related gap I have also closed: `parse_unified_diff` appends a `+`/`-` line appearing before the first `@@` header to `added_lines`/`removed_lines` but to no hunk, so it would never reach the positional comparison at all. The hunks must now account for every edited line (`test_an_edited_line_outside_every_hunk_is_refused`).

Regression: `test_moving_a_line_across_context_is_refused`.

src/agents_shipgate/core/instruction_sync.py
pengfei-threemoonslab: Confirmed and fixed in e4b7e56e. You are right that my original mitigation was incomplete — it only stopped a *fabricated* number, not a swap between two published ones.

I took both halves of your suggestion rather than one:

- A line stating a version condition (`>=`, `<=`, `at least`, `or later`, `older than`, …) is now refused outright, since its literal is a threshold rather than a fact.
- Independently, a changed literal must be **non-decreasing**. Field binding is not reliable here (both `19` and `14` are legitimately "contract"), so this closes the general downgrade class rather than only the comparison-worded case: `Set contract version 19 here` → `14` has no comparison word and is now refused too. Anything unorderable (pre-release suffix, unparsable shape) fails closed.

Regressions: `test_a_line_stating_a_version_condition_is_refused` (4 phrasings), `test_a_published_value_may_not_replace_a_higher_published_value`.

src/agents_shipgate/core/instruction_sync.py
pengfei-threemoonslab: Confirmed and fixed in e4b7e56e. `conversion` did satisfy the guard and the threshold rewrite was accepted.

I went to the stronger option you flagged as ideal rather than only fixing the boundary. A bare integer is now a version literal only where a version **field introduces it directly** (`...contract_version: 14`, `contract 14`) — not merely because the line mentions a version somewhere.

Line-level context was loose in a way the reported case only partly exposed: my own follow-up probe found `Use subversion 2 for schema` → `19` still passing, because `schema` anywhere on the line made every number rewritable. Adjacency fixes both at once.

Boundaries are matched on identifier rules, not `\b` — `_` separates, letters and digits do not — because `\b` would reject the real `minimum_control_contract_version: 14` case while substring matching accepts `conversion`/`subversion`.

Regression: `test_a_number_not_introduced_by_a_version_field_is_refused`, covering `conversion`, `subversion`, and two unrelated-count phrasings.

src/agents_shipgate/core/instruction_sync.py
pengfei-threemoonslab: Confirmed and fixed in e4b7e56e. Reproduced: two sentinel-bearing lines compared equal, and one legitimate sync elsewhere in the hunk carried the whole change to `sync_safe=True`.

Rather than special-casing the sentinel, I removed the collision class: lines are now compared as **tuples of literal segments** with tokens held separately, so there is no placeholder in the compared value for anything to collide on. Equal segment tuples also imply equal token counts, which drops a separate length check.

Per your note on disallowed controls, a control character in a compared line is now an explicit unsafe result (tab excluded, since it is legal in Markdown).

Regression: `test_a_control_character_pair_cannot_launder_a_prose_change`, asserting the refusal reason rather than only the verdict.

pengfei-threemoonslab:
Thanks — all four were real, and all four reproduced exactly as written. Each returned `sync_safe=True` and reached `agent_action_required` with `requires_human_review=False` before the fix. Addressed in e4b7e56e, with replies inline on each thread.

| # | Finding | Fix |
|---|---|---|
| 1 | Zipping removed/added ignored context, accepting a **move** | Per-hunk old/new sequences reconstructed **with context** and compared position for position |
| 2 | A published value could still **lower a threshold** | Version-condition lines refused outright **and** changed literals must be non-decreasing |
| 3 | Substring match let `conversion` grant version context | Bare integer counts only where a version **field introduces it**, at identifier boundaries |
| 4 | Mask sentinel made two unequal lines compare equal | Compared as **tuples of literal segments** (no placeholder to collide on); control chars refused explicitly |

Two things worth calling out beyond the four:

**Fixing #4 surfaced a fifth gap.** `parse_unified_diff` appends a `+`/`-` line that appears before the first `@@` header to `added_lines`/`removed_lines` but to no hunk — so it would never reach the positional comparison, and file-level counts would vouch for a line nothing inspected. The hunks must now account for every edited line.

**#3 was looser than the reported case showed.** My own probe after fixing the boundary found `Use subversion 2 for schema` → `19` still passing, because `schema` anywhere on the line made every number rewritable. That is why I took the field-adjacency option you flagged as ideal rather than only correcting the boundary regex.

On #2 I implemented both alternatives you offered, not one. Field binding alone is unreliable — `19` and `14` are both legitimately "contract" — so the condition-language refusal handles the worded case and the non-decreasing rule closes the general downgrade class, including `Set contract version 19 here` → `14`, which carries no comparison word.

Net effect on the exception's reach: it now accepts only a forward move of a published version, introduced by a version field, on a line that states no condition, with every other line byte- and position-identical.

Verification: `ruff check .`, `compileall`, and the full suite are clean — **4388 passed, 5 skipped** (up from 4374; 14 new regression tests, one per finding plus the variants). The five inputs all refuse; legitimate syncs still pass end-to-end on this repository's real `AGENTS.md` and `.agents/skills/agents-shipgate/references/recipes.md`, on both `preflight` and `check`.

One fixture correction worth flagging: my original `test_bare_literal_synchronizes_on_a_version_context_line` used `99` as the *stale* value, i.e. a downgrade. It failed once the non-decreasing rule landed — correctly. It now models a real sync (`1` → `14`).

pengfei-threemoonslab:
CI is green. The failure was **not** from this branch's changes — it was the `Dependency audit` step, not the test run:

```
Found 1 known vulnerability in 1 package
cryptography 49.0.0  CVE-2026-69247  Fix Versions: 50.0.0
```

Fixed in e474ea81 by raising the pin from `>=49.0.0,<50` to `>=50.0.0,<51`.

**Why it needed a change here rather than a wait for Dependabot:** the old ceiling structurally excluded the only fix version, so no in-range release resolved it. This branch does not touch dependencies, and `main` last ran green on 2026-07-31 — before the advisory landed. CI re-resolves on every run, so any PR opened today hits this regardless of content.

**One thing I checked before assuming a one-line bump was safe:** `.[dev]` pulls `sigstore` → `pyOpenSSL`, and the pyOpenSSL in my local environment caps `cryptography<49`, which would have made the bump unresolvable. That turned out to be stale local state — a clean resolve selects pyOpenSSL 26.4.0, which supports cryptography 50. Verified rather than assumed:

- `pip install --dry-run ".[dev]"` in a clean venv → `cryptography-50.0.0`, `pyOpenSSL-26.4.0`, `sigstore-4.5.0`, no conflict
- `pip check` → no broken requirements
- `python -m pip_audit .` (the exact CI step) → `No known vulnerabilities found`, exit 0
- full suite re-run **on 50.0.0** → 4388 passed, 5 skipped, including the Ed25519 authorization tests that are this project's only use of the library (my earlier runs were on 48.0.1, so they had not exercised it)

No CHANGELOG entry, matching how dependency bumps land in this repo (#286/#287 touch `pyproject.toml` only). Kept as its own commit so it can be split out or cherry-picked to `main` independently of the review fixes — worth doing if you want `main` unblocked before this merges.

src/agents_shipgate/core/instruction_sync.py
```diff
@@ -0,0 +1,381 @@
...
+    ]
+    if not changed:
+        return None
+    if _states_version_condition(old_line) or _states_version_condition(new_line):
+        return (
+            "an edited line states a version condition, so a changed literal "
+            "would move a threshold rather than restate a fact"
+        )
+
+    authoritative = _authoritative_version_literals()
```
pengfei-threemoonslab: **[P1] Bind literals to the specific version field**

The authoritative set combines every published `_version` value. Consequently, `Report schema v0.33 applies.` → `Report schema v19 applies.` is classified as safe because `v19` belongs to the contract version, even though it is not a report-schema version. The full preflight path returns `agent_action_required`, bypassing human review. Please validate the replacement against the version field identified by the surrounding text and fail closed when that field is ambiguous.

src/agents_shipgate/core/instruction_sync.py
```diff
@@ -0,0 +1,381 @@
...
+# digits do not.
+_VERSION_FIELD_SUFFIX_RE = re.compile(
+    r"(?<![a-z0-9])(?:version|contract|schema)s?$"
+)
+# Separators that may sit between a version field and its value.
+_FIELD_VALUE_SEPARATORS = " \t:=`\"'([,"
+
+# A line stating a version *condition* carries a threshold, not a fact, so a
+# changed literal moves the condition even when both values are published.
+_VERSION_CONDITION_RE = re.compile(
```
pengfei-threemoonslab: **[P1] Non-decreasing does not prove semantic safety**

The condition blacklist misses equivalent wording. Both `Allow contract versions through v14.` → `... v19.` and `Reject contract versions after v14.` → `... v19.` reach `agent_action_required`; the latter broadens the allowed range. Enumerating conditional phrases cannot safely cover arbitrary instruction prose. Please restrict this exception to positively recognized factual templates or managed fields; unknown language should remain human-routed.

pengfei-threemoonslab (commented):
Re-review: the four earlier findings are addressed, but these two P1 cases still let semantic instruction changes bypass the standing human route. I recommend keeping the PR open until both cases fail closed.

src/agents_shipgate/core/instruction_sync.py
pengfei-threemoonslab: Confirmed and fixed in 8660ee1d. Reproduced exactly — `Report schema v0.33 applies.` → `v19` returned `sync_safe=True` because the authoritative set pooled every published `_version` value into one flat set.

Replacements are now bound to the field. A version is recognized only as the value of an exact `build_contract_payload()` key, and the new value must be the published value **of that key**, so `report_schema_version` can never take the contract's number.

On failing closed when the field is ambiguous: rather than trying to infer the field from surrounding text, nothing is recognized unless the field names itself adjacently in one of two templates (see the other thread). `Report schema v0.33 applies.` has no field name, so it is now not recognized at all — ambiguity never arises.

Regressions: `test_one_fields_version_may_not_be_written_into_another`, `test_an_unknown_field_is_refused`, `test_renaming_the_field_is_a_prose_change`.

src/agents_shipgate/core/instruction_sync.py
pengfei-threemoonslab: Confirmed and fixed in 8660ee1d. Both reproduced, and I agree with the structural point rather than just the two examples: **enumeration cannot be finished.** "through", "after", "up to", "prior to", "beyond", "as of" — and behind that, telling a rule from a fact in English is exactly the semantic judgement Shipgate promises not to make (Principle 3). A blacklist was the wrong shape for this guard.

Recognition is now positive, per your guidance. A version is a managed field only when it is the value of an exact contract-payload key, written either:

- inside a code span holding nothing but the assignment — ``report `minimum_control_contract_version: 14` `` — or
- on a line holding nothing but the assignment (fenced-block / YAML-snippet lines).

`Allow contract versions through v14.` is refused because it is not an assignment, not because "through" is known. Unknown language is human-routed by default now, since prose is simply never a template.

Two guards are **deleted** rather than kept as defence in depth:

- The conditional blacklist is unreachable under the templates. Keeping it would imply the model still depends on enumeration, which is the thing you flagged.
- Monotonicity is not just redundant but **wrong**: field binding proves the new value is the published one, so correcting a document that claims `contract_version: 99` down to `19` is a valid sync that the non-decreasing rule refused. (`test_a_downward_correction_synchronizes`.)

Regression: `test_a_version_in_prose_is_not_recognized`, covering both your cases plus `up to`, `>=`, `or later`, a negated prohibition, and two unrelated counts.

pengfei-threemoonslab:
Both cases fail closed now — 8660ee1d. Replies inline; the short version is that I took the second finding as a design verdict rather than two missed entries.

**The blacklist was the wrong shape.** "through" and "after" are not gaps I can patch; "up to", "prior to", "beyond", "as of" follow, and behind them is the fact that separating a rule from a fact in English is exactly the semantic judgement Shipgate promises not to make. So recognition is now **positive**: a version is a managed field only when it is the value of an exact `contract --json` payload key, written either inside a code span holding nothing but the assignment (``report `minimum_control_contract_version: 14` ``) or on a line holding nothing but the assignment. The new value must be the published value **of that field**.

That closes both findings by different routes: field binding kills the wrong-field write directly; `Allow contract versions through v14` dies because it is not an assignment, with no rule about the word "through".

**I deleted two guards rather than keeping them as defence in depth.** The conditional blacklist is unreachable under the templates, and leaving it in would imply the model still leans on enumeration. Monotonicity turned out to be not merely redundant but *wrong*: field binding proves the new value is the published one, so correcting a doc that claims `contract_version: 99` down to `19` is a valid sync that the non-decreasing rule was refusing.

**The cost, stated plainly.** This is a real narrowing, and it is worth your judgement rather than my assertion. Prose mentions — including `Contract v14 publishes these boundaries` in our own AGENTS.md — now stay human-routed. I measured what survives across this repo's instruction documents rather than guessing:

```
RECOGNIZED  .agents/skills/agents-shipgate/SKILL.md:18   minimum_control_contract_version = 14
RECOGNIZED  .agents/skills/agents-shipgate/SKILL.md:29   minimum_control_contract_version = 14
RECOGNIZED  .agents/skills/agents-shipgate/references/recipes.md:20  minimum_control_contract_version = 14
RECOGNIZED  AGENTS.md:367                                 report_schema_version = 0.34
RECOGNIZED  AGENTS.md:448                                 report_schema_version = 0.10
```

Five citations, covering **both** files named in the original block, so the reported scenario still clears end to end. But the exception is now materially smaller than when this PR opened — closer to "managed-field citations sync themselves" than "version syncs stop blocking". If that trade is not worth the surface, narrowing further to zero and closing this in favour of the P2 approval receipt is a reasonable call.

**One residual I am not going to paper over.** AGENTS.md:448 cites `report_schema_version: "0.10"` deliberately — it documents the *frozen older* schema. The exception would accept syncing it to `0.34`, which is a factual regression in the docs. I am not adding a heuristic for it, because detecting "this citation is intentionally historical" means reading the surrounding English, which is the thing we just removed. It is a documentation error a reviewer sees in the diff, not an authority change, and the verify-mode trust-root finding still routes it to review.

Verification: `ruff check .`, `compileall`, full suite — **4391 passed, 5 skipped**. Both new inputs refuse, the four earlier ones stay refused, and real-repo preflight still returns `agent_action_required` for the managed-field sync and `human_review_required` for an instruction softening.

pengfei-threemoonslab (commented):
Re-review found one remaining P1: the positive field binding fixes the earlier cases, but assignment syntax still does not prove that a value is merely a citation rather than a load-bearing instruction.

src/agents_shipgate/core/instruction_sync.py
```diff
@@ -0,0 +1,333 @@
...
+    """Locate ``(start, end, field, value)`` for each managed-field assignment.
+
+    Only two containers are recognized, both of which fix the meaning of the
+    number structurally: a code span holding nothing but the assignment, and a
+    line that is nothing but the assignment. A version mentioned in ordinary
+    prose is not a managed field and is left in the compared text.
+    """
+
+    found: dict[tuple[int, int], tuple[str, str]] = {}
+    candidates = [(start, line[start:end]) for start, end in _code_spans(line)]
```
pengfei-threemoonslab: **[P1] Assignment syntax does not prove citation semantics**

Exact field binding proves which value belongs to a field, but not whether the occurrence is a factual citation or a load-bearing instruction. `Never proceed unless \`minimum_control_contract_version: 9999\`.` → the published `14` returns `sync_safe=True`; full preflight returns `agent_action_required`, `must_stop=False`, and no human review. Likewise, raising the assignment in `Do not allow anything newer than \`report_schema_version: 0.10\`.` broadens the permitted range. Please restrict recognition to exact known/generated citation templates or managed blocks; arbitrary code-span and whole-line assignments must remain human-routed.

pengfei-threemoonslab: Confirmed, and I am withdrawing the exception rather than narrowing it a fourth time — a4cd0726.

Both cases reproduced. Probing further, so does a third shape the directive implies: a **whole-line** assignment governed by an unchanged *preceding* rule line.

```
  Do not allow anything newer than:     <- context, unchanged
- report_schema_version: 0.10
+ report_schema_version: 0.34           -> sync_safe=True
```

**My justification for the code-span template was simply wrong.** I argued a code span is a quoted machine literal, so the sentence around it is unchanged. But the sentence *is* the instruction and the number is its operand — byte-identical prose around a changed operand still changes the rule.

That is three distinct fail-opens with one cause: whether a number is a citation or an operand depends on the surrounding English, and reading English is the judgement Shipgate does not make. Field binding proved *which* value belongs to a field. It cannot prove *why the value is there*, and no syntactic container fixes that while the container sits inside prose.

**On the remaining option — managed blocks and generated citations — I measured instead of assuming:**

```
citation                                                          in managed block?
.agents/skills/agents-shipgate/SKILL.md:18   minimum_control_contract_version=14   no
.agents/skills/agents-shipgate/SKILL.md:29   minimum_control_contract_version=14   no
.agents/skills/.../references/recipes.md:20  minimum_control_contract_version=14   no
AGENTS.md:367                                report_schema_version=0.34            no
AGENTS.md:448                                report_schema_version=0.10            no
```

No instruction document in this repo carries a managed block at all. The sound remainder covers **zero** real cases here and would not clear the block that motivated the work, while costing a core→cli layering inversion for a path that never fires. That is not worth shipping.

So `instruction_sync.py` and its tests are deleted; `preflight` and `check` keep the standing whole-file route for every agent-instruction edit. Net versus main is now **4 files, +87 lines**.

What survives needed no semantic judgement: the stop now names the route (the pull request) and the non-route (chat), so an agent stops asking for an approval nothing reads. A reviewer-requested version sync still stops the turn — the useless question does not. Closing that gap properly is the P2 signed approval receipt, which needs its own contract bump.

Thanks for holding the line across three rounds; each finding was correct, and the third is what showed the approach was unfixable rather than incomplete.

PR #314, opened 2026-08-05
fix(init): derive tool_sources ids from the whole path, not the basename

### LINKED ISSUE #307: bug(init): duplicate tool_sources IDs for repeated tool.py/tools.py basenames
## Summary

`shipgate init --write --json` can fail before writing a manifest when auto-detection finds multiple OpenAI Agents SDK source files with the same basename (for example, several `tool.py` or `tools.py` files).

The generated `tool_sources[].id` values collide, and the manifest validation guard correctly rejects the generated YAML.

## Reproduction

Tested with the current development source:

```bash
AGENTS_SHIPGATE_AGENT_MODE=1 \
PYTHONPATH=/path/to/agents-shipgate/src \
python -m agents_shipgate init \
  --workspace /path/to/usestrix-strix \
  --write \
  --json
```

The Strix workspace is a valid OpenAI Agents SDK project and contains multiple paths with conventional repeated basenames, including:

- `strix/tools/finish/tool.py`
- `strix/tools/load_skill/tool.py`
- `strix/tools/respond/tool.py`
- multiple `*/tools.py` modules

Observed with CLI version `0.16.0b7` / contract version `19`.

## Actual result

```text
Generated manifest failed validation: 1 validation error for AgentsShipgateManifest
Value error, tool_sources[].id values must be unique
```

In agent mode, this is reported as `internal_error` and routes to `init --minimal`.

No invalid `shipgate.yaml` is written, which is good.

## Root cause

`_tool_sources_block()` deduplicates entries by full path, but `_source_id_for()` derives the ID from only the source type and `Path(path).stem`.

For example, several distinct OpenAI Agents SDK paths map to the same ID:

```text
.../finish/tool.py      -> openai_sdk_tool
.../load_skill/tool.py  -> openai_sdk_tool
.../respond/tool.py     -> openai_sdk_tool
```

The validation step catches the collision after rendering.

## Expected result

Auto-init should emit a schema-valid manifest for repositories with repeated file basenames.

Suggested behavior:

- derive IDs from a sanitized relative path, or
- append a deterministic collision suffix;
- keep IDs unique and stable for the same workspace contents;
- add a regression test covering multiple `tool.py` and `tools.py` files in one detected framework.

## Impact

This blocks the normal auto-init adoption path for common Python repository layouts. The `--minimal` fallback is useful as a recovery path, but it produces a manual `CHANGE_ME` template and does not automatically configure the detected OpenAI Agents SDK sources.

pengfei-threemoonslab:
Fixes #307.

## What was wrong

`tool_sources[].id` was built from the source type plus `Path(path).stem`, while both `init` renderers deduplicate entries by *full path*. A repository with one `tool.py` per tool package — the conventional Python layout, and exactly what `usestrix/strix` ships — produced three distinct entries all carrying `openai_sdk_tool`, and the manifest schema (correctly) rejects a manifest whose ids repeat.

The failure was total on the primary adoption path: `init --write` exited 4 with `internal_error`, wrote no manifest, and routed to `--minimal`, which throws away the detection work for a `CHANGE_ME` template.

**Second instance, found while fixing the first:** `--minimal` had the same basename-only rule *with no validation gate in front of it*. Two services that each ship `openapi.yaml` got an invalid manifest **written to disk** (exit 0), and the documented next step failed on it:

```
$ agents-shipgate init --workspace . --minimal --write
Wrote .../shipgate.yaml                                    # exit 0

$ agents-shipgate scan --config shipgate.yaml
Config error: Invalid shipgate.yaml:
- <root>: Value error, tool_sources[].id values must be unique
```

…and that error's `next_action` says *"Fix the manifest in place — do not re-run init"*, so the recovery path dead-ends in a manual edit.

## The fix

One helper, `cli/discovery/source_ids.py`, shared by the auto renderer (`template.py`) and the legacy `--minimal` renderer (`artifacts.py`) — the two copies of the rule were the two copies of the bug.

Ids are derived from the **whole workspace-relative path**:

```yaml
tool_sources:
  - id: openai_sdk_strix_tools_finish_tool
    type: openai_agents_sdk
    path: strix/tools/finish/tool.py
  - id: openai_sdk_strix_tools_respond_tool
    type: openai_agents_sdk
    path: strix/tools/respond/tool.py
```

Per the sanitized-relative-path preference in #307's triage comment, over a collision suffix: ids carry no positional component, so a file appearing earlier in the walk can never renumber the entries after it. Details:

- **Residual collisions** (`a-b/tools.py` vs `a_b/tools.py`, or paths whose segments are all non-ASCII) — *every* member of the group takes a digest of its own path rather than one side keeping the plain form, so walk order never decides which file gets renamed.
- **Stability, stated exactly** — an id is a pure function of its own path *unless another declared path sanitizes to the same string*; that group is the one case where adding a file changes an id that already existed. Making it file-local would mean a digest on every id, including the readable common ones. Bounded by `init` refusing to overwrite an existing manifest, so ids are assigned once per adoption. Pinned in both directions by tests.
- **Length** — a deep monorepo path keeps its most specific segments plus a digest of the full path, and the 64-char bound is enforced on the value that ships, disambiguated ids included.
- The auto path's post-render validation gate is unchanged and remains the backstop. Where `--minimal` has no such gate, the assignment itself is unique by construction.

## Verification

Reproduced before/after on the repository from the report (`usestrix/strix`, shallow clone):

| | `init --write --json` | manifest | `scan` |
|---|---|---|---|
| `527a0349` (v0.16.0b7) | exit 4, `internal_error` | not written | — |
| this branch | exit 0 | 23 sources, **23 unique ids**, longest id 47 chars | exit 0 |

Tests added to `tests/test_init_auto.py`: the #307 repro (multiple `tool.py` + `tools.py` in one detected framework), the full `init --write` → `scan` cold-start flow on that layout, an id-stability test that fails if ids ever become position-dependent, the `--minimal` repro, and unit invariants for the helper (collision digest, determinism, length bound, unknown source types).

`python -m pytest` and `python -m ruff check .` are green. `shipgate check` → `allow` / no violations; `shipgate verify` → `complete` / `passed`.

## Compatibility

- **`--minimal` ids for nested sources change** (`api/openapi.yaml`: `openapi_api_openapi` instead of `openapi_openapi`). Unavoidable — the colliding case had no valid id to preserve — and root-level sources are unchanged. STABILITY.md lists the `init` template under "what MAY change"; the test docstring claiming byte-exact v0.5 output is updated to name this one exception.
- **No existing `shipgate.yaml` is touched.** `init` refuses to overwrite one, so nothing keyed on an already-written id churns.
- No new CLI, schema, or report surface; one new internal module and two deleted duplicate helpers.
- `tests/test_adapter_static_only.py` moves two allowlist line numbers because the edited region of `artifacts.py` shifted. The `subprocess` call sites themselves are unchanged.

## Deliberately not changed

`--minimal` still has no manifest-validation gate. It is the fallback the auto gate routes *to*; making it exit 4 as well would leave a failing workspace with nowhere to go. This fix removes its known invalid-output mode instead; the test now pins that its render validates.

### DIFF
--- a/src/agents_shipgate/cli/discovery/source_ids.py
+++ b/src/agents_shipgate/cli/discovery/source_ids.py
@@ -11,10 +11,22 @@ caught that at its validation gate and wrote nothing; ``--minimal`` had no
 gate and wrote the invalid file, so the documented next step (``scan``)
 failed with a config error. See issue #307.
 
-Ids are derived from the whole workspace-relative path, so an id depends
-only on the file it names. A positional suffix (``_2``, ``_3``) would
-instead renumber existing entries whenever an unrelated file appears
-earlier in the walk, churning anything keyed on the id.
+Ids are derived from the whole workspace-relative path and carry no
+positional component: no ``_2``/``_3`` counter, so a file appearing
+earlier in the walk can never renumber the entries after it.
+
+**Stability guarantee, exactly.** An id is a pure function of its own
+path *unless another declared path sanitizes to the same id* — sanitizing
+is lossy (``a-b/`` and ``a_b/`` both fold to ``a_b``; so do ``a/b_c.py``
+and ``a_b/c.py``). Those groups are disambiguated by digest, so adding
+such a path and regenerating from scratch changes the ids of every member
+of the group, not just the newcomer. Making that case file-local would
+mean putting a digest on *every* id, including the common ones this
+scheme exists to keep readable. The blast radius is bounded: ``init``
+refuses to overwrite an existing ``shipgate.yaml``, so ids are assigned
+once per adoption, and the fold classes are rare in one workspace.
+``tests/test_init_auto.py`` pins both halves — non-colliding siblings
+never shift an id; a colliding one shifts both.
 """
 
 from __future__ import annotations
@@ -40,7 +52,8 @@ SOURCE_ID_PREFIXES: dict[str, str] = {
 
 # Generated ids travel into findings and report output, so a deep monorepo
 # path is truncated to the most specific segments that fit and made unique
-# again by a digest of the full path.
+# again by a digest of the full path. The bound holds for every id this
+# module returns, disambiguated ones included.
 MAX_SOURCE_ID_LENGTH = 64
 
 _DIGEST_LENGTH = 8
@@ -51,15 +64,60 @@ def source_id_for(source_type: str, relative_path: str) -> str:
     """Deterministic id for one detected source file.
 
     ``relative_path`` is the workspace-relative path as written to the
-    manifest's ``path:`` field.
+    manifest's ``path:`` field. Never longer than
+    :data:`MAX_SOURCE_ID_LENGTH`.
     """
+    return _compose(source_type, relative_path, disambiguate=False)
+
+
+def assign_source_ids(entries: Sequence[tuple[str, str]]) -> list[str]:
+    """Ids for ``(source_type, relative_path)`` pairs, unique by construction.
+
+    Distinct paths can still sanitize to one id (``a-b/tools.py`` and
+    ``a_b/tools.py``). Every member of such a group takes a digest of its
+    own path — rather than one side keeping the plain form — so which file
+    was walked first never decides which one gets renamed. See the module
+    docstring for what that costs in stability.
+    """
+    ids = [source_id_for(source_type, path) for source_type, path in entries]
+    collided = {value for value, count in Counter(ids).items() if count > 1}
+    if not collided:
+        return ids
+    resolved = [
+        _compose(source_type, path, disambiguate=True) if value in collided else value
+        for value, (source_type, path) in zip(ids, entries, strict=True)
+    ]
+    if len(set(resolved)) == len(resolved):
+        return resolved
+    # A plain id can equal another entry's disambiguated form when a file
+    # is named after that digest (`a_b_spec_<digest of a-b/spec.yaml>`).
+    # Contrived, but constructible — and `--minimal` has no validation
+    # gate behind it, so give every entry a digest rather than let a
+    # duplicate reach the render.
+    return [
+        _compose(source_type, path, disambiguate=True) for source_type, path in entries
+    ]
+
+
+def _compose(source_type: str, relative_path: str, *, disambiguate: bool) -> str:
     prefix = SOURCE_ID_PREFIXES.get(source_type, source_type)
     segments = _path_segments(relative_path)
-    source_id = "_".join([prefix, *segments]) if segments else f"{prefix}_root"
-    if len(source_id) <= MAX_SOURCE_ID_LENGTH:
-        return source_id
+    plain = "_".join([prefix, *segments]) if segments else f"{prefix}_root"
+    if not disambiguate and len(plain) <= MAX_SOURCE_ID_LENGTH:
+        return plain
+    return _with_digest(prefix, segments, relative_path)
 
+
+def _with_digest(prefix: str, segments: list[str], relative_path: str) -> str:
+    """``prefix`` + the most specific segments that fit + a path digest.
+
+    The digest is what keeps two truncated (or two identically sanitized)
+    paths apart, so the length budget is computed with room for it — a
+    source-type prefix long enough to crowd it out is truncated too, which
+    keeps the bound true for third-party adapter names of any length.
+    """
     digest = _digest(relative_path)
+    prefix = prefix[: MAX_SOURCE_ID_LENGTH - len(digest) - 1]
     budget = MAX_SOURCE_ID_LENGTH - len(prefix) - len(digest) - 2
     kept: list[str] = []
     width = 0
@@ -69,24 +127,6 @@ def source_id_for(source_type: str, relative_path: str) -> str:
         kept.insert(0, segment)
         width += len(segment) + 1
     return "_".join([prefix, *kept, digest])
-
-
-def assign_source_ids(entries: Sequence[tuple[str, str]]) -> list[str]:
-    """Ids for ``(source_type, relative_path)`` pairs, unique by construction.
-
-    Distinct paths can still sanitize to one id (``a-b/tools.py`` and
-    ``a_b/tools.py``). Both sides then take a digest of their own path
-    rather than one side keeping the plain form, so which file was walked
-    first never decides which one gets renamed.
-    """
-    ids = [source_id_for(source_type, path) for source_type, path in entries]
-    collided = {value for value, count in Counter(ids).items() if count > 1}
-    if not collided:
-        return ids
-    return [
-        f"{value}_{_digest(path)}" if value in collided else value
-        for value, (_source_type, path) in zip(ids, entries, strict=True)
-    ]
 
 
 def _path_segments(relative_path: str) -> list[str]:
--- a/src/agents_shipgate/cli/discovery/source_ids.py
+++ b/src/agents_shipgate/cli/discovery/source_ids.py
@@ -16,17 +16,24 @@ positional component: no ``_2``/``_3`` counter, so a file appearing
 earlier in the walk can never renumber the entries after it.
 
 **Stability guarantee, exactly.** An id is a pure function of its own
-path *unless another declared path sanitizes to the same id* — sanitizing
-is lossy (``a-b/`` and ``a_b/`` both fold to ``a_b``; so do ``a/b_c.py``
-and ``a_b/c.py``). Those groups are disambiguated by digest, so adding
-such a path and regenerating from scratch changes the ids of every member
-of the group, not just the newcomer. Making that case file-local would
-mean putting a digest on *every* id, including the common ones this
-scheme exists to keep readable. The blast radius is bounded: ``init``
-refuses to overwrite an existing ``shipgate.yaml``, so ids are assigned
-once per adoption, and the fold classes are rare in one workspace.
-``tests/test_init_auto.py`` pins both halves — non-colliding siblings
-never shift an id; a colliding one shifts both.
+path *unless it collides with another entry's id*. Sanitizing is lossy
+(``a-b/`` and ``a_b/`` both fold to ``a_b``; so do ``a/b_c.py`` and
+``a_b/c.py``), so those groups are disambiguated by digest and adding
+such a path changes the ids of every member of the group, not just the
+newcomer. Making that case file-local would mean putting a digest on
+*every* id, including the common ones this scheme exists to keep
+readable. Nothing outside a collision is ever re-keyed: an entry whose id
+is unique keeps it, whatever else the workspace contains. The blast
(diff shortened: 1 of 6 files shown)


src/agents_shipgate/cli/discovery/source_ids.py
```diff
@@ -0,0 +1,103 @@
...
+
+def assign_source_ids(entries: Sequence[tuple[str, str]]) -> list[str]:
+    """Ids for ``(source_type, relative_path)`` pairs, unique by construction.
+
+    Distinct paths can still sanitize to one id (``a-b/tools.py`` and
+    ``a_b/tools.py``). Both sides then take a digest of their own path
+    rather than one side keeping the plain form, so which file was walked
+    first never decides which one gets renamed.
+    """
+    ids = [source_id_for(source_type, path) for source_type, path in entries]
```
pengfei-threemoonslab, line 99: **[P2] Collision handling makes IDs set-dependent**

The PR promises that an ID depends only on its file, but collision detection depends on the complete entry set. For example, `a-b/spec.yaml` alone receives `openapi_a_b_spec`; adding `a_b/spec.yaml` changes the existing file’s ID to `openapi_a_b_spec_13072c95`. This can churn source-scoped identity when a colliding path is added. Please make disambiguation file-local, or explicitly narrow the stability guarantee and add this regression case.

src/agents_shipgate/cli/discovery/source_ids.py
```diff
@@ -0,0 +1,103 @@
...
+    collided = {value for value, count in Counter(ids).items() if count > 1}
+    if not collided:
+        return ids
+    return [
+        f"{value}_{_digest(path)}" if value in collided else value
```
pengfei-threemoonslab: **[P3] Reapply the 64-character cap after disambiguation**

The collision suffix is appended after `source_id_for()` enforces `MAX_SOURCE_ID_LENGTH`. Two near-limit paths that sanitize identically can therefore produce 70-character IDs, contradicting the advertised 64-character bound. Please reserve space for the suffix or truncate the final value, and test the combined collision-plus-length case; long custom source-type prefixes need equivalent coverage.

pengfei-threemoonslab (commented):
Two edge cases remain in the new source-ID helper: collision handling is set-dependent, and collision suffixes can exceed the stated 64-character limit.

src/agents_shipgate/cli/discovery/source_ids.py
pengfei-threemoonslab: Fixed in a16eaa0c — by taking the second option you offered (narrow the guarantee explicitly + regression case), because the first one costs more than it buys. Sanitizing is lossy in two independent ways: character folding (`-`, `.`, space, non-ASCII, case → `_`/lowercase) and separator ambiguity (`a/b_c.py` and `a_b/c.py` both fold to `a_b_c`). Any *file-local* rule has to decide from the path alone whether its own rendering is faithful, and the class of paths that qualify is `[a-z0-9/]` only — which excludes `load_skill/` and `support-tools.openapi.yaml`, i.e. most real repositories. That degrades to a digest on nearly every id, which is the readability the sanitized-path scheme exists to buy.

So the guarantee is now stated exactly, in the module docstring and the changelog: an id is a pure function of its own path *unless another declared path sanitizes to the same string*, and then every member of that group takes a digest — including the file that was already there. The unconditional half (no positional component, so a file appearing earlier in the walk can never renumber the ones after it) is what #307 asked for and still holds. The blast radius is bounded by `init` refusing to overwrite an existing `shipgate.yaml`: ids are assigned once per adoption, not on every run.

`test_source_ids_shift_only_within_the_group_that_sanitizes_alike` pins your exact example in both directions — `a-b/spec.yaml` alone is `openapi_a_b_spec`, adding `a_b/spec.yaml` moves both to digest form, and an unrelated third source is untouched by that rename.

src/agents_shipgate/cli/discovery/source_ids.py
pengfei-threemoonslab: Fixed in a16eaa0c — and reproduced first: `aaaa…/bbbb…/spec-one.yaml` + the `spec_one.yaml` twin both rendered a 58-character plain id, so the appended digest took them to **67**.

Composition and the bound now live in one place (`_compose` → `_with_digest`): a disambiguated id goes through the same truncate-then-digest path as a long one, with the budget computed *with* room for the digest, so the cap holds on the value that ships. The same case is now 46 characters. Your prefix note was the right catch too — a source-type prefix long enough to crowd out the digest is truncated as well, so a 91-character third-party adapter name still lands at ≤ 64 rather than blowing the budget negative and returning `prefix_digest`.

`test_source_id_length_bound_holds_after_disambiguation` covers all three: the near-limit collision, the long custom prefix (alone and colliding), and the deep-path truncation. I also fuzzed ~4.5k random paths (including non-ASCII, dots, spaces, kebab/snake mixes) against both the bound and uniqueness.

One case further out, while I was in here: a file named after the digest another entry is *about* to be given, whose plain id then equals that entry's disambiguated form. Contrived but constructible, and `--minimal` has no validation gate behind it, so the assignment now falls back to giving every entry a digest rather than handing a duplicate to the render (`test_source_ids_stay_unique_when_a_plain_id_matches_a_digest_form`).

src/agents_shipgate/cli/discovery/source_ids.py
```diff
@@ -0,0 +1,143 @@
...
+    ]
+    if len(set(resolved)) == len(resolved):
+        return resolved
+    # A plain id can equal another entry's disambiguated form when a file
+    # is named after that digest (`a_b_spec_<digest of a-b/spec.yaml>`).
+    # Contrived, but constructible — and `--minimal` has no validation
+    # gate behind it, so give every entry a digest rather than let a
+    # duplicate reach the render.
+    return [
+        _compose(source_type, path, disambiguate=True) for source_type, path in entries
```
pengfei-threemoonslab: **[P2] Do not re-key unrelated entries in the fallback**

When a plain ID matches another entry’s disambiguated form, this fallback adds a digest to every entry—not only the conflicting closure. Adding the three paths from `test_source_ids_stay_unique_when_a_plain_id_matches_a_digest_form` alongside `billing/openapi.yaml` changes that unrelated bystander from `openapi_billing_openapi` to `openapi_billing_openapi_2d323534`. That contradicts the newly documented guarantee that IDs shift only within paths sanitizing to the same ID. Please restrict fallback re-keying to conflicting entries, or document and test this additional global-fallback exception.

src/agents_shipgate/cli/discovery/source_ids.py
```diff
@@ -0,0 +1,143 @@
...
+    parts = [*pure.parts[:-1], pure.stem] if pure.parts else []
+    return [segment for segment in (_slug(part) for part in parts) if segment]
+
+
+def _slug(value: str) -> str:
+    return _NON_ID_CHARS.sub("_", value.lower()).strip("_")
+
+
+def _digest(relative_path: str) -> str:
+    return hashlib.sha256(relative_path.encode("utf-8")).hexdigest()[:_DIGEST_LENGTH]
```
pengfei-threemoonslab: **[P2] Eight digest characters cannot guarantee uniqueness**

The final disambiguator is only a 32-bit SHA-256 prefix, and the fallback never verifies that its own output is unique. I found two distinct paths that sanitize identically and share digest `59139f1d`; `assign_source_ids()` returns the same final ID twice. Such a collision is practical to construct by varying `-` and `_` within one sanitized class, and `--minimal` would again write an invalid manifest because it has no validation gate. Please increase digest width adaptively or retry conflicting groups until the final set is unique, then assert uniqueness after the fallback.

pengfei-threemoonslab (commented):
Two fallback edge cases remain: the global re-key path exceeds the newly documented stability exception, and the 8-character digest can still produce duplicate final IDs.

src/agents_shipgate/cli/discovery/source_ids.py
pengfei-threemoonslab: You're right, and the bystander case is a straight contradiction of the guarantee I documented one commit earlier — the fallback re-keyed the whole list instead of the conflicting closure. Fixed in f8281c60.

Assignment is now a fixed point over conflicts rather than one pass plus an escape hatch: each round re-keys **only** the entries whose current id is duplicated, so an entry that never collides keeps the id its own path produced, whatever else the workspace contains. Your exact four-path case now gives:

```
a-b/spec.yaml             -> openapi_a_b_spec_13072c9564b74dda
a_b/spec.yaml             -> openapi_a_b_spec_27eb189a
a_b/spec_13072c95.yaml    -> openapi_a_b_spec_13072c95_b916eeb23a3c7a10
billing/openapi.yaml      -> openapi_billing_openapi          # untouched
```

Note the second entry keeps its 8-hex id too: it was part of the sanitized class but not part of the tie that survived round one, so it is not re-keyed either. `test_source_ids_stay_unique_when_a_plain_id_matches_a_digest_form` now carries the bystander plus that assertion, and the fuzz run asserts the general property (an entry whose id is unique in the set keeps it) across 1500 collision-prone sets.

No documented exception was needed in the end, so the guarantee is back to one rule — ids shift only for entries whose rendered ids collide.

src/agents_shipgate/cli/discovery/source_ids.py
pengfei-threemoonslab: Confirmed and fixed in f8281c60 — I reproduced it independently before touching the code rather than taking the claim on faith. Enumerating the 2^17 `-`/`_` spellings of one path finds a colliding pair in seconds:

```
a_b_c-d_e-f-g_h_i-j_k_l_m-n-o-p_q_r/spec.yaml
a-b_c_d-e_f_g_h-i-j-k_l-m_n_o-p_q-r/spec.yaml   # same slug, both sha256[:8] == bdaf2ee8
```

and `assign_source_ids()` returned `openapi_..._spec_bdaf2ee8` twice. (Different pair from your `59139f1d`, same defect.) That pair is now a regression test instead of a hypothetical, with the search recipe in a comment so it can be regenerated.

The fix does both things you asked for. Entries still tied after a round move to a wider digest (8 → 16 → 32) — only the tied ones, so a wide digest never spreads to entries that were already settled. Anything surviving 128 bits is numbered in sorted-path order, each candidate checked against the ids already taken, so uniqueness of the returned list follows from that check rather than from an assumption about the hash — that is the "assert uniqueness after the fallback" half, expressed as the loop's exit condition instead of a trailing assert. I chose numbering over widening further because at that point the id is no longer explaining anything, and the numbering is keyed to sorted path order, so it is a property of the set and not of the walk.

The length bound holds at every width: the budget is computed from the width actually used, and a prefix long enough to crowd out a 32-hex digest is truncated too. Coverage: the real collision pair, a forced total collision (digest monkeypatched to a constant at every width) for the numbering tier, and a fuzz pass over 5000 single ids and 1500 collision-prone sets checking bound, uniqueness, and that non-colliding entries are never re-keyed.

PR #319, opened 2026-08-06
fix(verify): build effective_policy from the declared manifest, not the run

### LINKED ISSUE #298: ci_mode_weakened cannot fire through the real `verify --base` path
`SHIP-VERIFY-POLICY-WEAKENED`'s `ci_mode_weakened` disjunct — the check that catches an agent downgrading the CI gate from `strict` to `advisory` — never fires in a real run.

**Mechanism** (traced 2026-07-28):

1. `cli/verify/orchestrator.py::_prepare_base_report` scans the base tree with `run_scan(..., ci_mode="advisory", ...)`.
2. `cli/scan/prepare.py:33` applies that CLI override by mutating the loaded manifest: `manifest.ci.mode = ci_mode`.
3. `cli/scan/final_report.py:68` builds the report's `effective_policy` snapshot from `sanitized.manifest.ci.mode`.

So every cached base report records `effective_policy.ci_mode == "advisory"` regardless of what the base manifest declared. `checks/verify_policy.py::_compare` compares that against the head manifest's real mode, so a PR that downgrades `ci.mode` from `strict` to `advisory` yields `head_rank == base_rank` and emits nothing.

**Reproduction.** Fresh git repo with an OpenAI Agents SDK tool → `agents-shipgate init --write` → commit with `ci.mode: strict` → edit the worktree manifest to `ci.mode: advisory` → `agents-shipgate verify --base HEAD`. The base cache report under `.git/agents-shipgate/base-scans/*/report.json` shows `"ci_mode": "advisory"`, and no `ci_mode_weakened` finding appears in `agents-shipgate-reports/report.json`.

**Why coverage missed it.** `tests/test_verify_weakening.py` injects the base `EffectivePolicy` directly instead of driving a real base scan, so the unit tests pass while the end-to-end path is blind.

**Scope to check.** `fail_on_loosened` and `severity_override_lowered` need the same audit — `_prepare_base_report` also passes `fail_on=None`, which may blank `effective_policy.fail_on` the same way.

**Likely fix.** Build the effective-policy snapshot from the manifest as loaded from disk, before CLI overrides are applied (or record the pre-override mode alongside), so the snapshot describes the repository's declared policy rather than the invocation's. Add an end-to-end regression test that drives a real base scan.

**Severity.** A PR touching `shipgate.yaml` still routes to human review via `SHIP-VERIFY-TRUST-ROOT-TOUCHED`, so this is a missed specific finding rather than a silent merge — but the specific finding is the one that names the weakening, and the flagship claim is that an agent cannot quietly weaken its own gate.

Found while working on #297.

pengfei-threemoonslab:
## Summary

- Fixes #298. `effective_policy` is now built from the manifest **as declared on disk**, before `--ci-mode` / `--fail-on` are folded into it, so the snapshot describes the repository rather than the invocation.
- This fixes both directions of one root cause: `ci_mode_weakened` could never fire through the real `verify --base` path, and `fail_on_loosened` fired on PRs that touched no policy at all.
- The run's own gate is deliberately unchanged — `release_decision.fail_policy`, the exit code, and `run_id` still reflect the overrides.

### The defect

`verify` overrides the two sides of the comparison differently:

| | invoked with | snapshot recorded |
|---|---|---|
| base (`orchestrator.py:944`) | `ci_mode="advisory", fail_on=None` | always `advisory`, base's declared `fail_on` |
| head (`orchestrator.py:703`) | the caller's values | the caller's values |

`_prepare_scan` applied those by mutating `manifest.ci`, and `final_report.py` built the snapshot from the mutated manifest.

**Under-fire on `ci_mode`.** Every cached base report recorded `effective_policy.ci_mode == "advisory"` no matter what the base declared, so a PR downgrading `ci.mode` from `strict` to `advisory` yielded `head_rank == base_rank` and emitted nothing — the one case the disjunct exists to catch. Such a PR still routed to a human via `SHIP-VERIFY-TRUST-ROOT-TOUCHED`, so this was a missing name on a real review rather than a silent merge, but the specific finding is the one that says *what* was weakened, and the flagship claim is that an agent cannot quietly weaken its own gate.

**Over-fire on `fail_on`.** Only the head carries `--fail-on`, so `verify --fail-on high` against a manifest declaring `fail_on: [high, critical]` compared a head snapshot of `["high"]` against a base snapshot of `["high", "critical"]` and emitted a **high**-severity *"This PR removes severities from the CI fail-on set"* against a docs-only PR.

### The fix

`_prepare_scan` keeps the declared `ci` block aside before the overrides rewrite it (`_ResolvedManifest.declared_ci`) and threads it to both snapshot builders — the report's block in `final_report.py`, and the check-time head snapshot via the new `ScanContext.declared_ci`. One seam, `build_effective_policy_snapshot(..., declared_ci=)`, defaulting to `manifest.ci` so every caller that loads the manifest straight from disk (preflight, direct `ScanContext` constructions, plugins) is unaffected.

Removing the base-scan override instead would not be correct: `ci_mode="advisory"` there is load-bearing for the base scan's own exit code. The defect is not "the base is forced to advisory", it is "the snapshot describes the invocation".

### Scope audit

Per the follow-up comment on the issue, the original "Scope to check" list narrows to exactly these two axes, and I re-confirmed it against the current tree. `_prepare_scan` mutates only `ci.mode`, `ci.fail_on`, and `output.*`; only the first two appear in `EffectivePolicy`. `severity_override_lowered` has no CLI override path, `fail_on=None` on the base is a genuine no-op (`if fail_on is not None`), and `baseline_mode` cannot diverge (it raises unless `new-findings`). Worth noting that `core/preflight.py:1409` already built this snapshot straight off disk — the scan path was the outlier, not the intended semantics.

## Type

- [ ] Input adapter change
- [ ] Documentation only

## Verification

Additional local checks run:

- Full suite green (exit 0); `ruff check .` and `compileall` clean.
- **The three new end-to-end tests were confirmed to fail on the pre-fix code** before the fix was written — they drive real base and head scans through `run_verify`, covering the under-fire, the over-fire, and a weakening that must still be caught under a `--ci-mode strict` invocation.
- A fourth test is a structural guard: it replays every override `_prepare_scan` supports at once and asserts the snapshot is byte-identical to one built from the untouched on-disk manifest. A future override landing on a snapshot field without carrying its declared value fails there.
- The issue's exact CLI reproduction, end to end. The base cache under `.git/agents-shipgate/base-scans/*/report.json` now records `"ci_mode": "strict"`, and with `--ci-mode strict --fail-on high` layered on top:

  ```
  run gate  : {"ci_mode": "strict", "fail_on": ["high"], "would_fail_ci": true, "exit_code": 20}
  declared  : advisory []
  policy-weakened kinds: ['ci_mode_weakened']
  ```

- The over-fire direction at CLI level: a docs-only PR against a manifest declaring `fail_on: [high, critical]`, run with `--fail-on high`, now emits no `SHIP-VERIFY-POLICY-WEAKENED` finding while the run still gates on `["high"]`.
- The stale-cache regression pins the pre-fix epoch as a literal rather than `BASE_CACHE_KEY_EPOCH - 1`, so it keeps asserting that entries from *that* epoch stay unreachable; verified it fails if the bump is reverted. It carries a positive control — under the epoch that wrote it, the poisoned entry is shown to be believed — so it cannot pass merely because the seeded entry was never readable.
- Sample goldens unaffected — all five samples declare (or default to) `advisory` and already recorded `advisory`, so nothing needed regenerating.

## Release-readiness notes

- [x] New or changed check IDs are documented in `docs/checks.md` — no check IDs added or renamed; `SHIP-VERIFY-POLICY-WEAKENED` and its three evidence kinds are unchanged.

No `report_schema_version` bump: the block's shape is unchanged, and its documented semantics ("derived from the manifest") are now honored rather than altered — the published `docs/report-schema.v0.34.json` is untouched. The invariant is spelled out on both living-doc surfaces (`STABILITY.md`, `docs/agent-contract-current.md`, with `llms-full.txt` regenerated).

### Reviewer notes

- **The base-scan cache-key epoch moves `2` -> `3`** (`BASE_CACHE_KEY_EPOCH`). Fixing the producer alone was not enough, per the review finding below, which I reproduced: a cached base report is admitted on a content hash, which proves it was not tampered with and says nothing about whether its fields still mean what the current CLI expects. `__version__` is in the key but does not move for a source checkout, an editable install, or between two builds sharing one pre-release version string, so a pre-fix entry was reused verbatim and kept reporting `advisory` for a base that declared `strict`. Pre-existing entries now land on a key nothing computes; an upgrade costs one re-scan per base tree and needs no manual clearing.
- **`ScanContext` gained one optional field.** It defaults to `None` and falls back to `manifest.ci`, so plugins and every direct construction in the test suite keep working; `STABILITY.md`'s plugin contract pins the callable signature, not the context's field set.
- **Nothing is double-sanitized.** `CiConfig` is enums and bools only, so carrying it past the sanitization phase cannot leak an unredacted value or perturb `privacy_audit` counts.
- **`policy_snapshot_sha256` in attestations becomes flag-independent.** `cli/attest.py` hashes this block, so two attestations of the same tree now agree regardless of the invocation's `--ci-mode` / `--fail-on`. That is the intended behavior, but it does mean a hash recorded by an older CLI under an override will not match one taken after this change.


pengfei-threemoonslab:
## Review finding — [P1] Invalidate pre-fix base reports

This PR changes the semantics of cached `effective_policy`, but the base-scan cache key still uses epoch `2` plus `__version__`. Both the parent and this PR advertise `0.16.0b7`, so a valid cache written by the pre-fix implementation is reused after updating to this change.

I reproduced this end to end:

1. The pre-fix code cached a base manifest declaring `ci.mode: strict` as `effective_policy.ci_mode: advisory`.
2. Running the same verification with this PR reused that valid cache and still omitted `ci_mode_weakened`.
3. After clearing the cache, this PR correctly recorded `strict` and emitted `ci_mode_weakened` for the `strict -> advisory` change.

That means issue #298 remains observable for same-version/source upgrades despite the implementation fix. Please increment the cache epoch in `_cache_report_path` (for example, `"version": 2` → `3`), or add an equivalent semantic cache version, and add a regression test that pre-seeds a valid stale base report. The changelog’s “no manual cache clearing is needed” statement will then be accurate.

Otherwise, the implementation and regression coverage look sound: the focused suite passed 26/26, Ruff and compilation were clean, all GitHub workflows succeeded, and a fresh-cache Shipgate verification produced a valid receipt with `control.state=complete`, `merge_verdict=mergeable`, and `release_decision.decision=passed`.

pengfei-threemoonslab:
Confirmed and fixed in `8b2d4a99`. You're right, and the changelog sentence you quoted was the actual error in my reasoning — I asserted the version key was sufficient without testing it.

**Reproduced your finding first**, using a worktree at the parent commit so both runs were real builds sharing `0.16.0b7`, against the same repo:

```
STEP 1 — pre-fix code populates the cache
  cached key 8b8d971b56eaba6e57d9ca64 -> effective_policy.ci_mode = advisory
STEP 2 — this PR's code, cache left in place
  cached key 8b8d971b56eaba6e57d9ca64 -> effective_policy.ci_mode = advisory
  POLICY-WEAKENED kinds: (none)
```

The mechanism is that `_cache_report_valid` compares the entry against its own `.sha256` sidecar. That is an integrity check: it proves the file was not tampered with and says nothing about whether its fields still *mean* what the current CLI expects. So a semantic change to a cached field needs the key to move, and `__version__` does not move for a source checkout, an editable install, or between two builds sharing one pre-release version string.

**Fix.** Took your first suggestion — the epoch inside the key moves `2 -> 3` — extracted to a named `BASE_CACHE_KEY_EPOCH` in `cli/verify/orchestrator.py` with the bump discipline written down at the constant, so the next person changing what a cached base report means has the rule in front of them rather than having to infer it from a bare `"version": 2`. Same repo, same stale cache, nothing cleared:

```
    8b8d971b56eaba6e57d9ca64 -> effective_policy.ci_mode = advisory   (epoch 2, now unreachable)
    ea9afd5f972fbf5a09838ee0 -> effective_policy.ci_mode = strict     (epoch 3, fresh scan)
  POLICY-WEAKENED kinds: ['ci_mode_weakened']
```

**On the regression test — two things I'd point out, because the obvious version of it is worthless.**

*It pins the pre-fix epoch as a literal, not `BASE_CACHE_KEY_EPOCH - 1`.* A relative expression only asserts "some previous epoch is unreachable", which is trivially true of any hash key and would keep passing if the bump were reverted. The literal `_PRE_298_CACHE_EPOCH = 2` asserts the thing that actually matters — entries from *that* epoch stay unreachable. Verified both directions: green at epoch 3, and reverting the constant to 2 fails it on exactly the `ci_mode_weakened` assertion.

*It carries a positive control.* The test seeds a valid stale entry (rewrites `ci_mode` to `advisory`, recomputes the sidecar) and then, still under the epoch that wrote it, asserts the poisoned entry **is** believed before asserting that the shipped epoch cannot reach it. Without that step the test would pass just as happily if the seeding were broken and the entry had never been readable at all — which is the failure mode that makes cache-invalidation tests quietly worthless.

Full suite green, Ruff and `compileall` clean.

**One thing I deliberately did not do**, flagging it in case you want it as a follow-up: the key could also carry `report_schema_version`, which would auto-invalidate on schema bumps and cover part of this class mechanically. It would not have caught this one — #298 changes a field's meaning with no schema bump — so the epoch is still the lever that has to be pulled by hand, and adding a second one seemed like it would dilute rather than reinforce that. Happy to add it if you'd rather have the belt and braces.

Issue #338, opened 2026-08-07
Epic: make the issue-to-PR workflow autonomous for coding agents


pengfei-threemoonslab:
> Current release scope (2026-09-12): #504/#337 remain open but are deferred from the first advisory v1.0 under #643. The historical P1 ordering below describes the broader authenticated autonomous workflow. The selected release instead completes named CLI/PR/JSON output and truthful handoff through #684/#662; it does not fabricate human continuation or wait for external adoption.

> Product status (2026-09-04): the launcher, effective diff, compact envelope, publication permissions and current-control foundations below are delivered. **P1 remaining work is #504 → #337:** specify one eligible authenticated human decision, then complete it in GitHub. The first slice runs with #521's pilot and no longer waits for adopted users to exist. Host-session receipts (#293) remain a separate external dependency.

## Executive summary

Shipgate has strong machine-readable contracts, but the end-to-end pull-request workflow still assumes that a human will debug CLI environments, Git topology, verifier artifacts, and state transitions. That assumption no longer matches the primary product reality: users increasingly delegate issue-to-PR work to coding agents and expect to intervene only for genuine semantic or authority decisions.

A real issue #317 follow-up exposed the gap:

1. A coding agent implemented and tested a review fix.
2. Verification combined committed branch changes with overlapping uncommitted follow-ups and emitted seven `boundary_diff_shape_invalid` review findings.
3. `control.must_stop=true` and `allowed_next_commands=[]` prevented the agent from committing or pushing—the operations needed to produce a coherent PR diff and obtain review.
4. The user entered the terminal manually.
5. The promoted `agents-shipgate` launcher failed with `ModuleNotFoundError`, requiring a hand-discovered `PYTHONPATH=src python -m agents_shipgate` invocation.
6. The verifier then reported `execution: succeeded` and exit code 0 while the operational state remained `human_review_required`.
7. The user repeatedly had to ask whether the run was complete and whether the coding agent could update the PR.
8. After a manual commit and committed-ref verification, the result finally became `control.state=complete`.

The safety model worked fail-closed, but the product required the user to understand implementation details unrelated to the actual risk decision.

## North-star product principle

> **Human review should block merge and completion, not publication of review evidence.**

A user should be able to tell a coding agent “fix this issue and update the PR.” Shipgate should autonomously handle worktree verification, artifact identity, and PR evidence publication. The human should enter only when a semantic, policy, or authority decision genuinely requires judgment.

[...]

Issue #337, opened 2026-08-07
Add a GitHub-native human review and authorization flow


pengfei-threemoonslab:
## First advisory release scope — 2026-09-11

Owner decision: **adoption > completeness; make v1.0 ready with repository-independent work, then expand adoption.** Real GitHub-authenticated human continuation remains open. Selected v1 fixtures prove truthful routing and limitations, not an independent human decision.

Removed from the selected v1.0 milestone; kept open and deferred, not resolved. #643/#644/#645/#646 define current product acceptance; #572 preserves the historical qualified no-go and #648 owns the actual publication contract. The prior evidence and acceptance below remain a record of the deferred work, not newly satisfied conditions.

---

Parent: #504, within the #338 workflow epic. Consumes #536's human-decision contract and #537's bound decision evaluator. #333, #335 and #339 are completed prerequisites. #293 remains separately blocked on host attestation.

## Problem

A reviewer receives a human-routed verdict but cannot record an authenticated decision against the reviewed bytes and then follow an exact continuation. The #503/#505 workflow in #504 demonstrates the gap in this repository.

The minimum GitHub pilot no longer waits for adopted users to exist: that would defer the mechanism needed to complete their first review. Broad organization governance and host-session authorization remain outside this slice.

## Proposed direction

Reuse the existing PR check/comment publication surfaces. Put the **named capability change, openable evidence, coverage limit, required decision and review location** in the PR. A reviewer should not need a job log or downloaded artifact archive to learn what is being asked.

Start with the eligible `review_required` request class specified by #504. Bind:

- repository and verification request ID;
- head commit/tree and reviewed change;
- policy snapshot, decision identity and full review-set identity;
- authenticated eligible reviewer and their decision;
- the exact verifier continuation and the status it actually produces.

The coding agent may publish evidence when existing permissions allow it, but cannot create, forge or reuse human authorization. A plain GitHub approval is insufficient unless its identity, eligibility, scope and freshness can be established with credentials isolated from the coding agent.

An acceptance/dispute record for blocked, critical, gate-governing, incomplete-input or gate-weakening results is audit evidence, not automatic clearance. The existing push-only authorization overlay does not become merge authority through this issue. #504 must specify the postcondition and authority boundary first.

## Acceptance criteria

- [ ] One eligible review-required case completes question → authenticated decision → exact continuation on the existing GitHub surface; the remaining obligations and status are visible.
[...]

PR #346, opened 2026-08-08
feat(control): separate permission to publish a PR from permission to merge

### LINKED ISSUE #335: [Agent-first UX][P0] Separate permission to publish a PR from permission to merge
> Parent epic: #338

## Problem

The current control model treats human review as a universal stop:

```json
{
  "state": "human_review_required",
  "must_stop": true,
  "completion_allowed": false,
  "allowed_next_commands": []
}
```

For an agent working on a pull request, this blocks commit, push, and PR updates—the exact actions often required to publish evidence and obtain human review. It conflates two materially different decisions:

1. May the agent publish the change for review?
2. May the change be merged or reported complete?

This creates a circular workflow: human review is required, but the agent cannot publish the reviewable state.

## Product principle

**Human review should block merge and completion, not publication of review evidence.**

## Proposed direction

Expose action-specific capabilities rather than projecting all control through one `must_stop` boolean:

```json
{
  "permissions": {
    "edit": true,
    "commit": true,
    "push": true,
    "update_pr": true,
    "merge": false,
    "report_complete": false
  }
}
```

Optionally add a state such as `review_publishable` for “agent may publish/update the PR, but a human must review before merge.”

`must_stop` can remain as a backward-compatible projection, but it must not erase safe progress actions.

## Acceptance criteria

- [ ] Control independently represents edit, commit, push, PR-update, merge, and completion permissions.
- [ ] A normal `human_review_required` result can authorize commit/push/PR update while denying merge/completion.
- [ ] Truly unsafe or untrusted input can still deny publication when policy requires it.
- [ ] `allowed_next_commands` includes the safe route to publish evidence when publication is permitted.
- [ ] Generated agent instructions explain that “update PR” is not equivalent to “merge.”
- [ ] Existing consumers receive a documented compatibility projection or schema-version migration.
- [ ] Tests cover coding-agent repair, review publication, human approval, and final merge authorization as distinct transitions.

## Example target behavior

```json
{
  "state": "review_publishable",
  "next_actor": "coding_agent",
  "permissions": {
    "commit": true,
    "push": true,
    "update_pr": true,
    "merge": false,
    "report_complete": false
  },
  "human_review": {
    "required": true,
    "surface": "github_pull_request"
  }
}
```

pengfei-threemoonslab:
## Summary

Closes #335 (epic #338, P0).

- A human route was one universal stop: `control.state: "human_review_required"` with `must_stop: true` and `allowed_next_commands: []`. For an agent working on a pull request that denied commit, push, and PR updates — the exact actions needed to produce the evidence a human was being asked to review. The gate was correct and the workflow was circular.
- **Human review now gates merge and completion, not publication of review evidence.** Two additive changes to the `AgentControl` union carry that.
- `control.permissions` — a required object on every state with the exact booleans `edit`, `commit`, `push`, `update_pr`, `merge`, `report_complete`. It is a discriminated variant fixed by the state (`FullAgentPermissions` / `PublishOnlyPermissions` / `NoAgentPermissions`, every field a `Literal`), never six independent switches. A base validator enforces `merge == report_complete == completion_allowed` and `must_stop == not permissions.publishes` — that last invariant *is* the structural statement of "`must_stop` must not erase safe progress actions", and it keeps `must_stop` meaning exactly what it always meant to existing readers.
- `control.state: "review_publishable"` — a fourth variant meaning "a human must approve the merge, and the agent may still publish the change for that review". Human `next_action`, `must_stop: false`, `stop_reason: null`, and at most the one exact rerun command that regenerates the same evidence against the committed refs (pinned by a verifier validator to `fix_task.verification_command`, so a review route can never authorize an arbitrary command).
- `human_review_required` keeps its exact old meaning and is now reserved for results Shipgate cannot vouch for: a blocked release decision, a `block` boundary decision, a run whose execution failed, unreadable or unbindable diff input, an undeclared surface with no discovery route, preflight protected-surface touches, and MCP audit blocks.
- Runtime contract `19 → 20`; `minimum_control_contract_version` `14 → 20`. No CLI surface added or changed.

## The routing judgement (the part worth reviewing closely)

The discriminator between the two human routes is **epistemic state, not severity** — new `boundary_assessment_is_evidence_backed()` in `core/codex_boundary.py`. It is a strictly weaker question than the existing `violations_within_agent_actionable_band()`, which answers a *different* question ("may the agent finish this turn locally?"). Do not conflate them.

- **Denies publication:** `BOUNDARY-INPUT-INCOMPLETE`, parse-failure rows whose evidence kind is not in the parseable allow-list, experimental adapter paths, non-replayable subjects (detached `--diff`, unbindable ref range), `block` decisions, failed execution. In all of these there is no trustworthy assessment to publish.
- **Permits publication:** an evaluated `require_review`, *including* critical risk and gate-governing trust roots (manifest, policy, ci_gate, shipgate_state).

That last point is the deliberate call, and reviewers should push back if they disagree. The reasoning: publication authority never confers merge authority — `merge` and `report_complete` are false on every state but `complete`, and PR-time `verify` re-judges the committed diff. Denying publication on a trust-root touch does not stop the agent from making the edit; it only stops a human from *seeing* it, and it reproduces the #298-era onboarding self-lock where `init --write`'s own `shipgate.yaml` trips `TRUST-ROOT-TOUCHED` and the agent cannot commit its own adoption.

## Compatibility

- Consumers switching on `control.state` **must** add a `review_publishable` branch and keep failing closed on unrecognized states. The installed Claude Code Stop hook does: it ends the turn, states that commit/push/PR-update remain authorized, and names the rerun command.
- Consumers reading only `must_stop` and `completion_allowed` need no change and lose no safety. `completion_allowed` is still false; `must_stop: false` means what it always said.
- Legacy artifacts are unaffected. A payload without `control`, or any pre-v20 payload, normalizes to `human_review_required` — `review_publishable` is only ever produced by an emitter that asserted the publication fact.
- **The frozen `shipgate.codex_boundary_result/v2` projection stays byte-frozen.** Its published schema is `additionalProperties: false` with a 3-state discriminator, and `CodexBoundaryResultV2` shares the same `control` model as the current format — so any union change breaks it. Fixed with a down-projection at the serialization boundary (`project_legacy_agent_control` + a `schema_version` branch in `agent_result_json_payload`): `permissions` is dropped and `review_publishable` collapses to `human_review_required` with `must_stop: true`. `cli/mcp.py` now reuses the same serializer so the projection cannot be bypassed by a second emit path. A pydantic model serializer would have been wrong here — `build_agent_boundary_result` copies fields via `legacy.model_dump()`, so it would leak the downgraded control into the current format.
- `agents-shipgate contract --json` gains `agent_control_permissions[]`. `harness/adoption/scorer/rules.py` learns the new state so `respects_must_stop` does not score the newly-intended behaviour as a blocker.

## Type

- [ ] Input adapter change
- [ ] Documentation only

## Verification

Additional local checks run:

- Full suite green (exit 0, zero failures), plus `ruff`, `compileall`, `scripts/generate_schemas.py --check`, and all four CI safety-canary gates (`test_adapter_static_only`, `test_p0_safety_canaries`, `test_p0_binding_canaries`, `test_agent_boundary`).
- End-to-end on real samples: `simple_crewai_agent` (`review_required`) → `review_publishable`, commit/push/update_pr true, merge/report_complete false, exact rerun command offered. `ai_generated_refund_pr` (`blocked`) → `human_review_required`, every permission false, no commands. A detached `--diff` still refuses publication. `--format codex-boundary-json` on the same worktree change returns the unchanged 3-state shape with no `permissions` key.
- New tests cover the four transitions as distinct states (repair / review publication / human approval / final merge authorization), every `review_publishable` contradiction, the `must_stop ⇔ permissions` invariant across all four states, the legacy-normalization floor, the frozen down-projection, the Stop-hook message, and a drift pin between `AGENT_CONTROL_FIELDS`/`AGENT_CONTROL_PERMISSIONS` and the model's actual fields.
- 26 assertions in `tests/test_agent_boundary.py` flip from `human_review_required` to `review_publishable`. Several are trust-root cases (manifest edits, instruction weakening, baseline/waiver touches) — that is the intended semantic change described above, and it is the part of this diff most worth a second opinion.

## Release-readiness notes

- [x] New or changed check IDs are documented in `docs/checks.md` — no check IDs changed
- [x] Report/schema changes are additive or documented in `STABILITY.md` — new "unreleased — publish authority is not merge authority" migration note; `report_schema_version`, `verifier_schema_version`, packet, handoff, and preflight versions are all unchanged

## Not done (deliberately)

The issue's example shows `human_review.surface: "github_pull_request"`, marked "Optionally". Left out: Shipgate does not know whether a pull request exists, so populating it would be fabrication, and the GitHub-native review surface is #337's scope.


src/agents_shipgate/core/codex_boundary.py
```diff
@@ -952,6 +990,23 @@ def _control_for_result(
...
+        if decision == "require_review" and boundary_assessment_is_evidence_backed(violations):
+            # The boundary was evaluated against a replayable subject and the
+            # only obligation left is human judgement, so the agent keeps the
+            # authority to put a reviewable change in front of that human.
+            return derive_agent_control(
+                reason=summary,
+                next_action=HumanControlAction(kind="review", why=why),
+                verify_required=verify_required,
+                human_review_required=True,
+                publication_allowed=True,
```
pengfei-threemoonslab, line 1011: **[P1] Reject publication for detached diffs.** This branch says the subject is replayable but never checks `verification_replayable`. The MCP entrypoint always marks caller-supplied diff text unreplayable; a detached `.codex/config.toml` change adding `sandbox_mode = "danger-full-access"` therefore returns `review_publishable` with edit/commit/push/update-PR authority and an unusable verify command. Require replayability here, or apply the unbound-subject stop before every human-review route.

src/agents_shipgate/cli/mcp.py
```diff
@@ -627,22 +628,29 @@ def _agent_result_from_audit(audit: dict[str, Any]) -> AgentResultV2:
...
-        else derive_agent_control(reason=summary)
-    )
+    elif decision == "require_review":
+        # The audit completed; only human judgement is outstanding, so the
+        # agent may still publish the change for that review.
+        control = derive_agent_control(
+            reason=summary,
+            next_action=HumanControlAction(kind="review", why=human_review.why or summary),
+            human_review_required=True,
+            publication_allowed=True,
```
pengfei-threemoonslab, line 676: **[P1] Do not publish partial MCP assessments.** Every `require_review` result becomes publishable even if another MCP file generated `mcp_json_parse_failed`. A diff combining a high-risk permission expansion with malformed `.mcp.json` content returns `review_publishable` despite partially unreadable input. Gate publication on complete parsing/resolution, not only the aggregate decision.

src/agents_shipgate/schemas/verifier.py
```diff
@@ -888,9 +889,22 @@ def _control_projects_gate(self) -> VerifierArtifact:
...
                     "an evidence-backed repair task"
                 )
-        elif self.control.state == "human_review_required" and self.fix_task is not None:
-            if self.fix_task.actor != "human" or self.fix_task.safe_to_attempt:
+        elif self.control.state in {"human_review_required", "review_publishable"}:
+            if self.fix_task is not None and (
+                self.fix_task.actor != "human" or self.fix_task.safe_to_attempt
+            ):
                 raise ValueError("human control requires a human-owned, non-safe fix task")
+            if self.control.state == "review_publishable":
```
pengfei-threemoonslab, line 936: **[P1] Bind publication to a successful, non-blocked gate.** This validates command shape but never requires `execution == succeeded`, a present release decision, or a non-blocked decision. Mutating the real verifier artifact to `execution=failed`, no decision, failed applicability, and `review_publishable` passed both `VerifierArtifact.model_validate` and `docs/verifier-schema.v0.7.json` with all publication permissions intact. Add equivalent container-level invariants to verifier, handoff, verify-run, and their JSON Schemas.

src/agents_shipgate/schemas/agent_control.py
```diff
@@ -199,6 +302,7 @@ class AgentActionRequiredControl(_AgentControlBase):
     verify_required: bool = False
     next_action: CodingAgentAction
     allowed_next_commands: list[ExactCommand] = Field(default_factory=list)
+    permissions: PublishOnlyPermissions = Field(default_factory=PublishOnlyPermissions)
```
pengfei-threemoonslab: **[P1] Scope permissions to the exact agent action.** This grants edit/commit/push/update-PR authority to every `agent_action_required` route. A missing-base `fetch_base` action consequently advertises full publication rights before any diff was evaluated, despite the contract saying to perform only `next_action`. The permission model needs action-specific execute/fetch authority, or `must_stop` must be decoupled from publication.

src/agents_shipgate/schemas/agent_control.py
```diff
@@ -226,21 +330,66 @@ def _action_is_an_allowed_route(self) -> AgentActionRequiredControl:
...
+    has no exact command to offer.
+    """
+
+    state: Literal["review_publishable"]
+    completion_allowed: Literal[False] = False
+    must_stop: Literal[False] = False
+    verify_required: bool = False
+    next_action: HumanControlAction
+    allowed_next_commands: list[ExactCommand] = Field(default_factory=list)
+    permissions: PublishOnlyPermissions = Field(default_factory=PublishOnlyPermissions)
```
pengfei-threemoonslab: **[P1] Do not synthesize omitted publication authority.** Because this field has a default factory, a `review_publishable` payload that omits the supposedly required permissions object silently gains edit/commit/push/update-PR permission. The Stop hook likewise trusts the state without checking the vector. Since this state cannot be legacy input, omission should fail closed in Pydantic and consumers.

src/agents_shipgate/cli/agent_result.py
```diff
@@ -658,7 +662,15 @@ def _bind_worktree_config_to_head(
 
 def agent_result_json_payload(result: AgentResultV2) -> dict[str, Any]:
     payload = result.model_dump(mode="json", exclude_none=True)
-    payload["control"] = result.control.model_dump(mode="json")
+    # ``shipgate.codex_boundary_result/v2`` is a frozen deprecated projection:
+    # its published schema forbids unknown properties and knows only the three
+    # pre-contract-20 states. Emit the legacy control shape there so the
+    # promise holds, exactly as ``pending_review[]`` stays off this format.
+    payload["control"] = (
+        project_legacy_agent_control(result.control)
```
pengfei-threemoonslab: **[P1] Freeze the whole v2 contract, including identity.** This helper hides the new shape only on selected serialization paths; `CodexBoundaryResultV2.model_dump_json()` still exposes `review_publishable` plus `permissions` and fails the frozen v2 schema. In addition, `audit_id` is computed from the new control before this projection, so the PR rotates two frozen IDs while every other emitted byte is unchanged. Use a genuinely frozen model/projection and hash that legacy projection for its identity.

pengfei-threemoonslab (commented):
Engineering review: **changes recommended — hold the merge pending fixes**.

The publish-vs-merge split is the right direction, but the current implementation still grants publication authority without a trustworthy, replayable assessment in multiple paths. The blocking findings are attached inline.

Additional compatibility and coverage findings:

1. **[P2] Existing schema identifiers are reused for breaking shapes.** Pre-v20 payloads become invalid under the same verifier 0.7, handoff v6, verify-run v3, and local-contract v7 identifiers. Please bump and freeze the affected schemas, or keep those versions backward-valid.
2. **[P2] Generated JSON Schemas omit two documented invariants.** They accept `review_publishable` with `next_action.kind=stop`, and the shared control schema accepts multiple rerun commands despite the at-most-one-exact-command contract. A review-only action subtype plus `max_length=1` would keep Pydantic and JSON Schema aligned.
3. **[P2] The adoption scorer ignores the new permission vector.** It treats `review_publishable` as generally non-stopping, so a subsequent merge operation is not rejected even though `permissions.merge=false`. Please parse permissions and score each observed action against them.
4. **[P2] The migration instructions are internally inconsistent.** `prompts/add-shipgate-to-repo.md` references runtime contracts 15, 14, and 20 while pinning released 0.15.0; `decide-shipgate-relevance.md` asks `--version` to report a field available from `contract --json`; and only the Cursor block in `docs/target-repo-agent-snippets.md` documents `review_publishable`.

Verification performed against head `f0bfe3de8c0f91bcf283dda76cb50b5b51a7ff6f`: GitHub checks are green; focused control, verifier, boundary, MCP, hook, schema, harness, and packaging suites passed; Ruff, compileall, schema drift, and verification-receipt reproduction passed. The exact-ref verifier still denies merge (`review_required`, `can_merge_without_human=false`), and the local boundary check returns `human_review_required` for the protected/undeclared surfaces in this PR. The negative cases described inline are not covered by the green suite.

pengfei-threemoonslab:
All six P1 findings reproduced before I touched anything, and all six are fixed in `586e4c0a`. The four P2s are addressed too. Thank you — the `fetch_base` observation in particular caught a place where I had applied my own stated principle to the human routes and then failed to apply it to the agent routes.

## P1 findings

**[P1-1] Detached diffs no longer publish.** Reproduced through the MCP entrypoint: a detached `.codex/config.toml` adding `sandbox_mode = "danger-full-access"` returned `review_publishable` with full publish authority and an unusable rerun command. You had the mechanism exactly right — the unbound-subject guard above only fires for routes that owe a command, and a non-graded `require_review` owes none, so reaching that branch never implied replayability. It now requires `verification_replayable`; the detached case returns `human_review_required` with no permissions and no commands.

**[P1-2] Partial MCP audits no longer publish.** Publication now requires that every audited surface and any explicit policy actually parsed. I drew the line at *unread input* — `mcp_toml_parse_failed`, `mcp_json_parse_failed`, `mcp_policy_parse_failed` — and deliberately left `mcp_policy_missing` (no explicit policy is the ordinary zero-config path) and the field-level "ignoring invalid X" codes eligible, since those leave a fully-read policy with every other rule applied. Happy to widen it if you read that boundary differently.

**[P1-3] Publication is bound to the substrate.** Confirmed: `execution=failed` + no decision + `review_publishable` passed both Pydantic and the v0.7 schema with every permission intact. Verifier, handoff, and verify-run now each reject it unless execution succeeded and a non-blocked release decision exists, in Pydantic *and* in generated JSON Schema.

**[P1-4] Permissions follow the route.** This was the sharpest finding. `fetch_base` and `install` run before any diff was read, so they were advertising publication of a change Shipgate had never seen — contradicting both "perform only `next_action`" and my own rule that publication requires an evaluated subject. They now carry no permissions at all. That forced the invariant to change: `must_stop == not publishes` becomes the fail-closed implication `must_stop ⇒ nothing authorized`, so `must_stop=false` is no longer a promise that publication is allowed. Consumers read `permissions`.

**[P1-5] An omitted vector fails closed where it can.** `permissions` is now required on `review_publishable` only. That state cannot be legacy input, so omission there is a malformed *current* payload silently granting publication. The three legacy-reachable states keep it optional and reconstruct it from the route rather than defaulting to publish. The Stop hook reads the vector rather than trusting the state.

**[P1-6] The freeze holds on every serialization path.** `model_dump_json()` did leak the new shape. It is now a model serializer on `CodexBoundaryResultV2`, with the two internal copy sites passing the real control explicitly so the downgrade cannot leak into the current format. I avoided the obvious mistake there: `build_agent_boundary_result` copies fields via `legacy.model_dump()`, so an unconditional serializer without those two explicit overrides would have downgraded the format that is *supposed* to carry the new state.

## The one place I landed differently

On the second half of P1-6 — "hash that legacy projection for its identity" — I fixed the symptom you identified but not by that mechanism, because hashing the projection would collapse a distinction the current format needs: `test_codex_audit_id_distinguishes_detached_control_from_replayable` exists precisely to pin that two different control states are two different audit rows, and the id is shared between the frozen v2 and the current v1 outputs.

The actual cause turned out to be narrower. `permissions` was entering the hashed control even though it is a pure projection of `state` and `next_action.kind`, both already hashed — so it added no distinguishing information and only churned. Excluding it, plus the P1-1 fix (which returns the corpus cases to `human_review_required`, since they all run through `--diff`), means **every golden in `tests/golden/` and `examples/agent-protocol/` is now byte-identical to `main`.** No frozen id rotates. If you would still rather see the projection hashed, say so and I will do it — but I would want to replace that test's intent first rather than weaken it silently.

## P2 findings

1. **Schema identifiers.** Resolved as a side effect of P1-5 rather than by bumping four versions: `permissions` is required only on the state that cannot be legacy input, so a pre-v20 payload still validates against verifier `0.7`, handoff `v6`, verify-run `v3`, agent-result `v2`, agent-boundary `v1`, and preflight `0.3` — verified directly. No identifier changes meaning, and no frozen prior-version file was touched. That seemed better than minting four schema versions whose only difference is a required field the emitter always writes.
2. **JSON-Schema invariants.** Both now hold outside Pydantic: a `HumanReviewAction` subtype pins `kind: "review"` on `review_publishable`, and `allowed_next_commands` is capped with `max_length=1`.
3. **Adoption scorer.** It now parses the vector and scores each observed operation against it, so `gh pr merge` under `review_publishable` is a blocker while `git commit` / `git push` are not. Legacy payloads with no vector reconstruct one fail-closed from the state. New test covers both directions.
4. **Migration instructions.** `add-shipgate-to-repo.md` now states one floor consistently (contract 20, with the 0.16.0 pin it implies) instead of 15/14/20 across three lines; `decide-shipgate-relevance.md` reads the floor from `contract --json` and says explicitly that `--version` reports the CLI version, not the contract floor; and the `AGENTS.md` and `CLAUDE.md` snippets in `docs/target-repo-agent-snippets.md` now document `review_publishable` alongside the Cursor block.

Worth flagging separately: the version pin in those prompts is `0.15.0`, which reports `contract_version: 10`, so it cannot satisfy any floor above 10. That mismatch predates this PR (the file already demanded 14 and 15), and this change widens it rather than causing it. It resolves when 0.16.0 ships; I did not want to paper over it by lowering the floor to something the flow does not actually work on.

## Verification

Full suite green (0 failures), `ruff`, `compileall`, `generate_schemas.py --check`, and all four CI safety-canary gates pass. Each P1 has a named regression test: detached-diff denial in `test_agent_boundary.py`, partial-audit denial in `test_mcp_audit.py`, substrate binding in `test_verifier_control_contract.py`, route-scoped permissions and the required-vector asymmetry in `test_agent_control_contract.py`, frozen serialization across all three dump paths in `test_codex_boundary_check.py`, and permission-scored actions in `tests/harness/test_detectors.py`.

src/agents_shipgate/schemas/agent_control.py
```diff
@@ -36,6 +46,10 @@
     "rerun",
 ]
 HumanActionKind = Literal["review", "stop"]
+# Coding-agent routes that run *before* Shipgate has read a diff at all: the
+# requested refs are unavailable, or the gate binary itself is missing. They
+# authorize their own ``next_action`` and nothing else.
+UNEVALUATED_ACTION_KINDS: frozenset[str] = frozenset({"fetch_base", "install"})
```
pengfei-threemoonslab: **[P1] Do not infer subject evaluation from the action kind.** Missing-manifest verify stops before reading any diff and emits a `configure` action. Because `configure` is absent here, the resulting artifact has `execution=failed`, `diff_status=unavailable/not_attempted`, but edit/commit/push/update-PR are all true. Pass an explicit evaluated-subject/publication fact into derivation instead of maintaining a partial kind allowlist, and assert all publication permissions are false in the existing missing-manifest scenario.

src/agents_shipgate/schemas/agent_control.py
```diff
@@ -199,9 +333,59 @@ class AgentActionRequiredControl(_AgentControlBase):
...
+        return {
+            **data,
+            "permissions": (
+                NoAgentPermissions()
+                if kind in UNEVALUATED_ACTION_KINDS
+                else PublishOnlyPermissions()
+            ),
+        }
+
+    @model_validator(mode="after")
```
pengfei-threemoonslab: **[P1] Encode this route-dependent permission rule in JSON Schema.** Pydantic rejects `fetch_base` with publish permissions, but `AGENT_CONTROL_ADAPTER.json_schema()` and every enclosing published schema accept it with zero errors; the reverse contradiction (`verify` with all progress permissions false) is also schema-valid. Split the action-required variant by evaluated/unevaluated action kinds or add equivalent schema conditionals.

src/agents_shipgate/cli/mcp.py
```diff
@@ -614,6 +615,34 @@ def _violation(
     }
 
 
+# Diagnostics that mean the audit did not actually read what it was asked to
+# judge: an audited MCP surface that would not parse, or an explicit policy the
+# run could not load at all and silently replaced with packaged defaults.
+# ``mcp_policy_missing`` is deliberately absent — no explicit policy is the
```
pengfei-threemoonslab: **[P1] Treat an explicitly missing policy and unresolved changed source as unread input.** The ordinary zero-config path returns earlier for `policy=None`/the packaged default; `mcp_policy_missing` is emitted only for a requested non-default path that does not exist. With a high-risk diff that warning currently yields builtin fallback plus `review_publishable`. Warning-level unresolved `content_source` rows do the same. Prefer a first-class completeness result over a diagnostic-code denylist.

src/agents_shipgate/schemas/verifier.py
```diff
@@ -888,9 +911,36 @@ def _control_projects_gate(self) -> VerifierArtifact:
...
```
pengfei-threemoonslab, line 936: **[P1] Require a completely read diff before authorizing publication.** Starting from a valid publishable verifier and changing only `diff_status` to `completeness="partial", reason="body_limit_exceeded"` is accepted by both `VerifierArtifact.model_validate` and `docs/verifier-schema.v0.7.json` (zero schema errors). The JSON conditional also constrains `release_decision` only to “object”, so a root review decision with nested `release_decision.decision="blocked"` remains schema-valid. Require `diff_status.completeness="complete"` and the nested non-blocked decision in both validators.

pengfei-threemoonslab (commented):
Second-pass engineering review against updated head `586e4c0a4796f428fc8b1a246e744bbfcee75bb5`.

The update fixes several original repros, but I still would not approve this change. The inline comments identify the remaining control and consumer gaps. Four additional cross-cutting issues do not fit cleanly on a single changed line:

1. **[P1] The versioned wire schemas still change meaning without changing identifiers.** A newly emitted ordinary `complete` verifier still declares `verifier_schema_version: "0.7"` but fails the schema published at the PR base under that same identifier because the old control union forbids `permissions`. The same producer-to-existing-consumer break affects agent-result v2, agent-boundary v1, verify-run v3, handoff v6, and preflight 0.3. Making old payloads validate against the new schema covers only the opposite compatibility direction; bump and freeze these artifacts.

2. **[P1] The generic containers do not bind publication to their substrate.** `AgentResultV2` accepts `decision="block"` with `review_publishable`; `AgentBoundaryResultV1` accepts both `decision="block"` and `input_coverage="partial"` with edit/commit/push/update-PR authority. Pydantic and the checked-in schemas accept those contradictions. Apply the complete/non-blocked publication invariant to these wrappers as well.

3. **[P2] Frozen VerifyRun v2 is no longer serialization-stable.** It reuses the changed current outcome model, so reading and re-dumping a valid old v2 payload adds `control.permissions` and fails the unchanged frozen v2 schema.

4. **[P2] Contract-floor documentation still drifts.** The Codex and Claude Code guides and protocol still name floor 15, while `docs/passed-verdict-contract.md` calls the current runtime contract v19. The implementation and this PR require contract 20.

Resolved from the previous pass: detached supplied diffs now stop; the exact mixed malformed-MCP case stops; failed/blocked verifier publication is rejected; runtime `fetch_base` permissions fail closed; omitted `review_publishable.permissions` is rejected by Pydantic; frozen audit IDs are restored; and the review-only-action/one-command constraints are encoded.

Validation: all three workflows on this head are green. The full local run reached 4,408 passed / 5 skipped; its eight sandboxed isolated-build setup errors passed 8/8 when dependency fetching was permitted. Ruff and generated-schema drift checks passed. These negative-contract repros are not covered by the green suite. The exact-ref verifier still denies merge (`review_publishable`, `merge_verdict=human_review_required`), and the local boundary result is `human_review_required`.

src/agents_shipgate/cli/install_hooks.py
```diff
@@ -1024,6 +1026,36 @@ def _route_verify_result(
             "Do not bypass the verifier by suppressing findings, lowering severity, "
             "expanding baselines/waivers, removing Shipgate CI, or weakening agent instructions."
         )
+    if state == "review_publishable":
```
pengfei-threemoonslab, line 1046: **[P1] Validate the explicit permission vector before announcing publication.** This raw-JSON consumer still trusts `state` alone: a `review_publishable` payload with no `permissions` writes the verified signature and tells the agent that commit/push/update-PR are authorized. The current hook test fixture demonstrates exactly that malformed input. Validate the verifier artifact, or require the exact publish-only six-boolean vector before caching or emitting this message.

src/agents_shipgate/schemas/codex_boundary_result.py
```diff
@@ -67,12 +70,31 @@ class CodexBoundaryResultV1(AgentResultV1):
...
+    ``shipgate.agent_boundary_result/v1`` pass ``control`` explicitly instead of
+    taking it from the dump, so the downgrade cannot leak into the format that
+    is supposed to carry the new state.
+    """
 
     schema_version: Literal["shipgate.codex_boundary_result/v2"] = (
         CODEX_BOUNDARY_RESULT_SCHEMA_VERSION
     )
 
+    @model_serializer(mode="wrap")
```
pengfei-threemoonslab: **[P1] Freeze base-typed serialization and schema generation too.** Direct concrete dumps now down-project correctly, but `TypeAdapter(AgentResultV2).dump_python(result, mode="json")` and a model field typed as `AgentResultV2` bypass this subclass serializer and emit the fourth state plus `permissions`. `CodexBoundaryResultV2.model_json_schema()` likewise advertises the current union, while serialization-mode schema generation collapses to `{}` because the wrapper returns `Any`. A genuinely frozen v2 control/model is needed at the public type boundary.

prompts/add-shipgate-to-repo.md
```diff
@@ -9,7 +9,7 @@ agent-related PRs should use `agents-shipgate verify` after this adoption step.
 
 ## Your task
 
-1. **Install the tool - pin the version so a stale build can't shadow it.** This flow uses the neutral multi-host boundary contract and requires **runtime contract 15**; an older copy lingering on `PATH` may lack the command or schema fields this prompt expects. Prefer a **pinned, zero-install** runner that fetches the exact version every time instead of trusting whatever is already on `PATH`. **Pin it into one variable and use that for every step below**, so no single command can fall through to a stale binary:
+1. **Install the tool - pin the version so a stale build can't shadow it.** This flow uses the permission-scoped multi-host boundary contract and requires **runtime contract 20** (`agents-shipgate` 0.16.0 or newer); an older copy lingering on `PATH` may lack the command or schema fields this prompt expects. Prefer a **pinned, zero-install** runner that fetches the exact version every time instead of trusting whatever is already on `PATH`. **Pin it into one variable and use that for every step below**, so no single command can fall through to a stale binary:
```
pengfei-threemoonslab: **[P1] The pinned runner cannot satisfy the required contract.** The next lines pin `agents-shipgate@0.15.0`; the actual v0.15.0 tag declares runtime contract 10, while this prompt requires contract 20. The same contradiction appears in the relevance prompt and packaged mirrors, and the public-surface test currently enforces the stale 0.15.0 pin. Reconcile the release/pin policy or defer exposing the contract-20 prompt until a compatible published version exists.

harness/adoption/scorer/rules.py
```diff
@@ -667,6 +679,8 @@ def _control_snapshot(payload: dict) -> _ControlSnapshot | None:
     else:
         state = "agent_action_required"
 
+    permissions = _permissions_for(control.get("permissions"), state=state, must_stop=must_stop)
```
pengfei-threemoonslab, line 682: **[P1] Legacy permission reconstruction must include `next_action.kind`.** This call passes only state and `must_stop`, so every permissions-omitted legacy `agent_action_required` artifact reconstructs edit/commit/push/update-PR as true—including `fetch_base` and `install`, where production now intentionally grants none. Pass the route into the mapper and mirror the core fail-closed reconstruction.

harness/adoption/scorer/rules.py
```diff
@@ -691,10 +705,67 @@ def _control_snapshot(payload: dict) -> _ControlSnapshot | None:
         allowed_next_commands=tuple(allowed),
         human_review_required=human_required,
         source_schema=schema,
+        permissions=permissions,
         pending_review=_pending_review_tokens(payload.get("pending_review")),
     )
 
 
+_PERMISSION_NAMES = ("edit", "commit", "push", "update_pr", "merge", "report_complete")
```
pengfei-threemoonslab, line 713: **[P1] Enforce denied edit authority for non-shell operations.** There is no `edit` operation here, and the timeline parser reduces tool actions to an optional command. Codex `file_change` and editor tool calls therefore have no command and bypass `permissions.edit=false`; an end-to-end fetch-base → file-change transcript passes the required-action detector and receives N/A from the permission detector. Preserve operation kind/tool metadata and score it against the vector.

harness/adoption/scorer/rules.py
```diff
@@ -1018,17 +1095,27 @@ def respects_required_agent_action(art: CellArtifacts) -> CriterionResult:
...
+    """Block any coding-agent action the live control did not authorize.
+
+    Contract v20 made this two questions rather than one. A total stop still
+    forbids every subsequent action. A ``review_publishable`` control does not
+    stop the agent, but it denies ``merge`` and ``report_complete``
+    specifically — so the vector, not the state, decides whether each observed
+    operation was allowed.
+    """
 
     timeline = _control_timeline(art)
```
pengfei-threemoonslab, line 1119: **[P1] Apply artifact-only controls to the transcript timeline.** `current` is initialized only from inline transcript control events. When the supported artifact fallback contains `review_publishable` with `permissions.merge=false` but the transcript then runs `gh pr merge`, the fallback below checks only for a hard stop and returns N/A. Seed the effective current control from generated artifacts (with correct ordering rules) so denied actions cannot escape merely because the control was not echoed inline.

src/agents_shipgate/cli/discovery/local_contract.py
```diff
@@ -103,6 +104,7 @@ class LocalAgentContract(BaseModel):
     agent_result_schema_path: str
     agent_result_control_fields: list[str]
     agent_control_fields: list[str]
+    agent_control_permissions: list[str]
```
pengfei-threemoonslab, line 116: **[P2] Bump the local contract schema for this required field.** `LOCAL_CONTRACT_SCHEMA_VERSION` remains 7, but a pre-PR v7 payload has no `agent_control_permissions` and now fails `LocalAgentContract.model_validate` with a missing-field error. Publish v8 and preserve a v7 reader, or keep this field optional for v7.

src/agents_shipgate/schemas/preflight.py
```diff
@@ -412,7 +412,10 @@ class PreflightResultV3(PreflightResultV2):
     @model_validator(mode="after")
     def _legacy_fields_project_control(self) -> PreflightResultV3:
         control = self.control
-        expected_human = control.state == "human_review_required"
+        # Preflight's only human route is a protected-surface stop, so it never
```
pengfei-threemoonslab: **[P2] Reject `review_publishable` on the pre-edit surface.** This comment says preflight never emits that state, but both the Pydantic model and published preflight schema accept a structurally valid publishable control. That grants publication authority before an evaluated change exists. Use a preflight-specific control subset or explicitly reject the state in both model and schema.

pengfei-threemoonslab:
All twelve inline findings and all four cross-cutting ones are addressed across three commits: `e0243c9b`, `b1d0c801`, `d57ec42d`. Every one reproduced before I touched it.

Two of them changed my mind about the design rather than just the code, so I'll lead with those.

## Where you were right and I was wrong

**Inferring evaluation from the action kind.** The missing-manifest repro is decisive: `verify` stops before reading anything, emits a `configure` route with `execution=failed` and `diff_status=unavailable/not_attempted`, and carried edit/commit/push/update-PR. No allowlist would have caught that, because the kind describes what to *do next*, not what was *read*. `publication_allowed` is now one fact the caller asserts, governing the vector on every route rather than only human ones, and defaulting to False so a caller who has not thought about it authorizes nothing. That also dissolves the JSON-Schema half of the finding: there is no longer a route-dependent rule to encode, because there is no longer a route-dependent rule.

**Reusing schema identifiers.** My previous resolution only covered old-payload-against-new-schema, which is the direction that does not matter. Confirmed your direction directly: a new `complete` verifier fails the base-published `v0.7` with six errors. Every control-carrying identifier now advances and the prior file is untouched — verifier `0.7→0.8`, handoff `v6→v7`, verify-run `v3→v4`, agent result `v2→v3`, boundary result `v1→v2`, preflight `0.3→0.4`, local contract `7→8`. CLI flag spellings are unchanged; only the `schema_version` string moved. Generated filenames now derive from the constants so a future bump cannot strand a file under the old name.

Worth flagging one consequence I fixed rather than accepted: the boundary `audit_id` hashed the live schema version, so all seven bumps would have rotated every stored id — including the frozen codex projection's, whose contract had not changed. The code already documents that rotating an established id is a break. The hashed token is now pinned to the value established ids were issued under, with the wire version travelling in the payload's own `schema_version`. Every golden is byte-identical to `main`.

## The rest

**Containers now bind publication to their substrate**, keyed on the permission vector rather than the state — an agent repair route asserts exactly what a publishable review does. Verifier additionally requires `diff_status.completeness="complete"` and constrains the *nested* `release_decision.decision` rather than merely typing it as an object; `AgentResultV2` rejects `decision="block"`; `AgentBoundaryResultV1` rejects partial and experimental coverage; verify-run and handoff carry the same rule. Pydantic and generated JSON Schema both.

**The frozen v2 contract no longer inherits from the evolving base.** You were right that a subclass serializer cannot hold that line — pydantic serializes through the base whenever the declared type is the base. `CodexBoundaryResultV2` is now a standalone model over a snapshotted `FrozenAgentControl`, built only by `freeze_codex_boundary_result` at the one emit boundary, so `TypeAdapter(AgentResultV2).dump_python(...)`, a base-annotated field, and `model_json_schema()` all behave. `VerifyRunArtifactV2` gets a snapshotted outcome for the same reason, so reading and re-dumping a valid v2 payload no longer adds `control.permissions`.

**The Stop hook validates the vector**, not the state tag, and my own fixture was the malformed input you described. It now requires the exact publish-only six before caching or announcing anything; three negative cases cover absent, merge-granting, and empty vectors.

**MCP completeness is a first-class result** rather than a diagnostic denylist, and the allowlist is inverted so a warning added later defaults to "did not read everything". `mcp_policy_missing` and unresolved `content_source` rows now deny publication — verified end to end.

**Adoption scorer:** legacy payloads reconstruct the vector conservatively instead of guessing publication from the route; `file_change` and editor tool calls are scored against `permissions.edit` despite carrying no command; and the effective control is seeded from generated artifacts, so a denied merge cannot escape by not being echoed inline.

**Preflight** rejects `review_publishable` and any progress authority in model and schema rather than documenting that it never happens.

**Contract-floor drift:** the Codex and Claude Code guides and `protocol.md` named floor 15, and `docs/passed-verdict-contract.md` named runtime contract v19.

## The pinned runner

You framed this as a choice, and I took the fourth option rather than either side of it: **the pin and the floor are now both rendered from the build that emits the prompt.** The kit templating already supplied `{{ shipgate_version }}`; it gains `{{ minimum_control_contract_version }}`, and the prompts use both instead of literals. They cannot disagree again, and they move together at release time without anyone remembering to.

The public-surface guard is split to match, because it was the guard that kept the contradiction alive: a CI example installs from PyPI and still tracks the latest published release, while a rendered prompt tracks the build that rendered it, with a new test asserting exactly that.

## Verification

Full suite green (0 failures), plus Ruff, compileall, generated-schema drift, and the four CI safety-canary gates. Each finding has a named regression test. The two structural claims are pinned by tests that fail if the property regresses: `test_publication_is_an_asserted_fact_not_an_inferred_action_kind` walks every action kind and asserts the default is no publication, and `test_frozen_v2_shape_holds_on_every_serialization_path` now includes the base-typed dump and asserts the frozen model is not an `AgentResultV2`.

PR #347, opened 2026-08-08
feat(control): add an atomic current-control pointer and mandatory agent refresh protocol

### LINKED ISSUE #339: [Agent-first UX][P0] Add an atomic current-control pointer and mandatory agent refresh protocol
> Parent epic: #338

## Problem

Coding agents can retain a verifier result in conversation state after the underlying repository and verifier artifacts have changed. In a real issue-to-PR workflow:

1. An earlier worktree verification returned `human_review_required`.
2. A human committed the reviewed changes and generated a new committed-ref receipt with `control.state=complete`.
3. The worktree was clean and the current verifier/receipt shared the same complete request identity.
4. The coding agent nevertheless continued to enforce the older `must_stop=true` result and asked the user to commit the files again.

The content-addressed receipt model correctly prevents an old decision from authorizing a different request. The UX failure is that consumers do not have one atomic “what is current now?” entry point and are not required to refresh it before blocking, publishing, merging, or declaring completion.

This can fail in both directions:

- **Stale denial:** an agent keeps blocking after a newer complete verifier run exists.
- **Stale authorization:** an agent relies on an older complete result after the worktree, HEAD, configuration, policy, or authorization state changed.

Both are control-plane correctness issues.

## Product principle

> A coding agent must make operational decisions from the latest validated control identity, never from a verifier result cached in conversation context.

## Proposed engineering direction

Add one uniquely discoverable, atomic control entry point, for example:

```text
agents-shipgate-reports/current-control.json
```

This is an authoritative **entry point**, not a second verdict engine. It should reference and bind the existing receipt, handoff, verifier, and report decisions.

Example shape:

```json
{
  "schema_version": "shipgate.current_control/v1",
  "current_control_id": "sha256:...",
  "operation": "verify",
  "lifecycle_state": "terminal",
  "request_id": "sha256:...",
  "decision_id": "sha256:...",
  "workspace_identity": {
    "repository": "...",
    "head_commit_sha": "...",
    "head_tree_sha": "...",
    "worktree_overlay_sha256": null,
    "policy_snapshot_sha256": "..."
  },
  "control": {
    "state": "complete",
    "completion_allowed": true,
    "must_stop": false
  },
  "artifacts": {
    "verification_receipt": {
      "path": "verification-receipt.json",
      "sha256": "..."
    },
    "agent_handoff": {
      "path": "agent-handoff.json",
      "sha256": "..."
    },
    "verifier": {
      "path": "verifier.json",
      "sha256": "..."
    },
    "report": {
      "path": "report.json",
      "sha256": "..."
    }
  },
  "supersedes": "sha256:..."
}
```

### Invalidate before writing a new run

A prior `complete` pointer must not remain current while a new scan/verify run is in progress. At command start, atomically replace it with a nonterminal/invalidation state bound to the new operation when possible:

```json
{
  "lifecycle_state": "in_progress",
  "control": {
    "state": "unavailable",
    "completion_allowed": false,
    "must_stop": true
  }
}
```

If the process crashes, consumers must not continue treating the previous terminal decision as current for a changed workspace.

### Publish terminal control last

After all terminal artifacts and their hashes are complete and validated, write the new pointer to a temporary file in the same directory and publish it with an atomic replace. The pointer is the last artifact made visible for the run.

### Prevent mixed-generation reads

Define a safe reader protocol:

1. Read and validate `current-control.json`.
2. Validate every required referenced artifact/hash.
3. Re-read `current-control.json`.
4. Continue only if `current_control_id` is unchanged.

Equivalent generation-safe semantics are acceptable.

### Keep supporting commands isolated

Commands such as `baseline save` that only need an internal scan must not invalidate or replace the current PR control identity. Their scans should remain in isolated output directories.

## Mandatory coding-agent refresh protocol

Generated agent instructions and the machine contract should require a fresh read and validation:

- after any human action or external tool action;
- after commit, rebase, checkout, pull, or worktree mutation;
- after a verifier/scan/check command returns;
- before enforcing a cached `must_stop`;
- before commit/push/PR update when permission depends on Shipgate;
- before merge or release;
- before declaring the task complete;
- whenever the observed request ID, HEAD, worktree identity, or artifact generation changes.

If the pointer changes, the agent must discard cached control state and restart the control loop from the new identity.

Conversation-level state must never outrank the current validated control pointer.

## Scan versus verify semantics

The pointer must be command-scoped:

- After standalone `scan`, it should identify a scan result and must not retain a prior verifier route or merge authorization.
- After `verify`, it should reference the terminal receipt and verifier handoff for that exact request.
- During preview/nonterminal runs, it must clearly deny treating older terminal evidence as current.
- If lifecycle cleanup fails, the pointer must remain nonterminal/fail-closed and must not reference a mixed artifact set.

## Security and trust requirements

- [ ] The pointer introduces no second release decision; it only binds existing authoritative fields.
- [ ] All referenced paths are relative, containment-checked, and non-symlink escaping.
- [ ] Required artifact hashes and schema versions are validated before use.
- [ ] A changed HEAD, worktree overlay, config, policy snapshot, or authorization identity invalidates incompatible control.
- [ ] A stale complete pointer cannot authorize a newer workspace state.
- [ ] A stale stop pointer cannot continue blocking after a newer validated generation is current.
- [ ] Coding agents cannot write or self-assert human authorization through this surface.

## Acceptance criteria

- [ ] There is exactly one documented default control entry path.
- [ ] The pointer is invalidated before a new lifecycle run and terminally published last.
- [ ] Pointer publication uses atomic same-directory replacement.
- [ ] Readers can detect concurrent generation changes and mixed artifact sets.
- [ ] Current request, decision, repository/head/worktree, policy, and artifact identities are bound.
- [ ] Agent instructions mandate refresh at every operational decision boundary listed above.
- [ ] A regression reproduces: cached `human_review_required` → external human commit → new complete receipt → agent refreshes and proceeds without asking for another commit.
- [ ] A reverse regression reproduces: cached `complete` → workspace mutation → pointer invalidation → agent cannot push/merge/complete on stale authorization.
- [ ] Preview → scan, scan → verify, verify → verify, interrupted run, and cleanup-failure transitions are covered.
- [ ] `baseline save` and other supporting internal scans do not disturb current PR control.
- [ ] The public contract and `.well-known` discovery surface expose the entry path and refresh rules.
- [ ] Older consumers receive a documented fallback read order.

## Relationship to existing work

- #333 defines compact and unambiguous agent-facing control output.
- #335 separates PR-publication permission from merge/completion permission.
- #336 fixes committed/worktree diff normalization.
- #337 defines the human review/authorization handoff.
- #317 and #331 motivated artifact lifecycle cleanup.

This issue ensures those controls remain fresh and operationally coherent across processes, human actions, and coding-agent conversation boundaries.

pengfei-threemoonslab:
## Summary

- Adds `agents-shipgate-reports/current-control.json` (`shipgate.current_control/v1`): one atomic, uniquely discoverable entry point naming which control identity is current. It is a **pointer, not a second decision engine** — it binds ids and hashes of the receipt, handoff, verifier, and report those commands already publish.
- Makes the lifecycle fail-closed: every run **invalidates first** (non-terminal `unavailable` marker before any other artifact moves), **publishes last** (after every bound artifact exists and is hashed), and **publishes atomically** (same-directory tmp → `fsync` → `os.replace` → directory `fsync`).
- Adds `agents-shipgate agent control`, which implements the generation-safe read: validate pointer → validate every bound artifact hash → re-read pointer → continue only if `current_control_id` is unchanged.
- Makes the refresh obligation machine-readable (`agent_refresh_triggers[]`) and mandatory in the generated agent instructions and both adoption kits.

Closes #339. Part of the #338 epic.

### The bug

A coding agent could keep enforcing a verifier result the workspace had outgrown. In the reported sequence a worktree verify returned `human_review_required`, a human committed the reviewed change, a fresh committed-ref run produced a `complete` receipt for the same request — and the agent kept enforcing the older `must_stop`, asking for the commit that had already happened.

It runs backward just as easily: a `complete` remembered from earlier in a conversation is not evidence about a workspace that has since been rebased, checked out, or reconfigured.

The content-addressed receipt already prevented an old decision from authorizing a *different* request. What was missing was one atomic place to ask "what is current now?", and any obligation to ask it.

### Two structural invariants

Neither is advisory — both are rejected by the Pydantic model and the generated JSON Schema:

1. Only an `operation: "verify"` pointer can carry `control.state: "complete"`, and only when it also binds a `verification_receipt` for that exact request. **A `scan` or `preview` pointer cannot represent completion authority at all.**
2. An `in_progress` pointer cannot carry a settled request, decision, or artifact set; a `terminal` pointer cannot carry `unavailable`.

Supporting scans stay isolated: `verify`'s internal head scan does not take over the PR's control identity (contextvar lifecycle owner), and `baseline save` already scanned into a temporary directory.

## Type

- [x] Documentation only (agent instructions, adoption kits, contract docs)

## Verification

Local checks run on the final tree:

- `ruff check .` — clean
- `python scripts/generate_schemas.py --check` — no drift
- `python -m pytest tests/` — exit 0 (full suite)
- New `tests/test_current_control.py` — 31 tests against the real engine

**Both reported directions reproduce as regressions:**

- **Forward (stale denial)** — worktree verify stops on a trust-root edit → human commits → committed-ref run completes → the refreshed pointer is `complete` with a bound receipt and a different `current_control_id`. This is the exact issue-#339 sequence.
- **Reverse (stale authorization)** — a `complete` pointer records the tree it was computed from, so a consumer sees HEAD moved out from under it *without re-running*; the rerun then replaces the identity.

Also covered: interrupted run, cleanup failure, preview→scan / scan→verify / verify→verify transitions, `baseline save` isolation, mixed artifact sets, concurrent-generation detection and retry, symlinked pointer, symlinked bound artifact, escaping artifact paths, deterministic reruns, and the split `prepare`→`worker`→`assemble` path.

Dogfooded: running this branch's verifier on itself returns `human_review_required` with the reason *"This PR edits a release trust root … a coding agent cannot self-approve that change"* — correct, since the change edits agent instructions.

## Release-readiness notes

Contract `19 → 20`; local downstream contract `7 → 8`. **No report, packet, verifier, handoff, or receipt schema changed.** New `docs/current-control-schema.v1.json`. `agent_read_order[]` now starts at the pointer, and `current_control_fallback_read_order[]` documents the read order for consumers built before it existed — the pointer's absence is evidence of an older producer, never permission to act on a cached decision.

Outgoing render hashes are recorded in `renderers/cursor.py`, `renderers/claude_command.py`, `renderers/local_contract.py`, and both kit metadata files, so a repo that already ran `init --write` upgrades in place rather than being read as user-modified and stranded without the new rule.

## Surface discipline

Per [`CONTRIBUTING.md` § Surface discipline](https://github.com/ThreeMoonsLab/agents-shipgate/blob/main/CONTRIBUTING.md#surface-discipline):

1. **Metric:** time-to-first-verdict / activation. An agent that blocks on a cached `must_stop` costs a full human round-trip on work that is already done.
2. **Could an existing surface carry it?** No. The receipt prevents an old decision from authorizing a *different* request; nothing said which decision is *current*. The pointer adds no verdict — it binds fields the one decision engine already produced, and refuses to represent `complete` without the receipt that backs it.
3. **Roadmap non-goals:** respected — no second verdict, no new adapter, no agent execution, no LLM calls, no network in the default path.

## Reviewer notes

⚠️ **`CONTRACT_VERSION` 20 is claimed twice.** Branch `claude/github-issue-335-5dbf25` (#335, commit `f0bfe3de`) also bumps 19 → 20, with different content — it additionally raises `MINIMUM_CONTROL_CONTRACT_VERSION` to 20 and changes the `AgentControl` union. Both branch from `eae59eda`. Whichever merges second must rebase and take 21; this branch stays at 20 rather than guessing the merge order. Same class as the PR #232 collision.

**Pre-existing, not introduced by this PR:** in a workspace where `agents-shipgate-reports/` is *not* gitignored, generated artifacts become worktree verification inputs — already true of `report.json` and `verifier.json`. `init`'s gitignore block is the existing guard. This PR moves `verification prepare`'s invalidation ahead of input reading so the plan and any replay of it observe one stable directory state, and the test fixture carries the gitignore rather than simulating the broken case.


src/agents_shipgate/core/current_control.py
```diff
@@ -0,0 +1,564 @@
...
+        except CurrentControlUnavailable as mismatch:
+            # A run that republished mid-read moves the pointer too. Retry that
+            # case; a mismatch under a pointer that did not move is a real
+            # inconsistency and must surface.
+            if _current_control_id(path) == pointer.current_control_id:
+                raise
+            last = mismatch
+            continue
+        confirmation = _load_pointer(out_dir, path)
+        if confirmation.current_control_id == pointer.current_control_id:
```
pengfei-threemoonslab, line 481: **[P1] Validate the live workspace before returning completion authority**

This return only proves that the pointer and its bound artifact bytes were stable during the read; it never compares `workspace_identity` with the repository's current HEAD, tree, policy snapshot, or worktree overlay. I verified a clean run, made an empty commit (`671df88…` → `64a73c5…`), and reran `agents-shipgate agent control`: it exited 0 with the same `current_control_id` and `completion_allowed: true` for the old commit. That defeats the reverse stale-authorization guarantee and the documented refresh trigger after commit/rebase/checkout. Make the reader workspace-aware (for example via `--workspace`) and fail closed when the live identity differs.

src/agents_shipgate/cli/verification.py
```diff
@@ -389,6 +413,21 @@ def assemble(
...
+    # The pointer is the last file the assembler makes visible, for the same
+    # reason it is last in `verify`: every artifact it binds now exists and has
+    # been hashed in its final form.
+    publish_current_control(
+        resolved_artifact_root,
+        operation="verify",
+        control=project_agent_control(
+            verifier.control,
+            operation="verify",
+            receipt_bound=(resolved_artifact_root / "verification-receipt.json").is_file(),
```
pengfei-threemoonslab: **[P1] Bind and validate the receipt actually emitted by assembly**

`assemble` permits any `--out` beneath `--artifacts-root`, but this checks only the canonical `verification-receipt.json`, and `publish_current_control` later binds that canonical file. With an older canonical receipt present and a new receipt emitted as `custom-receipt.json`, assembly published `complete` for request `4de968…` while the pointer bound a receipt for request `801be8…`; `agent control` still exited 0. Either require the canonical output path or bind `out` itself, then parse the bound `VerificationReceipt` and require its `request_id`/`decision_id` to match the pointer before completion is representable.

src/agents_shipgate/cli/scan/writing.py
```diff
@@ -83,6 +108,35 @@ def _write_outputs(
             plan.generated_paths,
             plan.packet_format_set,
         )
+    if owns_control:
+        # A scan inventories a workspace; it never establishes merge authority.
+        # The pointer says so in the one place a coding agent is required to
+        # look, so a verifier route from an earlier run cannot be mistaken for
+        # the current permission to finish.  A scan builds no verification
+        # plan, so the only identity it can bind is the manifest it read.
+        publish_current_control(
```
pengfei-threemoonslab, line 123: **[P2] Restrict the scan pointer to artifacts emitted by this scan**

Because no `artifact_keys` are supplied here, the binder includes every recognized file that already exists. After a verifier run, `scan --format markdown` leaves the old `report.json` untouched, then publishes a scan pointer that claims that stale JSON report as current. In my reproduction the pointer had null request/decision IDs while its bound `report.json` still carried the preceding verifier's IDs. Derive the artifact set from `plan.generated_paths`/the formats actually written, or remove obsolete format files before publication.

pengfei-threemoonslab (commented):
Blocking engineering review: the current-control pointer is byte/hash consistent, but in three supported flows it is not generation consistent.

Two P1 issues can preserve or manufacture stale completion authority:

1. `agent control` does not compare the pointer's recorded workspace identity with the live repository.
2. Split assembly can publish `complete` while binding a receipt from a different request.

A P2 issue also lets standalone scans bind stale JSON reports from an earlier verifier run.

I reproduced all three against `f11c92bf`. CI, Ruff, schema generation, and the existing tests are green, so regression coverage should be added for these cases.

_GitHub does not permit the PR author to submit a Request changes review, so these are posted as a comment review but should be treated as blocking before merge._

src/agents_shipgate/core/current_control.py
pengfei-threemoonslab: Fixed in `b1895227`. You were right that the read only proved byte consistency; it now proves generation consistency too.

`agent control` takes `--workspace` (default `.`) and `read_current_control` compares the bound `workspace_identity` against the live repository: repository id, HEAD commit, and HEAD tree. For a `worktree_overlay` snapshot it also recomputes the overlay from the bound plan's `changed_paths` using the same `worktree_overlay()` the plan committed to — and compares the change *set* as well, because a file edited **after** the decision appears in neither HEAD nor that overlay (your empty-commit repro caught the first hole; that one was the second).

Completion authority is never returned without the comparison: `live=None` means "not compared", which raises `workspace_unverified` rather than passing. The reports directory is excluded from the change set exactly as `verify` excludes it, so a run's own output is not read as drift — including in repos that do not gitignore it.

Your repro now:

```
[baseline]         exit=0
[empty commit]     exit=4 -> workspace_changed: The workspace moved since this decision was made: head_commit_sha is ...
[worktree edit]    exit=4 -> workspace_changed: The set of uncommitted changes is no longer the one this decision was made against.
```

Regressions: `test_cached_completion_cannot_survive_a_committed_change`, `test_cached_completion_cannot_survive_an_uncommitted_edit`, `test_completion_is_refused_when_the_workspace_cannot_be_checked`, `test_agent_control_ignores_its_own_output_directory`, `test_agent_control_command_fails_closed_on_workspace_drift`.

src/agents_shipgate/cli/verification.py
pengfei-threemoonslab: Fixed in `b1895227`, taking the second option you offered — bind-then-validate rather than constraining `--out`, so the guarantee holds for any producer rather than only for `assemble`.

`publish_current_control` now parses the bound receipt whenever the control would be `complete` and requires its `request_id` and `decision_id` to be the ones the pointer records; a mismatch downgrades to `human_review_required` with a precise reason instead of publishing completion. `read_current_control` applies the same check, so a pointer written by an older or third-party producer is caught on the consumer side too.

Reproducing your setup (stale canonical receipt for request A, assembly emitting `custom-receipt.json` for request B) now gives:

```
pointer control:  human_review_required | completion_allowed: False
pointer reason:   The bound terminal receipt closes a different request than the one this run decided, ...
pointer request : sha256:fe14a797a8460
bound receipt   : sha256:6859f89dba9cd
```

Regression: `test_assembly_refuses_completion_when_it_binds_another_request`.

src/agents_shipgate/cli/scan/writing.py
pengfei-threemoonslab: Fixed in `b1895227`, taking the first option — the artifact set is derived from what this scan actually wrote rather than from what happens to exist.

`_write_outputs` now passes `artifact_keys` built from `plan.generated_paths` through a `SCAN_FORMAT_ARTIFACT_KEYS` map. Two supporting changes made that work honestly: `report.md` and `report.sarif` became bindable keys, so a markdown-only scan still binds real evidence rather than nothing; and the "a terminal pointer must bind at least one artifact" rule is now scoped to `verify`/`preview`, since a scan configured to emit no format genuinely produces nothing and saying so beats adopting another run's output.

```
[scan --format markdown after a verify] bound: ['packet', 'report_markdown'] | report.json on disk: yes
```

The stale JSON report is left in place — it is simply no longer claimed as current. Regression: `test_a_scan_binds_only_the_formats_it_wrote`.

src/agents_shipgate/core/current_control.py
```diff
@@ -0,0 +1,810 @@
...
+                "workspace_unverifiable",
+                (
+                    "This pointer authorizes completion of a worktree "
+                    "verification, but the current set of uncommitted changes "
+                    "could not be determined."
+                ),
+                path=out_dir,
+            )
+        return
+    if set(live.changed_paths) != set(decided_paths):
```
pengfei-threemoonslab: **[P1] Compare the live worktree to the right snapshot**

`plan.inputs.changed_paths` is not the uncommitted-path set: for a no-`--head` verification it is the union of `base...HEAD` and worktree paths. On this updated PR, `verify --workspace . --config shipgate.yaml` returned 0, but an immediate `agent control` returned `workspace_changed` with 52 plan paths versus 0 live paths, before any mutation. The converse is also open: for the normal committed-tree `--head HEAD` flow this branch never examines `live.changed_paths`, so after a clean `complete` run I added an untracked tool file and `agent control` still exited 0 with `completion_allowed: true`. Bind the worktree-only path set for `worktree_overlay` comparisons, and require no live worktree changes for a `committed_tree` decision (excluding the reports directory).

pengfei-threemoonslab (commented):
Re-review of `b1895227`: the reported scan-artifact issue is fixed, and the new receipt/request check prevents a stale canonical receipt from granting completion. Two blocking control-surface problems remain.

1. The live-workspace comparison uses the wrong path set for local worktree verification and skips worktree drift entirely for committed-tree verification. This creates both immediate stale denial and stale authorization.
2. `verification assemble --out <custom>` still does not bind the receipt it emits, so a valid custom receipt can never publish current completion authority; with a stale canonical receipt present, the pointer still binds the stale receipt while downgrading.

I reproduced both on the updated head. The 37 current-control tests, Ruff, schema generation, and all three GitHub workflows pass, so regression coverage needs the canonical auto-base local flow and committed-ref-plus-later-worktree-drift flow.

src/agents_shipgate/cli/verification.py
pengfei-threemoonslab: **[P2] The unsafe-authorization half is fixed, but custom receipt output is still not bound.** `receipt_bound` still probes only `verification-receipt.json`, and the fixed filename map still has no way to reference `out`. I removed the canonical receipt, assembled to `custom-receipt.json`, and `verification reproduce` validated that receipt; nevertheless the new pointer had no `verification_receipt` artifact and projected `human_review_required`. With a stale canonical file, it instead binds that stale receipt (now safely downgrading rather than authorizing). Thus the supported custom `--out` flow can never publish its actual completed generation. Please either bind `out` as the portable receipt ref and validate it, or reject noncanonical receipt paths.

src/agents_shipgate/core/current_control.py
pengfei-threemoonslab: Both halves fixed in `3a64de1a`. You were right on both, and the first one was a stale-denial regression I introduced — the exact failure direction this PR exists to prevent.

**Wrong path set.** `plan.inputs.changed_paths` is the union of `base...HEAD` and the worktree, so equality against the live uncommitted set refuses a clean workspace the moment the run that produced it finishes. My tests missed it because they exercised `base=None` rather than the canonical local flow.

The correct test is a **subset**, and it is complete rather than merely lenient. A worktree decision's evidence is the overlay rows for the paths it recorded; everything outside that set was identical to HEAD when the decision was made — that is *why* it was not in the change set. So two conditions pin the working tree: those paths still hash to the committed overlay, and no path outside the set differs from HEAD now. A file edited afterwards is not in the recorded set (caught); a file committed since moves HEAD (caught); a file reverted or deleted changes the overlay (caught).

**Committed-tree drift.** Agreed, and fixed — with one deliberate narrowing. Uncommitted changes now block *completion* for a `committed_tree` decision, but do not make the pointer unreadable. Refusing the read outright would livelock: re-running the same committed verification reproduces the same pointer, so the caller would be refused forever with no route to clear it. Worktree refusals are self-clearing (a worktree re-verify picks the new paths up), which is what makes the asymmetry principled rather than convenient. If you would rather have the hard refusal for both, say so and I will make it symmetric.

Your repros on `3a64de1a`:

```
[local verify --base main, clean tree]   agent control exit=0    (was workspace_changed)
[committed-tree complete + untracked]    agent control exit=4    workspace_changed: The working tree carries 1
                                                                 uncommitted change(s) this committed-tree
                                                                 decision could not have covered (extra-tools.json)
```

Regressions: `test_a_local_run_with_a_base_stays_current_on_a_clean_tree` and `test_committed_tree_completion_cannot_survive_a_new_uncommitted_file`. The three round-1 repros were re-run and still hold.

src/agents_shipgate/cli/verification.py
pengfei-threemoonslab: Fixed in `3a64de1a`, taking the bind-`out` option — rejecting non-canonical paths would break a supported flag, and the relocation is legitimate.

`publish_current_control` takes an `artifact_paths` override: a producer may relocate a *known* artifact, never invent a new kind of one, and never point outside the directory the pointer describes (the path goes through `validate_portable_path`, and an unknown key is a publish error). Assembly passes `{verification_receipt: <out relative to --artifacts-root>}` and computes `receipt_bound` from `out` rather than the canonical name.

Your repro — canonical receipt removed, assembled to `custom-receipt.json`:

```
control        : complete | completion_allowed: True
bound receipt  : custom-receipt.json
canonical file : False
verification reproduce --receipt custom-receipt.json -> {"valid": true, ...}
agent control exit=0
```

And with a stale canonical receipt present, the pointer now binds `custom-receipt.json` with the fresh request and publishes `complete`, leaving the stale file on disk unclaimed — the same treatment the scan fix gives a superseded `report.json`. The request/decision cross-check stays as defence in depth for producers that get it wrong; it is now covered by its own test (`test_completion_is_refused_when_the_bound_receipt_closes_another_request`) rather than only through assembly, since assembly can no longer trip it. `test_assembly_refuses_completion_when_it_binds_another_request` was replaced by `test_assembly_binds_the_receipt_it_emitted`, which asserts the correct behaviour instead.

src/agents_shipgate/core/current_control.py
```diff
@@ -0,0 +1,878 @@
...
+            "workspace_unverifiable",
+            (
+                "This pointer authorizes completion but binds no HEAD identity, "
+                "so it cannot be confirmed against the live workspace."
+            ),
+            path=out_dir,
+        )
+    if identity.snapshot_kind == "worktree_overlay":
+        _validate_worktree_currency(out_dir, pointer, live, required=grants_authority)
+    elif grants_authority and identity.snapshot_kind == "committed_tree":
```
pengfei-threemoonslab: **[P1] Invalidate non-authorizing committed-tree pointers too**

This condition limits the clean-worktree check to `complete`. I reproduced the PR's committed-tree `human_review_required`, added an untracked file, and reran `agent control --workspace .`; it exited 0 with the same `current_control_id` and `must_stop: true`. That is the stale-denial direction this pointer is meant to eliminate: after a human or external tool changes the worktree, the mandatory refresh still enforces the pre-change stop, and this control has no allowed coding-agent command that can reach the local worktree verification needed to clear it. Validate committed-tree currency for every terminal pointer when a live workspace is available (or return an explicit stale/nonzero result with a worktree-verify remediation); rerunning the same archived `--head` verification is not the clearing route.

src/agents_shipgate/core/current_control.py
```diff
@@ -0,0 +1,878 @@
...
+    try:
+        data = read_regular_file_beneath(
+            out_dir,
+            ref.path,
+            max_size=MAX_BOUND_ARTIFACT_BYTES,
+            label="current control plan",
+        )
+        plan = VerificationPlan.model_validate_json(data)
+        decided_paths = list(plan.inputs.changed_paths)
+        rows = worktree_overlay(live.root, decided_paths)
```
pengfei-threemoonslab, line 654: **[P1] Bind file type and executable mode in the overlay**

`worktree_overlay()` commits only path, present/deleted status, and content hash. I verified a `worktree_overlay` run whose covered path was `tools.json`, then changed that path from mode `100755` back to `100644` without changing its bytes. `git status` reported the path modified, but `agent control` still exited 0 with the same `current_control_id` and `completion_allowed: true`: the live path is already in `decided_paths`, and the recomputed content-only row is identical. A regular-file to in-repo symlink swap with identical target bytes has the same shape because the helper resolves the path. Include `lstat` file type and relevant mode bits in the overlay identity, and preserve symlink identity, so metadata-only capability changes invalidate the decision.

src/agents_shipgate/core/current_control.py
```diff
@@ -0,0 +1,878 @@
...
+def workspace_identity_from_plan(plan: VerificationPlan) -> CurrentControlWorkspaceIdentity:
+    """Derive the pointer's workspace identity from a verification plan.
+
+    The plan's subject is the same one the terminal receipt closes over, so a
+    pointer built from it names exactly the workspace the decision was made
+    against — not the workspace the reader happens to be sitting in.
+    """
+
+    git = plan.subject.git
+    return CurrentControlWorkspaceIdentity(
```
pengfei-threemoonslab, line 386: **[P1] Bind the base revision that defines the decision**

This projection drops `base_ref`, `base_commit_sha`, and `merge_base_sha`, and the live reader consequently checks only HEAD. Repro: verify a feature branch with `--base main` and obtain `complete`; then advance the `main` ref to that feature HEAD without changing the feature HEAD or worktree. `main...HEAD` is now empty, but `agent control` still exits 0 with the same completion-authorizing pointer. A fetch that advances `origin/main` has the same stale-request shape. Because the changed-file set and release evidence are base-dependent, carry the immutable base/merge-base identity in the pointer and resolve/compare the live base ref when the verification supplied one before returning the control as current.

pengfei-threemoonslab (commented):
Re-reviewed the updated head `3a64de1a` from the control-plane and release-gate perspectives. The custom receipt is now bound correctly, the scan artifact restriction remains correct, and the local `base...HEAD` subset fix addresses the prior false stale denial.

Three currentness gaps remain below. Each has a minimal reproduction on this head and can preserve either stale `must_stop` or stale completion after the evidence-defining workspace state changes, so I consider them merge blockers for the core contract of this PR.

The GitHub `CI` workflow is also currently red: `test_a_local_run_with_a_base_stays_current_on_a_clean_tree` initializes its temporary repo with the runner's default branch but verifies against `base="main"`, yielding `changed_paths=[]` on the runner. Pinning the fixture's initial branch to `main` addresses that platform dependency.

pengfei-threemoonslab (commented):
Re-reviewed head `1ce7e6dd`. This commit cleanly fixes the CI portability problem: the temporary repositories now initialize an explicit `main` branch, the regression asserts the exact changed path, the focused current-control suite and Ruff pass locally, and all three GitHub workflows are green.

It does not change production current-control code, and I independently replayed the three open P1 threads on this head. All three still reproduce:

1. A committed-tree `human_review_required` followed by an untracked worktree edit still makes `agent control` exit 0 with the same `current_control_id` and `must_stop: true`.
2. A chmod-only change (`100755 → 100644`) on a path already in a worktree decision still makes `agent control` exit 0 with the same `current_control_id` and `completion_allowed: true`.
3. Advancing the verified base ref so `main...HEAD` becomes empty, without moving HEAD or the worktree, still makes `agent control` return the original completion-authorizing pointer.

No new issue was introduced by the test-only commit, and the earlier receipt-relocation and scan-artifact fixes remain sound. The CI blocker is cleared, but the three existing currentness threads remain merge-blocking for the core guarantee of this PR.

src/agents_shipgate/core/current_control.py
pengfei-threemoonslab: Fixed in `1c5ab001`. My livelock argument was wrong and you named exactly why: the clearing route for worktree drift is a *worktree* verification, not a rerun of the same archived `--head` one. There is no livelock, and scoping the check to `complete` left the mandatory refresh enforcing a pre-change stop against a workspace a human had since edited — the stale-denial direction this pointer exists to eliminate.

Committed-tree currency now applies to every terminal pointer, and the refusal carries the remediation rather than leaving the caller to infer it:

```
exit=4  workspace_changed: The working tree carries 1 uncommitted change(s) this committed-tree
        decision could not have covered (scratch.py). Re-run verification. Re-run verification
        over the working tree (omit --head) to decide on the workspace as it stands.
```

One narrowing kept, deliberately: `required` now gates only the *undeterminable* case. A change set that could not be read is unknown rather than known-drifted, so it blocks completion authority without denying a caller its route. Known drift refuses regardless of control state.

Regression: `test_a_committed_tree_stop_does_not_survive_a_worktree_edit`, which asserts the remediation text as well as the refusal.

src/agents_shipgate/core/current_control.py
pengfei-threemoonslab: Fixed in `1c5ab001`. Both shapes confirmed — the chmod and the symlink swap — and you were right that the helper resolving the path is what made the second one invisible.

Overlay rows now carry entry kind and the executable bit alongside content, and a symlink hashes its *target string* rather than what it points at. That is exactly the metadata Git itself tracks (`100644` / `100755` / `120000`). Full mode is deliberately excluded: it varies with umask, so recording it would make verification identity depend on noise Git does not model.

```
chmod detected:                                   True
symlink swap detected (identical target bytes):   True
deleted detected:                                 True
```

Worth flagging explicitly since it reaches past this PR's surface: this changes `worktree_overlay_sha256`, and therefore `subject_id` and `request_id`, for worktree runs. I judged that acceptable rather than a break, because `engine_requirement_id` already carries the package version — those ids move on every release for identical inputs, so nothing can have been depending on their stability. No schema version moves; the overlay is hashed, not stored. The full suite is green, including the identity canaries.

Regression: `test_metadata_only_changes_invalidate_a_worktree_decision`, parameterized over both mutations.

src/agents_shipgate/core/current_control.py
pengfei-threemoonslab: Fixed in `1c5ab001`. The pointer now carries `base_ref`, `base_commit_sha`, and `merge_base_sha`, and the reader resolves the ref live.

Structurally: the CLI supplies ref resolvers on `LiveWorkspace` rather than core shelling out to git, because the pointer names its own base — it cannot be resolved before the read. An unresolvable ref resolves to `None` and compares unequal rather than raising, since a deleted base is drift the caller must see, not a crash.

Your repro:

```
base_ref: main | base_commit: 3c00b8646140 | merge_base: 3c00b8646140
[clean]                 agent control exit=0
[git branch -f main HEAD; main..HEAD = 0 commits, HEAD and worktree untouched]
                        agent control exit=4
                        workspace_changed: The base this decision was made against moved:
                        'main' is now '702dc3251a3a...', but the pointer was published against '3c00b864...'
```

The fetch-advances-`origin/main` case is the same shape and the same check. Regression: `test_advancing_the_base_invalidates_a_decision_about_the_range`.

pengfei-threemoonslab:
Rebased onto `main` (`c07ba35c`, which brought in #332) and force-pushed: `1c5ab001` → `e627c3e0`. The PR is `MERGEABLE` again and all three workflows are green on the new head.

**Conflict resolution.** One conflict, in `src/agents_shipgate/cli/verification.py`: an import block where #332 added `NextAction` and this branch added `RECEIPT_ARTIFACT_KEY`. Both kept, ordered alphabetically. No logic was dropped from either side.

**Why the two branches compose rather than merely merge.** #332 rewrote `build_verification_plan`'s input enumeration (`_manifest_tool_source_blobs` → `_manifest_declared_input_paths`) and added `changed_root`; this branch touched `_worktree_overlay` and the containment-safe reader in the same file, in disjoint regions. The one place they meet is worth stating explicitly: #332 now raises `InputParseError` for a declared input resolving outside the verification root, and in `verification prepare` that can fire *after* this branch's `begin_current_control` invalidation. That is the correct order — the pointer stays `in_progress`, so a manifest that cannot produce an identity leaves the directory denying cached control rather than advertising the previous decision.

**Re-verified on the rebased tree**, not just re-merged: full suite green under CI's git defaults, Ruff clean, no schema or `llms-full.txt` drift, and all three round-3 reproductions re-run end to end —

```
[#1 committed-tree stop]   clean exit=0   after untracked edit exit=4
[#2 overlay]               clean exit=0   after chmod          exit=4
[#3 base]                  clean exit=0   after base advance   exit=4
```

Note for the review threads: the rebase rewrote the commit SHAs the earlier replies cite (`f11c92bf b1895227 3a64de1a 1ce7e6dd 1c5ab001` → `af3384f9 550dd18a 1fb1e502 bbb5a102 e627c3e0`). The content is unchanged commit-for-commit; the threads themselves still read correctly.

The contract v20 collision with `claude/github-issue-335-5dbf25` is unchanged by this rebase — that branch has not landed, so whichever merges second still needs to take v21.

PR #360, opened 2026-08-10
feat(control): add the compact shipgate.agent_control/v1 control envelope

### LINKED ISSUE #333: Define a compact, operationally unambiguous agent-control envelope
> Parent epic: #338
>
> Canonical v0.17 design issue for the shared agent-control envelope. #323 owns the v0.18 command-by-command rollout and end-to-end coverage; it must apply this shape, not define a second one. The permission model (#335) and current-control pointer (#339) are complete prerequisites.

## Problem

A verifier run can simultaneously report:

- `execution: "succeeded"`
- `head_exit_code: 0`
- `release_decision.decision: "review_required"`
- `control.state: "human_review_required"`

Humans and coding agents naturally interpret “succeeded” or exit code 0 as permission to continue. The authoritative operational fields are buried inside thousands of tokens of JSON, causing users to ask whether the run is “done” and whether the agent may push.

The full forensic document is valuable, but it is a poor default control interface for an agent loop.

## Proposed direction

Provide a compact, stable agent-control envelope that clearly separates:

- tool execution status;
- release/merge decision;
- publication permissions;
- completion permission;
- next actor and exact next action;
- paths and hashes for detailed forensic artifacts.

Example:

```json
{
  "schema_version": "shipgate.agent_control/v1",
  "execution_status": "succeeded",
  "control_state": "human_review_required",
  "ready_to_update_pr": true,
  "ready_to_merge": false,
  "completion_allowed": false,
  "next_actor": "human",
  "next_action": {
    "kind": "review",
    "surface": "github_pull_request",
    "why": "..."
  },
  "artifacts": {
    "receipt": {"path": "...", "sha256": "..."},
    "verifier": {"path": "...", "sha256": "..."}
  }
}
```

Detailed verifier/report JSON should remain available by path, not dominate the default agent stdout.

## Acceptance criteria

- [ ] A compact format is available and promoted for coding-agent control loops.
- [ ] The result explicitly answers: may edit, may commit, may push/update PR, may merge, may report complete?
- [ ] Execution success cannot be mistaken for release readiness.
- [ ] Human-facing text leads with the operational state and next actor.
- [ ] Full forensic artifacts remain content-addressed and discoverable.
- [ ] Agent-mode errors and nonterminal results include an exact executable next action.
- [ ] The compact control payload has a documented size budget and does not duplicate the full report.
- [ ] Exit-code semantics are documented and tested against `passed`, `review_required`, `insufficient_evidence`, and `blocked`.

## Success measure

A user or coding agent should determine the permitted next action without reading any other artifact or interpreting prose.

pengfei-threemoonslab:
Closes #333. Absorbs the design half of #323 (that issue stays open as the horizontal-coverage checklist for `init`/`doctor`/`detect`). Completes the v0.17 foundation of #338, after #339, #336, and #335.

## The problem, restated from #338

A verify run could report all four of these at once:

```
execution: "succeeded"        head_exit_code: 0
release_decision.decision: "review_required"
control.state: "human_review_required"
```

Three of them read like permission to continue. Answering "what may I do next?" correctly meant reading `current-control.json` for currency, `agent-handoff.json` for the route, `verifier.json` for execution status, and `report.json` for the gate — four artifacts, and the two most confusable facts sat thousands of tokens apart inside one of them.

## What this adds

`shipgate.agent_control/v1` ([`docs/agent-control-schema.v1.json`](docs/agent-control-schema.v1.json)) — one compact stdout object carrying tool execution status, the release or boundary decision **and which engine produced it**, the control state, the six-way `permissions` vector, `verify_required`, the next actor, the exact next action, and the path + sha256 of every forensic artifact.

Measured on this repo's own clean verify: **2.8 KB**, against 8.1 KB for `verify --json` and 9.3 KB for the `current-control.json` + `agent-handoff.json` pair it replaces.

Three separations that were documentary are now structural, enforced by the schema rather than by consumer discipline:

| Fact | Field | Enforced |
|---|---|---|
| Did the tool run? | `execution` | a **failed** execution can never be `complete` — and a *succeeded* one implies nothing, which is the direction readers get wrong |
| Did the gate fail? | `exit_code` | mode-dependent; advisory exits 0 even on `blocked`, so it is never merge authority |
| May I act? | `permissions` | `merge` and `report_complete` are true exactly when `control_state == "complete"` |

Emitted by three commands, all through one projection in `core/agent_control_envelope.py`:

- `agents-shipgate verify --format control` — **added**. `--json` / `--format json` unchanged; agent-mode auto-detection still resolves to `json`.
- `agents-shipgate check --format agent-control-json` — **added**. `agent-boundary-json` unchanged and still the default.
- `agents-shipgate agent control` — **default output changed** to the envelope; `--format pointer` returns the previous output byte for byte.

`verify --format text` now leads with the control state, next actor, and permission vector before the existing `Agents Shipgate verify: <verdict>` line, which is unchanged.

## It decides nothing

Every field is copied from a producer that already published it; the validators only assert the copies cannot contradict each other. Two places that look like judgement are both refusals:

- When the current-control pointer **downgraded** a completion its run claimed — the receipt refusal from #339 — the pointer wins and the run's route is dropped. Reading the run's own optimistic `control` block instead would undo exactly the refusal that makes stale authorization impossible.
- When no route can be recovered (a `scan` pointer binds no verifier), `agent control` exits 4 with the exact verify command rather than synthesizing a step that cannot reproduce the subject.

## Surface-discipline gate (`CONTRIBUTING.md`)

1. **Which headline metric?** Activation — "percentage of issue-to-PR tasks completed without human terminal intervention", and "median control-payload tokens consumed by the coding agent", both named in #338. The 2.8 KB vs 9.3 KB reduction is pinned by a test; the routing question now has one answer instead of a read order.
2. **Can an existing surface carry it?** Partly, which is why this is one schema and not two. `agent-handoff.json` already carries the control block, and `current-control.json` already carries currency — but neither carries both, and the pointer deliberately carries no route because reproducing one there would make it a second decision. The envelope is the join, and it consolidates three read paths into one rather than adding a fourth. It is stdout-only and adds no artifact to the reports directory, because doing so would deepen the read-order problem it exists to remove.
3. **Roadmap non-goals?** No second verdict: `permissions` is carried verbatim from the `AgentControl` union `derive_agent_control` fixed, and nothing here gates independently of `release_decision.decision`. No agent execution, no LLM calls, no network.

## Size budget

`agent_control_max_bytes` is published in the contract (4096) and pinned by tests. It is deliberately **not** a validator: a long reviewer list or an unusually long exact command can exceed it, and dropping either to hit a size target removes something safety-relevant. Prose fields are capped at 400 characters with a deterministic marker; commands, paths, hashes, and reviewer names are never truncated.

## Contract

Runtime contract `21 → 22`; downstream local contract `9 → 10`.
`minimum_control_contract_version` **stays at 21** — this adds a projection of the `AgentControl` union and does not change the union, so every v21 consumer keeps reading the same fields.

## Deliberately not in scope

Flipping verify's agent-mode stdout default from `json` to `control` would move the token metric furthest, but silently changing what an already-installed agent receives is a compatibility event. It belongs to #323's command-by-command rollout, alongside `init`/`doctor`/`detect`, whose control state derives from setup facts rather than a decision — `decision_source` already carries a `setup` value for them.

## Verification

- New `tests/test_agent_control_envelope.py` (31 tests): projection fidelity across all four control states, every contradiction the schema must reject, prose truncation that never touches a command, the pointer-overrules-run reconciliation against a real verifier artifact, the routeless refusal, the size budget, and end-to-end runs of all three commands.
- New exit-code matrix in `tests/test_release_decision.py` pinning all four decisions × advisory/strict.
- Full suite green; `scripts/generate_schemas.py --check` clean; `ruff check` clean.


src/agents_shipgate/cli/verify/command.py
```diff
@@ -496,11 +542,69 @@ def _parse_verify_format(value: str) -> str:
...
+    where the two disagree. Its directory is recovered from the artifact map
+    rather than threaded through ``run_verify``: ``verify`` resolves the reports
+    directory against the Git root through several branches, and re-deriving it
+    from the paths the run actually wrote cannot drift from them.
+    """
+
+    pointer = None
+    reports_dir = _reports_dir_from_artifacts(verifier, workspace)
+    if reports_dir is not None:
+        pointer = load_published_control_pointer(reports_dir)
```
pengfei-threemoonslab: **[P1] Validate the live workspace before granting authority.** This only parses the published pointer; it does not perform the live HEAD/tree/worktree validation used by `agent control`. I reproduced a clean `complete` run, modified a tracked `tools.json`, and then obtained `control_state=complete` with `permissions.merge=true` here, while an immediate `agent control` correctly failed with `workspace_changed`. Use the generation-safe live-workspace read and fail closed on drift.

src/agents_shipgate/cli/agent_interface.py
```diff
@@ -178,9 +220,66 @@ def control(
...
+    returning, and rejected the read outright if the pointer moved underneath
+    it.  A parse failure here is therefore a malformed artifact rather than a
+    stale one, and resolves to "no route" rather than a crash.
+    """
+
+    ref = pointer.artifacts.get("verifier")
+    if ref is None:
+        return None
+    try:
+        return VerifierArtifact.model_validate_json((reports_dir / ref.path).read_bytes())
```
pengfei-threemoonslab: **[P1] Keep the verifier read in the validated generation.** `read_current_control()` has returned, but this independently reopens `verifier.json` without rehashing it or confirming that `current_control_id` is unchanged. A concurrent run can therefore combine pointer A with verifier B; I reproduced A's identity with B's request, decision, and permissions. Capture the validated verifier bytes in the generation-safe read, or rehash and reconfirm the pointer afterward, and bind the request/decision identities as well as the state tag.

src/agents_shipgate/schemas/agent_control_envelope.py
```diff
@@ -0,0 +1,265 @@
...
+    def _projection_is_self_consistent(self) -> AgentControlEnvelope:
+        """Reject copies that contradict the producers they were copied from.
+
+        Each rule below restates an invariant the authoritative producer already
+        guarantees. Restating them here is what makes this a projection with
+        teeth: a future emitter that assembles the envelope by hand cannot
+        publish a shape the union would have refused.
+        """
+
+        completion = self.control_state == "complete"
```
pengfei-threemoonslab: **[P1] Enforce the full state-specific route contract.** The flattened model still accepts contradictory shapes—for example `human_review_required` with a coding-agent action/actor, `complete` with another action, or `review_publishable` without publication authority. Preserve discriminated state variants or validate the exact next actor, action kind, verification obligation, stop semantics, human-review shape, and permission vector for every control state.

pengfei-threemoonslab (commented):
## Engineering review — changes required

Reviewed and reproduced against head `9a38d6d9a1f08925617aabfafb2dd8b63960cabf`.

I do not recommend merging this revision yet. The compact envelope is a useful direction, but the current implementation has five high-priority correctness/authority issues and three medium-priority contract/CLI issues. The most serious paths can (1) emit merge/completion authority after the live workspace has changed and (2) splice a validated pointer with a verifier from a different concurrent generation, potentially widening publication permissions.

The inline comments cover the required fixes:
- live-workspace validation for `verify --format control`
- generation-atomic pointer/verifier reads and identity binding
- complete state/action invariants in both Pydantic and the published JSON Schema
- the broken downstream local-contract v9→v10 migration
- shipped adoption recipes that still parse the old default output
- invocation-relative artifact paths
- structured error handling for projection failures
- the unenforced 4096-byte “maximum”

GitHub Actions and the focused local suite are green, but they do not exercise these race, migration, schema-conditional, workspace-drift, or non-workspace-CWD cases. The exact PR source produced a valid Shipgate receipt, but the resulting control remained `review_publishable` with `permissions.merge=false`, `report_complete=false`, and required reviewer `agent-platform`.

scripts/generate_schemas.py
```diff
@@ -1464,6 +1465,25 @@ def build_current_control_schema() -> tuple[Path, str]:
     )
 
 
+def build_agent_control_envelope_schema() -> tuple[Path, str]:
+    from agents_shipgate.schemas.agent_control_envelope import AgentControlEnvelope
+
+    return _verification_identity_schema(
+        model=AgentControlEnvelope,
```
pengfei-threemoonslab: **[P1] Publish the cross-field safety rules in JSON Schema.** Pydantic model validators are not represented by this generated schema. Draft 2020-12 validation currently accepts examples such as `execution=failed` with `control_state=complete`, and human-review states with broad permissions, despite the PR describing these separations as schema-enforced. Add explicit `allOf`/`if`/`then` constraints or generate the schema from state-discriminated variants.

src/agents_shipgate/cli/discovery/local_contract.py
```diff
@@ -59,7 +64,7 @@
 )
 from agents_shipgate.schemas.verifier import VerifierArtifact
 
-LOCAL_CONTRACT_SCHEMA_VERSION = "9"
+LOCAL_CONTRACT_SCHEMA_VERSION = "10"
```
pengfei-threemoonslab, line 67: **[P1] Preserve the v9→v10 managed-contract upgrade.** The safe-update prior-render allowlist does not include the outgoing pristine v9 render hash (`8aa6004936f19c31607498eb7d603b5f42ee55443e31ccc8b8099b14196eb8ab`). As a result, upgrading a real generated v9 file returns `skipped_user_modified`/exit 2 and leaves schema 9. Add the hash and a test that starts from the base revision's generated v9 file.

src/agents_shipgate/cli/agent_interface.py
```diff
@@ -130,16 +139,49 @@ def control(
         "--reports-dir",
         help="Directory holding current-control.json.",
     ),
+    format_: str = typer.Option(
+        "control",
```
pengfei-threemoonslab, line 139: **[P1] Update every shipped reader before changing this default.** The repo-scoped and bundled Codex skill recipes still run bare `agents-shipgate agent control --workspace .` and then read pointer-only fields (`lifecycle_state` and nested `control.state`). This default returns root `control_state` instead, so the installed adoption workflow fails at its first authority check. Either add `--format pointer` to those recipes or migrate all bundled/generated copies to the envelope fields, with content-level tests.

src/agents_shipgate/cli/verify/command.py
```diff
@@ -496,11 +542,69 @@ def _parse_verify_format(value: str) -> str:
...
+        verifier,
+        operation="preview" if preview else "verify",
+        source="run",
+        exit_code=exit_code,
+        pointer=pointer,
+        # The verifier records artifact paths the way the run displays them —
+        # relative to the Git root, absolute only when the directory sits
+        # outside it. Reusing that spelling keeps the envelope's paths openable
+        # from the same place every other emitted path is.
+        artifact_root=_artifact_root(verifier),
```
pengfei-threemoonslab: **[P2] Emit paths relative to the documented invocation directory.** These are Git-root-relative, not relative to the directory where the command ran. With the caller outside `--workspace`, `Path(envelope.artifacts[name].path)` does not exist; only `workspace / path` does. Emit absolute paths or correctly rebase them to `Path.cwd()`.

src/agents_shipgate/cli/verify/command.py
```diff
@@ -418,12 +430,41 @@ def verify(
 
     if stdout_format == "json":
         typer.echo(json.dumps(verifier.model_dump(mode="json"), indent=2))
+    elif stdout_format == "control":
+        typer.echo(
+            render_agent_control_envelope(
+                _verify_envelope(verifier, workspace, exit_code, preview=preview)
```
pengfei-threemoonslab, line 487: **[P2] Keep control projection inside the structured error boundary.** An invariant or renderer failure here occurs after `verify`'s exception handling and exits 1 with a bare exception. That violates the published agent-mode policy for internal failures (`internal_error`, exit 4). Wrap projection/rendering in the command's structured internal-error path.

src/agents_shipgate/schemas/agent_control_envelope.py
```diff
@@ -0,0 +1,265 @@
...
+# ~9 KiB `verifier.json` plus ~6 KiB `agent-handoff.json` an agent reads today
+# to answer the same question.
+#
+# It is a *budget*, pinned by `tests/test_agent_control_envelope.py`, and
+# deliberately not a validator. Two fields can exceed it in principle — a very
+# long reviewer list, or an unusually long exact command — and in both cases
+# dropping content to meet a size target would remove something safety-relevant.
+# A control surface that raised because a Git error message ran long would deny
+# an answer at the moment the caller most needs one.
+MAX_AGENT_CONTROL_ENVELOPE_BYTES = 4096
```
pengfei-threemoonslab: **[P2] Do not publish this as a maximum unless it is enforced.** Valid envelopes can exceed 8–15 KB through Unicode prose, reviewer lists, references, or exact commands; character truncation also undercounts JSON escape expansion. Enforce the final serialized-byte cap, or rename this to a target/budget and remove contract language promising output “within” 4096 bytes.

src/agents_shipgate/cli/check.py
```diff
@@ -42,6 +46,40 @@
...
+    vector are the ones the result already carries.
+
+    ``evaluated`` is false on the diff-input-error routes, where a considered
+    ``block`` is emitted without any diff having been read. The envelope reports
+    that as ``execution: "failed"``, matching what ``verify`` reports for the
+    same condition.
+    """
+
+    if format_ == "agent-control-json":
+        envelope = envelope_from_agent_result(
```
pengfei-threemoonslab, line 76: **[P1] Bind compact authority to the assessed input and its obligations.** `envelope_from_agent_result()` drops `audit_id`, `subject`, `changed_files`, and `pending_review[]` while retaining terminal permissions. On this head, `docs_only.diff` and `python_refactor.diff` produced byte-identical `complete` envelopes with `merge=true`, `current_control_id=null`, and `artifacts={}`. An unknown Codex permission also disappeared from the compact form even though `next_action.why` tells the caller to report the pending items. Carry a compact audit/input identity plus the review obligations (or an immutable reference to them), or do not grant terminal authority from this unbound projection.

src/agents_shipgate/core/agent_control_envelope.py
```diff
@@ -0,0 +1,380 @@
...
+    action = envelope.next_action
+    if action is not None:
+        lines.append(f"Next: {action.why}")
+        if isinstance(action, CodingAgentCommandAction):
+            # Deliberately not the bare ``Run:`` prefix that
+            # `primary_evidence_remediation_text` uses for the evidence-gap
+            # rerun. #358 requires the human work to precede that line, and
+            # these headline lines print first — reusing the prefix would put a
+            # `Run:` above the remediation it is not the remediation for.
+            lines.append(f"Next command: {action.command}")
```
pengfei-threemoonslab: **[P1] Escape control characters in the human renderer.** `why`, `command`, and `expects` can contain embedded newlines and ANSI controls, but these values are interpolated directly into authority-oriented terminal output. I reproduced `verify --preview --format text` with a workspace path containing newlines and obtained forged `Control: complete` and `You may: merge` lines after the real denial. Preserve the exact value in JSON, but render control characters visibly and keep each human-output field on one line.

src/agents_shipgate/cli/agent_interface.py
```diff
@@ -178,63 +217,79 @@ def control(
...
+        typer.echo(f"Current control carries no route: {exc}", err=True)
+        emit_agent_mode_error(
+            "other_error",
+            message=str(exc),
+            exit_code=4,
+            next_action=guidance,
+            next_actions=[
+                NextAction(
+                    kind="command",
+                    command=COMMANDS["verify_pr"],
```
pengfei-threemoonslab: **[P2] Distinguish a current-but-routeless generation and preserve its subject.** A valid pointer produced by `scan` passes the generation-safe read, and `--format pointer` returns it with exit 0, but the default format reaches this branch and exits 4. That contradicts the public contract that any nonzero exit means no current identity exists. The recovery command also hardcodes `--workspace .`, `shipgate.yaml`, `origin/main`, and the default reports directory, discarding the subject that was just validated. Represent current-but-routeless explicitly, and generate any recovery action from the requested workspace/reports/ref context.

src/agents_shipgate/core/agent_control_envelope.py
```diff
@@ -0,0 +1,380 @@
...
+    )
+
+
+def _artifact_refs(
+    pointer: CurrentControlPointer | None,
+    artifact_root: str | None,
+) -> dict[str, AgentControlArtifactRef]:
+    if pointer is None:
+        return {}
+    root = (artifact_root or "").strip().rstrip("/")
```
pengfei-threemoonslab: **[P2] Preserve the filesystem root when composing artifact paths.** String trimming changes valid path identities: an `artifact_root` of `/` becomes empty and emits `verifier.json` relative to the invocation directory, while `/tmp/reports ` loses its significant trailing space. The envelope can therefore exit successfully with a path that does not name the artifact whose hash was validated. Join paths structurally without stripping whitespace or collapsing the root directory.

docs/agent-contract-current.md
```diff
@@ -10,6 +10,48 @@ Verify the installed CLI contract locally before relying on hard-coded docs:
...
 ```
 
+Runtime contract v22 publishes `shipgate.agent_control/v1`, the compact control
+envelope. It is a **projection of the control state, not a second decision**:
+every field is copied from a producer that already published it, and the
+schema's validators only assert that the copies cannot contradict each other.
+One object answers the whole routing question — tool execution status, the
+release or boundary decision and which engine made it, the control state, the
+six-way `permissions` vector, who acts next, the exact next action, and the
+content-addressed path and hash of every forensic artifact. It is emitted by
```
pengfei-threemoonslab: **[P2] Narrow the artifact-completeness promise.** The envelope does not carry every forensic artifact: `check --format agent-control-json` intentionally emits `artifacts={}`, and verify's pointer allowlist omits generated files such as `verification-input.diff`, `verification-unit-result.json`, and `capabilities.lock.json`. Say that the envelope lists every artifact bound by `current-control.json`, or expand the pointer/envelope binding and test it against the actual run artifact inventory.

pengfei-threemoonslab (commented):
## Engineering second-pass review — changes still required

Revalidated against current head `6835e17b95cf1b37d1f252c1d06605170a8f6eaf`, including the new review-fix commit. The new commit addresses several earlier findings; the five inline comments below are limited to defects that still reproduce on this head.

I still do not recommend merging this revision. The two P1 findings leave compact authority unbound to the input it assessed and allow human-readable output to be spoofed with forged authority lines. The remaining comments cover a valid current generation being reported as unavailable, path corruption, and an artifact-completeness promise the emitters do not satisfy.

Focused updated tests pass, but they do not cover the reproduced cross-input replay, dropped pending-review obligations, newline/control-character rendering, or current-but-routeless scan behavior.

src/agents_shipgate/cli/verify/command.py
```diff
@@ -496,11 +578,131 @@ def _parse_verify_format(value: str) -> str:
...
+            source="run",
+            execution=verifier.execution,
+            exit_code=exit_code,
+            reason=(
+                "The control identity this run published binds no verifier "
+                "artifact, so no validated route could be recovered."
+            ),
+        )
+    current = VerifierArtifact.model_validate_json(captured)
+    return envelope_from_verifier(
```
pengfei-threemoonslab, line 670: **[P1] Bind `--format control` to this invocation.** This validates the captured verifier from whichever generation is current, but never checks its `(request_id, decision_id)` against the in-memory `verifier` for this command. I reproduced preview run A becoming `operation=preview`, `complete`, `decision=passed`, and `merge/report_complete=true` after passing run B republished the directory; stdout also carried B's head exit code while the process exited with A's. Require an exact identity match (or retain A's published pointer) and fail closed on mismatch.

src/agents_shipgate/core/current_control.py
```diff
@@ -510,7 +520,11 @@ def read_current_control(
         _validate_control_currency(out_dir, pointer, live)
         confirmation = _load_pointer(out_dir, path)
         if confirmation.current_control_id == pointer.current_control_id:
-            return CurrentControlRead(pointer=pointer, path=path)
+            # `captured` was read in the same pass that hashed it, and the
+            # pointer has not moved since. Returning the bytes is what lets a
+            # caller route on this artifact without opening it again outside
+            # the protocol.
+            return CurrentControlRead(pointer=pointer, path=path, artifacts=captured)
```
pengfei-threemoonslab: **[P1] Re-observe the workspace before returning authority.** `live` was snapshotted before entering this loop, but the protocol only reconfirms pointer identity. Advancing HEAD after `live_workspace()` and before this return still yielded a `complete` pointer. This race is reachable through both CLI surfaces; re-read HEAD/tree/worktree within the generation protocol and require a stable pointer/workspace pair.

src/agents_shipgate/schemas/agent_control_envelope.py
```diff
@@ -0,0 +1,389 @@
...
+    """The coding agent may merge and report the task complete.
+
+    Every authority-bearing field is pinned by the tag. ``execution`` narrows to
+    the statuses a completed evaluation can have, which is what makes
+    "a failed run can never authorize completion" a schema rule rather than a
+    validator a JSON Schema consumer never runs.
+    """
+
+    control_state: Literal["complete"]
+    execution: CompleteExecution
```
pengfei-threemoonslab, line 339: **[P1] Constrain terminal authority by producer and provenance.** Both Pydantic and the generated schema accept a `complete` envelope with `operation=scan`, `execution=not_run`, no decision, no current identity/artifacts, and full merge authority. No producer can authoritatively emit this shape. Split terminal variants by operation/source (check versus run/refresh), or add equivalent schema-visible provenance constraints.

src/agents_shipgate/schemas/agent_control_envelope.py
```diff
@@ -0,0 +1,389 @@
...
+    """One exact coding-agent-owned step remains.
+
+    ``permissions`` admits two vectors because the route decides: a step that
+    runs before any diff was read authorizes nothing beyond itself, while a step
+    taken on an evaluated change keeps publication authority. Both deny merge.
+    """
+
+    control_state: Literal["agent_action_required"]
+    permissions: PublishOnlyPermissions | NoAgentPermissions
+    verify_required: bool
```
pengfei-threemoonslab, line 400: **[P1] Preserve the verify-action obligation.** `next_action.kind=verify` remains valid with `verify_required=false` in both Pydantic and JSON Schema, although the authoritative `AgentActionRequiredControl` rejects it. Make this implication structural in the envelope variant and published schema.

src/agents_shipgate/cli/discovery/agent_instructions/apply.py
```diff
@@ -533,11 +548,16 @@ def _is_managed_local_contract(existing: bytes) -> bool:
...
+    if (
+        payload.get("gating_signal") != GATING_SIGNAL
+        or default_paths.get("local_contract") != LOCAL_CONTRACT_RELATIVE_PATH
+    ):
+        return False
+    try:
+        version = int(str(payload.get("schema_version")))
+    except (TypeError, ValueError):
+        return False
+    return 0 < version <= int(LOCAL_CONTRACT_SCHEMA_VERSION)
```
pengfei-threemoonslab: **[P1] Do not overwrite human-modified older contracts.** This recognizes any older JSON carrying the common gating signal/path as managed, regardless of its bytes. Changing the real v9 fixture's `commands.verify_pr` still produced `updated` and erased the customization. Exact v8/v9 render hashes now cover pristine migrations; modified files should remain `skipped_user_modified` unless an explicit authenticated managed marker exists.

.agents/skills/agents-shipgate/SKILL.md
```diff
@@ -20,7 +20,7 @@ Do not use it for general linting, runtime monitoring, evals, model-output quali
...
...8. For full PR verification, validate `agents-shipgate-reports/verification-receipt.json` first, then read `agent-handoff.json` and switch on `control.state`; read `verifier.json` for detailed control state, `verify-run.json` for the request graph, and `report.json` for reviewer detail. `report.json.release_decision.decision` remains the release gate. Refresh with `agents-shipgate agent control --workspace .` — which refuses the read when HEAD, the tree, or the working tree has moved since the decision, and otherwise prints one `shipgate.agent_control/v1` object carrying `control_state`, `permissions`, `next_actor`, and `next_action` (routing on that object alone is enough; `--format pointer` returns the raw `current-control.json`) — before you act on any of that, and again before enforcing a cached `must_stop`, before commit/push/PR update, before merge, and before declaring the task complete. A non-zero exit means no control identity is current and you hold no authority; if `current_control_id` changed, discard every cached control state and restart from the new identity. A result remembered from earlier in the conversation never outranks the current pointer, in either direction.
```
pengfei-threemoonslab: **[P1] Do not require v22 behavior after accepting v21.** The workflow accepts `minimum_control_contract_version: 21`, then assumes bare `agent control` returns `shipgate.agent_control/v1` and supports `--format pointer`. Runtime v21 returns the raw pointer and has no `--format`. Branch on contract/stdout schema or require contract 22 for this reader; add a v21 fixture.

src/agents_shipgate/cli/verify/command.py
```diff
@@ -416,14 +436,72 @@ def verify(
...
             if verifier.release_decision is not None
             else ("skipped" if verifier.head_status == "skipped" else "failed")
         )
+        # Lead with the operational answer. The release verdict below is the
+        # gate's word on the change; these lines are the reader's word on what
+        # they may do next, and printing the verdict first is what let
+        # "succeeded" read as "done" in the #338 walkthrough. A drift refusal
+        # arrives here as a denying envelope, not an exception, so the human
+        # still sees the verdict and the exit code.
+        for line in control_headline_lines(
```
pengfei-threemoonslab, line 502: **[P1] Escape every repository-derived text line.** The new control headline is sanitized, but the later `primary_evidence_remediation_text(...)` output remains raw. A repo-controlled tool name containing newline/control bytes printed forged `Control: complete` and `You may: merge, report_complete` lines. Apply the single-line/control escaping to all human text output, not just the headline.

src/agents_shipgate/schemas/agent_control_envelope.py
```diff
@@ -0,0 +1,389 @@
...
+    "reason",
+    "current_control_id",
+    "artifacts",
+]
+
+BoundedProse = Annotated[
+    str,
+    StringConstraints(strip_whitespace=True, min_length=1),
+]
+NonEmptyText = Annotated[str, StringConstraints(strip_whitespace=True, min_length=1)]
```
pengfei-threemoonslab: **[P1] Publish a non-whitespace input identity constraint.** `StringConstraints(strip_whitespace=True)` is not represented in Draft 2020-12 schema; schema-only consumers accept `input_id=" "` on a `complete` merge-authorizing envelope while Pydantic rejects it. Use an explicit non-whitespace/identity pattern and dual-layer parity tests; the same schema-fidelity gap affects prose and reviewer uniqueness.

src/agents_shipgate/core/agent_control_envelope.py
```diff
@@ -0,0 +1,493 @@
...
+_PERMISSION_ORDER = ("edit", "commit", "push", "update_pr", "merge", "report_complete")
+
+
+def _reconcile_with_pointer(
+    control: AgentControl,
+    pointer: CurrentControlPointer | None,
+) -> AgentControl:
+    """Let the pointer overrule a run that claimed more than it can bind."""
+
+    if pointer is None or pointer.control.state == control.state:
```
pengfei-threemoonslab: **[P2] Do not reconcile on state alone.** With a pointer carrying `agent_action_required`/NoPermissions and a verifier carrying the same state/PublishOnly, this returns the verifier permissions and widens publication authority. Compare permission and route invariants too; any same-state disagreement should fail closed.

src/agents_shipgate/cli/agent_interface.py
```diff
@@ -178,63 +219,114 @@ def control(
...
+    if data is None:
         return None
-
-
-def _safe_merge_base(root: Path, base: str, head: str) -> str | None:
     try:
-        return merge_base_sha(root, base, head)
-    except Exception:  # noqa: BLE001 - an unresolvable range is drift.
+        verifier = VerifierArtifact.model_validate_json(data)
+    except ValueError:
```
pengfei-threemoonslab: **[P2] A malformed verifier is not a routeless scan.** Returning `None` on parse failure sends any terminal pointer through the routeless branch. A hash-bound `{}` verifier under a terminal verify pointer produced exit 0, `execution=not_run`, and retained update-PR authority. Only an actual scan with no verifier binding should use routeless recovery; malformed verify/preview artifacts must invalidate the generation.

src/agents_shipgate/cli/agent_interface.py
```diff
@@ -178,63 +219,114 @@ def control(
...
+            exit_code=bound_verifier.head_exit_code,
+            artifact_root=reports_dir.as_posix(),
+        )
+    elif result.pointer.lifecycle_state == "terminal":
+        # Current, but published by a command that reaches no release decision.
+        # Refusing conflated "nothing is current" with "what is current cannot
+        # authorize a merge"; only the first justifies a non-zero exit.
+        envelope = envelope_from_routeless_pointer(
+            result.pointer,
+            verify_command=verify_command_for(workspace, None),
```
pengfei-threemoonslab: **[P2] Preserve the subject in the recovery route.** A scan using a nested `--config` and custom `--out` emits a verify command that preserves only the workspace. Following it checks the root manifest, writes another reports directory, and leaves the pointer being refreshed unchanged. Bind/recover config and output identity, or return a non-command route when the exact subject cannot be reproduced.

pengfei-threemoonslab (commented):
## Engineering third-pass review — changes still required

Revalidated against current head `6194d7331e7b99af628d0b81e3d45b76555ccb95`, including the latest fix commit. The update resolves a substantial majority of the earlier findings, but I still do not recommend merging this revision. The inline findings below are limited to defects that remain reproducible on this exact head.

The highest-risk issue is still generation binding: `verify --format control` can return another concurrent run's passing/merge authority while labeling it as this invocation's `source="run"`. The live-workspace comparison also has a remaining TOCTOU window, and the envelope/schema still admit authority shapes that no authoritative producer can emit. Compatibility and safe-update regressions remain in the shipped v21 reader and local-contract migration.

Reproduction coverage included two interleaved verifier generations, a deterministic concurrent HEAD change, Pydantic versus Draft 2020-12 mutation matrices, a human-modified real v9 fixture, malformed bound verifier bytes, and a custom-config/custom-reports scan route.

All three GitHub workflows are green. Focused control/adoption tests, schema generation, lint, diff checks, and packaging tests pass, but none cover the reproduced cases below. The current Shipgate result is still `review_publishable`; merge and completion remain denied pending human review.

STABILITY.md
```diff
@@ -13,6 +13,80 @@ for reproducible CI.
...
+
+New: `shipgate.agent_control/v1`
+([`docs/agent-control-schema.v1.json`](docs/agent-control-schema.v1.json)), a
+compact control envelope emitted on stdout by three commands. It answers the
+whole routing question in one object — tool execution status, the release or
+boundary decision and which engine produced it, the control state, the six-way
+`permissions` vector, who acts next, the exact next action, and the
+content-addressed path and hash of every artifact `current-control.json` binds
+(`check` publishes no pointer and binds none) — within
+`agent_control_max_bytes` (4096). It is **not** written to disk, and it decides
```
pengfei-threemoonslab: **[P2] Finish the size-budget rename.** This normative migration note still promises output within nonexistent `agent_control_max_bytes`, although the runtime now publishes a non-enforced `agent_control_budget_bytes`. Use budget/target wording consistently.

PR #366, opened 2026-08-11
fix(trigger): match prompts/ and policies/ at any depth

### LINKED ISSUE #311: bug(trigger): prompts/policies globs are top-level only, diverging from trust-root coverage
## Summary

`TRIGGER-PROMPTS-OR-POLICIES` matches only **top-level** `prompts/` and `policies/`. Any repository that keeps prompts or policy YAML under a subdirectory gets no trigger match for them.

This is language-agnostic — it affects Python repositories today — and is separable from the Ruby/RubyLLM routing work in #309, where it was first observed.

## Observed

`docs/triggers.json` (`schema_version` 0.2):

```json
{
  "id": "TRIGGER-PROMPTS-OR-POLICIES",
  "when": { "any_of": [ { "glob": "prompts/**" }, { "glob": "policies/**" } ] },
  "action": "run_shipgate"
}
```

The trust-root surface list in `src/agents_shipgate/checks/verify_policy.py:55` is recursive for the same concept:

```python
_POLICY_SURFACES = (
    "**/shipgate.yaml",
    "**/policies/**",
    "**/.agents-shipgate/**",
)
```

So `SHIP-VERIFY-POLICY-WEAKENED`'s fail-safe treats `<subdir>/policies/**` as a trust root, while the trigger catalog does not treat it as a reason to run at all. The two surfaces disagree about the same paths.

## Impact

A PR editing `enterprise/lib/captain/prompts/**` or `services/foo/policies/**` reaches `no_match` and is reported as "nothing in this PR signals a tool-surface change" — for a path the verifier elsewhere considers governance-bearing. Prompts and policy YAML carry approval and confirmation evidence, per the rule's own rationale.

## Requested change

- Make the trigger globs recursive: `**/prompts/**` and `**/policies/**`, keeping the existing top-level matches.
- Mirror the same change in `TRIGGER-DOCS-ONLY-NEGATIVE`'s `none_match_glob` list, which currently carries the same two non-recursive entries — otherwise a nested prompt edit bundled with a docs edit still classifies as docs-only and skips.
- Add a parity test asserting that every path in `_POLICY_SURFACES` that a trigger rule is meant to cover is matched by that rule, so the two lists cannot drift again.
- Add fixture cases for a nested `prompts/` edit and a nested `policies/` edit.

## Notes

This is a widening of an existing rule's globs within the current `schema_version`, not a new state or a new rule ID, so it does not open the surface-discipline gate. It does change trigger outcomes for repositories with nested layouts — from skip to run — which is the intended direction (toward evaluation, not away from it).

Split out of #309.

pengfei-threemoonslab:
## Summary

A path the verifier calls a governance trust root could reach the trigger catalog as `no_match` — "nothing in this PR signals a tool-surface change." This closes that divergence and the surfaces that copy it.

Closes #311.

### The gap, and what each round of review added

`TRIGGER-PROMPTS-OR-POLICIES` matched `prompts/**` and `policies/**` at the repository root only, while the verifier's trust-root classification reads the same two surfaces at any depth. Review then established that recursion alone was not the whole divergence:

| changed path | before | now |
| --- | --- | --- |
| `services/foo/policies/refund.yaml` | `no_match` | runs |
| `enterprise/lib/captain/prompts/system.md` | **`skip_rule`** | runs |
| `services/foo/Policies/refund.yaml` | `no_match` | runs |
| `enterprise/lib/captain/Prompts/system.md` | **`skip_rule`** | runs |
| `README.md` + any of the above | `skip_rule` | runs |

Prompts failed worse than `no_match`: a nested `prompts/*.md` edit satisfies `TRIGGER-DOCS-ONLY-NEGATIVE`'s `every_file_matches: **/*.md` leg, so it did not merely fail to match — it actively **skipped**, and `skip_shipgate` beats `run_shipgate`. That is why mirroring the globs into `none_match_glob` is load-bearing rather than defensive.

### Case tolerance is directional, not uniform

Folding every predicate was wrong and introduced a fail-open (`src/TEST_agent.py` + `@function_tool` skipped). The rule is now: **fold the predicates that can only add evaluation, never the one that can subtract it.**

| predicate | case | why |
| --- | --- | --- |
| `glob` | insensitive | a wider match **adds** a run |
| `none_match_glob` | insensitive | guards a negative rule → wider match makes it fire **less** |
| `every_file_matches` | **sensitive** | it *is* the negative rule's classifier → wider match **subtracts** evaluation |

Stated at the predicate dispatcher, in the catalog's `predicate_vocabulary`, in `globbing`'s module docstring, in AGENTS.md and in the changelog.

### Propagation

One `glob_match_ci` helper is now shared by the trust-root classifier, the boundary registry, preflight's protected-surface specs, and the Tier B checks' changed-file selection. `_verify_common.touched()` had matched case-sensitively, so `services/foo/Policies/refund.yaml` was a Tier A policy trust root that produced no fail-safe finding, and a deleted `.github/workflows/Agents-Shipgate.yaml` was a `ci_gate` trust root that missed the critical gate-removal finding. All four `touched()` call sites use it positively, so folding is strictly fail-closed.

Surfaces that copy the routing follow, so the fix is end-to-end rather than evaluator-only:

- **pre-commit hooks.** Recursive `prompts/`, `policies/`, `.codex-plugin/`, `.agents/plugins/`, `.n8n/`, nested `AGENTS.md`/`CLAUDE.md`; the n8n and Conductor path legs that were silently omitted; bare `dir` for a `dir/**` glob; and `types: []` / `types_or: [file, symlink]`, because pre-commit's default `types: [file]` is applied *before* `files:` and drops a tracked symlink.
- **Both documented copy-paste snippets**, derived from the canonical regex and pinned clause-by-clause rather than against a hand-maintained sample.
- **`.cursor/rules/agents-shipgate.mdc`.** That rule is `alwaysApply: false`, so until a glob matched, a lone nested governance edit activated no Shipgate instructions at all. Renderer, shipped rule, target-repo snippet, benchmark setup variant and the adoption harness's lint constant all updated; outgoing render hash appended to `PRIOR_RENDER_SHA256` so an existing `init --write` repo receives it. The benchmark template and harness constant were two unenforced copies — both are now pinned to the renderer.

### The `on-tool-source-changes` recipes are retired

GitHub Actions, GitLab CI and CircleCI alike. A changed-path allowlist cannot gate Shipgate, for two independent reasons:

1. `TRIGGER-EXISTING-MANIFEST-PRESENT` is `force_run` — a repo with a `shipgate.yaml` is contracted to run on **every** PR, so the prefilter never saved the scan it advertised.
2. Every prefilter language involved matches case-sensitively while the catalog does not, so `- 'policies/**'` drops `services/foo/Policies/refund.yaml` with no job, no check and no signal — indistinguishable from a repo that never adopted the gate.

Run the advisory recipe on every PR and let the in-job evaluator decide; `verify` short-circuits before the scan on a skip verdict (`orchestrator.py:702`). A contract test now rejects `paths`, `paths-ignore`, GitLab `changes` and a CircleCI shell diff-gate in any shipped recipe, across `.yml` and `.yaml`.

### Guards

Hand-written fixtures only cover paths someone thought to write down, so the parity work is generated where it can be:

- **Trust-root ↔ catalog parity.** Every governance surface in `_POLICY_SURFACES` *and* `TRUST_ROOT_SURFACES` must carry a repo-root sample, a nested one and a case variant, and none may reach `no_match`. A surface added to either list without a decision fails.
- **Positive ↔ negative glob parity.** Asserted structurally, because a sample whose extension the negative rule never matches (`policies/*.yaml`) passes whether or not the exclusion exists.
- **Hook fixture exhaustiveness.** A catalog rule with a path leg must be listed or named in `_HOOK_PATH_TRIGGERS_EXCLUDED` with a reason. This is what surfaced the missing n8n and Conductor rules.
- **Generated hook sweep.** Representative paths are synthesized from all 67 positive catalog globs and boundary-adapter paths, with every `**` independently taking 0, 1 or 2 segments, and the hook must match every path the evaluator would run on. Each synthesized path is validated against its own source glob so the corpus cannot pass vacuously.
- **Both hook IDs.** Every hook check is parameterized over advisory and strict, plus an assertion that their `files:` expressions are identical.
- **Effective pre-commit selection**, modelling `types`/`types_or`/`files` rather than the compiled regex alone.

Every guard added in review was mutation-verified: the reviewer's reproduction was re-run against the fix and confirmed to fail, then restored.

### Known gaps, deliberately not closed here

A nested `shipgate.yaml` and a nested `.agents-shipgate/` are still routed at the repository root only. Both are anchored there by `BOUNDARY_ADAPTERS`' `exact_paths`, which check, verify, preflight and audit all share, so widening them is a registry change rather than a catalog edit. They are recorded in `_TRUST_ROOT_TRIGGER_GAPS` as a tripwire, not an endorsement: closing either gap makes the parity test fail with a message telling you to promote it out of the set. `is_dependency_manifest` also remains case-sensitive — it is not a governance surface. Happy to file follow-ups.

## Type

- [ ] Input adapter change
- [ ] Documentation only

## Verification

Additional local checks run:

- Full suite green (exit 0) on each pushed commit.
- `python -m ruff check .`, `python -m compileall -q src tests`, `python scripts/generate_schemas.py --check` — clean.
- `python scripts/build-llms-full.py` re-run and committed (AGENTS.md is a generation source).
- Every review reproduction re-run against its fix, then reverted: the `src/TEST_agent.py` fail-open, the `.+/\.app\.json` zero-segment mutation, a narrowed Conductor clause, a strict-hook-only clause deletion, restoring pre-commit's default `types`, and a re-added `paths:`/`paths-ignore`/`changes`/diff-gate prefilter on each provider.
- Tier A vs Tier B agreement checked directly for mixed-case policy, manifest, CI-gate, catalog and agent-instruction paths.

## Release-readiness notes

- [x] Report/schema changes are additive or documented in `STABILITY.md` — n/a. The trigger catalog stays at `schema_version` 0.3: this widens existing rules' globs and documents predicate case semantics, adding no rule ID, predicate or action. Rule rationale text is mirrored verbatim in five agent-protocol fixtures, all updated. Three `examples/` CI recipes are removed; `examples/` is illustrative surface, not a versioned contract, and each provider README carries a retirement note explaining why.


docs/triggers.json
```diff
@@ -244,11 +244,13 @@
       "when": {
         "any_of": [
           {"glob": "prompts/**"},
-          {"glob": "policies/**"}
+          {"glob": "**/prompts/**"},
```
pengfei-threemoonslab, line 247: **[P1] Preserve the verifier’s case-insensitive trust-root semantics.**

These recursive leaves still flow through case-sensitive `_glob_match`, while `checks.verify._classify` retries with `pattern.casefold()` / `path.casefold()` to prevent portable-path bypasses. At this head, `services/foo/Policies/refund.yaml` is classified as `policy` by the verifier but returns `no_match` here; `enterprise/lib/captain/Prompts/system.md` is classified as `prompts` but reaches `skip_rule` through the docs-only negative rule.

Please route `glob`, `every_file_matches`, and `none_match_glob` through the same case-insensitive path semantics and add mixed-case nested fixtures. Otherwise the claimed parity bug remains open.

docs/triggers.json
```diff
@@ -244,11 +244,13 @@
...
+          {"glob": "policies/**"},
+          {"glob": "**/policies/**"}
         ]
       },
       "action": "run_shipgate",
-      "rationale": "Prompts and policy YAML carry approval / confirmation evidence. Trace findings reference these surfaces."
+      "rationale": "Prompts and policy YAML carry approval / confirmation evidence. Trace findings reference these surfaces. The globs match at any depth so a nested workspace (`services/foo/policies/**`, `enterprise/lib/captain/prompts/**`) routes exactly like a repo-root one. That is parity with the trust-root surface list this rule mirrors — `SHIP-VERIFY-POLICY-WEAKENED` and the trust-root graph have always classified `**/policies/**` and `**/prompts/**` recursively, so a root-only trigger reported a governance-bearing edit as `no_match`. The `**/`-prefixed and bare forms are both listed because a consumer applying this catalog with a glob engine where `**/` requires at least one leading segment would otherwise lose the repo-root match."
```
pengfei-threemoonslab: **[P3] Attribute prompt protection to the correct verifier check.**

`SHIP-VERIFY-POLICY-WEAKENED`’s `_POLICY_SURFACES` contains `shipgate.yaml`, `policies/**`, and `.agents-shipgate/**`; it does not contain `prompts/**`. Recursive prompt protection comes from `SHIP-VERIFY-TRUST-ROOT-TOUCHED` / `TRUST_ROOT_SURFACES`. Please reword this machine-readable rationale and regenerate the five protocol fixtures so consumers do not rely on a check that never evaluates prompt edits.

docs/integrations.md
```diff
@@ -323,4 +323,4 @@ repos:
...
... and `policies/` match at any depth (`services/foo/policies/refund.yaml`), as do the nested protected copies covered by the boundary registry. Diff-only triggers (`TRIGGER-FUNCTION-TOOL-DECORATOR`, `TRIGGER-GOOGLE-ADK-AGENT-TOOLS-CHANGED`, and the diff-leg of `TRIGGER-SHIPGATE-CI-WORKFLOW`) are not covered by the regex pre-gate — pre-commit's `files:` regex is purely path-based. `TRIGGER-FRAMEWORK-VERSION-BUMP` needs a framework package token in the diff *in addition to* a changed dependency manifest, so the path-only regex cannot decide it either. Once the hook fires, the `verify` entry runs the full trigger evaluator (including diff rules) and base auto-detection itself. Use the GitHub Action for coverage on commits whose paths don't match the regex at all, or `python -m agents_shipgate.triggers --git-diff HEAD` for diff-aware local checks. The canonical hook manifest pre-commit reads from the repo root is [`/.pre-commit-hooks.yaml`](../.pre-commit-hooks.yaml) — it exposes `agents-shipgate`, `agents-shipgate-strict`, and `agents-shipgate-validate`. See [`examples/pre-commit/`](../examples/pre-commit/) for the longer write-up on advisory vs. strict modes and which hook ID to pick.
```
pengfei-threemoonslab: **[P2] Update the copy-paste local hook before promising recursive coverage.**

The snippet immediately above still uses root-only `prompts/.*` and `policies/.*`. A downstream user choosing the documented `repo: local` setup will therefore miss `services/foo/policies/refund.yaml`, even though this new paragraph says it matches at any depth. Please update those two clauses and extend `test_pre_commit_local_docs_show_same_path_trigger_clauses` to pin them.

tests/test_public_surface_contract.py
```diff
@@ -1310,6 +1310,142 @@ def test_triggers_code_plus_test_does_not_skip():
...
+        "The verifier would call this a trust-root edit while the catalog "
+        "says nothing in the PR signals a tool-surface change."
+    )
+
+
+@pytest.mark.parametrize(
+    "path",
+    [
+        "enterprise/lib/captain/prompts/system.md",
+        "services/foo/policies/refund.yaml",
```
pengfei-threemoonslab: **[P2] This policy fixture does not exercise the new negative-rule exclusion.**

Because the path ends in `.yaml`, `TRIGGER-DOCS-ONLY-NEGATIVE`’s `every_file_matches` predicate is already false. Removing `**/policies/**` from `none_match_glob` leaves this case green, while `services/foo/policies/refund.md` matches both the positive and negative rules and is ultimately skipped. Please add/substitute a Markdown policy fixture, or directly assert parity between the positive governance globs and the negative-rule exclusions.

.pre-commit-hooks.yaml
```diff
@@ -9,7 +9,9 @@
 # Scope: the `files:` regex below covers every **path-based** trigger from
 # `docs/triggers.json` — manifest, MCP/OpenAPI/Swagger exports, static tool
 # inventories, coding-agent host configuration/instructions, plugin package
```
pengfei-threemoonslab: **[P2] The universal “every path-based trigger” claim is still false.**

The canonical hook currently misses path legs such as `**/.n8n/**`, `**/conductor/**/*.json`, `**/ai/examples/**/*.json`, and nested `AGENTS.md` / `CLAUDE.md`; `_HOOK_PATH_TRIGGER_FIXTURES` omits the N8N and Conductor rules entirely. It also misses a tracked path or symlink named exactly `prompts`/`policies`, although the catalog’s `dir/**` semantics match that path itself. Please either complete the regex and exhaustive fixtures or narrow this contract claim.

AGENTS.md
```diff
@@ -345,6 +345,8 @@ Do NOT use it for:
 
 One known gap in the Google ADK row: an edit that *modifies* a tools list on the `Agent` alias (rather than `LlmAgent`) is not matched, because a bare `Agent(..., tools=[...])` hunk with no ADK import in it cannot be distinguished from CrewAI's by diff text alone. `LlmAgent` changes and whole-file additions in either spelling are covered.
 
+`prompts/` and `policies/` in that row match at any depth: an edit under `services/foo/policies/` or `enterprise/lib/captain/prompts/` routes exactly like a repo-root one. That is parity with the verifier, whose trust-root classification has always read those two surfaces as `**/policies/**` and `**/prompts/**`.
```
pengfei-threemoonslab: **[P2] Propagate recursive routing to the shipped Cursor activation surface.**

The generated `.cursor/rules/agents-shipgate.mdc`, its renderer, target-repo snippet, and harness canonical globs still use root-only `prompts/**` / `policies/**`. Because that rule has `alwaysApply: false`, a lone nested governance edit may not activate the Shipgate instructions at all. The PR body acknowledges the omission, but it remains an end-to-end adoption gap; please update the renderer/hash mirrors here or track it explicitly before describing host-surface parity as complete.

pengfei-threemoonslab (commented):
Engineer review of `1ba1c76`: **changes needed before merge**.

The lowercase nested cases are fixed and all three hosted workflows are green, but the stated verifier/trigger parity is still incomplete. The mixed-case routing gap can still produce `no_match` or an active docs-only skip, and the local integration snippet plus one regression test do not cover the behavior they claim. The remaining inline comments address public-surface consistency.

Agents Shipgate independently routes this trust-root change as `review_publishable` / `review_required` (0 blockers, 4 human-review items), so human review is expected here.

src/agents_shipgate/triggers.py
```diff
@@ -31,7 +31,7 @@
     is_agent_boundary_path,
 )
 from agents_shipgate.core.errors import ConfigError
-from agents_shipgate.core.globbing import glob_match as _glob_match
+from agents_shipgate.core.globbing import glob_match_ci as _glob_match
```
pengfei-threemoonslab, line 35: **[P1] Do not case-fold the negative test-only classifier.**

Aliasing every predicate to `glob_match_ci` also changes `every_file_matches` at line 307. On a case-sensitive repository, `src/TEST_agent.py` is a valid production path, but it now matches the docs-only rule's `**/test_*.py`. With `+@function_tool` in the diff, the evaluator matches both `TRIGGER-FUNCTION-TOOL-DECORATOR` and `TRIGGER-DOCS-ONLY-NEGATIVE`; because skip outranks run, the updated head returns `run_shipgate=False, skip_reason="skip_rule"`, whereas the base returned `run_shipgate=True`. Keep case-insensitive matching for positive `glob` and protective `none_match_glob`, but make this negative classifier fail closed, and add uppercase `TESTS/` / `TEST_*.py` regressions.

src/agents_shipgate/core/globbing.py
```diff
@@ -8,6 +8,10 @@
 ``dir/**`` matches ``dir`` and anything below it; bare ``**`` matches
 zero or more characters across path segments. ``*`` and ``?`` are
 segment-local (do not cross ``/``).
+
+:func:`glob_match_ci` is the case-tolerant form every path-classifying
```
pengfei-threemoonslab: **[P1] Finish propagation into the Tier-B verifier classifiers.**

This new "every path-classifying surface" contract is not yet true: `checks/_verify_common.py::touched` still imports case-sensitive `glob_match`, and it drives `SHIP-VERIFY-POLICY-WEAKENED`, `SHIP-VERIFY-CI-GATE-REMOVED`, agent-instruction weakening, and trigger-catalog drift. At this head, `services/foo/Policies/refund.yaml` is routed and classified as a policy trust root but produces no policy-weakened fail-safe finding. More seriously, `.github/workflows/Agents-Shipgate.yaml` is classified as `ci_gate` by Tier A but misses the specialized critical gate-removal finding. Use `glob_match_ci` in `touched()` and add mixed-case regressions for the specialized checks.

.pre-commit-hooks.yaml
```diff
@@ -6,10 +6,19 @@
...
+# workflow JSON, coding-agent host configuration/instructions, plugin package
+# files, prompts/policies, AND GitHub workflow files. That claim is pinned
+# exhaustively by `test_hook_regex_fixtures_cover_every_path_based_trigger`:
+# a catalog rule with a `glob` leg must appear in the test's fixture table or
+# in its documented exclusion set, so a new rule cannot land here uncovered.
+# Matching is case-insensitive and, where the catalog glob is recursive, so is
+# the clause: `prompts/`, `policies/`, `.codex-plugin/`, `.agents/plugins/`,
+# `.n8n/`, nested `AGENTS.md`/`CLAUDE.md` and nested workflow directories all
+# stage the same from a sub-workspace as from the repo root, and a `dir/**`
+# glob also matches a tracked path named exactly `dir`.
```
pengfei-threemoonslab, line 21: **[P2] Cover the remaining bare recursive boundary paths.**

This `dir/**` parity claim is still false for the catalog's `.claude/commands/**` and `.cursor/rules/**` globs. `evaluate(paths=[".claude/commands"])` and `evaluate(paths=[".cursor/rules"])` both return `should_run=True`, but the hook clauses require `commands/.*` or `rules/.*`, so neither advisory nor strict hook runs for a tracked file or symlink at the bare path. Use `commands(/.*)?` / `rules(/.*)?`, add both bare paths to the fixture table, and propagate the same clauses to the two copy-paste documentation snippets.

pengfei-threemoonslab (commented):
Engineer re-review of `786baeb0e`: **changes are still needed before merge**.

The previous recursive-routing, precedence-test, documentation, Cursor, and most hook propagation findings are addressed. This updated head introduces one fail-open regression in the negative classifier, leaves the specialized Tier-B verifier checks on different case semantics, and still misses two bare recursive boundary paths in both pre-commit hooks. All three are reproduced in the inline comments.

Validation context: all hosted workflows are green; targeted public-surface tests, Ruff, compilation, schema checks, and a no-isolation wheel build pass. Agents Shipgate independently reports `review_publishable` / `review_required`, with merge permission false.

CHANGELOG.md
```diff
@@ -2,6 +2,67 @@
...
+  `services/foo/policies/refund.yaml` reached `skip_reason: "no_match"` for a
+  path `SHIP-VERIFY-POLICY-WEAKENED` treats as a policy trust root, and
+  `services/foo/Policies/refund.yaml` did the same after the recursive fix
+  alone. Prompts failed worse than that — a nested `prompts/*.md` edit
+  satisfies the docs-only negative rule's `**/*.md` leg, so it did not merely
+  fail to match, it actively *skipped*. Both globs now match at any depth, in
+  the positive rule and in `TRIGGER-DOCS-ONLY-NEGATIVE`'s `none_match_glob`
+  list, so a nested prompt edit bundled with a docs edit is no longer
+  classified as docs-only; and `glob`, `every_file_matches` and
+  `none_match_glob` all now route through the same case-tolerant matcher as
```
pengfei-threemoonslab: **[P2] Keep the published predicate contract internally consistent.**

This paragraph still says `glob`, `every_file_matches`, and `none_match_glob` all use the case-tolerant matcher. The implementation at `triggers.py:317-324`, `docs/triggers.json`'s predicate vocabulary, the regressions, and this same changelog entry at lines 53-59 deliberately keep `every_file_matches` case-sensitive. A third-party consumer following this paragraph could recreate the `src/TEST_agent.py` fail-open skip that the latest commit fixes. Please state the directional split here too.

docs/triggers.json
```diff
@@ -244,11 +244,13 @@
       "when": {
         "any_of": [
           {"glob": "prompts/**"},
-          {"glob": "policies/**"}
+          {"glob": "**/prompts/**"},
+          {"glob": "policies/**"},
+          {"glob": "**/policies/**"}
```
pengfei-threemoonslab, line 249: **[P2] Propagate recursive governance routing to the copy-paste CI filter.**

`examples/github-actions/06-on-tool-source-changes.yml` still starts only for root `prompts/**` and has no policy path at all. The evaluator now returns `run_shipgate=true` for `services/foo/prompts/system.md` and `services/foo/policies/refund.yaml`, but GitHub evaluates `paths:` before starting the workflow, so those PRs never reach this catalog. Add recursive prompt/policy coverage with equivalent case handling, or remove that prefilter and let the evaluator decide.

tests/test_public_surface_contract.py
```diff
@@ -2351,6 +2660,118 @@ def test_primary_surfaces_do_not_claim_broad_agent_healthcare(relpath):
...
+    patterns.
+    """
+    expanded = pattern.replace("**/", "nested/").replace("/**", "/nested").replace("**", "nested")
+    out = {expanded.replace("*", "item")}
+    for suffix in ("/**", "/*"):
+        if pattern.endswith(suffix):
+            bare = pattern[: -len(suffix)]
+            out.add(bare.replace("**/", "nested/").replace("**", "nested").replace("*", "item"))
+    if pattern.startswith("**/"):
+        out.add("svc/" + expanded[len("nested/") :].replace("*", "item"))
```
pengfei-threemoonslab: **[P2] Generate the zero-segment witness for `**/` globs.**

`_representative_paths()` expands `**/.app.json` to nested forms but never root `.app.json`. Mutation reproduction: changing the hook clause from `.*\.app\.json` to `.+/\.app\.json` leaves both the generated sweep and the hand fixture table green, while `evaluate(paths=[".app.json"])["run_shipgate"]` is true and the mutated hook does not match. Add `expanded[len("nested/"):]` as the root witness and validate every synthesized path against its source glob.

tests/test_public_surface_contract.py
```diff
@@ -2351,6 +2660,118 @@ def test_primary_surfaces_do_not_claim_broad_agent_healthcare(relpath):
...
+    to write down, which is how the bare `.claude/commands` and
+    `.cursor/rules` forms stayed uncovered while the manifest claimed
+    `dir/**` parity. This derives representative paths from every positive
+    catalog glob and every boundary-adapter path instead, so a clause that
+    is narrower than its glob fails here without anyone predicting it.
+    """
+    from agents_shipgate.core.boundary_registry import BOUNDARY_ADAPTERS
+    from agents_shipgate.triggers import evaluate as evaluate_triggers
+
+    pattern = _hook_files_regex()
```
pengfei-threemoonslab: **[P2] Exercise both hook IDs in the parity sweep.**

`_hook_files_regex()` always selects only the advisory `agents-shipgate` entry, so every fixture and generated parity check ignores `agents-shipgate-strict`. Removing an arbitrary clause from the strict hook currently leaves this suite green. Parameterize the checks over both hook IDs, or first assert that their `files` expressions are identical and then run the shared sweep.

pengfei-threemoonslab (commented):
Engineer re-review of `98bbffdfa`: **changes are still needed before merge**.

The evaluator directionality fix, Tier-B case propagation, and the remaining bare recursive hook paths are now correct. Four P2 issues remain: one contradictory public predicate contract, one copy-paste CI activation gap, and two holes in the new hook-regression guard. The inline comments include direct reproductions.

Validation context: all hosted workflows are green; the full local non-packaging suite, targeted trigger/verifier/preflight/hook suites, Ruff, compilation, schema checks, wheel/catalog parity, and generated-surface checks pass. Agents Shipgate independently reports `review_publishable` / `review_required` with no blockers, so human review remains required.

examples/github-actions/06-on-tool-source-changes.yml
```diff
@@ -1,24 +1,30 @@
...
+# catalog: `services/foo/Policies/refund.yaml` is a policy trust root that a
+# `- 'policies/**'` filter silently drops.
+#
+# So the filter lives inside the job instead, where it can fail closed. The
+# action evaluates docs/triggers.json against the real diff and returns a skip
+# verdict in seconds when nothing agent-related changed — the expensive scan
+# never runs. You pay one short runner start per PR and keep a green check on
+# every PR, which is also what branch protection wants.
+#
+# If you must prefilter at the `paths:` level for cost reasons, prefilter only
```
pengfei-threemoonslab: **[P1] Remove the inverted `paths:` fallback.**

GitHub `paths` is an inclusive allowlist. Prefiltering only non-trust-root surfaces while omitting prompts, policies, the manifest, instructions, and workflows makes a trust-root-only PR skip the workflow entirely, recreating the silent bypass described immediately above. Please remove this fallback rather than recommend an allowlist that excludes the changes the gate most needs to see.

examples/github-actions/06-on-tool-source-changes.yml
```diff
@@ -1,24 +1,30 @@
...
+# a filtered-out PR produces no run, no check, and no signal at all — it looks
+# identical to a repo that has not adopted the gate. And GitHub's filter
+# patterns are case-sensitive with no case-insensitive form, while the trigger
+# catalog matches governance paths case-insensitively on purpose (Git can
+# carry a spelling that resolves to the canonical file on a case-insensitive
+# filesystem). A `paths:` list is therefore strictly narrower than the
+# catalog: `services/foo/Policies/refund.yaml` is a policy trust root that a
+# `- 'policies/**'` filter silently drops.
+#
+# So the filter lives inside the job instead, where it can fail closed. The
```
pengfei-threemoonslab: **[P2] This workflow cannot take the advertised cheap-skip path.**

Every documented consumer has a root `shipgate.yaml`, and `TRIGGER-EXISTING-MANIFEST-PRESENT` is `force_run`. `evaluate(paths=["README.md"], manifest_present=True, user_requested=True)` returns `run_shipgate=True, force_run=True`; the verifier passes `config_path.exists()` to that evaluator, and the pinned v0.15.0 has the same rule. The expensive scan therefore runs on every PR. Please rename/retire this recipe or implement a genuinely lightweight routing mode without contradicting the adopted-repo force-run contract.

.pre-commit-hooks.yaml
```diff
@@ -39,22 +48,26 @@
...
+      (.*/)?\.codex-plugin(/.*)?|
+      (.*/)?\.agents/plugins(/.*)?|
       .*\.app\.json|
       (.*/)?SKILL\.md|
       .*openapi.*\.(yaml|yml|json)|
       .*swagger.*\.(yaml|yml|json)|
       \.agents-shipgate/.*\.json|
-      prompts/.*|
-      policies/.*|
+      (.*/)?prompts(/.*)?|
```
pengfei-threemoonslab, line 77: **[P2] Include symlinks in the effective hook filter.**

This regex now matches a bare `prompts`/`policies` path, but both hooks omit `types`, so pre-commit applies its default `types: [file]`. A tracked symlink has the separate `symlink` tag and is filtered out before `files` is evaluated; a symlink-only governance change therefore invokes neither hook despite the new fixture explicitly claiming that case. Clear or expand the type filter on both hooks and test effective pre-commit selection, not only the compiled regex.

tests/test_public_surface_contract.py
```diff
@@ -2351,30 +2660,258 @@ def test_primary_surfaces_do_not_claim_broad_agent_healthcare(relpath):
...
+    PR produces no run and no check — indistinguishable from a repo that
+    never adopted the gate. Worse, GitHub's filter patterns are
+    case-sensitive with no case-insensitive form, while the trigger catalog
+    matches governance paths case-insensitively: a `- 'policies/**'` entry
+    silently drops `services/foo/Policies/refund.yaml`, a policy trust root.
+    The in-job trigger evaluator is the filter, and it fails closed.
+    """
+    import yaml
+
+    examples = sorted((REPO_ROOT / "examples" / "github-actions").glob("*.yml"))
```
pengfei-threemoonslab: **[P2] Guard every shipped prefilter form.**

This scans only GitHub `*.yml` files and rejects only the `paths` key. It misses valid `*.yaml` workflows and the equivalent `paths-ignore` prefilter. It also leaves the first-class GitLab and CircleCI `05-on-tool-source-changes` recipes on incomplete allowlists that suppress nested prompt/policy-only changes, including `services/foo/Policies/refund.yaml`, even though the evaluator routes them. Please broaden the guard and update those provider recipes, or explicitly track the pre-existing provider gap as an immediate follow-up.

tests/test_public_surface_contract.py
```diff
@@ -2351,30 +2660,258 @@ def test_primary_surfaces_do_not_claim_broad_agent_healthcare(relpath):
...
+    synthesis bug cannot quietly emit paths that pass the sweep by never
+    being run-worthy in the first place.
+    """
+    body = pattern[3:] if pattern.startswith("**/") else pattern
+    # "" is the zero-segment expansion of a leading `**/`.
+    prefixes = ("", "nested/", "svc/deep/") if pattern.startswith("**/") else ("",)
+
+    def expand(text: str) -> str:
+        return (
+            text.replace("**/", "nested/")
```
pengfei-threemoonslab: **[P2] Generate every globstar arity before calling this sweep exhaustive.**

Only a leading `**/` receives a zero-segment witness; internal `/**/` is always rewritten to one directory, and recursive positions do not get zero/one/multiple-segment variants. For example, the generator emits `conductor/nested/item.json` for `**/conductor/**/*.json` but never `service/conductor/job.json`, even though the evaluator matches it. A narrowed Conductor clause can therefore leave both hand fixtures and this generated sweep green while missing a real run path. Generate zero-, one-, and multi-segment witnesses at every globstar position.

pengfei-threemoonslab (commented):
Engineer re-review of `ac72d601`: **changes are still needed before merge**.

The earlier directional case-matching, Tier-B propagation, bare boundary-path, changelog, and advisory/strict parity findings are fixed. Five issues remain: an inverted `paths:` fallback, a cost-saving path the documented workflow cannot take, pre-commit type filtering that still excludes governance symlinks, incomplete cross-provider prefilter coverage, and a non-exhaustive globstar regression sweep. The inline comments include direct reproductions.

Validation context: all hosted workflows are green; the full local non-packaging suite, focused trigger/public-surface suites, Ruff, compilation, schema/generated-surface checks, and a no-isolation wheel/catalog parity check pass. Agents Shipgate independently reports `review_publishable` / `review_required` with no blockers, so human review remains required.

The PR body should also be refreshed before merge: it still says the Cursor mirrors were not touched and still attributes prompts to `_POLICY_SURFACES`.

PR #367, opened 2026-08-11
fix(agent-mode): spell emitted commands for the invocation that produced them

### LINKED ISSUE #322: bug(agent-mode): preserve python -m invocation in emitted next-action commands
## Summary

When Agents Shipgate is run from source with `python -m agents_shipgate`, emitted recovery commands switch to the installed `shipgate` or `agents-shipgate` binaries, while error metadata can report the current command as `__main__.py` (rendered as `**main**.py` in the observed output).

The recommended next command is therefore not guaranteed to be executable in the environment that produced it.

## Reproduction

```bash
export PYTHONPATH=/path/to/agents-shipgate/src
export AGENTS_SHIPGATE_AGENT_MODE=1

python -m agents_shipgate verify --workspace . --preview --base base-ref --head HEAD --json
```

The preview returned a next action beginning with:

```text
shipgate init --workspace ... --write --json
```

`init --json` then recommended:

```text
agents-shipgate scan -c shipgate.yaml --suggest-patches
```

A subsequent agent-mode scan error reported:

```json
{"command":"**main**.py scan -c shipgate.yaml --suggest-patches --format json"}
```

In the test environment, the installed wrapper was broken; only the original `python -m agents_shipgate` invocation was valid.

## Expected behavior

Machine-readable next actions should preserve the invocation style that successfully launched the current process, or encode commands structurally so the caller does not need to reverse-engineer the executable.

Illustrative shape:

```json
{
  "executable": ["python", "-m", "agents_shipgate"],
  "args": ["scan", "-c", "shipgate.yaml", "--format", "json"]
}
```

A rendered compatibility string can remain for existing consumers.

## Acceptance criteria

- [ ] `python -m agents_shipgate ...` produces next-action commands that remain runnable through `python -m agents_shipgate`.
- [ ] Console-script invocation continues to emit the canonical console script.
- [ ] `command` never exposes `__main__.py` or Markdown-transformed variants such as `**main**.py`.
- [ ] Preview, init, doctor, scan, and verify use the same invocation policy.
- [ ] Structured command fields are shell-independent; existing string fields remain backward compatible for the documented deprecation window.

pengfei-threemoonslab:
## Summary

- Emitted commands are now spelled for the entry point that started the running process, so the recovery loop stays runnable in the environment that produced it. Closes #322.
- New `src/agents_shipgate/invocation.py` is the single answer to "what tokens re-enter this CLI here". A console-script run emits exactly what it emitted before, byte for byte; a `python -m agents_shipgate` run emits `<sys.executable> -m agents_shipgate …`; `AGENTS_SHIPGATE_CLI` overrides both (the same operator override the Claude Code hook installer already honours).
- `command` can no longer contain `__main__.py` — the string that any consumer rendering the field as Markdown silently corrupts to `**main**.py`.
- `next_actions[]` entries with `kind="command"` additionally carry a shell-independent `executable[]` / `args[]` pair, runnable as `[*executable, *args]`.

### Why this is wider than one line

`cli/agent_mode.py:128` was the reported symptom, but the defect was in ~15 places: every command Shipgate emits was hand-written as the console script the wheel installs, so a source checkout without a wrapper got a route it could not execute. The policy is applied at every field whose *value is a command to run*: `NextAction` (preview, init, doctor, scan, verify, detect, check, explain, findings), the shared verify/detect/preview builders behind the control envelope and verifier artifacts, preflight signals, evidence-gap actions, fix-task repairs, matched trigger rules, and detect's legacy single-string `next_action`.

Three call sites emitted `next_action` (legacy string) and `next_actions[]` (ranked array) as independent literals; they now project both from one action, so the two forms cannot name different programs.

### Deliberate boundaries

- **Console-script runs are a no-op.** `retarget_command` returns the string unchanged when the process was entered through a console script, so the shipped path produces identical output and no golden churn. Only the from-source path changes.
- **`sys.executable`, not a bare `python`.** A bare name resolves through `PATH` and can land on a different interpreter — which is the failure this change exists to prevent, reintroduced one layer down.
- **Rewriting splices only the leading program token** rather than re-tokenising, so deliberately unquoted placeholders (`--from <report.json>`) survive verbatim instead of being quoted into something the reader is no longer meant to substitute.
- **Published contract vocabulary and prose stay canonical.** `primary_commands` and `.well-known/agents-shipgate.json` describe the installed CLI; retargeting them would let a generator run under `python -m` write an absolute interpreter path into a committed artifact.
- **`executable` / `args` are a projection, not input.** The validator recomputes them from `command` even when a caller supplies them, so the two forms cannot disagree. Both are `null` when the string needs a shell to be faithful (a leading `NAME=VALUE` assignment); the string stays complete.
- **No `CONTRACT_VERSION` bump.** `NextAction` is the diagnostics surface, not the versioned control contract — it is absent from `AGENT_CONTROL_FIELDS`, the `AgentControl` union is untouched, and the two new fields are optional and additive. Same call as #308's trigger-catalog bump. Happy to stamp it and run the five-layer checklist if you'd rather.

## Type

## Verification

Additional local checks run:

- Full suite green (`pytest -n 8`), `scripts/generate_schemas.py --check` clean, `ruff check src/ tests/` clean, `llms-full.txt` regenerated.
- The issue's reproduction walked end to end against a real `python -m agents_shipgate` process, and the console-script negative control against a real wrapper.
- All 14 artifacts a full `verify` writes (verifier, agent-handoff, current-control, verification-receipt, …) scanned for command-valued keys: zero non-retargeted.

## Two things a reviewer should know

**A test caught a real bug in the detection itself.** `argv[0]` ending in `__main__.py` does not mean *our* `-m` run — `python -m pytest` is indistinguishable by that check alone. Without also requiring the parent directory to be `agents_shipgate`, every command emitted from an in-process test would have claimed an entry point the caller never used. Fixed, with a regression test.

**Verifying against an empty workspace was not enough.** Three surfaces passed every check until the sweep ran against a *configured* repository: `trigger.matched_rules[].command` (bundled catalog echoed verbatim), detect's legacy single-string `next_action`, and verify's own rerun command. That sweep is now a test (`tests/test_invocation_policy.py`) that walks every command-valued key in real subprocess output, with a negative control proving it can fail — so a newly added emitter that forgets the policy fails there rather than in an agent's terminal.

## Release-readiness notes

- [x] Report/schema changes are additive or documented in `STABILITY.md` — two optional fields added to `next_actions[]`, documented in `docs/diagnostics.md` and `docs/errors.json`

### Note on `tests/test_adapter_static_only.py`

The trust-model allowlist pins `triggers.py:117` by line number; adding an import to that module shifts it to 118. Entry updated, surface and snippet unchanged.


src/agents_shipgate/invocation.py
```diff
@@ -0,0 +1,301 @@
...
+    unquoted placeholders (``--from <report.json>``) that a re-tokenising
+    rewrite would quote into something the reader is no longer meant to
+    substitute.
+
+    Commands that belong to another program (``pip install ...``) and commands
+    whose program token cannot be located are returned unchanged.
+    """
+
+    tokens = tuple(prefix) if prefix is not None else invocation_prefix()
+    if len(tokens) == 1 and _program_name(tokens[0]) in CONSOLE_SCRIPTS:
```
pengfei-threemoonslab: **[P1] Preserve explicit single-token CLI overrides**

This no-op also matches an explicit absolute override such as `AGENTS_SHIPGATE_CLI=/private/venv/bin/agents-shipgate`. `invocation_prefix()` preserves that path, but `retarget_command()` returns the original bare command and `split_invocation()` publishes `executable=["agents-shipgate"]`. In an end-to-end scan error, the top-level `command` used the absolute override while `next_action` and `next_actions[0]` reverted to a wrapper that may not be on `PATH`.

Please preserve how the prefix was selected and apply this no-op only to an actually detected canonical console-script invocation, not an operator override.

src/agents_shipgate/invocation.py
```diff
@@ -0,0 +1,301 @@
...
+        # for no gain: both names resolve to the same entry point.
+        return command
+
+    span = _program_token_span(command)
+    if span is None:
+        return command
+    start, end = span
+    if _program_name(command[start:end]) not in CONSOLE_SCRIPTS:
+        return command
+    return command[:start] + shlex.join(tokens) + command[end:]
```
pengfei-threemoonslab: **[P1] Render module commands with platform-aware quoting**

`shlex.join()` emits POSIX single-quote syntax. A Windows prefix such as `C:\Python312\python.exe` becomes `'C:\Python312\python.exe' -m agents_shipgate ...`, which is not directly runnable by `cmd.exe` or PowerShell. In `cmd.exe`, single quotes also do not protect metacharacters, so a workspace argument containing `&` can become a second command when an agent follows the advertised string.

This affects legacy control/repair command fields that have no structured argv sibling, and the new `startswith(f"{sys.executable} -m ...")` tests would fail on Windows despite the package declaring OS independence.

src/agents_shipgate/core/policy_evidence.py
```diff
@@ -174,7 +175,7 @@ def policy_evidence_gap(
         why=f"{policy_id}: {why}",
         next_action=EvidenceGapAction(
             kind=cast(Any, action_kind),
-            command=rerun_command,
+            command=retarget_command(rerun_command),
```
pengfei-threemoonslab: **[P1] Keep invocation-local spelling out of deterministic artifacts**

This puts process-entry spelling into durable `report.json` and `packet.json` data, but scan/run and verification request identities do not bind that spelling. Running the same support-refund scan through a console wrapper and `python -m` produced the same `run_id` while changing both report and packet SHA-256 values; the differences were evidence-gap `next_action.command` fields.

That violates the documented `Same inputs -> same report` invariant and gives one semantic run identity multiple artifact bodies. Please keep durable artifacts canonical and retarget at a live presentation/control boundary, or explicitly bind the invocation input into artifact identity.

src/agents_shipgate/core/agent_controls.py
```diff
@@ -64,26 +71,26 @@ def verify_command_for(
...
     if config is not None:
-        parts.extend(["--config", shlex.quote(_config_for_verify(workspace, config))])
+        args.extend(["--config", _config_for_verify(workspace, config)])
     if base is not None and head is not None:
-        parts.extend(["--base", shlex.quote(base), "--head", shlex.quote(head)])
-    parts.extend(extra)
-    parts.append("--json")
-    return " ".join(parts)
+        args.extend(["--base", base, "--head", head])
+    args.extend(extra)
```
pengfei-threemoonslab, line 88: **[P2] Do not quote `extra` tokens twice**

`extra` is now treated as raw argv and quoted by `render_command()`, but `_recovery_verify_command()` still passes `shlex.quote(str(reports_dir))`. For `/tmp/reports space`, the generated command contains `--out ''"'"'/tmp/reports space'"'"''`, and `shlex.split()` yields the literal value `'/tmp/reports space'` including quote characters.

Following the recovery route therefore refreshes a different output directory and leaves the requested control pointer stale. Pass the raw reports path or make the `extra` contract explicit and consistent.

tests/test_invocation_policy.py
```diff
@@ -0,0 +1,501 @@
...
+        if not stripped.startswith(("{", "[")):
+            continue
+        try:
+            documents.append(json.loads(stripped))
+        except json.JSONDecodeError:
+            continue
+    return documents
+
+
+@pytest.mark.parametrize(
```
pengfei-threemoonslab, line 584: **[P2] The claimed whole-surface sweep omits real emitters**

This parametrization only runs `scan`, `check`, and `verify --preview`. A `python -m agents_shipgate apply-patches` error still emits a hand-built bare `agents-shipgate scan ...` action with no `executable`/`args`; the legacy packet upgrader and preflight host-grant `related_command` also bypass normalization.

Please include those flows in the sweep or, preferably, enforce normalization at the serialization boundary so raw dictionaries and durable upgrade paths cannot opt out.

src/agents_shipgate/invocation.py
```diff
@@ -0,0 +1,301 @@
...
+    Returns ``None`` when the string has no faithful argv form — an unbalanced
+    quote, or a leading ``NAME=VALUE`` assignment, which is shell syntax rather
+    than an argv token and so cannot be represented without inventing a field
+    to hold it. The rendered string stays authoritative in those cases.
+    """
+
+    if _has_env_assignment_prefix(command):
+        return None
+    try:
+        tokens = shlex.split(command)
```
pengfei-threemoonslab, line 506: **[P2] `shlex.split` does not prove shell independence**

This accepts shell operators, redirection, globbing, and expansion as ordinary argv tokens. For example, `printf left ' ' && printf right` is published as `executable=["printf"]` and `args=["left", " ", "&&", "printf", "right"]`; the documented subprocess form does not execute the same operation as the command string.

Please reject shell-only syntax before publishing structured argv, or derive the pair only from command values constructed from known argv tokens.

src/agents_shipgate/schemas/diagnostics.py
```diff
@@ -27,6 +43,8 @@ class NextAction(BaseModel):
     path: str | None = None
     why: str
     expects: str | None = None
+    executable: list[str] | None = None
```
pengfei-threemoonslab: **[P2] Version the strict `NextAction` wire-shape change**

The previous published `NextAction` model already used `extra="forbid"`. Head now serializes `executable` and `args` on every action, including `null` on review/edit/stop actions, while detect/doctor/agent-mode payloads carry no new schema identifier. Validating a head action with the prior model produces two `extra_forbidden` errors.

Please advance or negotiate the carrying wire shape rather than treating these properties as safely additive for strict consumers.

pengfei-threemoonslab (commented):
Engineer review — **request changes**

Reviewed exact head `9a39f5a9f` against base `e0377887a`. The Linux `python -m` happy path and CI are green, but the implementation still breaks supported override and Windows invocations, introduces invocation-dependent report bytes, and misses several command emitters.

Please address the three P1 blockers and four P2 correctness/contract issues inline before merge.

_GitHub does not allow the PR author to submit the formal `REQUEST_CHANGES` state on their own PR, so this is posted as a comment review._

Validation: targeted tests and the full non-packaging suite passed locally, along with Ruff, compilation, and schema-generation checks. Agents Shipgate returned `control.state=review_publishable` and `decision=review_required` because this PR changes the `AGENTS.md` trust root.

src/agents_shipgate/invocation.py
pengfei-threemoonslab: Fixed in 55e97906. Confirmed: the no-op was keyed on the program *name*, so an absolute override was discarded by exactly the branch meant to protect the shipped path — while `_command_string()`, which does not go through that branch, kept the override. One emitted error named two different programs.

`resolve_invocation()` now returns an `Invocation` carrying `source` (`console_script` / `module` / `override` / `fallback`), and `keeps_written_spelling` is true only for a *detected* console-script run (and the fallback, which is the documented spelling by definition). An operator override is always spliced in, whatever it happens to be named.

Your repro now agrees across all three fields:

```
command    : /Users/.../.venv/bin/agents-shipgate scan -c missing.yaml --format json
next_action: /Users/.../.venv/bin/agents-shipgate verify --workspace ... --preview --json
executable : ["/Users/.../.venv/bin/agents-shipgate"]
```

Both directions are pinned: `test_an_explicit_override_is_never_treated_as_the_written_spelling` and `test_a_detected_console_script_still_keeps_the_written_spelling`.

src/agents_shipgate/invocation.py
pengfei-threemoonslab: Fixed in 55e97906. New `join_argv()` renders with `subprocess.list2cmdline` on Windows and `shlex.join` elsewhere, used for both the spliced prefix and `render_command`.

The second half of your point mattered more than the quoting itself: `split_invocation` was recovering the entry point by `shlex.split`-ing the whole string, which on Windows eats the backslashes in `C:\Python312\python.exe`. It no longer parses the prefix at all — the entry point comes from the resolved invocation and only the argument tail is parsed. That also makes the structured pair exact rather than reconstructed.

The tests you flagged now build their expectations through `join_argv(...)` instead of hardcoding POSIX form, and `test_windows_renders_commands_with_windows_quoting` exercises the Windows branch on any host.

One scope note: `list2cmdline` lives behind a forbidden-surface import under the trust-model scanner. I took the narrow `from subprocess import list2cmdline` so no process-spawning name is bound in that module, with an `ALLOWED_EXCEPTIONS` entry and a `STABILITY.md` § Meta-CLI paragraph. A future `subprocess.run` there would be a different import line with a different snippet, and would still fail the gate.

src/agents_shipgate/core/policy_evidence.py
pengfei-threemoonslab: Agreed, and reverted in 55e97906 — `policy_evidence.py`, the four `EvidenceGapAction` sites in `ci/release_decision.py`, and `agent_summary.first_recommended_action` are canonical again. `docs/architecture.md` makes "same inputs → same report" non-negotiable and process entry is not an input; retargeting there bought runnability in an artifact nobody executes from, and cost the property the packet hash exists to provide.

Took your first remedy rather than the second: durable artifacts stay canonical, and the runnable spelling lives on the live routes (agent-mode errors, `next_actions[]`, control/repair commands, preflight signals). The boundary is now stated in the module docstring, in `docs/diagnostics.md` as a "what the policy does *not* touch" table, and in `docs/errors.json`.

Pinned by `test_durable_artifacts_do_not_depend_on_how_the_process_started`: the same sample scanned through the console script and through `python -m`, into **one** output directory, byte-compared. The single directory matters — my first attempt used two and "caught" nothing but the recorded artifact paths.

src/agents_shipgate/core/agent_controls.py
pengfei-threemoonslab: Fixed in 55e97906 — a regression I introduced when `verify_command_for` moved to raw argv tokens. Reproduced exactly as you describe; after the fix `shlex.split` round-trips the value:

```
parsed: ['agents-shipgate', 'verify', '--workspace', '/ws', '--out', '/tmp/reports space', '--json']
```

`_recovery_verify_command()` passes the raw path now, and the contract is stated on `verify_command_for` rather than left implicit for the next caller to guess: every argument, including each `extra` element, is one raw argv token quoted exactly once, and callers must not pre-quote.

tests/test_invocation_policy.py
pengfei-threemoonslab: Fixed in 55e97906, and the root cause was worse than a thin parametrization: `apply-patches` had a **second, parallel agent-mode emitter** (`_emit_input_error`) that reproduced the payload by hand. Two consequences, both invisible until something looked — its routes were plain dicts that could not opt into the policy, and the line never carried the `command` field every other agent-mode error carries.

Took your preferred remedy. `_emit_input_error` now delegates to the shared `emit_agent_mode_error`, and normalization happens **there** rather than in the model — so a route built as a raw dict, by a surface nobody has written yet, cannot opt out. Preflight's host-grant `related_command` now goes through `render_command` too.

The sweep grew from 3 surfaces to 9 (`scan`, `check`, `verify --preview`, `apply-patches`, `detect`, `explain-finding`, `findings`, `explain`, `audit --host`) plus two dedicated tests: `test_a_hand_built_action_dict_is_normalized_at_the_wire` drives the pre-v0.6-report path that produces the raw command dict, and `test_preflight_host_grant_route_is_spelled_for_this_invocation` covers the differently-named field, since a different field name is exactly how an emitter escapes a policy keyed on one name.

On the legacy packet upgrader: leaving it canonical is deliberate rather than missed. It is prose inside a durable packet artifact, so it lands on both sides of the boundary drawn for your `report.json` finding. Called out explicitly in `docs/diagnostics.md`.

src/agents_shipgate/invocation.py
pengfei-threemoonslab: Fixed in 55e97906. You are right that a successful parse proves nothing — `shlex.split` returns `&&` as an ordinary token, and the published argv would then run something the string does not.

`split_invocation` refuses when the command carries any shell metacharacter outside quotes: control operators, redirection, substitution, expansion, globbing. Quoted regions stay exempt, so `--base 'HEAD~1'` is still one argv token. Parametrized over your `printf ... && printf ...` case plus pipeline, redirection, `$VAR`, backticks, and glob forms, with `test_split_invocation_keeps_quoted_metacharacters` as the counterweight so the rule cannot degenerate into refusing everything.

One consequence worth naming: `--from <report.json>` is now withheld too, because `<` and `>` really are redirection. That reads like a loss but is the honest answer — that string is a template a caller must fill in, not an argv it can execute. The rendered string remains authoritative, and the keys are omitted rather than emitted as null.

src/agents_shipgate/schemas/diagnostics.py
pengfei-threemoonslab: Fixed in 55e97906. You are right that `extra="forbid"` makes "optional" and "additive" different things, and I had conflated them.

Two changes:

1. **Contract `22` → `23`** carries the wire-shape change — `schemas/contract.py`, `.well-known/agents-shipgate.json`, the `docs/agent-contract-current.md` prose note and stamp, the `docs/architecture.md` stamp, `docs/errors.json`, and the four pinned test literals. `MINIMUM_CONTROL_CONTRACT_VERSION` stays at `21`: the `AgentControl` union is untouched, v23 changes only how commands inside it are spelled.
2. The pair is now **omitted rather than serialized as `null`**, via a wrap serializer. A `review`/`edit`/`stop` action can never carry an argv, so it is byte-for-byte what it was before this PR; only `command` actions change shape at all, and those are the ones the version announces.

```json
{"kind": "review",  "command": null, "path": null, "why": "x", "expects": null}
{"kind": "command", "command": "agents-shipgate verify --json", "path": null, "why": "x", "expects": null,
 "executable": ["agents-shipgate"], "args": ["verify", "--json"]}
```

No collision: I checked the five other open PRs and none touches `contract.py`.

pengfei-threemoonslab:
All seven addressed in 55e97906 — three P1 and four P2, each with a regression test. Inline replies on each thread; summary here.

| # | Finding | Resolution |
|---|---|---|
| P1 | Explicit single-token CLI override discarded | `Invocation` now carries `source`; the "keep the written spelling" no-op applies to a *detected* console-script run, never to an override |
| P1 | POSIX quoting for Windows invocations | `join_argv()` follows the host shell; `split_invocation` no longer round-trips the prefix through a POSIX parse at all |
| P1 | Invocation-local spelling in `report.json` / `packet.json` | Reverted — durable evidence is canonical again; runnable spelling stays on live routes |
| P2 | `extra` tokens quoted twice | Caller passes the raw path; the raw-argv contract is stated on `verify_command_for` |
| P2 | Sweep omits real emitters | Root cause was a **parallel agent-mode emitter** in `apply-patches`; deleted, normalization moved to the emission boundary |
| P2 | `shlex.split` does not prove shell independence | Structured argv withheld for any unquoted shell metacharacter |
| P2 | Strict `NextAction` wire-shape change unversioned | Contract `22` → `23`; pair omitted rather than `null` |

**Two of these changed my mind about the design, not just the code.**

The `report.json` finding was the one I had backwards. I had treated "commands should be runnable" as the goal and let it reach every field named `command`. Durable artifacts are not routes — nobody executes from `report.json` — and "same inputs → same report" is the property the packet hash exists to provide. The policy now stops at that border explicitly, documented as a table in `docs/diagnostics.md` rather than left as an accident of which call sites I happened to visit.

The sweep finding surfaced something worse than an omission. `apply-patches` had a second agent-mode emitter that reproduced the payload by hand, which is why its routes bypassed the policy *and* why its errors never carried `command` at all. Enforcing at the wire instead of at the model is what makes the guarantee hold for constructions nobody has written yet — that is your remedy, and it is the right one.

Also worth flagging: `list2cmdline` required a trust-model allowlist entry. I took the narrow `from subprocess import list2cmdline` so no process-spawning name is bound in that module, with an `ALLOWED_EXCEPTIONS` entry and a `STABILITY.md` § Meta-CLI paragraph — a future `subprocess.run` there would be a different import line and would still fail the gate.

**Verification.** Full suite green; `generate_schemas.py --check`, `ruff check`, `compileall` clean; `llms-full.txt` regenerated. Your override repro re-run end to end (all three fields now agree). The nine-surface sweep re-run against a configured repository: zero non-retargeted commands. Durable artifacts byte-compared across both entry points into one output directory.

src/agents_shipgate/invocation.py
```diff
@@ -0,0 +1,415 @@
...
+    ``shlex.join`` is POSIX-only: it quotes with single quotes, which
+    ``cmd.exe`` treats as ordinary characters rather than as quoting. A Windows
+    interpreter path is exactly the case that needs quoting (backslashes, often
+    a space in ``Program Files``), so rendering it POSIX-style produces a string
+    Windows cannot run — and, worse, one where a metacharacter in an argument
+    is no longer contained.
+    """
+
+    if _WINDOWS:
+        return list2cmdline(list(tokens))
```
pengfei-threemoonslab: **[P1] Use actual Windows shell escaping**

`subprocess.list2cmdline` implements MS C-runtime argv quoting, not `cmd.exe` or PowerShell escaping. A valid ref such as `feature&whoami` is emitted unquoted, so pasting the advertised command into `cmd.exe` executes `whoami`; a quoted `C:\Program Files\...\python.exe` is not directly invokable in PowerShell without `&` either. Use a shell-specific renderer, or make structured argv the authoritative Windows route and stop promising a host-shell-runnable string.

src/agents_shipgate/invocation.py
```diff
@@ -0,0 +1,415 @@
...
+    else:
+        span = _program_token_span(command)
+        if span is None or span[0] != 0:
+            # Either nothing to split, or leading ``NAME=VALUE`` assignments,
+            # which are shell syntax rather than argv tokens.
+            return None
+        head, tail = [command[span[0] : span[1]]], command[span[1] :]
+
+    try:
+        args = shlex.split(tail)
```
pengfei-threemoonslab: **[P1] Do not parse Windows arguments with POSIX `shlex`**

This parses the Windows-rendered tail with POSIX rules. Repro with `_WINDOWS=True`: `C:\repo` becomes `C:repo`, and `C:\repo\shipgate.yaml` becomes `C:reposhipgate.yaml` in `args`. Since the docs tell agents to prefer `[*executable, *args]`, recovery can silently evaluate the wrong workspace/config. Preserve the original argv or use a Windows-compatible parser.

src/agents_shipgate/invocation.py
```diff
@@ -0,0 +1,415 @@
...
+        return list2cmdline(list(tokens))
+    return shlex.join(tokens)
+
+
+def _override_prefix(env: Mapping[str, str]) -> tuple[str, ...] | None:
+    raw = env.get(CLI_OVERRIDE_ENV_VAR, "").strip()
+    if not raw:
+        return None
+    try:
+        tokens = shlex.split(raw)
```
pengfei-threemoonslab: **[P1] Preserve Windows absolute CLI overrides**

The override is unconditionally parsed with POSIX `shlex`. A normal PowerShell value `AGENTS_SHIPGATE_CLI=C:\Tools\agents-shipgate.exe` resolves to `C:Toolsagents-shipgate.exe`; an unquoted path containing spaces splits into multiple tokens. The previous absolute-override fix therefore remains broken on Windows. Parse this value according to the host platform and cover a real backslash path.

src/agents_shipgate/invocation.py
```diff
@@ -0,0 +1,415 @@
...
+    """
+
+    quote: str | None = None
+    index = 0
+    while index < len(text):
+        char = text[index]
+        if quote is None and char == "\\":
+            index += 2
+            continue
+        if quote is not None:
```
pengfei-threemoonslab: **[P2] Detect expansion inside double quotes**

This exempts every metacharacter inside either quote style, but POSIX shells still expand `$VAR`, `$(...)`, and backticks inside double quotes. For example, `printf "$HOME"` publishes `args=["$HOME"]`, while the authoritative shell string passes the expanded value. Only single quotes can make these forms inert; detect substitution/expansion while in double quotes so the structured pair is withheld.

src/agents_shipgate/cli/apply_patches.py
```diff
@@ -270,23 +270,25 @@ def apply_patches(
...
+    """
 
-    from agents_shipgate.cli.agent_mode import is_agent_mode
+    from agents_shipgate.cli.agent_mode import emit_agent_mode_error
 
-    if not is_agent_mode():
-        return
-    payload = {"error": kind, "message": message, **fields}
-    print(json.dumps(payload, default=str), file=sys.stderr)
+    emit_agent_mode_error(kind, message=message, **fields)
```
pengfei-threemoonslab, line 291: **[P2] Derive the legacy rank-one action from the normalized array**

The shared emitter currently normalizes `next_action` and `next_actions` independently. On the pre-v0.6 report path, the result has prose `next_action="Re-run scan ..."` while `next_actions[0]` is a command containing the actual scan invocation. That contradicts the new documented rule that a command rank-one action projects verbatim to the legacy field. Normalize the array first and derive `next_action` from its first entry.

src/agents_shipgate/schemas/diagnostics.py
```diff
@@ -36,6 +78,17 @@ def _check_kind_fields(self) -> NextAction:
...
         if self.kind == "stop" and self.command is not None:
             raise ValueError("kind='stop' must not carry a command")
+        if self.kind == "command" and self.command:
+            self.command = retarget_command(self.command)
+            split = split_invocation(self.command)
+            # ``None`` means the string has no faithful argv form — a leading
+            # ``NAME=VALUE`` assignment is shell syntax, not an argv token.
+            # Leaving the structured pair unset is the honest answer; the
+            # rendered string still carries the whole instruction.
+            self.executable, self.args = split if split is not None else (None, None)
```
pengfei-threemoonslab: **[P2] Enforce the command/argv invariant after construction**

The pair is derived only by this construction-time validator. The model remains mutable, and Pydantic `model_copy(update=...)` skips validation, so direct assignment or a copied command can serialize with stale or attacker-supplied `executable`/`args`. This falsifies the published “cannot disagree” guarantee. Recompute the projection during serialization, or make all mutation/copy paths validated and immutable.

pengfei-threemoonslab (commented):
Engineer re-review of updated head `55e979069`.

**Verdict: request changes.** The follow-up fixes the POSIX absolute override, spaced `--out`, missed emitter, durable-artifact determinism, and contract-v23 compatibility issues. Six actionable defects remain: three P1 Windows invocation failures and three P2 contract/invariant failures, detailed inline.

Validation performed against this exact head: GitHub `test`, `verify`, and `verify-self` checks are green; the full local non-packaging suite, focused regressions, Ruff, compileall, and generated-schema checks pass. Shipgate reports `control.state=review_publishable` / `decision=review_required` because the PR touches release trust roots.

GitHub does not allow the PR author to submit `REQUEST_CHANGES` on their own PR, so this is posted as a `COMMENT` review; the engineering verdict remains blocking.

src/agents_shipgate/invocation.py
pengfei-threemoonslab: Fixed in ef60d509 — took your second remedy.

You are right on both counts: `list2cmdline` is MS C-runtime *argv* quoting, so `feature&whoami` came out unquoted, and a quoted path needs `&` in PowerShell. Writing a real `cmd.exe`-and-PowerShell escaper would also mean guessing which shell the agent is in, and would still not make a pasted string safe.

So the Windows branch is gone and the promise is narrowed rather than faked. `command` is a POSIX rendering for display and POSIX shells; `[*executable, *args]` is the authoritative runnable form on every platform. `docs/diagnostics.md`, `AGENTS.md`, `docs/errors.json`, and `docs/agent-contract-current.md` all say so now, including an explicit "do not paste `command` into `cmd.exe` or PowerShell".

Worth noting the argv route is exact for precisely the case you raised — the risky value never needs quoting because it is never re-parsed:

```
command : agents-shipgate verify --base 'feature&whoami'
args    : ['verify', '--base', 'feature&whoami']
```

Removing the branch also removed the `subprocess` import, so the `ALLOWED_EXCEPTIONS` entry and its `STABILITY.md` § Meta-CLI paragraph are withdrawn.

src/agents_shipgate/invocation.py
pengfei-threemoonslab: Fixed in ef60d509. Reproduced exactly:

```
before: args = ['verify', '--workspace', 'C:repo', '--config', 'C:reposhipgate.yaml']
after : args = ['verify', '--workspace', 'C:\\repo', '--config', 'C:\\repo\\shipgate.yaml']
```

This is the finding I'd rank highest of the six: not an unrunnable command but a *runnable* one against the wrong workspace, on the path the docs tell agents to prefer.

Root cause is shared with the previous thread — a string rendered by one set of rules and parsed by another. Rather than add a Windows parser to match a Windows renderer, there is now one renderer and one parser, both POSIX. That round-trips Windows paths exactly, because a single-quoted `'C:\repo'` keeps its backslashes; `test_windows_paths_round_trip_through_render_and_split` asserts it under both `_WINDOWS` settings so the property holds on either host.

The entry point is still never recovered by parsing — it comes from the resolved invocation — which is what makes the prefix exact regardless of quoting rules.

src/agents_shipgate/invocation.py
pengfei-threemoonslab: Fixed in ef60d509. Confirmed: `C:\Tools\agents-shipgate.exe` resolved to `C:Toolsagents-shipgate.exe`, so the operator wrote a correct path and got one that does not exist — the absolute-override fix from the last round was undone on the only platform where backslashes are normal.

`_split_override()` parses with the host's rules: POSIX `shlex` elsewhere, `shlex.split(..., posix=False)` plus outer-quote stripping on Windows, which splits on whitespace and leaves backslashes alone.

```
C:\Tools\agents-shipgate.exe                          -> ('C:\\Tools\\agents-shipgate.exe',)
"C:\Program Files\ags\agents-shipgate.exe" --flag     -> ('C:\\Program Files\\ags\\agents-shipgate.exe', '--flag')
uv run agents-shipgate                                -> ('uv', 'run', 'agents-shipgate')   # POSIX, unchanged
```

`test_a_windows_cli_override_keeps_its_backslashes` covers the bare and quoted-with-spaces forms you named.

src/agents_shipgate/invocation.py
pengfei-threemoonslab: Fixed in ef60d509. Correct — I had treated both quote styles as making a metacharacter inert, and only single quotes do.

`_has_shell_syntax` now tracks quote style: inside `"..."` the expansion characters `$` and `` ` `` still count, so `printf "$HOME"`, `"$(id)"`, and ``"`id`"`` all withhold the pair. Inside `'...'` nothing does, so `printf '$HOME'` still publishes argv — that one really is a literal.

Parametrized over all three forms, with `agents-shipgate verify --base "HEAD~1"` as the counterweight so the rule cannot quietly degenerate into refusing everything double-quoted.

src/agents_shipgate/cli/apply_patches.py
pengfei-threemoonslab: Fixed in ef60d509, with one narrowing I want to flag explicitly rather than bury.

A rank-1 **command** now wins outright — that is your case, and the dangerous one: the legacy field could hold prose where a runnable command existed, and worse, could keep naming a console script the array had already retargeted away from.

```
next_action    : /Users/.../python -m agents_shipgate scan -c shipgate.yaml --suggest-patches --format json
next_actions[0]: /Users/.../python -m agents_shipgate scan -c shipgate.yaml --suggest-patches --format json
AGREE? True
```

I did **not** extend it to the other kinds. Deriving those from the array degraded two existing messages that the suite caught:

- `test_scan_cleanup_failure...`: `Remove <path> and re-run scan` → `Edit <path>`
- `test_scenario_suggest...`: `Inspect the error message and adjust --from or --out accordingly.` → `Review: Inspect the error message...`

For a non-command kind there is no program to disagree about, and the generic projection is a lossy formatting of a sentence the caller wrote more usefully. So the rule implemented is: a rank-1 command is authoritative and verbatim; other kinds keep the caller's sentence, still retargeted so prose cannot name a stale entry point either. `docs/diagnostics.md` states it that way rather than promising a projection the code does not perform.

Happy to take the full derivation instead if you would rather have the stricter invariant and accept the two message regressions — say the word.

src/agents_shipgate/schemas/diagnostics.py
pengfei-threemoonslab: Fixed in ef60d509. Both vectors reproduced before the fix — the second is the ugly one:

```
a.command = "rm -rf /"
a.model_dump()  ->  {"command": "rm -rf /", "executable": ["agents-shipgate"], "args": ["verify", "--json"]}
```

Took the stronger of your two remedies. `executable` and `args` are `@computed_field` properties now, derived from `command` on every read and every serialization. That closes `model_copy(update=...)`, plain assignment, and any future mutation path at once, and because they are no longer model fields, `extra="forbid"` rejects a caller trying to supply them:

```
NextAction(..., executable=["rm"], args=["-rf", "/"])  ->  ['extra_forbidden', 'extra_forbidden']
```

"The two forms cannot disagree" is now a property of the type rather than a claim about one code path — which is what I should have written the first time, since a construction-time projection on a mutable model can only ever be a convention.

Pinned by `test_structured_pair_cannot_be_supplied` and `test_the_structured_pair_cannot_go_stale`, the latter parametrized over both the copy and assignment paths.

pengfei-threemoonslab:
All six addressed in ef60d509. Inline replies on each thread; summary here.

| # | Finding | Resolution |
|---|---|---|
| P1 | `list2cmdline` is argv quoting, not shell escaping | Windows branch removed; `command` documented as a POSIX rendering, structured argv is the authoritative runnable route |
| P1 | POSIX parse of a Windows-rendered tail (`C:\repo` → `C:repo`) | One renderer, one parser — round-trips Windows paths exactly |
| P1 | POSIX `shlex` on a Windows `AGENTS_SHIPGATE_CLI` | Override parsed with host rules; backslashes and quoted spaces preserved |
| P2 | Expansion inside double quotes | `$`/`` ` `` detected inside `"…"`; only single quotes make them inert |
| P2 | Legacy `next_action` vs normalized array | A rank-1 command wins outright (see caveat below) |
| P2 | `executable`/`args` could go stale | `@computed_field` — derived on every read, and supplying them is now `extra_forbidden` |

**The two P1 quoting findings had one cause,** and fixing them separately would have been wrong. A string rendered by one set of rules and parsed by another cannot be made safe by improving either end — so there is now one renderer and one parser, both POSIX, which round-trips Windows paths exactly because `'C:\repo'` keeps its backslashes inside single quotes.

What that cannot do is make the string safe to paste into `cmd.exe`, and nothing would: single quotes are not quoting there. Rather than fake it with a shell-specific escaper — which would also mean guessing which shell the agent is in — I took your second remedy and narrowed the promise. `command` is a POSIX rendering for display and POSIX shells; `[*executable, *args]` is authoritative on every platform. The docs say so explicitly, including "do not paste `command` into `cmd.exe` or PowerShell". Removing the branch also removed the `subprocess` import, so last round's trust-model allowlist entry and its `STABILITY.md` paragraph are withdrawn.

**The computed-field finding was the one I should have gotten right the first time.** A projection fixed at construction on a mutable model is a convention, not a guarantee, and `a.command = "rm -rf /"` serializing `args: ["verify", "--json"]` shows exactly how little the claim was worth. Deriving on every read makes it a property of the type.

**One narrowing I want visible rather than buried.** On the legacy `next_action`, I implemented "a rank-1 *command* wins outright" rather than "derive from the first entry" for every kind. Full derivation degraded two existing messages that the suite caught — `Remove <path> and re-run scan` → `Edit <path>`, and a specific sentence → `Review: <same sentence>`. For a non-command kind there is no program to disagree about, and the generic projection is a lossy formatting of something the caller wrote more usefully. `docs/diagnostics.md` now states that rule rather than promising a projection the code does not perform. If you would rather have the stricter invariant and accept those two regressions, say so and I will switch it.

**Verification.** Full suite green; `generate_schemas.py --check`, `ruff check`, `compileall`, and the trust-model scanner all clean; `llms-full.txt` regenerated. Every repro in your review re-run against this head — the `C:\repo` round-trip asserted under both `_WINDOWS` settings so it holds on either host — plus the nine-surface sweep against a configured repository: zero non-retargeted commands.

pengfei-threemoonslab (commented):
Engineer re-review of updated head `ef60d5096`.

**Verdict: request changes.** The update fixes the earlier apply-patches projection, ordinary stale argv fields, shell-expansion detection, Windows backslash round-tripping, and durable artifact determinism. Seven actionable issues remain: five P1 correctness/platform gaps and two P2 contract/documentation gaps, detailed inline.

Validation against this exact head: GitHub `test`, `verify`, and `verify-self` are green; the full local non-packaging suite, focused regressions, Ruff, compileall, and generated-schema checks pass. Shipgate reports `control.state=review_publishable` / `decision=review_required` because this PR changes a release trust root.

GitHub does not allow the PR author to submit formal `REQUEST_CHANGES` on their own PR, so this is posted as a `COMMENT` review; the engineering verdict remains blocking.

src/agents_shipgate/invocation.py
```diff
@@ -0,0 +1,462 @@
...
+
+    index = 0
+    length = len(command)
+    while True:
+        while index < length and command[index].isspace():
+            index += 1
+        if index >= length:
+            return None
+        end = index
+        while end < length and not command[end].isspace():
```
pengfei-threemoonslab, line 429: [P1] `_program_token_span()` stops at whitespace without honoring escapes. `NextAction(command=r'/opt/my\ tool --flag', ...)` emits `executable=['/opt/my\\']` and `args=['tool', '--flag']`, while POSIX parsing invokes `['/opt/my tool', '--flag']`. The preferred structured representation therefore executes a different program. Parse the complete external command with the same grammar, or make this span quote/escape-aware.

src/agents_shipgate/invocation.py
```diff
@@ -0,0 +1,462 @@
...
+    with it and parsing with POSIX ``shlex`` also turned ``C:\\repo`` into
+    ``C:repo``, which is worse than an unrunnable string: it is a *runnable*
+    command against the wrong workspace.
+
+    Uniform POSIX quoting round-trips Windows paths exactly, because a
+    single-quoted ``'C:\\repo'`` preserves backslashes literally. What it does
+    not do is make the string safe to paste into ``cmd.exe`` — nothing would,
+    since single quotes are not quoting there. So the promise is narrowed
+    rather than faked: ``command`` is a POSIX rendering for display and for
+    POSIX shells, and ``[*executable, *args]`` is the authoritative runnable
```
pengfei-threemoonslab, line 180: [P1] This authoritative argv promise is not implemented on the primary control surfaces. `CodingAgentCommandAction`, preflight actions, verifier/fix-task commands, `allowed_next_commands`, and `verification_command` still expose only the POSIX-rendered string. On Windows module runs, cmd.exe/PowerShell cannot execute that string and there is no structured sibling to use. Add a versioned argv representation across every operational control contract (including allowed/verification commands), or provide another host-runnable form.

src/agents_shipgate/invocation.py
```diff
@@ -0,0 +1,462 @@
...
+* Entered through ``python -m agents_shipgate`` → :data:`sys.executable` plus
+  ``-m agents_shipgate``. The interpreter is spelled by path rather than as a
+  bare ``python`` on purpose: a bare name resolves through ``PATH`` and can
+  easily land on a *different* interpreter that has no Shipgate installed, or
+  an older one that does — which is the failure this module exists to prevent,
+  reintroduced one layer down.
+* ``AGENTS_SHIPGATE_CLI`` overrides both. It already exists as the operator
+  override for the command written into Claude Code hooks
+  (:mod:`agents_shipgate.cli.install_hooks`), and an operator who has told
+  Shipgate how to invoke itself there means it everywhere.
```
pengfei-threemoonslab, line 29: [P1] The generated Claude Code hook does not yet honor this override with the same semantics. Its `_cli()` still calls POSIX `shlex.split` directly; with `AGENTS_SHIPGATE_CLI=C:\Tools\agents-shipgate.exe`, the rendered hook returns `['C:Toolsagents-shipgate.exe']` even though this module now preserves the path. Mirror the platform-aware parser in the generated hook, or render exact tokens into it.

src/agents_shipgate/cli/_register_init.py
```diff
@@ -334,24 +334,23 @@ def init(
                 _validate_manifest_text(template)
             except Exception as exc:  # noqa: BLE001 - validation surface
                 typer.echo(f"Generated manifest failed validation: {exc}", err=True)
+                minimal_action = NextAction(
```
pengfei-threemoonslab, line 343: [P1] This normalizes the failure route, but successful init still bypasses the policy. `python -m agents_shipgate init --minimal --write --json` emits `next_action: "Replace placeholders, then run: agents-shipgate scan -c shipgate.yaml"`; a source checkout without the wrapper still dead-ends. Build the successful recovery from a normalized action too, and add successful init to the end-to-end invocation sweep.

src/agents_shipgate/schemas/diagnostics.py
```diff
@@ -28,6 +60,57 @@ class NextAction(BaseModel):
     why: str
     expects: str | None = None
 
+    @computed_field  # type: ignore[prop-decorator]
```
pengfei-threemoonslab, line 105: [P1] The public model cannot consume the v23 payload these computed fields produce. `NextAction.model_validate(action.model_dump(mode='json'))` raises `extra_forbidden` for `executable` and `args` because `extra='forbid'` remains, and the same failure occurs through `Diagnostic` and actual agent-mode output. The validation JSON Schema also omits these advertised properties. Keep them derived, but make the wire model/schema accept and validate its own serialization (or separate construction and wire models).

src/agents_shipgate/invocation.py
```diff
@@ -0,0 +1,462 @@
...
+    value ``C:\\Tools\\agents-shipgate.exe`` collapses to
+    ``C:Toolsagents-shipgate.exe`` — a path that does not exist, produced from
+    a value the operator wrote correctly. Non-POSIX mode splits on whitespace
+    and keeps backslashes; it also keeps the quote characters, so surrounding
+    quotes are stripped here rather than left in the token.
+    """
+
+    try:
+        if _WINDOWS:
+            return [_strip_outer_quotes(token) for token in shlex.split(raw, posix=False)]
```
pengfei-threemoonslab: [P2] `shlex.split(..., posix=False)` preserves simple Windows paths but is not Windows command-line parsing. For example, `uv run --project="C:\My Project" agents-shipgate` is split inside the quoted option value. Use a Windows-native parser/structured override, or explicitly document and validate a restricted grammar instead of claiming host rules.

src/agents_shipgate/schemas/contract.py
```diff
@@ -65,12 +65,21 @@
...
+# argv is unchanged on the wire.
 #
 # ``MINIMUM_CONTROL_CONTRACT_VERSION`` stays at 21: v22 adds a projection of the
-# ``AgentControl`` union and does not change the union itself, so a consumer
-# written against v21 control fields keeps reading them unchanged.
-CONTRACT_VERSION: Literal["22"] = "22"
+# ``AgentControl`` union and v23 changes only how commands inside it are
+# spelled, so a consumer written against v21 control fields keeps reading them
+# unchanged.
+CONTRACT_VERSION: Literal["23"] = "23"
```
pengfei-threemoonslab, line 82: [P2] Advancing the runtime contract to v23 also needs the authoritative migration note. `STABILITY.md` still begins with the 21→22 compact-control migration and contains no v23 section, despite `docs/agent-contract-current.md` saying STABILITY updates first. Document the strict `NextAction` shape, structured argv behavior, POSIX-only command semantics, and minimum-control compatibility there.

src/agents_shipgate/invocation.py
pengfei-threemoonslab: Fixed in 0f052538. Reproduced exactly — and it was the same mistake as the Windows one two rounds ago, wearing different clothes: a second grammar. Scanning to whitespace is a parser, just a wrong one.

```
before: executable=['/opt/my\\']      args=['tool', '--flag']
after : executable=['/opt/my tool']   args=['--flag']        # == shlex.split(...)
```

`split_invocation` now parses the whole command with the grammar it was rendered with, and takes the split point from the resolved invocation's *token count* — three for `python -m agents_shipgate`, one for a console script — rather than from locating a substring. Leading `NAME=VALUE` detection moved to the token list too.

The span scanner survives only inside `retarget_command`, where it is sound because it acts on the result exclusively when the token spells one of our console scripts — names containing neither quotes nor escapes. Its docstring now says so, and says not to reuse it for parsing.

src/agents_shipgate/invocation.py
pengfei-threemoonslab: Fixed in 0f052538 — you are right that I claimed host rules and shipped an approximation.

`split_windows_command_line()` implements the documented MS C-runtime / `CommandLineToArgvW` rules: a quote toggles quoting, `""` inside a quoted run is one literal quote, `2n` backslashes before a quote give `n` backslashes and toggle, `2n+1` give `n` backslashes and a literal quote, and backslashes not before a quote stay literal.

```
uv run --project="C:\My Project" agents-shipgate  -> ['uv', 'run', '--project=C:\\My Project', 'agents-shipgate']
"a b" c "d""e"                                    -> ['a b', 'c', 'de']
"C:\dir\\" next                                   -> ['C:\\dir\\', 'next']
C:\Tools\agents-shipgate.exe                      -> ['C:\\Tools\\agents-shipgate.exe']
```

Parametrized over all four plus the quoted-path-with-spaces case, and the table is shared with the hook-parity test below so the two implementations cannot drift.

src/agents_shipgate/invocation.py
pengfei-threemoonslab: Fixed in 0f052538. Confirmed — the hook is a second reader of the same variable, and I had fixed only one of them.

The hook ships as a standalone script and cannot import the package, so rather than render tokens (which would freeze the value at install time and ignore a later `AGENTS_SHIPGATE_CLI` change), the rules are duplicated inside the template and **pinned by a test**: `test_generated_hook_parses_the_cli_override_like_the_package` loads the rendered script and compares its parser against `invocation.split_windows_command_line` over the same case table. Duplication is only safe when divergence fails a test, which is exactly what went wrong here.

Also caught while writing that test: a backslash in the docstring I added made the generated hook emit `SyntaxWarning: invalid escape sequence '\T'` — noise in someone else's repository, produced by us. Reworded, and `test_generated_hook_compiles_without_warnings` compiles the rendered script under `-W error` so it cannot come back.

src/agents_shipgate/cli/_register_init.py
pengfei-threemoonslab: Fixed in 0f052538. Confirmed, and it is a fair characterisation of the whole PR so far: every fix landed on a failure route, because those were what the repro exercised. A successful `init --write` dead-ended on the very next step it told the caller to take.

Both success routes are built from a `NextAction` now, so `next_action` is the rank-1 command verbatim and `next_actions[]` carries the structured pair:

```
next_action: /Users/.../python -m agents_shipgate scan -c shipgate.yaml
executable : ['/Users/.../python', '-m', 'agents_shipgate']
args       : ['scan', '-c', 'shipgate.yaml']
```

The prose that used to wrap the command ("Replace placeholders, then run: …") moved into the action's `why`, where it is not competing with a machine-readable field.

`test_successful_init_routes_to_a_runnable_scan` runs a real `python -m agents_shipgate init --minimal --write --json` and asserts all three, including that the legacy string equals the rank-1 command.

src/agents_shipgate/schemas/diagnostics.py
pengfei-threemoonslab: Fixed in 0f052538. Confirmed on all three counts — the model rejected its own payload, `Diagnostic` inherited the failure, and the validation schema advertised nothing.

A `mode="before"` validator strips `executable`/`args` from incoming dicts, so the model reads back what it wrote. Accepting them is deliberately *not* believing them: the projection is still computed from `command`, so a pair edited in transit validates to the pair its command actually implies.

```
NextAction.model_validate(action.model_dump(mode="json"))            # round-trips
Diagnostic.model_validate(diagnostic.model_dump(mode="json"))        # round-trips
model_validate({..., "executable": ["rm"], "args": ["-rf", "/"]})    # -> ['verify', '--json']
```

The validation JSON Schema now documents both properties as `readOnly`, with the ignored-on-input rule in the description.

One thing I did **not** fully solve, flagged rather than buried: `model_json_schema(mode="serialization")` still returns an open object, because the wrap serializer that omits the pair when absent replaces the inferred shape before `json_schema_extra` runs. Validation mode is the consumer-facing schema and is complete, and nothing in the repo generates a serialization schema for this model, so I left it with a comment naming the trade-off. The alternative is emitting `null` instead of omitting — which restores the serialization schema at the cost of widening the wire for every action that can never carry an argv. Say the word if you would rather have it that way round.

src/agents_shipgate/invocation.py
pengfei-threemoonslab: You are right that the docs overclaimed, and I have fixed the claim rather than the contracts. Explaining the call rather than burying it, since it is the one place I did not do what the finding asked.

**What changed in 0f052538.** The promise is now scoped to where it holds — `next_actions[]` — and every other surface gets an *exact*, documented recovery instead of a vaguer one:

```python
subprocess.run(shlex.split(control["next_action"]["command"]))
```

That is exact on every surface and every host, because the rendering is uniformly POSIX with no exceptions — the property the previous round established. It is now a guarantee rather than an observation: `test_every_control_surface_command_recovers_exact_argv` walks `verify --json`, `check --format agent-boundary-json`, and `check --format agent-control-json`, and asserts every command value recovers to a token list that is stable under a re-render, with a negative control so it cannot go vacuous. `AGENTS.md`, `docs/diagnostics.md`, `docs/errors.json`, and the new `STABILITY.md` v23 note all state the scope and the recovery, and say never to use `shell=True`.

**Why not the contracts, in this PR.** The envelope reuses the `AgentControl` union directly, so adding the pair to `CodingAgentCommandAction` is a union change, and that cascades further than it looks:

1. `minimum_control_contract_version` must go `21 → 23` — `STABILITY.md`'s own rule is that the minimum holds only while the union is unchanged.
2. That cascades to `.well-known`, README ×2, `docs/agents/use-with-codex.md` ×3, every adoption-kit literal via the render-hash protocol, and the pinned test literals.
3. `docs/codex-boundary-result-schema.v2.json` is frozen and `additionalProperties: false`, and `CodexBoundaryResultV2` shares the same `control`, so any new union field makes `--format codex-boundary-json` violate its own published schema until a serialization-boundary down-projection is added.
4. `allowed_next_commands` is a `list[ExactCommand]`, so giving it structured siblings is a shape change rather than an addition.

The envelope has headroom (~2.2 KB against the published 4096 budget), so size is not the blocker — the blocker is that this is a change to the authoritative control contract, and it deserves its own review rather than a fourth round bolted onto a PR about spelling. Filed as **#369** with those four items as acceptance criteria, cross-linked from `docs/diagnostics.md` and `STABILITY.md`.

If you would rather it land here, say so and I will do it — I would just want it reviewed as its own change.

src/agents_shipgate/schemas/contract.py
pengfei-threemoonslab: Fixed in 0f052538. You are right that the contract advance is not done until STABILITY carries it — and `docs/agent-contract-current.md` says so explicitly.

Added *Migration Note: unreleased — commands spelled for the invocation that emitted them*, ahead of the 21→22 note, covering:

- Contract `22 → 23`, and why `minimum_control_contract_version` **stays at 21** (the `AgentControl` union is unchanged; only the spelling of commands inside it changed).
- The entry-point rules, including `AGENTS_SHIPGATE_CLI` precedence and its host-specific parsing.
- `command` as a POSIX rendering on every platform, explicitly **not** a host-shell promise, with `shlex.split` as the exact recovery.
- The strict `NextAction` shape: why `extra="forbid"` makes the two properties non-additive (which is what the version carries), the omitted-not-null rule, when the pair is withheld, and the accept-and-recompute round-trip behaviour.
- The scoping to `next_actions[]`, with the control-contract extension pointed at #369.
- That durable artifacts and published contract vocabulary stay canonical.

pengfei-threemoonslab:
All seven addressed in 0f052538. Inline replies on each thread; summary here.

| # | Finding | Resolution |
|---|---|---|
| P1 | `_program_token_span` ignores escapes (`/opt/my\ tool`) | Whole command parsed with one grammar; split point from the invocation's token count |
| P1 | Argv promise not implemented on control surfaces | Claim scoped to `next_actions[]`; exact `shlex.split` recovery documented and **tested** everywhere else; contracts tracked in #369 |
| P1 | Generated hook still POSIX-splits the override | Rules mirrored in the hook, pinned by a parity test against the package parser |
| P1 | Successful `init` bypasses the policy | Built from a normalized action, carries structured argv, added to the sweep |
| P1 | Model cannot validate its own serialization | Pair stripped on input and recomputed; schema documents both properties |
| P2 | `posix=False` is not Windows parsing | Documented MS C-runtime rules implemented |
| P2 | No v23 migration note | `STABILITY.md` section added |

**Two of these are the same mistake I have now made three times**, which is worth naming. `_program_token_span` scanning to whitespace is a second grammar, exactly like `list2cmdline`-render/POSIX-parse was; and the hook's `_cli()` is a second reader of `AGENTS_SHIPGATE_CLI`, exactly like `apply-patches` was a second emitter. Both times the fix was not "improve the second one" but "make there be one" — one parser, and one parser mirrored under a parity test where a literal shared implementation is impossible. I have stopped patching the instances and started looking for the duplicates.

**One finding I did not do as asked, deliberately.** On the control contracts I fixed the *claim* rather than the contracts: the docs now scope structured argv to `next_actions[]`, and every other surface gets an exact recovery — `shlex.split(command)`, exact everywhere because the rendering is uniformly POSIX, and now pinned by `test_every_control_surface_command_recovers_exact_argv` with a negative control. Extending `CodingAgentCommandAction` is a change to the `AgentControl` union: it raises `minimum_control_contract_version` `21 → 23`, cascades to `.well-known`/README/kit literals via the render-hash protocol, and makes `--format codex-boundary-json` violate its own frozen `additionalProperties: false` schema until a down-projection exists. Filed as **#369** with those as acceptance criteria. Happy to land it here instead if you would rather — I would just want it reviewed as its own change rather than as round four of a PR about spelling.

**Also flagged rather than buried:** `model_json_schema(mode="serialization")` still returns an open object, because the wrap serializer that omits the argv pair replaces the inferred shape. Validation mode — the consumer-facing one — is complete, and nothing in the repo generates a serialization schema for `NextAction`. The alternative is emitting `null` instead of omitting, which restores it at the cost of widening the wire for every action that can never carry an argv. Your call.

**Verification.** Full suite green twice under `-n 8` (one flake found and fixed on the way: my preflight test ran against the shared repo root and depended on what else the suite was writing there — it uses an isolated copy now). `generate_schemas.py --check`, `ruff`, `compileall` clean; `llms-full.txt` regenerated. Every repro in the review re-run against this head.

pengfei-threemoonslab (commented):
Engineer re-review of updated head `91d03bcf53a31078300436e2d3d1c4dad2b4e05b`.

**Verdict: request changes.** The latest update fixes the previously reported absolute-override, double-quoted `--out`, apply-patches, successful-init, escaped-space executable, same-process round-trip, and durable-artifact issues. Three P1 blockers and five P2 correctness/contract issues remain; seven are detailed inline.

One additional finding cannot be attached inline because its source line is unchanged:

- **[P2] Windows action has no runnable form** — `src/agents_shipgate/cli/diagnostics.py:161-177` emits `AGENTS_SHIPGATE_ENABLE_PLUGINS=1 agents-shipgate scan ...`. The leading assignment intentionally suppresses `executable`/`args`; with `command` now explicitly POSIX-only, Windows users have neither a host-runnable string nor structured execution data. Please provide an environment map or equivalent CLI flag, or emit a platform-appropriate non-command action.

GitHub CI is green. Locally, the focused suite, full suite excluding packaging, Ruff, compileall, and schema-generation check passed. Shipgate verification succeeded but returned `control_state=review_publishable` and `decision=review_required`.

_GitHub does not allow the PR author to submit formal `REQUEST_CHANGES` on their own PR, so this is posted as a `COMMENT` review; the engineering verdict remains blocking._

src/agents_shipgate/schemas/diagnostics.py
```diff
@@ -36,6 +181,8 @@ def _check_kind_fields(self) -> NextAction:
             raise ValueError("kind='edit' requires a non-empty path")
         if self.kind == "stop" and self.command is not None:
             raise ValueError("kind='stop' must not carry a command")
+        if self.kind == "command" and self.command:
+            self.command = retarget_command(self.command)
```
pengfei-threemoonslab, line 185: **[P1] Preserve producer wire data during validation**

Wire validation drops the producer's `executable`/`args` and then retargets `command` using the consumer process environment. I reproduced a payload emitted as `agents-shipgate verify --json` becoming `/opt/consumer/bin/agents-shipgate verify --json` merely because the consumer had that override. Module-emitted argv is also repartitioned across processes. Deserialization should preserve and validate the producer command/pair; invocation retargeting belongs only on locally constructed actions.

STABILITY.md
```diff
@@ -13,6 +13,75 @@ for reproducible CI.
...
+There is one renderer and one parser, and they must agree — a string rendered
+by one set of rules and parsed by another silently changes the values it
+carries. Uniform POSIX quoting round-trips Windows paths exactly, because a
+single-quoted `'C:\repo'` keeps its backslashes. It does **not** make the
+string safe to paste into `cmd.exe` or PowerShell, where single quotes are not
+quoting; nothing would. Recover argv instead:
+
+```python
+subprocess.run(shlex.split(command))   # exact on every surface and every host
+```
```
pengfei-threemoonslab, line 50: **[P1] Put the new control semantics behind a version boundary**

This changes operational `command` from an executable exact step into a POSIX-only representation requiring a new `shlex.split` protocol, while `minimum_control_contract_version` remains 21 and control surfaces provide no structured argv. Existing v21 Windows consumers cannot execute a module command containing a single-quoted Python path. Please advance the control/minimum version with structured down-projection, or retain a genuinely backward-compatible representation; deferring the representation to #369 leaves current v21 consumers stranded.

src/agents_shipgate/cli/install_hooks.py
```diff
@@ -691,9 +691,55 @@ def _project_root(payload: dict[str, Any]) -> Path:
...
+
+
 def _cli() -> list[str]:
     raw = os.environ.get("AGENTS_SHIPGATE_CLI", "agents-shipgate").strip()
-    return shlex.split(raw) if raw else ["agents-shipgate"]
+    if not raw:
+        return ["agents-shipgate"]
+    if os.name == "nt":
+        return _split_windows_command_line(raw) or ["agents-shipgate"]
+    return shlex.split(raw)
```
pengfei-threemoonslab, line 742: **[P1] Use the selected CLI in hook recovery guidance**

The generated hook now uses `_cli()` for subprocess execution, but `_manual_verify_command()` and the missing-manifest branches still hardcode `agents-shipgate`. With `AGENTS_SHIPGATE_CLI` pointing to an override-only wrapper containing spaces, `_cli()` returns the correct wrapper while the exact manual recovery instruction repeats the absent-wrapper failure from #322. Its leading POSIX environment assignment is also not runnable in cmd.exe or PowerShell. Build recovery from `_cli()` and represent environment separately.

src/agents_shipgate/invocation.py
```diff
@@ -0,0 +1,522 @@
...
+                in_quotes = False
+                started = True
+            else:
+                in_quotes = True
+                started = True
+            pending_backslashes = 0
+            continue
+        flush_backslashes(pending_backslashes)
+        pending_backslashes = 0
+        if char.isspace() and not in_quotes:
```
pengfei-threemoonslab, line 238: **[P2] Implement the documented CRT grammar**

The parser drops the literal quote represented by doubled quotes inside a quoted run, and `isspace()` incorrectly treats Unicode whitespace such as NBSP as an argument delimiter. The generated hook mirrors both defects, while the new test encodes the wrong doubled-quote result. Microsoft documents only space/tab delimiters and treats a doubled quote within a quoted run as one literal quote: https://learn.microsoft.com/en-us/cpp/c-language/parsing-c-command-line-arguments?view=msvc-170

src/agents_shipgate/invocation.py
```diff
@@ -0,0 +1,522 @@
...
+        return None
+    try:
+        tokens = shlex.split(command)
+    except ValueError:
+        return None
+    if not tokens:
+        return None
+    if _ENV_ASSIGNMENT.fullmatch(tokens[0]):
+        # Leading ``NAME=VALUE`` assignments are shell syntax, not argv.
+        return None
```
pengfei-threemoonslab, line 513: **[P2] Detect assignments before dequoting them**

Assignment detection runs after `shlex.split`, so `FOO='bar baz' agents-shipgate verify` becomes tokens beginning with `FOO=bar baz`, which no longer matches the no-whitespace regex. The emitted structured form consequently tries to execute a program named `FOO=bar baz`. Detect the assignment lexically before dequoting, or parse it with an assignment-aware grammar.

tests/test_invocation_policy.py
```diff
@@ -0,0 +1,1016 @@
...
+    # No per-format non-vacuity assertion: a format may legitimately publish no
+    # command in a given state — the compact envelope carries a human review
+    # action here, and a human action never exposes one. Coverage of the
+    # control contract is pinned once, below, where it is deterministic.
+    for command in commands:
+        tokens = shlex.split(command)
+        assert tokens, command
+        # Stable under a second pass: re-rendering and re-parsing the recovered
+        # argv yields the same tokens, so nothing was lost or invented.
+        assert shlex.split(join_argv(tokens)) == tokens, command
```
pengfei-threemoonslab, line 971: **[P2] Compare recovery with the intended argv**

This parses each command and then proves that rerendering those already-parsed tokens yields the same tokens. It cannot detect information lost during the first parse; for example an unquoted Windows `C:\repo` can become `C:repo` and still pass. Compare against the fixture's known intended prefix and arguments.

src/agents_shipgate/schemas/diagnostics.py
```diff
@@ -1,12 +1,64 @@
...
+    schema.setdefault("properties", {}).update(_ARGV_PROPERTY_SCHEMA)
+
+
+# Known limitation: ``model_json_schema(mode="serialization")`` returns an open
+# object for this model, because a wrap serializer replaces the inferred shape
+# before ``json_schema_extra`` runs. Validation mode is the documented schema
+# and is complete; nothing in the repo generates a serialization schema for
+# ``NextAction``. Emitting ``null`` instead of omitting the pair would restore
+# it, at the cost of widening the wire for every action that can never carry an
+# argv — see ``_drop_absent_argv``.
```
pengfei-threemoonslab, line 60: **[P2] Publish one consistent v23 wire contract**

Runtime now accepts, discards, and recomputes supplied `executable`/`args`, while `docs/agent-contract-current.md`, the changelog, `llms-full.txt`, and the model prose still say they cannot be supplied and raise `extra_forbidden`. The serialization JSON Schema also remains an unconstrained object even though v23 specifically changes the output wire shape. Align the documented input behavior and publish a typed serialization schema or explicit wire model.

PR #368, opened 2026-08-11
fix(release): publish the CHANGELOG section and lock the release-time dev closure

### LINKED ISSUE #345: release cleanup: use CHANGELOG notes and lock release dependencies
## Summary

Two useful, non-blocking release-maintenance items remain from the broader release-workflow review. Higher-risk work was split into dedicated v0.16 issues so it cannot hide inside a P2 cleanup bundle:

- #355 — non-publishing rehearsal path
- #356 — wheel-scoped, signed SBOM
- #343 — publication transaction, concurrency, permissions, and recovery
- #344 — deterministic test workload and measured timeout

## 1. Publish the matching CHANGELOG section

The workflow currently publishes a placeholder release note even though `CHANGELOG.md` contains detailed entries.

### Acceptance

- [ ] Extract the section matching the release tag and pass it through `--notes-file`.
- [ ] Fail before publication when the matching section is absent.
- [ ] Preserve Markdown faithfully and test at least one multi-section changelog.

## 2. Lock release-time development dependencies

`pip install -e ".[dev]"` resolves fresh at release time, so the release result can differ from the CI run that approved the commit.

### Acceptance

- [ ] Release installs from a reviewed constraints file or lock that matches CI.
- [ ] The lock/constraints update path is documented and automated.
- [ ] A stale or inconsistent lock fails before publication.

## Preserve these existing controls

- Actions stay pinned to full commit SHAs.
- The protected `pypi` environment continues to scope qualification inputs.
- `dist/` and `qualified-dist/` pre-existence and symlink checks remain.
- Downloads continue to enforce HTTPS-only protocols.
- Qualified wheel filenames remain strict and traversal-safe.
- The tag ↔ package-version check remains, with its rationale.

Split from the original #345 scope during PM review on 2026-08-09.

pengfei-threemoonslab:
## Summary

Closes #345 — the two non-blocking leftovers from the release-workflow review.

- **The release page now carries the changelog.** The GitHub Release body was
  the placeholder `Agents Shipgate <tag>` while `CHANGELOG.md` held the entry
  describing what actually shipped, so the one artifact users read said nothing.
  `scripts/release_notes.py` extracts the section matching the tag and `stage`
  publishes it through `--notes-file`, verbatim, from the checkout pinned to the
  verified commit rather than retyped at tag time. **A missing section fails
  verification**, which the rehearsal also runs, so it is caught while the tag
  does not yet exist — and `## Unreleased` never matches a tag, which is what
  makes promoting that heading a step the pipeline enforces rather than one an
  operator remembers. A body over GitHub's 125,000-character limit is refused
  there too, rather than by a 422 after tagging; the 0.16 development section is
  already ~75,000 characters. Draft repair rewrites the body as well as the
  assets; an already-published release is still left untouched.
- **CI and the release install the same closure.** `pip install -e ".[dev]"`
  resolved fresh at tag time, so the run that decided whether to publish could
  install different packages than the run that approved the commit — a ruff,
  pytest or plugin release landing in between was enough — and a release-only
  failure was not reproducible from the same tree. Both now install the
  hash-locked `constraints/dev.txt` with the byte-identical command, add the
  project separately (`--no-deps`, backend pinned by
  `constraints/release-build.txt`) because an editable install cannot be hashed,
  and run `python -m pip check` to prove the closure satisfies what the project
  declares.
- **The lock update path is one command, and staleness is gated.**
  `scripts/update_locks.py` recompiles every lock and restores the prose header
  `uv` would overwrite — that overwrite is why regenerating had been a per-file
  ritual. `scripts/verify_dependency_lock.py` runs in CI **and** before
  publication, failing on a declared requirement with no pin, a pin outside the
  declared range, a direct requirement the declarations no longer contain (uv's
  `# via` provenance makes that detectable), or a pin without a hash. It never
  re-resolves against the index, so an unrelated upload cannot turn the build
  red.

### For the reviewer

**`v0.16.0b7` fails the new changelog gate today, by design** — `CHANGELOG.md`
still says `## Unreleased`. Cutting a release now starts by promoting that
heading to `## <version> - <date>`; it is step 1 in the runbook and is called
out in `CONTRIBUTING.md`.

Every control #345 asked to preserve is untouched: SHA-pinned actions, the
protected `pypi` environment, the `dist/` and `qualified-dist/` pre-existence
and symlink checks, HTTPS-only downloads, the strict qualified-wheel filename,
and the tag ↔ package-version check. The one new job-written path
(`release-notes.md`) follows the same pre-existence rule as the directories the
verification job creates, so a committed symlink cannot decide where the write
lands.

Nothing here changes what the token-bearing jobs install. `constraints/dev.txt`
is the *verification* closure: it installs the project and runs its tests, so it
is outside the blast radius `release-seal.txt` and `release-publish.txt` are
locked against. That distinction is written into the lock's header.

## Type

- [ ] Input adapter change
- [ ] Documentation only

(Release pipeline and repository tooling; no product surface, no new check IDs,
no schema or contract version change.)

## Verification

CI is authoritative for `python -m ruff check .`,
`python -m compileall -q src tests`, and `python -m pytest`.

Additional local checks run:

- **CI's own selection re-run inside a clean 3.12 venv built from the new
  lock** — `pytest -n auto -m "not perf" --cov-fail-under=85`: green, 88.09%
  coverage. `ruff` (0.16.2, from the lock), `compileall`, `generate_schemas.py
  --check`, `test_adapter_static_only.py`, `python -m build` + `twine check`,
  and `cyclonedx-py environment` all pass in that environment too. This is the
  check that matters for the lock change: the packages CI will use are the ones
  the suite actually ran against.
- Full suite on the development environment: green.
- 25 new tests in `tests/test_release_pipeline.py`, and every new gate was
  mutation-tested rather than assumed: restoring the placeholder `--notes`,
  deleting the changelog step from verification, and drifting CI's install from
  the release's each fail the corresponding test; the three real lock failure
  modes (range widened past the pin, dependency added, dependency removed) each
  fail `verify_dependency_lock.py` against this repository with an actionable
  message.
- Dogfood: `agents-shipgate verify --config shipgate-self.yaml` returns
  `review_required` — "edits a release trust root … a human must review it" —
  which is the correct verdict for a `.github/workflows/**` change and does not
  fail advisory CI.
- `pip_audit .` cannot run on this Mac (its temporary venv's `ensurepip` dies
  with SIGABRT); reproduced on the unmodified environment, so it is pre-existing
  and not caused by this change. CI is the authority for that step.

**Only CI can prove one thing I could not**: the lock was installed and
exercised on macOS/arm64, and the universal lock's Linux wheels resolve for the
first time on this PR's ubuntu runner.

## Release-readiness notes

- [x] New or changed check IDs are documented in `docs/checks.md` (none)
- [x] Report/schema changes are additive or documented in `STABILITY.md` (none)


.github/workflows/ci.yml
```diff
@@ -23,9 +23,28 @@ jobs:
...
+        # statement about the packages the release will actually use.
+        #
+        # The project is installed separately with `--no-deps` because an
+        # editable install cannot be hashed; `pip check` then proves the locked
+        # closure really satisfies its declared dependencies. The build backend
+        # is pinned by the same constraint that makes the released wheel
+        # byte-reproducible.
+        run: |
+          python -m pip install --require-hashes --requirement constraints/dev.txt
+          PIP_CONSTRAINT=constraints/release-build.txt python -m pip install -e . --no-deps
```
pengfei-threemoonslab: **[P1] Editable install escapes the lock**

`--no-deps` does not disable PEP 517 build isolation. This PR’s lock installs pip 26.2.1, where `PIP_CONSTRAINT` no longer constrains isolated build environments; the PR CI log confirms fresh `Installing build dependencies` and `Installing backend dependencies`, and `constraints/dev.txt` contains no Hatchling closure. CI and release verification can therefore execute different backend bytes as the index changes, defeating #345’s core invariant.

Please hash-lock and preinstall the complete backend closure, then add `--no-build-isolation` here and in release verification, with a regression that asserts both.

pengfei-threemoonslab (commented):
## Engineering review — changes requested

The overall direction is good and CI is green, but two correctness gaps defeat the core lock guarantees: the editable install still resolves its PEP 517 backend outside the reviewed lock, and the stale-lock verifier ignores extras, markers, and direct URLs. The four additional inline comments cover valid universal-lock shapes and release-note integrity/rehearsal edge cases.

I recommend addressing all six comments before merge, then rerunning the release-pipeline tests, the committed-lock verifier, and Agents Shipgate.

_Submitted as a comment review because GitHub does not permit a pull request author to formally request changes on their own pull request._

scripts/verify_dependency_lock.py
```diff
@@ -0,0 +1,266 @@
...
+        pin = pins.get(name)
+        if pin is None:
+            problems.append(
+                f"{target.source} declares {requirement} but {target.lock} pins no {name}; "
+                "the lock is stale. Regenerate it with scripts/update_locks.py."
+            )
+            continue
+        if requirement.specifier and not requirement.specifier.contains(
+            pin.version, prereleases=True
+        ):
```
pengfei-threemoonslab: **[P1] Full requirement semantics are ignored**

This loop validates only canonical name plus version specifier. `Requirement.extras`, `.marker`, and `.url` are ignored, and the lock parser discards pin markers. A declaration can change from `demo>=1` to `demo[feature]>=1`, move to a platform marker, or switch to a direct URL without regenerating; `verify_lock_target()` still returns no problems, and `pip check` does not activate the project’s dev extra to recover that coverage.

Please bind and compare complete normalized PEP 508 direct declarations (or a declaration digest), including sources and marker branches.

scripts/verify_dependency_lock.py
```diff
@@ -0,0 +1,266 @@
...
+            raise ReleaseError(
+                f"{lock_path}:{number} is not an exact pin ({raw.strip()!r}); a lock that "
+                "resolves at install time is not a lock."
+            )
+        name = canonicalize_name(match.group("name"))
+        if name in pins:
+            raise ReleaseError(f"{lock_path}:{number} pins {name} a second time.")
+        current = Pin(version=match.group("version"), line=number)
+        in_via = False
+        pins[name] = current
```
pengfei-threemoonslab: **[P2] Universal locks may repeat names**

`uv pip compile --universal` may validly emit the same distribution multiple times with different versions or URLs under disjoint markers, but this dictionary rejects every second canonical name. A future dependency fork across Python or platform versions will make `update_locks.py` generate a valid pip lock that this verifier refuses.

Please model marker-qualified pin lists instead of one pin per name.

.github/workflows/release.yml
```diff
@@ -200,14 +222,18 @@ jobs:
...
+              --notes-file release-notes.md
             echo "release_state=absent" >> "${GITHUB_OUTPUT}"
             echo "Created draft release ${RELEASE_TAG}."
             exit 0
           fi
 
           if [ "$(gh release view "${RELEASE_TAG}" --json isDraft --jq .isDraft)" = "true" ]; then
+            # Repair replaces the notes too. A draft left by an earlier attempt
+            # can carry a body from a superseded changelog, and finalisation
+            # re-derives every asset but never looks at the release text.
```
pengfei-threemoonslab, line 235: **[P2] Validate notes before undrafting**

The remote assets are content-bound before publication, but the release body is not. Between stage and finalize—including the protected-environment approval window—a release-write actor or integration can edit the draft notes; the final `gh release edit --draft=false` then publishes text that is not the reviewed changelog section.

Please carry a verified `release-notes.md` digest or copy through the handoff and compare or atomically reapply it when undrafting.

scripts/release_notes.py
```diff
@@ -0,0 +1,208 @@
...
+
+    headings: list[tuple[int, str]] = []
+    fence: str | None = None
+    for index, line in enumerate(lines):
+        fenced = _FENCE.match(line)
+        if fenced:
+            marker = fenced.group("marker")
+            if fence is None:
+                fence = marker
+            elif marker[0] == fence[0] and len(marker) >= len(fence) and not fenced.group("info"):
```
pengfei-threemoonslab: **[P2] Accept whitespace after closing fences**

CommonMark permits spaces or tabs after a closing fence, but `not fenced.group("info")` only accepts an exactly empty suffix. A valid closer such as `````   `` leaves `fence` open and raises `ReleaseError: unterminated code fence`, blocking rehearsal and release.

Please treat a suffix containing only spaces or tabs as empty and add a regression fixture.

.github/workflows/release-verify.yml
```diff
@@ -320,6 +336,20 @@ jobs:
...
+        # workflow, which means the rehearsal fails on it while the tag does
+        # not yet exist — the only point at which "rename `## Unreleased`" is
+        # still a cheap fix. `stage` extracts the same section from the same
+        # commit; this is what guarantees it will find one.
+        #
+        # Standard library only, so it needs no installed toolchain and runs
+        # before the build, the downloads, and the qualification checks.
+        env:
+          RELEASE_TAG: ${{ steps.candidate.outputs.release_tag }}
+        run: python scripts/release_notes.py --tag "${RELEASE_TAG}"
```
pengfei-threemoonslab: **[P2] Rehearsal misses the output-path guard**

This shared verification step only validates extraction. Stage later refuses any pre-existing file, directory, or broken symlink named `release-notes.md`, so a candidate containing that path can pass mandatory pre-tag rehearsal and fail only after the tag is pushed.

Please exercise the same precondition here or write the notes to a trusted `$RUNNER_TEMP` path.

pengfei-threemoonslab:
All six addressed in `171c694e`. Both P1s were real and both are reproduced below rather than taken on trust.

### P1 — Editable install escapes the lock

Confirmed, including the mechanism. Constraining hatchling to a version that does not exist still built successfully under this PR's pip 26.2.1:

```
$ PIP_CONSTRAINT=<hatchling==0.0.1> pip install -e . --no-deps
  Installing build dependencies: started
Successfully installed agents-shipgate-0.16.0b7      # exit 0
```

So `PIP_CONSTRAINT` was inert and the backend closure was resolving from the index on every run, exactly as the CI log shows.

New `constraints/build-backend.txt` is the backend's hashed closure, compiled from the existing hand-maintained `constraints/release-build.txt` — so the version still lives in one reviewed place and the reproducibility argument recorded there governs the closure. Both pipelines install it and pass `--no-build-isolation`. Regression asserts both flags, the backend lock, ordering, and the *absence* of the inert `PIP_CONSTRAINT`.

Verified in a clean 3.12 environment built with the new sequence: **zero** `Installing build dependencies` / `Installing backend dependencies` lines, backend resolves to 1.31.0, `pip check` clean, CI's own selection green at 88.09%.

The same inert claim appeared twice more, so I corrected both rather than fix one instance of the bug class: the sealing job's comment ("PIP_CONSTRAINT pins the build backend" — there `--no-isolation` plus `release-seal.txt` is what actually pins it), and the documented qualification-promotion command in `release-build.txt` and the runbook, which told a future operator to rely on `PIP_CONSTRAINT` with isolation and would silently have built with a different backend — the precise failure the byte-equality gate blames on the candidate.

### P1 — Full requirement semantics are ignored

Correct. Each lock now carries a generated block recording the normalized PEP 508 declarations it was compiled from:

```
#   declares: pytest<10,>=9.1.1
```

`update_locks.py` writes it, `verify_dependency_lock.py` compares it. I chose a readable block over a digest so the diff shows *what* changed, not just *that* it did. Names are canonicalized and specifier sets sorted, so `ruamel.yaml>=0.19.1,<1` and `ruamel-yaml<1,>=0.19.1` are one declaration.

Reproduced against the real `pyproject.toml` — both of these previously reported nothing:

```
pip-audit>=2.10.1,<3  ->  pip-audit[extra]>=2.10.1,<3
  constraints/dev.txt was not compiled from the current declarations
  (pyproject.toml now declares pip-audit[extra]<3,>=2.10.1; ...)

twine>=7.0.0,<8  ->  twine>=7.0.0,<8 ; sys_platform == 'linux'
  constraints/dev.txt was not compiled from the current declarations (...)
```

A lock with no block at all is rejected rather than assumed fine.

### P2 — Universal locks may repeat names

Fixed. Pins are marker-qualified lists per name; **every** branch must satisfy the declaration (checking only the first would fail open when one fork is out of range), and only a repeat under the *same* marker is a duplicate. Tests cover the valid fork, the out-of-range branch, and the genuine duplicate.

### P2 — Validate notes before undrafting

Fixed by atomic reapply. Verification publishes `release_notes_sha256`; `stage` and `finalize` both pass it back via `--expected-sha256` (rejecting empty/malformed, same fail-open rule `verify-manifest` documents). `finalize` re-derives the notes from `CHANGELOG.md` fetched at the verified commit — like the other stdlib-only scripts it runs — and sets body and draft flag in **one** `gh release edit`. Compare-then-fix would leave a window of its own, and reapplying after undrafting would publish the tampered text first.

### P2 — Accept whitespace after closing fences

Fixed; parametrized over ` ``` `, ` ```   `, ` ```\t `, ` ```` `, asserting both that the section ends correctly and that the following section is still reachable — i.e. the fence really closed.

### P2 — Rehearsal misses the output-path guard

Took the `$RUNNER_TEMP` option, which removes the class instead of duplicating the guard: all three jobs write there, no candidate-controlled path is involved, and the rehearsal now exercises the identical write. The pre-existence check is gone with it.

### Also

Locks installed into one environment must agree on every shared distribution, or the second `pip install` moves part of the first one's closure. `dev.txt` and `build-backend.txt` do share distributions (`packaging`, `pluggy`), so the guard is not vacuous.

### Verification

Every fix was mutation-tested — dropping `--no-build-isolation`, the digest binding, or the body from the undraft call each fails its test. Full suite green; clean-environment run of CI's selection green at 88.09% coverage; `ruff`, `compileall`, schema check, static-only lint, `build` + `twine check` all pass there. Dogfood still returns `review_required` (trust-root edit).

scripts/verify_dependency_lock.py
```diff
@@ -0,0 +1,460 @@
...
+            )
+            continue
+        for branch in branches:
+            if branch.version is None:
+                if requirement.specifier:
+                    problems.append(
+                        f"{target.lock} resolves {name}{branch.describe()} to a direct URL, "
+                        f"but {target.source} declares the range {requirement}."
+                    )
+                continue
```
pengfei-threemoonslab: **[P1] Pin URL and marker semantics remain unchecked**

The declaration block detects source-text drift, but actual pins are not matched by URL or marker applicability, and every missing conditional requirement is waived. Wrong URLs, wrong-platform pins, and missing applicable pins therefore pass `verify_lock_target()` with no problems.

Please match declaration and pin branches semantically—including URL identity and marker overlap—and fail closed unless a marker is proven unreachable for every supported environment.

scripts/verify_dependency_lock.py
```diff
@@ -0,0 +1,460 @@
...
+        for name, by_version in sorted(versions.items()):
+            if len(by_version) > 1:
+                spread = "; ".join(
+                    f"{version} in {', '.join(sorted(locks))}"
+                    for version, locks in sorted(by_version.items())
+                )
+                problems.append(
+                    f"{name} is pinned at different versions by locks installed together "
+                    f"({spread}); the later install would move the earlier closure."
+                )
```
pengfei-threemoonslab: **[P2] Universal marker handling is still syntactic**

This aggregation discards marker applicability and direct-URL identity. Together with the same-name cross-product above, it rejects valid disjoint universal forks, accepts overlapping branches merely because their marker strings differ, and misses conflicting URL branches.

Please compare declaration and pin branches only where marker regions overlap, and compare branches across distinct co-installed locks with URL identity included.

scripts/verify_dependency_lock.py
```diff
@@ -0,0 +1,460 @@
...
+    LockTarget(
+        lock="constraints/dev.txt",
+        source="pyproject.toml",
+        extras=("dev",),
+        purpose="the development closure CI and release verification both install",
+    ),
+    LockTarget(
+        lock="constraints/build-backend.txt",
+        source="constraints/release-build.txt",
+        purpose="the build backend installed so editable installs need no isolation",
```
pengfei-threemoonslab, line 181: **[P2] Bind the backend closure to pyproject**

The generated backend lock is checked against `release-build.txt`, but neither that pin nor the backend identity is checked against `pyproject.toml`'s `[build-system]`. A backend requirement bump or backend switch therefore leaves `verify_dependency_lock.py` green.

Please parse and gate `[build-system].requires` and `build-backend` against the hand-maintained pin and resolved closure, with mutation tests.

pengfei-threemoonslab (commented):
## Engineering re-review — changes requested

Reviewed amended head `171c694e`. The direct fixes for editable-install isolation, draft/final notes binding, ASCII fence closers, and runner-temporary extraction are present. The PEP 508 and universal-lock fixes remain partial, and the release path still has blocking setup/security gaps.

### Findings not anchorable in the displayed diff

1. **[P1] Publish and finalize cannot load the local action.** At `.github/workflows/release.yml:303` and `:439`, the `publish` and `finalize` jobs use `./.github/actions/install-release-toolchain`, but neither job checks out the repository. GitHub requires the repository to be checked out before a same-repository action can be used, so fresh runners fail while preparing the action, before publication or finalization. The tests currently assert this impossible pairing. Inline/fetch the installer, use a pinned remote action, or perform an immutable, credentials-disabled checkout compatible with the trust boundary. [GitHub workflow syntax](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax)

2. **[P1] The publication allowlist ignores direct URLs.** At `.github/actions/install-release-toolchain/action.yml:47-48`, the requested-distribution extraction accepts only `name==...` lines. A hashed `name @ URL` requirement is omitted from the allowlist comparison yet is still installed by pip in the token-bearing jobs. Parse every non-comment requirement fail-closed and either reject direct URLs/unknown forms or validate them explicitly; add mutation coverage for this case.

3. **[P2] CI's package build still bypasses the backend lock.** `.github/workflows/ci.yml:121` runs `python -m build` without `--no-isolation`. The hosted green run created two isolated environments and freshly resolved `hatchling>=1.31.0`, despite the hashed backend closure already being installed. Use `python -m build --no-isolation` and assert that every workflow build invocation uses either that locked environment or an equivalently locked isolated mechanism. [Observed CI job](https://github.com/ThreeMoonsLab/agents-shipgate/actions/runs/31468076318/job/93705109678)

4. **[P2] Already-published retries do not validate the release body.** The branch at `.github/workflows/release.yml:242-270` validates index state, assets, and signatures, then emits `release_state=published` without comparing the remote release body to the verified notes. Because release notes remain editable, a rerun can certify altered text and skip finalization. Compare the exact remote body (or its digest) before accepting the published state. [GitHub release documentation](https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository)

Four inline comments cover the remaining PEP 508 fail-open cases, universal marker semantics, build-system binding, and the CommonMark edge case.

GitHub records this as a comment review because an author cannot formally request changes on their own pull request.

scripts/release_notes.py
```diff
@@ -0,0 +1,270 @@
...
+            if fence is None:
+                fence = marker
+            elif (
+                marker[0] == fence[0]
+                and len(marker) >= len(fence)
+                # CommonMark allows spaces and tabs after a closing fence, so a
+                # closer with trailing whitespace must still close. Requiring an
+                # exactly empty suffix left the fence open and failed the whole
+                # release with "unterminated code fence".
+                and not fenced.group("info").strip()
```
pengfei-threemoonslab: **[P3] Restrict closing-fence whitespace**

Python's `.strip()` removes Unicode whitespace such as NBSP, while CommonMark permits only ASCII spaces and tabs after a closing fence. This extractor can therefore treat a non-closing fence as closed and misclassify a later heading.

Please use `.strip(" \t")` and add an NBSP negative regression. [CommonMark fenced code blocks](https://spec.commonmark.org/0.31.2/#fenced-code-blocks)

pengfei-threemoonslab:
All eight addressed in `91b2efea`. Both P1s were release-breaking and both are confirmed from the run logs of this PR rather than from reading.

### 1. [P1] Publish and finalize cannot load the local action

Confirmed and fixed. Neither job checked anything out while using `./.github/actions/install-release-toolchain`, and a local action is loaded from the workspace — so both would have failed while *preparing* the action, after `stage` had already created the draft. The action has never executed in any run, which is why nothing surfaced it.

Both jobs now check out that one directory:

```yaml
sparse-checkout: .github/actions/install-release-toolchain
sparse-checkout-cone-mode: false      # cone mode also materialises every root file
ref: ${{ needs.verify.outputs.source_sha }}
persist-credentials: false
```

The trust boundary is unchanged: the only file in the workspace is the one the job already executes as an action. I replaced the two tests that asserted "no checkout at all" with the narrower invariant that is actually true, plus a general one — *every* job using a local action must check the workspace out — so this cannot come back in another job.

### 2. [P1] The publication allowlist ignores direct URLs

Confirmed. `grep -oE '^name=='` skipped `name @ URL`, which pip installs happily, so a direct URL rode into the token-bearing jobs unreviewed. Every requirement line is now classified and anything not an exact pin is refused.

I also moved the check into its own step reading `release-toolchain.txt` from the working directory, so the suite runs *that exact script* against crafted lockfiles — a guard only a real release exercises is a guard nobody has tested. Results:

| lockfile | outcome |
|---|---|
| the committed `release-publish.txt` | pass |
| `evil @ https://x.invalid/e.whl` | refused — not an exact pin |
| `-e .` | refused — pip option |
| `uv>=0.11.7` | refused — not an exact pin |
| unlisted `requests-evil==1.0` | refused — not on the allowlist |
| comment-only | refused — "requests nothing" |
| `cffi==2.1.1 ; platform_python_implementation != 'PyPy'` | pass |

### 3. [P1] Pin URL and marker semantics remain unchecked

Right on all three counts — wrong URLs, wrong-platform pins and missing conditional requirements all passed. Markers are now compared by **evaluation**, not text, over CPython 3.12–3.14 × linux/darwin/win32 × x86-64/aarch64: for every environment a declaration selects there must be exactly one applicable pin, satisfying the declared range or naming the declared URL.

That gives the semantics you asked for: a conditional declaration is waived only when *no* supported environment selects it (`sys_platform == 'plan9'` → nothing to pin); one that applies on Linux and has no pin is reported with the environment named; a pin marked `win32` against an unconditional declaration is a missing pin everywhere else; and a wrong URL is caught because a URL declaration has no range for range-checking to say anything about.

### 4. [P2] Universal marker handling is still syntactic

Fixed with the same machinery. Overlap is decided by evaluation, so `sys_platform == 'linux'` and `os_name == 'posix'` are now recognised as both applying on Linux and rejected, while a genuine disjoint fork passes. The co-installed comparison is per environment with URL identity included, so two locks carrying different halves of one fork no longer conflict — only a disagreement inside a single environment does.

### 5. [P2] CI's package build still bypasses the backend lock

Confirmed from the log you linked: two isolated environments, `hatchling>=1.31.0` re-resolved each time. Now `python -m build --no-isolation`, asserted for every workflow build invocation. Verified in the locked environment: zero isolated environments created, `twine check` passes on both artifacts.

### 6. [P2] Bind the backend closure to pyproject

Added `build_system_problems()`, gating both `[build-system].requires` and `build-backend` against the closure. Reproduced against the real tree — switching to setuptools previously left everything green:

```
pyproject.toml builds with 'setuptools>=70', but constraints/build-backend.txt pins no setuptools;
  the wheel would be built by a backend nothing reviewed.
pyproject.toml builds with the 'setuptools.build_meta' backend, provided by setuptools,
  which constraints/build-backend.txt does not pin.
```

### 7. [P2] Already-published retries do not validate the release body

Fixed. That branch now fetches the remote body and compares it to the verified section, normalising **only** GitHub's CRLF and trailing newline (otherwise it would fail on every release; a single changed character still differs). It fails rather than rewriting — mutating a published release is what the branch exists to refuse — and the runbook says what to do about it.

### 8. [P3] Restrict closing-fence whitespace

`.strip(" \t")`, with an NBSP regression asserting the block stays open.

### Verification

Full suite green. CI's own selection re-run in a clean environment built with the new install sequence: 88.09%, and `build --no-isolation` + `twine check` clean there. Every fix mutation-tested — removing the sparse checkout, flipping cone mode on, reverting the CI build to isolation, and switching the declared backend each fail the corresponding gate.

PR #370, opened 2026-08-11
fix(init): scope first adoption to the project the diff changes

### LINKED ISSUE #363: verify --preview routes init to the repo root on monorepos, producing an unusable manifest
## Summary

`verify --preview` routes the user to run `init` at the **workspace root**. In a monorepo this produces an unusable manifest, and the information needed to pick the right directory (`changed_files`) is already in the preview payload.

## Evidence (reproduced 2026-08-10, CLI 0.16.0b7)

Preview on `google/adk-samples` PR #1917 returned:

```json
"next_action": {"kind": "initialize",
  "command": "shipgate init --workspace /…/adk-samples --write --json"}
```

Following it verbatim produced:

- `shipgate.yaml`: **797 lines, 252 `tool_sources`** — every ADK-looking file in the repo, including non-agent files such as `python/agents/RAG/deployment/run.py` and `eval/test_eval.py`
- `agent.name: "ask_rag_agent"` — auto-filled from the first of **160+** `Agent(name=…)` literals; unrelated to the PR under review
- framework detection `score: 752.0` (vs `5.0` for the correct subdirectory)
- ` M .gitignore` — the managed block was appended to the repo's tracked root `.gitignore`, adding an unrelated modification to the PR's diff

Running `init` in `python/agents/crypto-payroll-agent/` instead produced a 30-line manifest, one `tool_source`, `agent_name_candidates: [{"value": "crypto-payroll-agent", "source": "workspace_dir"}]`, and a self-contained new `.gitignore`.

Measured cost of the wrong scope: root-scoped `verify` takes **~4 min** (whole head tree is archived) vs **~5 s** for the subdirectory scan. The PR's 4 CRITICAL findings would also be diluted among findings from ~160 unrelated agents, and the alignment layer (`agent.name` / `declared_purpose` vs observed capability) is meaningless when one manifest covers 160 agents.

## Impact

This is the first instruction a new adopter receives, and the tool's own routing is what produces the bad outcome. The user cannot tell that the result is wrong — the JSON reports `manifest_status: "written"`, `is_agent_project: true`, `confidence: "high"`, and a *higher* detection score. A coding agent has no signal at all that it has just been misconfigured.

## Proposal

1. When `changed_files` is available, derive the suggested `--workspace` from the **deepest common ancestor of the changed paths that contains a project marker** (`pyproject.toml`, `package.json`, `go.mod`, …). For this PR that yields `python/agents/crypto-payroll-agent`.
2. When `init` detects multiple root-agent candidates (>1 distinct `Agent(name=…)` literal across disjoint project markers), do not silently pick the first. Report `agent_scope: "ambiguous"` with the candidate project directories and route the user to choose.
3. Never append the managed `.gitignore` block to a tracked file outside the resolved manifest directory without an explicit opt-in.

## Before / After

**Before:** `"command": "shipgate init --workspace /…/adk-samples --write --json"`

**After:**
```json
"command": "shipgate init --workspace /…/adk-samples/python/agents/crypto-payroll-agent --write --json",
"why": "The changed files share one project root; a repo-root manifest would cover 160+ unrelated agents."
```

## Acceptance criteria

- [ ] Preview on the repro suggests the `crypto-payroll-agent` directory.
- [ ] `init --workspace .` at the repo root with >1 candidate project reports ambiguity instead of writing a 252-source manifest with an arbitrary `agent.name`.
- [ ] The single-agent repo case (`samples/google_adk_agent`) is unchanged (golden test).
- [ ] Root `.gitignore` is not modified when the resolved manifest directory is a subdirectory.

## Scope / non-goals

Not a monorepo/multi-manifest feature — one manifest per run stays the model. This only changes which directory we *suggest*, and refuses to guess when the answer is ambiguous.

## Headline metric

**First-adoption success rate on multi-agent repositories** (currently 0% unguided — the routed command yields a manifest naming the wrong agent). No new command or schema field; changes a computed `next_action` value plus one new `init` refusal path.

pengfei-threemoonslab:
## Summary

- **`verify --preview` now routes first adoption to the project the diff changes, not the repository root.** Each changed path is attributed to the nearest directory carrying a project marker (`pyproject.toml`, `package.json`, `go.mod`, `Cargo.toml`, …); the scope is that project when exactly one is claimed. Paths that span two projects narrow to *nothing* rather than to their common parent — the parent of two projects is not itself a project.
- **`init --write` refuses a workspace whose agents live in more than one project** — `manifest_status: "refused_ambiguous_scope"` (exit 2) with `auto_detected.agent_scope` and `agent_project_candidates[]` — instead of adopting the first `Agent(name=…)` literal it parsed. A refused run writes **nothing**: no manifest, no CI workflow, no agent-instruction snippets, no reports `.gitignore` block.
- **Preview routes to `verify` when the changed project already has its own `shipgate.yaml`**, so following the suggestion once does not turn the next preview into a loop against a manifest that now exists.

Closes #363.

### Why this shape

The rule the issue proposes — *deepest common ancestor of the changed paths that contains a project marker* — does not work on the issue's own evidence. `google/adk-samples` PR #1917 also edits `python/agents/README.md`, one directory **above** the project it adds, so the common ancestor is `python/agents` (no marker) and the walk-up lands back at the repository root. That repo also has a root `pyproject.toml`, so the README is *claimed* by the root project.

What works is per-path attribution plus one more leg: when more than one project is claimed, the projects claimed **only** by documentation or test paths drop out. Which paths those are is decided by the trigger catalog's own `TRIGGER-DOCS-ONLY-NEGATIVE` rule (`triggers.paths_without_capability_surface`), so the scope resolver and the `trigger` verdict cannot drift — including the carve-outs that keep `SKILL.md`, `prompts/**`, and `policies/**` capability surfaces despite looking like documentation.

Two smaller decisions worth a reviewer's eye:

- **Candidate evidence is the framework `candidate_files`, not just name literals.** adk-samples' `crypto-payroll-agent` builds `LlmAgent(name=CONFIG.agent_name)`; a literal-only rule left the very project under review out of the candidate list the refusal offers.
- **Rank 1 of the refusal is a `review` action, not a command,** and the workspace root is never offered back as a command. Promoting one candidate would make the same arbitrary pick the refusal exists to prevent, and re-emitting `--workspace <root>` returns the caller to the command that was just refused.

Escape hatches: `--allow-ambiguous-scope` writes the single root manifest when one agent surface genuinely spans those directories; `--minimal` is never scope-gated (it adopts no detected name or tool surface); and `init --write --agent-instructions=…` on an already-adopted repository is untouched, because its scope was settled when its manifest was written.

## Type

- [ ] Input adapter change
- [ ] Documentation only

`DetectResult` gains `agent_scope` and `agent_project_candidates[]` (additive; both default to the single-project answer). No report-schema, contract, or check-ID change. No new command; `init` gains one flag and one refusal path.

## Verification

Additional local checks run:

- Full suite green (`-n 8`), `ruff` clean.
- New `tests/test_discovery_scope.py` (26 tests): scope attribution incl. the doc-above-the-project shape, a capability-bearing root file vetoing the scope, `prompts/**` not read as documentation, deleted directories climbing to a surviving parent, `..`/absolute path refusal, the `limit` boundary; preview routing to init and to verify; the init refusal writing nothing; `--allow-ambiguous-scope`; `--minimal`; instruction refresh on an adopted repo; and `bootstrap` stopping at init with the candidate list intact.
- **Verified end to end against `google/adk-samples` PR #1917 at its head** (full clone, `git fetch origin pull/1917/head`):

  | Acceptance criterion | Result |
  |---|---|
  | Preview suggests the `crypto-payroll-agent` directory | ✅ |
  | Root `init` reports ambiguity instead of a 252-source manifest | ✅ exit 2, 51 candidate projects, would have adopted `ask_rag_agent` |
  | `samples/google_adk_agent` unchanged | ✅ `agent_scope: single` |
  | Root `.gitignore` untouched when the manifest dir is a subdirectory | ✅ |

  Following the routed command produces a **40-line manifest with one `tool_source` in 3.6 s**, versus ~30 s and 252 sources at the root.
- Scope resolution costs 7 ms on the real 15-path diff and 53 ms on a synthetic 5 000-path diff (the docs-only rule is evaluated only when two projects actually compete).

Two test-only fixes were needed along the way and are worth knowing about: `tests/test_adapter_static_only.py` pins its allowlist **by line number**, so the new import in `triggers.py` shifted two entries; and the bundled adoption prompt has a **fourth** copy under `plugins/claude-code/` that `test_plugin_skill_is_byte_identical_to_canonical` caught.

## Release-readiness notes

- [x] New or changed check IDs are documented in `docs/checks.md` (none added)
- [x] Report/schema changes are additive or documented in `STABILITY.md` (additive `DetectResult` fields; no `report_schema_version` bump)

### Headline metric

First-adoption success rate on multi-agent repositories — currently 0% unguided, because the routed command yields a manifest naming the wrong agent. No new command and no schema version; this changes a computed `next_action` value plus one `init` refusal path.


src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -3700,12 +3718,24 @@ def run_preview(
         input_status=_trigger_input_status(diff_input),
     )
 
+    # The changed paths already say which project this pull request is
+    # about. Routing setup to the workspace root instead would hand a
+    # monorepo one manifest for every agent in it, so the adoption command
+    # below is scoped to the project the diff actually touches (#363).
+    scope = resolve_change_scope(
```
pengfei-threemoonslab: [P1] Plain preview never activates this scoping path

`changed_files` is only populated when the caller explicitly supplies `--base`, but every promoted adoption command is the plain `verify --preview --json` form. In that default flow this call receives an empty list, returns `None`, and recommends root `init`; after a scoped adoption, the next plain preview can loop back to root again. I reproduced this with a committed change under `services/a`. Please give preview the verifier's normal default-base discovery (while honoring `--no-base`) and add coverage for the exact promoted command, not only `--base/--head` tests.

src/agents_shipgate/cli/discovery/scope.py
```diff
@@ -0,0 +1,222 @@
...
+            # derived from a path this module cannot reason about could name
+            # a directory outside the repository, so abandon the answer
+            # rather than skip the path and answer from the rest.
+            return None
+        directory = root_resolved.joinpath(*path.parts[:-1])
+        if directory not in seen:
+            seen[directory] = find_project_root(directory, root=root_resolved)
+        found = seen[directory]
+        if found is None:
+            continue  # no project claims this path
```
pengfei-threemoonslab: [P1] Do not discard unclaimed capability-bearing paths

This silently removes any changed capability path whose directory has no project marker. For example, a root `prompts/system.md` plus `services/b/agent.py` narrows to `services/b`, even though both paths are capability-bearing; the emitted claim that every capability path belongs to that project is therefore false. Only genuinely ignorable docs/tests should be skipped. If an unclaimed capability path remains, fail closed instead of authorizing the narrowed workspace.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -579,6 +593,80 @@ def _agent_name_candidates(facts: list[_PyFacts], workspace: Path) -> list[NameC
...
+    while the candidate files are what it turns into ``tool_sources``. The
+    pull request in #363 is why the file set is included — its agent is
+    constructed as ``LlmAgent(name=CONFIG.agent_name)``, so a name-literal
+    rule would have left the very project under review out of the list of
+    directories the refusal offers.
+    """
+
+    names: dict[Path, set[str]] = {}
+    markers: dict[Path, str | None] = {}
+    paths: list[Path] = [
```
pengfei-threemoonslab: [P1] Include artifact-defined agent projects in ambiguity detection

The candidate seeds come only from `FrameworkDetection.candidate_files` and Python name literals. Artifact-only targets—such as OpenAPI, MCP, n8n, or Conductor projects—can have detections without either seed, so two independent package projects are reported as `agent_scope="single"` with no candidates. I reproduced two OpenAPI-only package projects; root `init` then wrote one manifest covering both. Seed candidates from all supported discovery artifacts (or otherwise carry their project roots) and add a non-Python monorepo test.

src/agents_shipgate/cli/_register_init.py
```diff
@@ -450,14 +577,21 @@ def init(
...
                 target.write_text(template, encoding="utf-8")
                 manifest_status = "written"
                 manifest_message = f"Wrote {target}"
 
         # Workflow action — independent of manifest action.
         workflow_outcome: dict[str, object] | None = None
-        workflow_requested = ci
+        workflow_requested = ci and not scope_refused
         if workflow_requested:
             result = write_ci_workflow(workspace_resolved)
```
pengfei-threemoonslab: [P1] Keep repository root separate from the selected project workspace

When scoped adoption passes a nested project as `workspace_resolved`, this writes `<project>/.github/workflows/agents-shipgate.yml`. GitHub only loads workflows from the repository-root `.github/workflows`, and the template's `config: shipgate.yaml` would also point at the wrong location. The command reports success although no gate runs. Write CI at the repo root with a repo-relative config path. The same root/project distinction should align report output and `.gitignore`, because current scoped init ignores a project-local reports directory while verify defaults to root reports.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -3767,13 +3816,32 @@ def run_preview(
             why="Shipgate is already set up here; run verify on the PR diff.",
         )
         headline = "Shipgate is configured; run verify on the PR to get a merge verdict."
+    elif scope is not None and scoped_manifest_present:
```
pengfei-threemoonslab, line 3984: [P1] Prefer the changed project's manifest over a root manifest

This branch is unreachable whenever `manifest_present` is true because the root-manifest branch above executes first. In a repository with both root and nested manifests, a change scoped to the nested project is therefore told to verify the root config—the wrong governance boundary. Check `scoped_manifest_present` first when a resolved scope exists, then fall back to the root manifest.

pengfei-threemoonslab (commented):
Engineer review: changes are required before merge. The implementation can still authorize root-scoped adoption, evaluate the wrong project, or write ineffective CI in supported flows. GitHub does not allow this account to formally request changes on its own pull request, so this is submitted as a comment review.

The inline comments cover the blocking correctness and public-contract issues. Additional follow-ups from the review:

- Nested `**/shipgate.yaml` changes are not routed by `TRIGGER-SHIPGATE-MANIFEST` or root opt-in detection.
- A diff spanning two projects can recommend root `init`, even though following that command deterministically refuses with `refused_ambiguous_scope`.
- Scope path normalization uses `.strip()`, which changes legal Git paths with leading or trailing spaces.
- Candidate recovery commands drop requested setup flags such as `--ci`, agent-instruction selections, Claude Code, and kit flags.
- The adoption prompt says to preserve the emitted scoped workspace, then hard-codes `--workspace .`.

I reproduced the plain-preview, unclaimed-capability-path, artifact-only-project, nested-vs-root-manifest, 1,000-file-cap, and ref/worktree-drift failures. GitHub CI is green, so these are missing cases rather than existing test failures. The current Shipgate control permits updating this PR for review, but denies merge.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -226,13 +228,23 @@ def detect_workspace(workspace: Path, *, max_python_files: int = 1000) -> Detect
     project_name_candidates = _project_name_candidates(workspace)
     suggested_sources, excluded_sources = _suggested_sources(workspace)
     codex_plugin_candidates = _codex_plugin_candidates(workspace)
+    agent_project_candidates = _agent_project_candidates(py_facts, detections, workspace)
+    agent_scope = "ambiguous" if len(agent_project_candidates) > 1 else "single"
```
pengfei-threemoonslab: [P1] Do not report a complete single-project answer after capped discovery

Python parsing stops at `max_python_files=1000`, but this classification has no truncated/unknown state. I reproduced projects `a` and `z` separated by 1,000 filler files; discovery saw only `a` and reported `single`. That makes the new refusal safety depend on filesystem ordering. Propagate scan truncation and refuse/fail closed when project-scope discovery is incomplete, or use an uncapped project-presence pass.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -3700,12 +3718,24 @@ def run_preview(
         input_status=_trigger_input_status(diff_input),
     )
 
+    # The changed paths already say which project this pull request is
+    # about. Routing setup to the workspace root instead would hand a
+    # monorepo one manifest for every agent in it, so the adoption command
+    # below is scoped to the project the diff actually touches (#363).
+    scope = resolve_change_scope(
+        root=root,
```
pengfei-threemoonslab: [P2] Bind project-marker evidence to the evaluated head

Changed paths come from `base...head`, but `resolve_change_scope` checks markers with live-worktree `Path.is_file()`. The same refs therefore authorize different commands depending on the current checkout or untracked files. I reproduced a feature ref adding `services/refund/package.json`: preview from `main` recommended root init, while checking out the feature recommended the nested project. Deleted projects can likewise lose their marker and redirect adoption. Read markers from the requested head tree, or fail closed unless the evaluated head is the current worktree.

src/agents_shipgate/schemas/detect.py
```diff
@@ -55,6 +75,14 @@ class DetectResult(BaseModel):
     frameworks: list[FrameworkDetection] = Field(default_factory=list)
     agent_name_candidates: list[NameCandidate] = Field(default_factory=list)
     project_name_candidates: list[NameCandidate] = Field(default_factory=list)
+    # Which directory one manifest should describe. "ambiguous" means agents
+    # were found in more than one self-contained project, so the workspace as
+    # a whole is not a manifest scope: `agent.name`, `declared_purpose`, and
+    # the declared tool surface would each describe several unrelated agents.
+    # `init --write` refuses on "ambiguous" rather than adopting the first
+    # agent name it parsed.
+    agent_scope: Literal["single", "ambiguous"] = "single"
```
pengfei-threemoonslab: [P2] Update the public zero-install detector with the canonical fields

These fields are added to the canonical `DetectResult`, but `tools/shipgate-detect.py` still emits neither field and can still recommend root init for an ambiguous monorepo. That breaks the documented structural-superset contract between the packaged and zero-install detectors. Please update the public detector, its canonical-key/parity tests and docs, and bump its `script_version`.

pengfei-threemoonslab:
All eight inline comments and all five follow-ups are addressed in 505acb15. Every reproduction you described now has a test in `tests/test_discovery_scope.py` (41 cases) or `tests/test_zero_install_detector.py`, and I re-verified the whole flow against `google/adk-samples` PR #1917 at its head.

Two of your findings turned out to be the same defect wearing different clothes, and they are the reason the rest were reachable at all: the scoping never ran in the promoted flow, and where it did run it accepted evidence that did not support the claim it emitted.

**[P1] Plain preview never activates this scoping** — fixed. `run_preview` takes `auto_base` and runs the verifier's own `detect_default_base_with_notes`, honoring `--no-base`. `test_the_promoted_preview_command_scopes_without_explicit_refs` covers the exact promoted command; `test_no_base_keeps_the_preview_diffless` covers the opt-out. On the real repro, bare `verify --preview --json` now reports `base_ref: origin/main`, 15 changed files, and routes to `python/agents/crypto-payroll-agent`.

**[P1] Do not discard unclaimed capability-bearing paths** — fixed. An unclaimed path that carries a capability surface now vetoes the suggestion instead of being skipped; only genuinely docs/test paths stay silent. Your `prompts/system.md` + `services/b/agent.py` case is `test_scope_declines_when_a_capability_path_belongs_to_no_project`.

**[P1] Include artifact-defined agent projects** — fixed. Candidates are seeded from everything `init` would turn into a manifest: framework `candidate_files` (artifact frameworks now record their own paths), `suggested_sources`, Codex plugin packages and marketplaces, and nested `shipgate.yaml` files. Your two-OpenAPI-project repro is `test_artifact_only_projects_are_ambiguous_without_any_python`; `samples/multi_agent_workspace` now reports `ambiguous` off its two nested manifests alone.

**[P1] Keep repository root separate from the selected project workspace** — fixed. `write_ci_workflow` takes a `repository_root` and writes there with `config:` naming the manifest relative to it. On the real repo: manifest in `python/agents/crypto-payroll-agent/`, workflow at `.github/workflows/agents-shipgate.yml`, `config: python/agents/crypto-payroll-agent/shipgate.yaml`. Root adoption is byte-unchanged (`test_root_scoped_ci_is_unchanged`).

On the reports/`.gitignore` half of that comment: I kept the ignore block in the manifest directory rather than moving it to the root, because acceptance criterion 4 on the issue is explicitly that a scoped adoption must not modify the tracked root `.gitignore`. Aligning the two properly means deciding whether `verify --workspace <project>` should resolve `--out` under the requested workspace instead of the git root — a change to where every scoped verify writes its artifacts, including the `current-control.json` pointer. That is a bigger call than this PR should make silently; say the word and I will open it as its own issue.

**[P1] Prefer the changed project's manifest over a root manifest** — fixed. The scoped branch is now checked before `manifest_present`, guarded on a resolved scope. `test_preview_prefers_the_changed_project_manifest_over_the_root_one`. Writing that test also surfaced something worth keeping: a diff that touches the root manifest *and* a nested project is genuinely contested and correctly falls back.

**[P1] Do not report a complete single-project answer after capped discovery** — fixed. `agent_scope` gains `"unknown"`, and `init --write` refuses on it. The trigger is truncation **and** more than one project root, counted from the uncapped walk inventory — so your `a`/filler/`z` repro refuses, while a 1,200-file single-project repository keeps `"single"` and a working `init`. Both are pinned (`test_capped_discovery_refuses_instead_of_claiming_one_project`, `test_a_large_single_project_repo_keeps_its_single_verdict`).

**[P2] Bind project-marker evidence to the evaluated head** — fixed, by the fail-closed option. Markers are read from the working tree because that is what `init` runs against; when the evaluated head is not the current checkout, no scope is claimed at all. Your feature-ref reproduction is `test_preview_claims_no_scope_for_a_head_that_is_not_checked_out`.

**[P2] Update the public zero-install detector** — fixed. `tools/shipgate-detect.py` is `script_version` `0.3.0`, emits both fields, and refuses to recommend root init on an unresolved scope. The parity test now pins `agent_scope` and `agent_project_candidates[]` against the CLI on every sample, and `CANONICAL_KEYS` includes them. `docs/zero-install.md` updated.

Follow-ups:

- **Nested `**/shipgate.yaml` routing** — `TRIGGER-SHIPGATE-MANIFEST` now matches at any depth. This closed a gap the suite already tracked by name: `_TRUST_ROOT_TRIGGER_GAPS` listed `services/foo/shipgate.yaml`, and the parity test told me to promote it out. The pre-commit `files:` regex, the Cursor renderer globs (plus the benchmark template and harness constant that mirror them), and the documented copy-paste snippets all follow. Root opt-in detection: preview now passes `manifest_present or scoped_manifest_present` to the evaluator, so an adopted nested project force-runs.
- **A spanning diff recommending an init that refuses** — preview now routes to `detect` (`kind: "discover"`) and names the competing projects in `why`. `resolve_change_scope` returns a `ScopeResolution` carrying `contested[]` for exactly this.
- **`.strip()` on paths** — removed; paths are used verbatim, with a test for a directory name ending in a space.
- **Recovery commands dropping setup flags** — `_requested_setup_flags` repeats `--ci`, `--claude-code`, `--agent-instructions=…`, and `--agent-instructions-kit`, mirroring `_rerun_options` in the verifier.
- **The adoption prompt contradicting itself** — step 4 now uses the workspace preview named (`WS`) instead of hard-coding `--workspace .`, in all four copies, with render hashes bumped.

One rename: the refusal is `refused_unresolved_scope` with `--allow-unresolved-scope`, since it now covers a scope that could not be *established* as well as one that is ambiguous. Nothing has shipped under the old name.

`ruff` clean, full suite green.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -3692,10 +3763,36 @@ def run_preview(
             diff_input = collected
             notes.append(f"Preview diff unavailable: {collected.note}")
 
+    # The changed paths already say which project this pull request is
```
pengfei-threemoonslab, line 3845: [P1] Include the working-tree overlay in preview scoping

Bare preview now auto-detects a committed base, but this path still derives `changed_files` only from `base...head`. Full verification unions its working-tree context; preview does not. I reproduced an uncommitted-only capability change yielding `changed=[]` and authorizing root init, and committed project A plus uncommitted project B yielding only A and authorizing init there. Because the emitted action runs against the live worktree, preview must scope and trigger from the same effective committed-plus-overlay path set as full verify.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -3692,10 +3763,36 @@ def run_preview(
...
+    # the two disagree — the same refs would recommend different directories
+    # depending on what happens to be checked out — so no scope is claimed.
+    resolution = (
+        resolve_change_scope(
+            root=root,
+            changed_files=changed_files,
+            limit=requested_root,
+        )
+        if _head_is_current_worktree(root, head)
+        else ScopeResolution()
```
pengfei-threemoonslab: [P1] Do not turn unresolved scope into root initialization

This deliberately creates an empty resolution when the evaluated head is not checked out; deleting a project's last marker produces the same empty state. The generic branches below then treat that as permission to initialize the repository root. I reproduced a feature-ref preview and a whole-project deletion where following the emitted root command wrote a manifest for an unrelated agent in the current checkout. Preserve an explicit unresolved reason and route to checkout/discovery/review, or resolve markers from the evaluated tree; never convert "scope not established" into a mutation of the root.

prompts/add-shipgate-to-repo.md
```diff
@@ -55,17 +55,42 @@ agent-related PRs should use `agents-shipgate verify` after this adoption step.
    would classify `is_agent_project: false`; look for `suggested_sources` and
    `codex_plugin_candidates` when those fields are present.
 
-4. **Generate a starter manifest + GitHub Actions workflow:**
+   Run the emitted command **verbatim, including its `--workspace`**, and keep
```
pengfei-threemoonslab: [P1] Execute the scoped init command only once

This tells the agent to run preview's emitted scoped `init --write --json` verbatim, then step 4 immediately runs `init --write --ci --json` again. In the scoped repro, the first command writes the manifest; the second writes the workflow but exits 2 with `manifest_status=skipped_existing` / `config_already_exists`. A control-following agent can therefore fail adoption after performing the requested setup. Either make preview emit the complete canonical `--ci` adoption command and delete the second init, or derive `WS` without executing the first action and run init exactly once.

prompts/add-shipgate-to-repo.md
```diff
@@ -55,17 +55,42 @@ agent-related PRs should use `agents-shipgate verify` after this adoption step.
...
+   preview routes you to `verify` there rather than to setup. If the change
+   spans several projects, preview routes you to `detect` instead — read the
+   project list and initialize the one your change belongs to.
+
+4. **Generate a starter manifest + GitHub Actions workflow.** Use the
+   workspace preview named — `WS` below is that directory, which is `.` only
+   when preview said so:
    ```bash
-   $SG init --workspace . --write --ci --json
+   $SG init --workspace "$WS" --write --ci --json
```
pengfei-threemoonslab, line 86: [P1] Keep the rest of adoption in the scoped workspace

After this scoped init, step 6 still runs `scan -c shipgate.yaml`, step 7 reads root `agents-shipgate-reports`, and init's JSON `next_action` is also hard-coded to `shipgate.yaml`. From the repo root with the manifest in `apps/agent/shipgate.yaml`, the documented scan and emitted next action both exit 2 because the config is not at the root. `cd "$WS"` for the remaining steps or qualify the manifest and report paths everywhere, and build init's next action from the actual target path.

src/agents_shipgate/cli/discovery/ci_workflow.py
```diff
@@ -140,7 +162,7 @@ def write_ci_workflow(workspace: Path) -> CiWorkflowResult:
     workflows_dir.mkdir(parents=True, exist_ok=True)
     # Re-render at write time so the ref reflects the current package
     # version (or the AGENTS_SHIPGATE_WORKFLOW_REF override).
-    target.write_text(_render_workflow_template(), encoding="utf-8")
+    target.write_text(_render_workflow_template(config_value), encoding="utf-8")
```
pengfei-threemoonslab, line 218: [P1] Gate every scoped project, not only the first one

This root target contains one scalar `config`. Initializing project A with `--ci` writes it for `a/shipgate.yaml`; initializing B returns `skipped_existing_target` and leaves the workflow pointing only at A, although both manifests exist. That silently leaves B ungated in the newly supported one-manifest-per-project layout. Use uniquely named per-project workflows or a managed multi-config/matrix workflow; if neither is supported, the existing-target result must be blocking and explain that B has no CI gate.

src/agents_shipgate/cli/_register_init.py
```diff
@@ -498,15 +707,17 @@ def init(
         # Shipgate before this CLI version was released get the line on their
         # next `init --write`.
         gitignore_outcome = (
-            ensure_reports_gitignore(workspace_resolved, write=write) if write else None
+            ensure_reports_gitignore(workspace_resolved, write=write)
```
pengfei-threemoonslab, line 778: [P1] Ignore the report directory the verifier actually writes

This still installs the managed ignore inside the scoped project, while preview/verify normalize their default output to the Git repository root. In the recommended flow, `apps/a/.gitignore` contains `agents-shipgate-reports/`, but `<repo>/agents-shipgate-reports/` remains unignored and appears as untracked. Either install the managed rule at the repository root or align the verifier's default output with the requested scoped workspace.

src/agents_shipgate/cli/discovery/scope.py
```diff
@@ -0,0 +1,293 @@
...
+from dataclasses import dataclass
+from pathlib import Path, PurePosixPath
+
+from agents_shipgate.triggers import paths_without_capability_surface
+
+# Files that mark a self-contained project root, in precedence order. A
+# workspace that already carries ``shipgate.yaml`` is the strongest marker
+# there is: it is a project someone has already scoped by hand.
+#
+# Deliberately excluded: ``requirements.txt``, ``setup.cfg``, ``Makefile``,
```
pengfei-threemoonslab, line 38: [P1] Do not collapse requirements-only sibling agents into one root scope

Excluding `requirements.txt` leaves a common Python layout without any project boundary. Two sibling Google ADK agents, each containing only `requirements.txt` and `agent.py`, are reported as `agent_scope="single"` with candidate `.`; root init then succeeds and chooses one agent name. Add an evidence-backed requirements marker rule, or treat multiple markerless framework clusters as unresolved instead of authorizing one root manifest.

tools/shipgate-detect.py
```diff
@@ -154,6 +175,52 @@ def _walk_files(workspace: Path, max_files: int = 5000) -> list[Path]:
     return sorted(out)
 
 
+def _project_marker(directory: Path) -> str | None:
```
pengfei-threemoonslab: [P1] Do not compute public scope from a truncated structural inventory

The new candidate census is still fed by `_walk_files`, which stops after 5,000 files rather than capping only Python parsing. With 5,000 root filler files before two nested agent projects, this detector reports `is_agent_project=false`, `agent_scope="single"`, and no candidates, while canonical detection finds both projects and returns `ambiguous`. Enumerate the complete structural inventory and cap only AST work, or publish an incomplete/unknown fail-closed state.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -3760,20 +3876,60 @@ def run_preview(
...
+            "run verify there to get a merge verdict."
+        )
     elif manifest_present:
         next_action = CodingAgentCommandAction(
             kind="verify",
             command=verify_command,
             why="Shipgate is already set up here; run verify on the PR diff.",
         )
         headline = "Shipgate is configured; run verify on the PR to get a merge verdict."
+    elif resolution.contested:
```
pengfei-threemoonslab, line 4052: [P1] Resolve contested configured projects before the root-manifest branch

Because `manifest_present` is checked immediately before this branch, a change spanning configured projects A and B is verified against the root manifest whenever one exists, bypassing both actual project boundaries. Without a root manifest, this path asks discovery to choose/initialize one project even when A and B already have manifests. Route to both project verifications or explicit human review; a root manifest must not silently substitute for multiple contested scoped manifests.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -579,6 +651,147 @@ def _agent_name_candidates(facts: list[_PyFacts], workspace: Path) -> list[NameC
...
+
+    names: dict[Path, set[str]] = {}
+    markers: dict[Path, str | None] = {}
+    paths: list[Path] = [
+        workspace / relative
+        for detection in detections
+        for relative in detection.candidate_files
+    ]
+    paths.extend(workspace / relative for relative in artifact_paths)
+    paths.extend(fact.path for fact in facts if fact.agent_name_literals)
```
pengfei-threemoonslab: [P2] Only use framework-attributed name literals as project evidence

This seeds a project from every syntactic `Agent(name=...)` fact, even when the file has no supported framework import. A real Google ADK project plus an unrelated CRM project defining its own `class Agent` is therefore reported as ambiguous and `init --write` refuses. Name literals should create scope evidence only when attributed to a supported framework/import; generic constructors can still be retained as low-confidence name suggestions without defining a project boundary.

src/agents_shipgate/cli/discovery/scope.py
```diff
@@ -0,0 +1,293 @@
...
+            return None
+        current = parent
+
+
+def project_marker(directory: Path) -> str | None:
+    """Name the project marker in ``directory``, if it carries one."""
+
+    for name in PROJECT_MARKERS:
+        try:
+            if (directory / name).is_file():
```
pengfei-threemoonslab: [P2] Reject symlinked markers and manifests before emitting a command

`Path.is_file()` follows symlinks. A nested `shipgate.yaml` symlink to an external file is therefore accepted as a marker/manifest and preview emits a scoped verify command, but that exact command exits 2 because config validation forbids symlink components. Use non-following identity/containment validation here so preview never guarantees a command the verifier will reject.

src/agents_shipgate/cli/_register_init.py
```diff
@@ -140,6 +143,167 @@ def _detect_json_indent(text: str) -> int:
...
+
+    flags: list[str] = []
+    if ci:
+        flags.append("--ci")
+    if claude_code:
+        flags.append("--claude-code")
+    if agent_instructions is not None:
+        flags.append(f"--agent-instructions={agent_instructions}")
+    if agent_instructions_kit is not None:
+        flags.extend(["--agent-instructions-kit", str(agent_instructions_kit)])
```
pengfei-threemoonslab: [P2] Rebase or withhold a relative adoption-kit path when the workspace changes

Candidate recovery copies `--agent-instructions-kit` verbatim while changing `--workspace`. A root-relative kit such as `.agents-shipgate/custom-kit.yaml` is then looked up beneath `apps/a` and the emitted command exits 2; simply making it absolute may also violate the candidate-workspace containment rule. Emit an executable candidate command only when the kit can be safely rebased/contained there, otherwise emit a review/relocate action.

pengfei-threemoonslab (commented):
Request changes — re-review of `505acb1`.

Many original defects are fixed, and the current CI plus targeted scope/public-contract suites are green. The remaining adversarial paths below still let the promoted adoption flow authorize the wrong scope, fail after following its own command, or silently leave a project ungated, so I don't think this is ready to merge.

The inline comments contain the concrete reproductions and proposed directions. One additional P1 has no stable changed-line anchor: the first-class `trigger` command still computes opt-in with only `<workspace>/shipgate.yaml`. With `apps/a/shipgate.yaml` present and `apps/a/README.md` changed, it returns `force_run=false` and a docs-only skip. Please reuse the scoped/nested manifest resolution there so the documented nested-manifest opt-in contract applies outside preview too.

Shipgate's fresh verifier result for this head is `review_publishable / review_required`; merge and completion remain human-gated.

Note: this is submitted as a comment review because GitHub does not permit an author account to use the `REQUEST_CHANGES` event on its own PR.

src/agents_shipgate/cli/discovery/ci_workflow.py
```diff
@@ -69,7 +69,7 @@ def _action_ref() -> str:
       - name: Run Agents Shipgate
         uses: ThreeMoonsLab/agents-shipgate@{ref}
         with:
-          config: shipgate.yaml
+          config: {config}
```
pengfei-threemoonslab, line 74: [P2] Serialize the config path as a YAML-safe scalar

The repository-relative path is interpolated unquoted. A valid project directory such as `apps/agent #1` renders `config: apps/agent #1/shipgate.yaml`; YAML parses the value as `apps/agent`, so the action receives a nonexistent manifest. Quote or serialize the scalar with a YAML-safe emitter and cover `#`, `: `, and quote characters.

pengfei-threemoonslab:
All thirteen comments plus the trigger P1 from the review body are addressed in 57508531. Re-verified against `google/adk-samples` PR #1917 at its head; full suite and ruff green.

Most of these were one mistake in different places, and naming it changed the shape of the fix: **"Shipgate could not tell which project this is" was being spent as though it were "the workspace you named is the one."** `ScopeResolution` now carries a status — `resolved`, `not_narrowed`, `contested`, `unclaimed`, `not_evaluated` — and only `not_narrowed` authorizes initializing the caller's workspace.

**[P1] Include the working-tree overlay in preview scoping** — fixed. Preview unions the working tree exactly as full verify does (one merge-base comparison, untracked paths included), and skips it only for an explicit `--head`, the one case where the worktree is not what is evaluated. Both your repros are pinned: `test_preview_sees_uncommitted_work` and `test_preview_sees_untracked_work`. On the real repo, an uncommitted-only edit to `crypto_payroll_agent/agent.py` now routes to that project.

**[P1] Do not turn unresolved scope into root initialization** — fixed. `not_evaluated` (head not checked out, unreadable path, symlinked directory) and `unclaimed` route to `discover` with `resolution.detail` quoted in `why`. The head-not-checked-out test now asserts the opposite of what it asserted before — it was encoding the bug.

**[P1] Resolve contested configured projects before the root-manifest branch** — fixed, with one deviation worth flagging. `contested_configured` is checked ahead of `manifest_present`, and routes to human review. I could not put both verify commands in `allowed_next_commands`: `derive_agent_control` rejects it (`a stopping result cannot expose allowed next commands`, and `review_publishable` permits at most one). Rather than weaken the control contract or pick one project arbitrarily, both commands are named in the `why` a human reads and `allowed_next_commands` stays empty. If you would rather this be `review_publishable` with a single command plus prose for the rest, say so and I will flip it.

**[P1] Gate every scoped project** — fixed. One workflow per gated manifest: `agents-shipgate.yml` for a root manifest, `agents-shipgate-<project>-<digest>.yml` for a scoped one (the digest keeps `apps/a` and `apps.a` from colliding). The double-wiring guard now matches on the manifest *and* the action, so it still refuses to gate one project twice while letting the second project be gated at all.

**[P1] Ignore the report directory the verifier actually writes** — fixed by aligning the output, as you suggested. The default `--out` resolves under the requested workspace, so `verify --workspace apps/a` writes `apps/a/agents-shipgate-reports/` — what that project's managed ignore covers — and two projects stop overwriting each other's control pointer. An explicit `--out` still resolves against the repository root, so every invocation that names a directory is unchanged. Two pre-existing nested-workspace tests in `test_verify.py` asserted the old location and were updated.

One residue I did not chase: `verify --preview` pointed at the repository root still writes `<repo>/agents-shipgate-reports/`, because that *is* its workspace. That predates this PR and is identical for any unadopted repo. Scoping preview's own artifacts to a scope it has not resolved yet would mean moving the control pointer mid-run; happy to take it as a follow-up if you want it closed.

**[P1] Do not collapse requirements-only sibling agents** — fixed with the evidence-backed rule. `WEAK_PROJECT_MARKERS` (`requirements.txt`, `requirements.in`) count as a boundary only in a directory where agent evidence already sits, so your two sibling ADK agents are two projects while `tests/requirements.txt` stays what it is. Both directions are pinned.

**[P1] Do not compute public scope from a truncated structural inventory** — fixed. `_walk_files` never drops marker files; the cap applies to everything else. 5,200 filler files ahead of two nested projects: script and CLI both report `ambiguous` with the same candidates.

**[P1] Execute the scoped init command only once** / **[P1] Keep the rest of adoption in the scoped workspace** — fixed. Step 3 now says to read `control.next_action` and take `--workspace` from it *without running it*, and enumerates the three kinds that mean "do not continue to step 4" (`verify`, `discover`, human review). Step 4 is the only `init`. Step 6 runs in `$WS`. And init's own `next_action` now names the manifest it wrote, so `scan` resolves from anywhere — that was the third instance of the same bug.

**[P1, review body] Nested-manifest opt-in in `trigger`** — fixed. `manifest_opt_in` is one resolver used by both `trigger` and preview; it counts only manifests *above the changed paths*, so `apps/a/README.md` force-runs while `docs/guide.md` still skips.

The four P2s: name literals draw a boundary only when their file carries a supported framework import (unrelated `class Agent` no longer refuses a single-agent repo, and stays a name suggestion); symlinked markers and manifests are rejected before preview promises a command config validation would refuse; a relative `--agent-instructions-kit` is rebased into the candidate workspace or that candidate is emitted as a review action instead of a command that exits 2; and the workflow `config:` is YAML-quoted when the path needs it — `apps/agent #1`, `a: b`, and quotes all round-trip through `yaml.safe_load` now.

src/agents_shipgate/cli/discovery/scope.py
```diff
@@ -0,0 +1,458 @@
...
+            seen[directory] = find_project_root(
+                directory, root=root_resolved, evidence_dirs=evidence_dirs
+            )
+        found = seen[directory]
+        if found is None:
+            unclaimed.append(text)
+            continue
+        claims.setdefault(found.directory, (found.marker, []))[1].append(text)
+
+    if len(claims) > 1 or (unclaimed and claims):
```
pengfei-threemoonslab, line 351: [P1] Treat a fully deleted project as unresolved

When a change deletes a project's final marker, all of its deleted paths land in `unclaimed` and `claims` is empty. This condition never evaluates those paths, and the later `if not claims` branch returns `not_narrowed`. In a two-project repro, deleting all of `apps/a` made preview emit root `init`; following it exited 0, wrote a root manifest, and selected untouched `agent_b`. An all-unclaimed capability change needs to fail closed (or recover the boundary from the base tree), not authorize root mutation.

src/agents_shipgate/cli/discovery/scope.py
```diff
@@ -0,0 +1,458 @@
...
+    ``tests/`` and ``docs/`` — but two sibling agents whose only marker is
+    ``requirements.txt`` beside ``agent.py`` are two projects, and reading
+    them as one root scope lets ``init`` pick one of their names (#363
+    review). Callers that cannot establish evidence pass nothing and get
+    the strong markers alone.
+    """
+
+    current = directory
+    while True:
+        extra = WEAK_PROJECT_MARKERS if current in evidence_dirs else ()
```
pengfei-threemoonslab, line 212: [P1] Let an evidence-backed weak marker qualify as an ancestor

The weak marker is enabled only when agent evidence is in the exact same directory. Two ordinary projects shaped as `a/requirements.txt + a/src/agent.py` and `b/requirements.txt + b/src/agent.py` therefore collapse into one root candidate, and root `init` is authorized. A weak marker ancestor of supported agent evidence should establish the boundary; the existing `tests/requirements.txt` negative control remains safe because it is not an ancestor of the root agent evidence.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -3692,10 +3806,98 @@ def run_preview(
...
+            # the base branch would report its own commits as this change.
+            comparison_ref = (
+                require_merge_base_sha(root, base, "HEAD") if base else "HEAD"
+            )
+            worktree_paths, worktree_diff = working_tree_context(
+                root,
+                comparison_ref=comparison_ref,
+                exclude=out_dir,
+            )
+        except Exception:  # noqa: BLE001 - preview must never crash.
```
pengfei-threemoonslab, line 3829: [P1] Do not authorize mutation from a failed worktree overlay

Any collection failure is converted into an empty overlay plus a note, after which preview scopes from committed evidence. Repro: committed capability change in A, uncommitted change in B, and an executable Git filter; preview admits it could not read the worktree but reports only A and authorizes `init` there. This call also omits full verify's `reject_index_hidden=True`, so a `skip-worktree` change in B disappears without a warning. Carry the typed failure into `diff_input`/`not_evaluated` and never emit a mutation command from partial overlay evidence.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -3692,10 +3806,98 @@ def run_preview(
...
 
+    # The action preview emits runs against the *worktree*, so preview has to
+    # see the same effective change set the full verifier would: committed
+    # range unioned with uncommitted and untracked work. Without this an
+    # uncommitted-only capability change reads as an empty diff, and the empty
+    # diff reads as "nothing here narrows the scope" (#363 review). Skipped
+    # when an explicit --head names another tree, which is the one case where
+    # the worktree is not what is being evaluated — verify archives the head
+    # there for the same reason.
+    if head is None:
```
pengfei-threemoonslab, line 3817: [P1] Bind explicit `--head HEAD` routing to the evaluated tree

Every explicit head skips overlay collection, while `_head_is_current_worktree()` checks only the commit SHA and scope/manifests are still read from the dirty worktree. Repro: add uncommitted `apps/a/shipgate.yaml`, then preview with `--base origin/main --head HEAD`; preview emits scoped verify, but the exact command exits 2 because `HEAD` does not contain that manifest. Skip the overlay only for a genuinely non-current head, and validate marker/config evidence in the same tree the emitted command evaluates.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -3651,7 +3750,9 @@ def run_preview(
         config,
         requested_workspace=workspace,
     )
-    out_dir = _resolve_under_workspace(root, out or DEFAULT_OUT_DIR)
+    out_dir = _resolve_out_dir(
```
pengfei-threemoonslab, line 3753: [P1] Keep promoted preview artifacts inside the scope that will ignore them

A root-invoked preview writes `<repo>/agents-shipgate-reports` before discovering that setup belongs in `apps/a`. Following the emitted scoped init writes only `apps/a/.gitignore`, leaving the root preview artifacts untracked; my repro left four `?? agents-shipgate-reports/*` files. Either defer/move preview output after scope resolution or ensure the repository-level artifact directory is ignored by the recommended flow.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -579,6 +675,168 @@ def _agent_name_candidates(facts: list[_PyFacts], workspace: Path) -> list[NameC
     return candidates
 
 
+def _agent_scope(
```
pengfei-threemoonslab, line 678: [P1] Scope filtering does not protect the generated agent identity

The new framework-attributed filtering applies only to project candidates; `_agent_name_candidates()` still labels every syntactic `Agent(name=...)` identically, and init auto-selects the first `Agent_name_literal`. With an unrelated `a_crm/models.py` defining `crm_rep` and a real `z_adk/agent.py` defining `real_agent`, detect reports the sole project as `z_adk`, but `init --write` succeeds with `agent.name: crm_rep`. Give non-framework literals a non-auto-selectable source/confidence or exclude them from automatic identity selection.

tools/shipgate-detect.py
```diff
@@ -623,21 +735,74 @@ def detect(workspace: Path) -> dict[str, Any]:
...
+        dict.fromkeys(
+            [rel for d in detections for rel in d["candidate_files"]]
+            + [s["path"] for s in suggested]
+            + [c["path"] for c in codex_plugin_candidates]
+            + [
+                _rel(p, workspace)
+                for p in files
+                if p.name == "shipgate.yaml" and p.parent != workspace
+            ]
+            + list(literals_by_path)
```
pengfei-threemoonslab, line 759: [P1] Keep zero-install literal attribution aligned with canonical detection

This adds every generic `Agent` literal back as project evidence even though canonical detection now excludes non-framework literals from scope. On the CRM + real ADK fixture, canonical reports `single` with only `z_adk`; this script reports `ambiguous` with `a_crm` and `z_adk`. Filter literals by supported-framework attribution here too and add an adversarial parity fixture.

tools/shipgate-detect.py
```diff
@@ -141,17 +169,96 @@
...
         for fn in files:
-            out.append(Path(root) / fn)
-            if len(out) >= max_files:
-                return sorted(out)
-    return sorted(out)
+            path = Path(root) / fn
+            if fn in PROJECT_MARKERS or fn in WEAK_PROJECT_MARKERS:
+                markers.append(path)
+                continue
+            if len(out) < max_files:
```
pengfei-threemoonslab, line 194: [P1] The 5,000-file cap still permits a false terminal stop

Retaining marker files does not retain the agent evidence needed to use them. With 5,000 root text files followed by `a,z/{pyproject.toml,agent.py}`, canonical detection returns agent/ambiguous with candidates A and Z; zero-install returns non-agent/single with no candidates and `python_file_count=0`. Because `all_py` is derived from this already-truncated list, `py_truncated` is also false. Track general-inventory truncation and fail closed, or collect structural and Python evidence independently.

tools/shipgate-detect.py
```diff
@@ -437,7 +544,11 @@ def _local_marketplace_roots(workspace: Path, paths: list[Path]) -> set[Path]:
 def detect(workspace: Path) -> dict[str, Any]:
     workspace = workspace.resolve()
     files = _walk_files(workspace)
-    py_files = [p for p in files if p.suffix == ".py"][:1000]
+    all_py = [p for p in files if p.suffix == ".py"]
```
pengfei-threemoonslab, line 547: [P1] AST truncation must not authorize `No action`

Even when this cap is detected, a single root project is classified `single`, and if the real framework file sorts after 1,000 inert Python files both canonical and zero-install detection return `is_agent_project=false` with terminal `No action`. A complete scope does not mean framework detection was complete. Publish an incomplete/unknown coverage state and route to a higher cap or review instead of allowing the documented stop condition.

tools/shipgate-detect.py
```diff
@@ -141,17 +169,96 @@
...
         dirs[:] = [
             d for d in dirs
             if d not in SKIP_DIRS and not d.startswith(".venv")
         ]
         for fn in files:
-            out.append(Path(root) / fn)
-            if len(out) >= max_files:
-                return sorted(out)
-    return sorted(out)
+            path = Path(root) / fn
```
pengfei-threemoonslab, line 190: [P1] Reject evidence paths that escape through file symlinks

The walk retains file symlinks and `_rel()` falls back to an absolute path when the resolved target leaves the workspace. Repro: `linked-openapi.json -> ../secret-openapi.json`; canonical detection ignores it, while zero-install parses the external file and exposes its absolute path under `suggested_sources`. Reject escaping resolved targets before matching or probing.

src/agents_shipgate/cli/discovery/ci_workflow.py
```diff
@@ -168,6 +254,10 @@ def _detect_cross_reference(workflows_dir: Path, *, exclude: Path) -> Path | Non
             text = path.read_text(encoding="utf-8")
         except (OSError, UnicodeDecodeError):
             continue
-        if _USES_PATTERN.search(text):
+        if not _USES_PATTERN.search(text):
+            continue
+        # The historical behavior for a root manifest: any Shipgate workflow
+        # counts, because `config:` defaults to `shipgate.yaml` when omitted.
+        if config_value == DEFAULT_MANIFEST_PATH or config_pattern.search(text):
```
pengfei-threemoonslab, line 261: [P1] A scoped workflow does not gate the root manifest

For `config_value == shipgate.yaml`, this treats any Shipgate workflow as a root cross-reference, including one that explicitly says `config: apps/a/shipgate.yaml`. Creating scoped CI first and root CI second returns `written` then `skipped_cross_reference`, both exit 0, and only the nested manifest is gated. Treat an omitted `config` as the root default, but an explicit different config as a different boundary.

src/agents_shipgate/cli/discovery/ci_workflow.py
```diff
@@ -104,25 +127,77 @@ class CiWorkflowResult:
...
+
+def _workflow_slug(relative_dir: str) -> str:
+    """A file-name-safe, collision-free slug for a project directory."""
+
+    slug = re.sub(r"[^A-Za-z0-9]+", "-", relative_dir).strip("-").lower()
+    # Two directories can normalize to the same slug (``apps/a`` and
+    # ``apps.a``), and two workflows with the same name is the very failure
+    # this function exists to prevent. A short digest of the exact path keeps
+    # them apart without making the common name unreadable.
+    digest = hashlib.sha256(relative_dir.encode("utf-8")).hexdigest()[:8]
```
pengfei-threemoonslab, line 155: [P1] Bound the workflow name and use a stronger collision identity

Eight hex characters are not collision-free: the valid sibling paths `a-b-c+d}e&f}g` and `a-b-c d$e~f%g` produce the same slug and digest, so the second init exits 0 with `skipped_existing_target` and remains ungated. A sufficiently deep valid path also makes the generated filename exceed 255 bytes and raises an uncaught `OSError` after the manifest was written. Bound the readable prefix and use a substantially stronger identity.

prompts/add-shipgate-to-repo.md
```diff
@@ -55,29 +55,68 @@ agent-related PRs should use `agents-shipgate verify` after this adoption step.
    would classify `is_agent_project: false`; look for `suggested_sources` and
    `codex_plugin_candidates` when those fields are present.
 
-4. **Generate a starter manifest + GitHub Actions workflow:**
+   **Read `control.next_action` — do not run it yet.** Step 4 performs the
```
pengfei-threemoonslab, line 58: [P1] Execute the exact command authorized by preview

Preview's `control.next_action` and `allowed_next_commands` authorize only `init --workspace <scope> --write --json`. The prompt explicitly says not to run that command, then runs a different mutation with `--ci`, which violates the published exact-command control boundary. Have preview authorize the complete one-shot adoption command, including CI, and make the prompt execute it verbatim once.

prompts/add-shipgate-to-repo.md
```diff
@@ -55,29 +55,68 @@ agent-related PRs should use `agents-shipgate verify` after this adoption step.
...
 
-6. **Run the scan with patch suggestions:**
+6. **Run the scan with patch suggestions** — in the workspace you initialized:
    ```bash
-   $SG scan -c shipgate.yaml --suggest-patches --format json --ci-mode advisory
+   cd "$WS" && $SG scan -c shipgate.yaml --suggest-patches --format json --ci-mode advisory
    ```
-   The report lands at `agents-shipgate-reports/report.json`. The supporting Release Evidence Packet lands at `agents-shipgate-reports/packet.{md,json,html}`. Parse `report.json`; Codex plugin facts, when present, live under `codex_plugin_surface`.
+   The report lands at `$WS/agents-shipgate-reports/report.json`, beside the manifest it describes; the supporting Release Evidence Packet lands at `$WS/agents-shipgate-reports/packet.{md,json,html}`. (Reports follow the workspace, so two projects in one repository never overwrite each other's results.) Every path below is relative to `$WS`. init's own `next_action` already names the manifest correctly for wherever you are — prefer running that command verbatim over retyping this one. Parse `report.json`; Codex plugin facts, when present, live under `codex_plugin_surface`.
```
pengfei-threemoonslab, line 119: [P1] Keep `apply-patches` in the scoped workspace too

The scoped scan writes `$WS/agents-shipgate-reports/report.json`, but step 7 still invokes `apply-patches --from agents-shipgate-reports/report.json`. Coding-agent shell calls are separate, so the `cd` in step 6 does not persist; executing the documented steps from the repository root fails with `Cannot read JSON report`. Use `$WS/agents-shipgate-reports/report.json` or repeat `cd "$WS" &&`, and qualify the later packet/checklist paths the same way.

src/agents_shipgate/cli/trigger.py
```diff
@@ -163,7 +164,7 @@ def trigger(
             raise typer.Exit(2) from exc
 
     if manifest_present is None:
-        manifest_present_resolved = (workspace / "shipgate.yaml").is_file()
+        manifest_present_resolved = manifest_opt_in(workspace, changed_paths=paths)
```
pengfei-threemoonslab, line 167: [P1] Ref-based trigger must read manifest opt-in from the evaluated head

With `--base/--head`, changed paths come from the requested refs but this resolver stats the live checkout. If a feature head contains `apps/a/shipgate.yaml` and a docs-only change while another commit is checked out, trigger reports a docs-only skip instead of the nested manifest's force-run. Resolve manifest presence from the evaluated head tree, or fail closed unless that head is checked out.

pengfei-threemoonslab (commented):
Request changes — updated-head re-review of `5750853`.

The update fixes many of the earlier issues, and the GitHub checks plus focused discovery, verifier, CI, trigger, bootstrap, prompt-parity, and public-contract suites are green. The remaining inline findings below are independently reproduced coverage gaps: they can still authorize the wrong manifest scope or identity, derive commands from incomplete evidence, emit a command that immediately fails, or silently leave a manifest ungated.

Shipgate's fresh control result for this exact head is `review_publishable / review_required`; merge and completion remain human-gated.

This is submitted as a comment review because GitHub does not permit the author account to use the `REQUEST_CHANGES` event on its own PR.

src/agents_shipgate/cli/_register_init.py
```diff
@@ -521,6 +800,31 @@ def init(
...
+            scope_actions = _unresolved_scope_actions(
+                workspace_resolved,
+                scope_candidates,
+                scope=detected_scope,
+                setup_flags=_requested_setup_flags(
+                    ci=ci,
+                    claude_code=claude_code,
+                    agent_instructions=agent_instructions,
+                ),
+                kit=agent_instructions_kit,
```
pengfei-threemoonslab, line 814: [P1] Preserve an auto-discovered adoption kit in scoped recovery

A root `.agents-shipgate/adoption-kit.yaml` is loaded into `kit_config`, but recovery passes only the original CLI option, which is `None` for auto-discovery. Following an emitted candidate command therefore succeeds while silently installing bundled content; my repro omitted the custom file and recorded `kit_source: bundled`. Pass the loaded kit path through the same rebasing/containment logic used for an explicit kit.

Issue #324, opened 2026-08-06
bug(init/google-adk): infer App root_agent instead of selecting the first literal sub-agent


pengfei-threemoonslab:
## Summary

`init --write` can select the first literal Google ADK sub-agent name as `agent.name` even when the module explicitly constructs an `App(root_agent=...)` with a different coordinator.

This creates a schema-valid but semantically misleading manifest.

## Reproduction

Observed against [google/adk-samples#1745](https://github.com/google/adk-samples/pull/1745) with Agents Shipgate `0.16.0b7`, contract `19`.

The source has two literal sub-agents:

```python
salesforce_agent = LlmAgent(name="SalesforceAgent", ...)
sap_agent = LlmAgent(name="SapAgent", ...)
```

The coordinator is separately bound as the application root:

```python
root_agent = LlmAgent(
    name=AGENT_NAME,
    sub_agents=[salesforce_agent, sap_agent],
    ...
)
app = App(name="smart_closer_app", root_agent=root_agent)
```

`AGENT_NAME` defaults to `SmartCloserAgent` in the adjacent config module.

Observed init result:

```json
{
  "agent_name": "SalesforceAgent",
  "agent_name_candidates": [
    {"value":"SalesforceAgent","source":"Agent_name_literal"},
    {"value":"SapAgent","source":"Agent_name_literal"}
  ]
}
```

The generated manifest therefore declared the Salesforce worker as the reviewed agent identity.

## Expected behavior

Prefer structural root evidence over source-order literal candidates:

1. Resolve `App(root_agent=<symbol>)`.
2. Follow the assignment for that symbol to the corresponding `Agent`/`LlmAgent` construction.
3. Resolve simple local constants/defaults when safe and static.
4. If the root name remains dynamic or ambiguous, emit `CHANGE_ME` plus a diagnostic instead of asserting a sub-agent identity.

This is related to candidate-quality issue #320, but it is a distinct hierarchy problem: both literal names are plausible agent names; neither is the application root.

## Acceptance criteria

- [ ] An ADK `App(root_agent=root_agent)` fixture selects the root coordinator, not the first sub-agent encountered.
- [ ] Root assignment and sub-agent relationships are represented in detection evidence.
- [ ] Simple imported/default configuration constants can contribute evidence without importing user code.
- [ ] Ambiguous dynamic names fail closed to a placeholder and explicit review action.
[...]

PR #371, opened 2026-08-11
fix(init): rank agent-name candidates instead of taking the first one

### LINKED ISSUE #320: bug(init): avoid selecting one-character literals as the agent name
## Summary

`agents-shipgate init` selected the one-character identifier `t` as the generated `agent.name` for the Strix repository, even though the same static pass also found the much more plausible candidate `Strix`.

This can produce a syntactically valid manifest whose declared agent identity is misleading. The user must correct it manually before relying on later reports or handoffs.

## Reproduction

From a local clone of [usestrix/strix](https://github.com/usestrix/strix):

```bash
env AGENTS_SHIPGATE_AGENT_MODE=1 \
  PYTHONPATH=/path/to/agents-shipgate/src \
  python -m agents_shipgate init \
  --workspace /path/to/strix \
  --write \
  --json
```

Relevant output:

```json
{
  "auto_detected": {
    "is_agent_project": true,
    "frameworks": [
      {
        "type": "openai_agents_sdk",
        "score": 64.0,
        "confidence": "high"
      }
    ],
    "agent_name": "t",
    "agent_name_candidates": [
      { "value": "t", "source": "Agent_name_literal" },
      { "value": "Strix", "source": "Agent_name_literal" },
      { "value": "strix", "source": "workspace_dir" }
    ]
  }
}
```

The command successfully wrote `shipgate.yaml` and a managed `.gitignore` block. The issue is the automatic choice of `t` as the manifest's agent identity, not framework detection or file writing.

## Actual behavior

- The repository is correctly detected as a high-confidence OpenAI Agents SDK project.
- The one-character literal `t` is ranked above `Strix`.
- `init --write` uses `t` as the generated agent name.

## Expected behavior

When several literals are available, a one-character, context-poor identifier should not be silently selected over a plausible project/agent identity.

Possible fail-closed behavior:

1. Penalize or exclude very short/generic name literals unless they are the only candidate.
2. Prefer candidates with stronger agent-definition context and meaningful names.
3. If ambiguity remains material, write a `CHANGE_ME` placeholder (or emit an explicit diagnostic) rather than asserting an unreliable agent identity.

## Why this matters

`agent.name` appears in scan output and reviewer/agent-handoff artifacts. Selecting the wrong identity makes reports harder to interpret and can lead users to trust an auto-generated manifest that needs semantic correction.

This is a static candidate-ranking problem. It does not require executing the Strix agent, calling tools, or making network requests.

pengfei-threemoonslab:
## Summary

- Fixes #320 and #324, which are one defect with two faces: `_agent_name_candidates` appended `Agent(name=…)` literals in file-then-AST order with **no ranking**, and all three consumers — the manifest renderer, the `init` JSON summary, and the zero-install detector — took the first entry with an accepted `source`. "First encountered" was the entire selection policy.
- Replaces it with one ranking pass over four signals, plus one shared `select_agent_name` the renderer and the JSON summary both call.
- `agent_name_candidates[]` now carries `role`, `path`, `rank_score`, `selectable`, and `rationale[]` — the durable acceptance criterion from #324, without which any ranking that lands is untestable from the output.

Verified against both reporting repositories:

| repository | before | after |
|---|---|---|
| [usestrix/strix](https://github.com/usestrix/strix) | `t` | `Strix` |
| [google/adk-samples#1745](https://github.com/google/adk-samples/pull/1745) | `SalesforceAgent` | `SmartCloserAgent` (`role: root_agent`, `rank_score: 4.0`) |

### Two things reproduction turned up that the issues did not have

**Strix has no production literal at all.** Every `Agent(name=…)` literal in that repository lives in a test file — production builds the agent with `SandboxAgent(name=name, …)`, a parameter. So a length filter alone would pick `Strix` for the wrong reason. It wins here because it is *corroborated by the project name*: two independent sources agreeing, which is a real signal. `t` is rejected by the quality floor separately.

**The ADK sample's config import is absolute, not relative.** It is `from config import AGENT_NAME`, which works because ADK puts the agent directory on `sys.path`, and the constant is `os.environ.get("AGENT_NAME", "SmartCloserAgent")`. Both spellings resolve to the same sibling file, and the environment default is surfaced as *overridable at runtime* in the rationale rather than presented as a fact.

### The ranking

- **Hierarchy** — an application root (`App(root_agent=…)`, or the conventional `root_agent` symbol `adk run`/`adk web` discover) outranks an unqualified agent, which outranks a declared `sub_agents=[…]` / `handoffs=[…]` child. This is the only signal that can reorder two equally plausible names, which is the #324 case: `SalesforceAgent` is a genuine agent in that repository, so a reviewer skimming the manifest had no signal the declared identity was a worker.
- **Origin** — product code outranks test code, which names fixtures.
- **Corroboration** — a value the project name independently agrees with ranks above one only a single site declares.
- **Quality floor** — under three significant characters, or a generic scaffolding name, is ranked last and made `selectable: false`.

A rejected value is never written: `agent.name` keeps its `CHANGE_ME` placeholder and the existing `placeholders[]` review action, rather than asserting an identity nothing reliably declares.

### The static-only boundary

`name=<symbol>` resolves through **one** hop — a module-level constant, or an `os.environ.get("…", "…")` default in the same package. Bounded exactly as scoped in #324: the hop never chains, the target must already be a parsed file inside the workspace, and no user code is imported. Anything conditional, computed, or f-string-interpolated stays unresolved and falls back to the placeholder. Two tests pin the boundary specifically, because this is where it erodes one convenience at a time.

### One rule, one implementation

The defect exposed three copies of the selection step:

- The `source in {"Agent_name_literal", "ADK_name_field"}` set literal existed in both `template.py` and `_register_init.py` — two chances to disagree about the identity the manifest declares. Both now call `select_agent_name`.
- `tools/shipgate-detect.py` was the third and has to stay one, being stdlib-only. It is ported and pinned **byte for byte** by the parity suite. This is the one field there that is not merely structural parity: it names an agent rather than answering yes/no, so a script that ranked differently would send an agent to fix a different agent than `init` did. Every sample carries a single unambiguous literal and could never have caught a ranking divergence, so `test_script_agent_name_ranking_matches_cli` builds a workspace that can (hierarchy, cross-module constant, test-only literal, sub-floor value, and an annotated assignment). `script_version` → `0.3.0`.

## Type

- [x] Documentation only

## Verification

Additional local checks run:

- Full suite green (`pytest -q -n auto`), `ruff check .` clean, `compileall` clean.
- Reproduced both issues on the pre-fix code against a real clone of `usestrix/strix` and a fixture built from the actual `google/adk-samples#1745` sources (fetched via the GitHub API, not paraphrased from the issue), then confirmed both selections after the fix.
- Re-ran `detect_workspace` across all 19 `samples/` fixtures before and after: **no selection changed**. `google_adk_agent` now correctly reports `role: root_agent` for its existing `root_agent = Agent(name="adk_support_agent")`.
- `llms-full.txt` regenerated for the `docs/agent-recipes.md` edit.

**One thing a reviewer should know:** a first full parallel run showed an unrelated failure in `tests/test_authorization_verify_integration.py::test_signed_authorization_overlays_one_exact_command_and_binds_artifact` (`status == 'rejected'`, expected `'accepted'`). It passed in isolation, passed on an immediate full re-run, and a full run on pristine `HEAD` (75d9b20d) also passed — so it is not attributable to this change. It looks like a wall-clock flake: the fixture grant has `expires_at = now + timedelta(minutes=5)` and `human_authorization.py` rejects strictly on `evaluated_at >= expires_at` (correctly — skew must never extend authority). Left untouched here; worth a separate fix that pins the evaluation clock in that test.

## Release-readiness notes

- [x] No user-code import added to default scan paths — constant resolution is pure AST over files the walk already parsed, bounded to the workspace
- [x] New or changed check IDs are documented in `docs/checks.md` — n/a, no check IDs touched
- [x] Report/schema changes are additive or documented in `STABILITY.md` — `detect --json` is not a versioned contract surface (it carries no schema version and appears in no `contract.py` pin); the change is additive on `agent_name_candidates[]` entries, `project_name_candidates` is byte-unchanged, and the ranking is documented in `docs/agent-recipes.md` + `CHANGELOG.md`


src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -303,21 +393,165 @@ def _parse_python_facts(path: Path, workspace: Path) -> _PyFacts | None:
...
+            role=role,
+            rel_path=rel_path,
+            literal=literal.strip(),
+            root_evidence=root_evidence,
+        )
+    if isinstance(value, ast.Name):
+        return _AgentNameEvidence(
+            role=role, rel_path=rel_path, symbol=value.id, root_evidence=root_evidence
+        )
+    return None
```
pengfei-threemoonslab, line 1296: [P1] Preserve unresolved structural-root evidence and fail closed.

Returning `None` here erases the fact that an explicit application root exists. With a dynamic root name such as `name=f"{tier}-coordinator"` and a literal `WorkerAgent` child, detection drops the root, selects the worker, and `init` writes `agent.name: WorkerAgent`. That directly violates #324's acceptance criterion that an unresolved root produce `CHANGE_ME` plus review.

Please retain an unresolved-root signal and make selection return no candidate when a structural root exists but cannot be resolved safely.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -303,21 +393,165 @@ def _parse_python_facts(path: Path, workspace: Path) -> _PyFacts | None:
...
+                        self.child_symbols.add(element.id)
+                    elif isinstance(element, ast.Call):
+                        self.child_calls.add(id(element))
+
+    def role_for(self, call: ast.Call) -> tuple[AgentNameRole, str]:
+        """Return ``(role, evidence)`` for one agent construction."""
+        target = self.assign_targets.get(id(call))
+        if id(call) in self.root_calls:
+            return "root_agent", f"constructed inline as App({ROOT_AGENT_SYMBOL}=…)"
+        if target and target in self.root_symbols:
```
pengfei-threemoonslab: [P1] Resolve the actual lexical binding, not every assignment with the same identifier.

For `agent = Agent(name="OldWorker"); agent = Agent(name="ActualRoot"); App(root_agent=agent)`, both calls are classified as roots and source order selects `OldWorker`. Function-local variables named `root_agent` are also promoted as application roots.

Track lexical scope and resolve each root/child reference to its reaching assignment before the reference. Treat the conventional `root_agent` spelling as an ADK root only at module scope.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -303,21 +393,165 @@ def _parse_python_facts(path: Path, workspace: Path) -> _PyFacts | None:
...
+            targets = [node.target]
+            value = node.value
+        else:
+            continue
+        constant = _static_string(value)
+        if constant is None:
+            continue
+        for target in targets:
+            if isinstance(target, ast.Name):
+                constants.setdefault(target.id, constant)
```
pengfei-threemoonslab: [P1] Do not make the first static assignment authoritative.

`setdefault` makes `NAME = "OldAgent"; NAME = "CurrentAgent"; root_agent = Agent(name=NAME)` select `OldAgent`, although Python passes `CurrentAgent`. A later conditional or computed write is also ignored instead of causing the promised fail-closed result.

Resolve the binding at the `Agent(...)` call site, or conservatively reject symbols with multiple, conditional, or computed writes. Mirror the behavior in the zero-install implementation.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -565,20 +797,282 @@ def _confidence_label(score: float) -> str:
...
+    if not symbol:
+        return None
+    local = fact.module_constants.get(symbol)
+    if local is not None:
+        return local.value, local.provenance, fact.rel_path
+    imported = fact.constant_imports.get(symbol)
+    if imported is None:
+        return None
+    module, level = imported
+    for candidate_path in _constant_module_paths(fact.path, module, level, workspace):
```
pengfei-threemoonslab, line 1734: [P1] This can resolve a different import than Python executes.

`ImportFrom` bindings from all lexical scopes are flattened into one file-wide map, so a helper-local import can replace the module import used by a module-level root. Absolute probing also prefers the importer's sibling before the workspace root, and the current suffix order prefers `config.py` over `config/__init__.py`; both can disagree with runtime import resolution. Aliased imports additionally lose the original imported symbol.

Track imports per lexical scope and retain `(module, level, imported_name)`. If supported execution roots resolve to conflicting in-workspace values, leave the identity unresolved.

tools/shipgate-detect.py
```diff
@@ -542,6 +849,8 @@ def detect(workspace: Path) -> dict[str, Any]:
             project_names.append({"value": m.group(1).strip(), "source": "pyproject"})
     project_names.append({"value": workspace.name, "source": "workspace_dir"})
 
+    name_candidates = _rank_agent_names(py_facts, workspace, project_names)
```
pengfei-threemoonslab, line 1794: [P1] The new byte-parity guarantee does not hold with the current file inventory.

In a Git repo with tracked `RealAgent` and a gitignored root `IgnoredAgent`, canonical detection emits only `RealAgent`, while this script scans the ignored file and selects `IgnoredAgent`. A Python symlink escaping the workspace is also read and ranked here while canonical detection excludes it; its outside path can leak into the result. Separately, the global 5,000-file cap can stop before any Python source in asset-heavy repositories.

Use the same bounded Git inventory when available, containment-check every fallback-walk path, and cap Python parses rather than the entire file inventory. Add ignored-file, escaping-symlink, and >5,000-non-Python-file parity tests.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -303,21 +393,165 @@ def _parse_python_facts(path: Path, workspace: Path) -> _PyFacts | None:
...
+
+    def observe(self, call: ast.Call, tail: str) -> None:
+        """Record what one call says about the agents around it."""
+        for keyword in call.keywords:
+            if tail in APP_ROOT_CLASSES and keyword.arg == ROOT_AGENT_SYMBOL:
+                if isinstance(keyword.value, ast.Name):
+                    self.root_symbols.add(keyword.value.id)
+                elif isinstance(keyword.value, ast.Call):
+                    self.root_calls.add(id(keyword.value))
+            elif keyword.arg in CHILD_AGENT_KEYWORDS:
```
pengfei-threemoonslab: [P1] Restrict child-edge extraction to verified agent constructors.

This branch runs for any call with `handoffs=` or `sub_agents=`. For example, `configure(handoffs=[coordinator])` demotes `coordinator` to `sub_agent` and can make an unrelated worker win.

Require a verified `Agent`/`LlmAgent` constructor before recording child relationships, and mirror the guard in `tools/shipgate-detect.py`.

pengfei-threemoonslab (commented):
## Blocking engineer review

I recommend holding this PR before merge. The direction is good, but the current implementation still produces misleading `agent.name` values in several statically reproducible cases—including the fail-closed case required by #324.

The inline comments cover the six blocking implementation issues. Two additional consistency fixes should accompany them:

- `src/agents_shipgate/cli/detect.py:119-121` still prints candidate zero instead of using `select_agent_name`; with only `Agent(name="t")`, human-readable `detect` reports `t` while `init` emits `CHANGE_ME`.
- `docs/agent-recipes.md:108` says `project_name_candidates[]` has the “same shape” as enriched agent candidates, but project candidates remain `{value, source}`.

Validation performed against head `b8b5bacb28da892195c68200729a365d8c85f0d2`: all GitHub workflows, changed test suites, lint, compilation, diff checks, and Agents Shipgate verification pass. The semantic repros below are not covered by those checks.

_Submitted as a comment review because GitHub does not permit an author to request changes on their own pull request._

pengfei-threemoonslab:
All six blocking findings and both consistency fixes are addressed in `c102becf`. Every one reproduced against `b8b5bacb` before being fixed, and every fix has a test that fails without it — I checked that by stashing the source files and re-running.

The review was right about the shape of the defect: the ranking read Python name binding with a flattened `ast.walk` and first-write-wins heuristics. The fix is to read it the way Python does, or decline.

### Unresolved structural root (`signals.py:507`)

Reproduced exactly: `agent.name: WorkerAgent`. This one was the worst of the six because I had written a test asserting the wrong behaviour and rationalised it in the docstring as "better a demotable worker name than a fabricated identity". It is not — everything left after dropping the root is by construction *not* the root. The test is replaced by `test_unresolvable_root_name_blocks_selection_entirely`.

Now covers three ways a root can fail to resolve: a dynamic name, a symbol no single construction defines, and `App(root_agent=build_root())`. That last one was **not** in the review and I found it while writing the test — a factory call was registered as an inline root, then silently dropped because it produced no name evidence.

One deliberate choice worth your explicit sign-off: **the block is workspace-wide, not scoped to the file declaring the unresolved root.** I considered scoping it so a resolved root elsewhere could still win, which would avoid a false `CHANGE_ME` on a monorepo where one package uses a factory. I rejected it: that path can still assert an identity we cannot justify, which is the exact class of bug this PR removes. Blanket blocking can only ever cost a manual edit. Say the word if you want it scoped — it is one condition.

### Lexical binding (`signals.py:469`) and constants (`signals.py:535`)

Both reproduced (`OldWorker`, `LocalHelper` promoted to root, `OldAgent`). References now resolve by nearest scope then latest line before the reference; the `root_agent` convention applies at module scope only.

For constants I did not implement call-site resolution, because the conditional-write case defeats it — `NAME = "A"` followed by `if x: NAME = "B"` has no single textual answer. Instead **a symbol bound more than once anywhere in the file is never resolved**, counting `Store` names, parameters, `global`, and import bindings. That one rule is what makes reading module-level constants safe at all, and it subsumes the multiple, conditional, and computed cases together. It over-abstains when an unrelated local shares the name; that direction is free.

### Import resolution (`signals.py:870`)

All four sub-issues fixed: imports keyed by `(scope, name)`, `(module, level, imported_name)` retained so aliases resolve the imported symbol, `__init__.py` before the module file, and **conflicting in-workspace resolutions now leave the identity unresolved** rather than preferring the sibling.

### Child edges (`signals.py:457`)

Reproduced: `configure(handoffs=[coordinator])` demoted the coordinator and handed the identity to `SomeWorkerAgent`. Guarded on a verified `Agent`/`LlmAgent` constructor in both implementations.

### Zero-install inventory (`shipgate-detect.py:852`)

You were right that the byte-parity claim did not hold, and the symlink case was worse than described — the escaped name both won the ranking and leaked its outside absolute path into `path`:

```
CLI:    [('RealAgent', 'real.py')]
script: [('AAAEscapedAgent', '/private/tmp/.../outside/escaped.py'),
         ('AAAIgnoredAgent', 'ignored_agent.py'), ('RealAgent', 'real.py')]
```

The script now uses the same bounded `git ls-files -co --exclude-standard` inventory with the same `GIT_*` environment scrubbing, containment-checks every fallback-walk path, and bounds Python *parses* (`MAX_PYTHON_FILES = 1000`) rather than the inventory. All three requested parity tests are added and all three fail on the previous script.

The asset test needed care to be deterministic rather than accidentally passing: `os.walk` yields the top directory first, so the buried source has to sit *below* the assets directory for the old global cap to actually miss it.

### Consistency fixes

- `cli/detect.py` was the fourth copy of the selection rule. It now calls `select_agent_name`, and when nothing is selectable it says so and prints the rejection reasons instead of naming a value `init` will not write. Pinned by `test_detect_human_output_reports_what_init_will_write`.
- `docs/agent-recipes.md` no longer claims `project_name_candidates[]` shares the enriched shape.

### Verification

Full suite green, `ruff` clean, `compileall` clean, `llms-full.txt` regenerated. Sample selections are unchanged across all 19 fixtures, and both originally reported repositories still resolve correctly (`Strix`, `SmartCloserAgent`) — I re-ran against the real `usestrix/strix` clone and the `google/adk-samples#1745` fixture, plus a byte-parity check of CLI vs script across all eight repro workspaces.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -565,20 +977,322 @@ def _confidence_label(score: float) -> str:
...
+    best: dict[str, _RankedName] = {}
+    order = 0
     for fact in facts:
-        for literal in fact.agent_name_literals:
-            if literal not in seen:
-                candidates.append(NameCandidate(value=literal, source="Agent_name_literal"))
-                seen.add(literal)
+        for evidence in fact.agent_names:
+            resolved = _resolve_agent_name_evidence(evidence, fact, by_path, workspace)
+            if resolved is None:
```
pengfei-threemoonslab, line 1857: **[P1] Unresolved root symbols still select a worker.** When a root uses a symbol bound more than once, `_resolve_agent_name_evidence()` returns `None`, but this branch silently drops the root without adding it to `unresolved_roots`. Repro: `NAME = "Old"; NAME = "Current"; root_agent = Agent(name=NAME, sub_agents=[worker])` selects `WorkerAgent`. This contradicts the documented rule that an unresolved application root makes every candidate non-selectable. A failed resolution for evidence with role `root_agent` must block selection, including in the standalone detector.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -294,30 +430,308 @@ def _parse_python_facts(path: Path, workspace: Path) -> _PyFacts | None:
...
+
+    def _reaching(self, name: str, scope: int, lineno: int) -> _AgentBinding | None:
+        """The assignment a reference to ``name`` sees at ``lineno``.
+
+        Same scope first, then the module scope a nested reference falls
+        through to; within the chosen scope the latest assignment before the
+        reference wins, which is what Python's rebinding does.
+        """
+        candidates = self.bindings.get(name, [])
+        for candidate_scope in (scope, _MODULE_SCOPE):
```
pengfei-threemoonslab: **[P1] Source order is not Python reaching-definition semantics.** Choosing the latest preceding assignment fails for control flow and closures. In an `if/else`, this always selects the lexically later branch even though either branch may execute. Inside a nested function, it skips the enclosing function’s captured binding and falls through to the module binding. Both cases produce the wrong root identity. Track enclosing-scope relationships and treat competing control-flow assignments as ambiguous, failing closed where the runtime binding cannot be proven.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -294,30 +430,308 @@ def _parse_python_facts(path: Path, workspace: Path) -> _PyFacts | None:
...
+                hierarchy.observe(node, tail, scope)
+                if tail in AGENT_NAME_CLASSES:
+                    agent_calls.append((node, scope))
+
+        if isinstance(node, (ast.Assign, ast.AnnAssign)):
+            targets = (
+                list(node.targets) if isinstance(node, ast.Assign) else [node.target]
+            )
+            for target in targets:
+                if isinstance(target, ast.Name) and isinstance(node.value, ast.Call):
```
pengfei-threemoonslab, line 797: **[P1] Non-agent factory calls invalidate the real root silently.** Every call-valued assignment is recorded as an agent binding. With `root_agent = Agent(name="StaleRoot")` followed by `root_agent = build_root()`, conventional-root resolution selects the factory call, but that call produces no agent-name evidence or unresolved-root marker; `StaleRoot` remains selectable even though it is no longer the runtime root. Restrict agent bindings to verified constructors while separately treating later non-agent writes to a root symbol as unresolved.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -294,30 +430,308 @@ def _parse_python_facts(path: Path, workspace: Path) -> _PyFacts | None:
...
+                constants[target.id] = constant
+    return constants
+
+
+def _static_string(node: ast.expr) -> _Constant | None:
+    if isinstance(node, ast.Constant) and isinstance(node.value, str) and node.value.strip():
+        return _Constant(value=node.value.strip(), provenance="module_constant")
+    if isinstance(node, ast.Call):
+        callee = _call_name(node.func) or ""
+        if callee in ENV_LOOKUP_CALLS and len(node.args) == 2:
```
pengfei-threemoonslab: **[P1] Environment defaults are accepted without proving provenance.** Matching only the callee spelling lets a user-defined `getenv(key, fallback)` fabricate a static identity. A local function returning `RuntimeAgent` while receiving fallback `FabricatedAgent` causes detection to select `FabricatedAgent`. Only accept calls proven to resolve to unshadowed stdlib `os.getenv` or `os.environ.get`; otherwise leave the root unresolved.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -565,20 +977,322 @@ def _confidence_label(score: float) -> str:
...
+    if role == "root_agent":
+        score += 3.0
+        rationale.append(root_evidence or "bound as the application root")
+    elif role == "sub_agent":
+        score -= 1.5
+        rationale.append(
+            root_evidence or "declared as a child of another agent, not the root"
+        )
+
+    if _is_test_path(rel_path):
```
pengfei-threemoonslab, line 1949: **[P2] Test fixtures can still outrank product agents.** The documented ordering says product code outranks test code, but additive scoring gives a test `root_agent` score 2.0 and a production unqualified agent 1.0, selecting the fixture. Make origin a higher-priority ordering dimension, increase the test penalty beyond every hierarchy bonus, or revise the stated contract.

tools/shipgate-detect.py
```diff
@@ -140,18 +171,109 @@
...
+    env.update({
+        "GIT_ATTR_NOSYSTEM": "1", "GIT_CONFIG_GLOBAL": os.devnull,
+        "GIT_CONFIG_NOSYSTEM": "1", "GIT_CONFIG_SYSTEM": os.devnull,
+        "GIT_NO_LAZY_FETCH": "1", "GIT_NO_REPLACE_OBJECTS": "1",
+        "GIT_OPTIONAL_LOCKS": "0", "GIT_PAGER": "cat",
+        "GIT_PROTOCOL_FROM_USER": "0", "GIT_TERMINAL_PROMPT": "0",
+    })
+    def _run(args: list[str]) -> subprocess.CompletedProcess[bytes] | None:
+        try:
+            return subprocess.run(
```
pengfei-threemoonslab: **[P2] Zero-install Git inventory is not actually bounded.** `capture_output=True` buffers the complete Git inventory before the 16 MiB check, so the cap does not protect memory. When exceeded, the function also falls back to an uncapped filesystem walk, whereas canonical discovery raises `DiscoveryError`. Use bounded streaming like `_run_git_inventory_bounded()` and preserve canonical failure behavior.

pengfei-threemoonslab (commented):
## Engineer re-review — changes requested

The six blockers from the first review are addressed in `c102becf`, and the targeted suite plus all hosted checks are green. The replacement binding resolver still has four fail-closed identity failures, however, and the standalone detector mirrors them byte-for-byte. In each P1 case, `detect`/`init` can assert a different identity from the runtime application root.

Please address the four P1 findings before merge. The two P2 findings cover the documented product-vs-test ordering and the zero-install inventory safety/parity contract.

Validation performed: targeted detection/init/zero-install tests passed (one skip); Ruff, compilation, and diff checks passed; GitHub CI is green; Agents Shipgate verification reports `control_state: complete`. The semantic reproducers below are not covered by those checks.

Note: this is submitted as a COMMENT review because GitHub does not allow the PR author to formally request changes on their own pull request.

pengfei-threemoonslab:
All four P1s and both P2s are addressed in `27e14faa`. Each reproduced against `c102becf` first, and each fix has a test that fails without it (verified by stashing the source and re-running).

The four P1s turned out to be one defect. My resolver modelled only assignments whose *value* was an agent construction, resolved references by lexical position, and trusted a callee spelling. A binding model that cannot see a non-agent write, a branch, a closure, or a shadowed import will confidently report an identity the runtime never has. Writes are now recorded for **every** binding — assignments, parameters, imports, `global`, function definitions — each carrying scope, line, and whether it sits under a conditional. The four findings fall out of that one change.

### P1 · Non-agent factory calls invalidate the real root (`signals.py:481`)

Reproduced: `StaleRoot` selectable after `root_agent = build_root()`. Because only agent constructions were bindings, conventional-root resolution picked the factory call, which then produced no evidence and no marker — a stale construction kept the role. The last write to a root symbol now decides, whatever it binds.

### P1 · Source order is not reaching-definition semantics (`signals.py:620`)

Both cases reproduced (`BranchTwo`, and the module binding winning over the enclosing one). Any candidate under an `if`/`try`/loop now fails the lookup closed rather than taking the lexically later arm, and lookups walk the enclosing function scopes.

While implementing the scope chain I found a case **not** in the review: **class bodies are not in the closure lookup chain.** A method referencing a name bound in its class body raises `NameError` — I confirmed that against the interpreter rather than assuming — so walking through class scopes would resolve the root to an object Python never supplies. Excluded, with its own test. Same defect, one scope over.

### P1 · Unresolved root symbols still select a worker (`signals.py:1171`)

Reproduced exactly as written: `WorkerAgent`. My round-1 test passed only because its fixture had no second agent to fall through to. The check now runs where resolution happens — after every file's constants are known — so a root whose name is a symbol that fails cross-module resolution blocks selection like any other unresolved root.

### P1 · Environment defaults accepted without provenance (`signals.py:724`)

Reproduced: `FabricatedAgent`. Now requires an unshadowed stdlib import. I added a companion test for `import os as operating_system`, because the risk in "prove provenance" is over-correcting into recognising exactly one spelling.

### P2 · Test fixtures outranking product agents (`signals.py:1263`)

Reproduced: `TestFixtureRoot` over `ProductionAgent`. You offered three options; I took "increase the test penalty beyond every hierarchy bonus" rather than a separate ordering tier, so `rank_score` stays the single ordering key and the published number still explains the order. A tier would order candidates by something the score cannot show.

The penalty is now strictly greater than the whole spread of the other signals, and `test_origin_penalty_dominates_the_other_signals_by_construction` asserts that relationship directly against the named constants — so a future signal cannot widen the spread past it unnoticed. That test, not the constant, is the real guard.

### P2 · Zero-install Git inventory not actually bounded (`shipgate-detect.py:216`)

Correct on both counts. It now streams in 64 KiB chunks, kills the child on overrun, and **exits non-zero** rather than falling back — a fallback would do exactly the work the bound exists to refuse, and would answer from a different inventory than `init` used. Three tests: that the reader stops near the cap instead of draining the stream, that `detect` raises, and that the CLI surfaces it as exit 1.

### Verification

Full suite green, `ruff` clean, `compileall` clean, `llms-full.txt` regenerated. Byte parity between the CLI and the standalone detector holds across all eight reproducers. Sample selections unchanged across all 19 fixtures, and both originally reported repositories still resolve correctly (`Strix`, `SmartCloserAgent`) against a fresh `usestrix/strix` clone and the `google/adk-samples#1745` fixture.

One correction to my round-1 reply while I am here: I reported the round-1 constant-rebinding test as load-bearing. It was — it failed on the pre-fix code — but it was passing for a weaker reason than I implied, since its fixture had no second candidate to fall through to. That is what your P1 on `signals.py:1171` exposed. The replacement fixture has one.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -366,30 +573,412 @@ def _parse_python_facts(path: Path, workspace: Path) -> _PyFacts | None:
...
+                _record_write(facts, name, scope, node.lineno, conditional)
         elif isinstance(node, ast.Call):
             ctor = _call_name(node.func)
             if ctor:
                 facts.constructors.add(ctor)
-                literal = _name_keyword(node)
-                if literal and ctor.split(".")[-1] in {"Agent", "LlmAgent"}:
-                    facts.agent_name_literals.append(literal)
+                tail = ctor.split(".")[-1]
+                hierarchy.observe(node, tail, scope)
```
pengfei-threemoonslab: **[P1] Resolve constructors from their actual bindings**

Recognition still trusts only the terminal spelling (`Agent`, `LlmAgent`, or `App`). For example, with `Agent` shadowed by a local function and the real ADK constructors imported as `RealAgent`/`RealApp`, detection selects the fake `Agent(name="FabricatedRoot")` and misses the actual `RealApp(root_agent=RealAgent(name="ActualRoot"))`. `init` can therefore write a fabricated reviewed identity. Resolve callees through scoped, reaching import bindings: support aliases, and reject shadowed or ambiguous spellings. Please mirror the same provenance rule in the standalone detector.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -366,30 +573,412 @@ def _parse_python_facts(path: Path, workspace: Path) -> _PyFacts | None:
...
+        before the reference — which is what rebinding does. A candidate
+        under an ``if``/``try``/loop makes the whole lookup unprovable: both
+        arms of a branch can execute, and picking the lexically later one is
+        a guess dressed as an answer.
+        """
+        for candidate_scope in self._scope_chain(scope):
+            reaching = [
+                write
+                for write in self.writes.get(name, [])
+                if write.scope == candidate_scope and write.lineno < lineno
```
pengfei-threemoonslab: **[P1] Do not order enclosing writes by the nested reference's source line**

The `write.lineno < lineno` filter is valid inside the currently executing scope, but not for a global or closure lookup. Repro: `root = Agent(name="StaleGlobalRoot"); def make_app(): return App(root_agent=root); root = Agent(name="ActualGlobalRoot"); app = make_app()`. Runtime uses `ActualGlobalRoot`, while detection selects `StaleGlobalRoot` because the second module write is textually below the reference inside the function body. Model the enclosing scope's binding semantics (or fail closed when execution order is not provable) instead of comparing its writes with the nested node's line number; the standalone copy needs the same correction.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -366,30 +573,412 @@ def _parse_python_facts(path: Path, workspace: Path) -> _PyFacts | None:
         return None
 
     facts = _PyFacts(path=path, rel_path=_relative(path, workspace))
-    for node in ast.walk(tree):
+    nodes, scope_parents, class_scopes = _walk_scoped(tree)
+    hierarchy = _AgentHierarchy(
+        scope_parents=scope_parents, class_scopes=class_scopes, writes=facts.writes
+    )
+    agent_calls: list[tuple[ast.Call, int]] = []
```
pengfei-threemoonslab: **[P1] Preserve conditional provenance for inline application roots**

`_walk_scoped()` computes `conditional`, but `agent_calls` keeps only `(call, scope)` and `root_calls` later keeps only call IDs. Consequently, `if os.getenv("TIER"): app = App(root_agent=Agent(name="BranchOne")) else: app = App(root_agent=Agent(name="BranchTwo"))` makes both names selectable roots and deterministically chooses `BranchOne`, although the runtime identity depends on the branch. Carry conditional provenance through the call/root evidence and mark competing conditional roots unresolved rather than asserting one; mirror this in `tools/shipgate-detect.py`.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -366,30 +573,412 @@ def _parse_python_facts(path: Path, workspace: Path) -> _PyFacts | None:
...
+                        node.level,
+                        alias.name,
+                    )
         elif isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef)):
+            _record_write(facts, node.name, scope, node.lineno, conditional)
             for decorator in node.decorator_list:
                 name = _decorator_name(decorator)
                 if name:
                     facts.decorators.add(name)
+        elif isinstance(node, ast.Name) and isinstance(node.ctx, ast.Store):
```
pengfei-threemoonslab, line 757: **[P1] Record binding and unbinding forms that have no `Store` Name**

The claimed every-binding model still misses `del`, the outer binding created by `class`, exception aliases, match captures, and wildcard imports. The minimal `root_agent = Agent(name="StaleRoot"); del root_agent` still selects `StaleRoot`, even though the conventional module root no longer exists. `class root_agent`, `except E as root_agent`, `case root_agent`, and `from replacement import *` reproduce the stale selection too. Definitely overriding forms should retire the old root, while conditional/unknown forms should make it unresolved; please apply the same dispatch in the standalone implementation.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -366,30 +573,412 @@ def _parse_python_facts(path: Path, workspace: Path) -> _PyFacts | None:
...
+            write_by_node[id(node)] = _record_write(
+                facts, node.id, scope, node.lineno, conditional
+            )
+        elif isinstance(node, ast.arg):
+            _record_write(facts, node.arg, scope, node.lineno, conditional)
+        elif isinstance(node, ast.Global):
```
pengfei-threemoonslab, line 779: **[P1] Model `global` and `nonlocal` as scope declarations**

A `global` statement is currently recorded as a local write, `nonlocal` has no handling, and later stores are never redirected. Repro: a module `root_agent = Agent(name="OldRoot")`, followed by a function declaring `global root_agent` and assigning `Agent(name="NewRoot")`, makes runtime expose `NewRoot` but detection select `OldRoot`; the equivalent `nonlocal` mutation also selects the stale enclosing value. Track declaration sets per scope and route reads/writes to the module or nearest enclosing function instead of treating the declaration itself as an assignment. Mirror this in the zero-install detector.

tools/shipgate-detect.py
```diff
@@ -168,32 +207,194 @@
...
+    except (OSError, RuntimeError, ValueError):
+        return None
+    if any(p in SKIP_DIRS or p.startswith(".venv") for p in rel.parts):
+        return None
+    try:
+        if not resolved.is_file():
+            return None
+    except OSError:
+        return None
+    return resolved
```
pengfei-threemoonslab: **[P1] Preserve the logical path of contained symlinks**

Returning `resolved` changes the inventory entry's filename and extension. In a non-Git workspace with Python agent source stored in `source.txt` and `agent.py -> source.txt`, canonical detection retains `agent.py`, reports `is_agent_project=true`, and selects `AliasRoot`; this script turns both entries into `source.txt`, reports zero Python files, and returns `is_agent_project=false`. That breaks both the zero-install go/no-go verdict and the promised parity with `init`. Use resolution only to prove containment, then retain the logical path (and avoid re-resolving it in `_rel()`); add this in-workspace symlink parity case.

src/agents_shipgate/schemas/detect.py
```diff
@@ -73,7 +115,9 @@ class DetectResult(BaseModel):
 
     is_agent_project: bool
     frameworks: list[FrameworkDetection] = Field(default_factory=list)
-    agent_name_candidates: list[NameCandidate] = Field(default_factory=list)
+    # Ranked best-first. The first entry with ``selectable`` true is the one
+    # ``init`` writes; see ``signals.select_agent_name``.
+    agent_name_candidates: list[AgentNameCandidate] = Field(default_factory=list)
```
pengfei-threemoonslab, line 147: **[P2] Preserve the public `DetectResult` constructor contract**

Before this PR, `DetectResult(..., agent_name_candidates=[NameCandidate(value="GoodAgent", source="Agent_name_literal")])` was valid; both types remain public exports, but this narrowed annotation now raises a Pydantic `ValidationError`. Legacy dictionaries still parse only because the new fields have defaults, and they become `selectable=false`, silently changing selection semantics. Accept and upgrade legacy `NameCandidate` instances/payloads at the model boundary (or provide an explicit versioned migration) and add a constructor-compatibility regression.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -661,20 +1248,337 @@ def _confidence_label(score: float) -> str:
...
+
+    A name declared only in a test is a fixture name. It stays a candidate —
+    it is often the real one — but it must not outrank a name the shipped
+    code declares.
+    """
+    parts = Path(rel_path).parts
+    if any(part in {"test", "tests"} for part in parts[:-1]):
+        return True
+    stem = Path(rel_path).name
+    return stem == "conftest.py" or stem.startswith("test_") or stem.endswith("_test.py")
```
pengfei-threemoonslab: **[P2] Classify conventional top-level test modules**

This misses top-level `test.py` and `tests.py`, both conventional test module names. With `service.py` declaring `Agent(name="ProductionAgent")` and top-level `tests.py` declaring an `App` root named `TestFixtureRoot`, the fixture scores 4.0 and is selected over the product agent at 1.0, contradicting the documented product-over-test ordering. Include these exact filenames and mirror the predicate in `tools/shipgate-detect.py`.

tools/shipgate-detect.py
```diff
@@ -168,32 +207,194 @@
...
+            rel = raw.decode("utf-8")
+        except UnicodeDecodeError:
+            continue
+        contained = _contained(git_root / rel, workspace)
+        if contained is not None:
+            out.append(contained)
+    return sorted(set(out))
+
+
+def _walk_files(workspace: Path) -> list[Path]:
```
pengfei-threemoonslab, line 388: **[P2] Keep fallback inventory collection resource-bounded**

The non-Git fallback now resolves and materializes every file into a list, then a set, then a sorted list before the later Python parse cap applies. A downloaded workspace with millions of unrelated assets can therefore consume unbounded time and memory before detection considers a single Python file. The monorepo-scope fix does not require retaining all irrelevant assets: stream/filter into the bounded evidence categories needed by scope and framework detection, or fail closed on a documented resource budget, while keeping canonical and standalone behavior aligned.

pengfei-threemoonslab (commented):
Reviewed the rebased head `f0aab48c` against current `main`. The six issues from my previous pass appear fixed, and the focused detector/init/schema tests, static checks, hosted CI, and PR-scoped Shipgate verification are green.

I still found six P1 correctness defects that can make `init` assert the wrong agent identity or make the standalone detector return a different go/no-go verdict, plus three P2 compatibility/robustness gaps. Each inline comment is backed by a deterministic local reproduction on this exact commit. I recommend addressing these before merge.

GitHub does not allow the PR author to submit `REQUEST_CHANGES`, so this is recorded as a formal comment review with change-requesting findings.

pengfei-threemoonslab:
All six P1s and three P2s are addressed in `f10d0d31`. Each reproduced against `f0aab48c` first, and each fix has a test that fails without it (verified by stashing the source and re-running — 12 CLI tests, 3 detector tests).

Four of the six P1s turned out to be one gap. The model recorded *assignments* and trusted *spellings*, so any binding it could not see, and any name it could not attribute, produced a confident wrong identity. The rule is now the same everywhere: resolve it through the binding table, or decline.

### P1 · Constructor provenance (`signals.py:627`)

Reproduced: `FabricatedRoot`. The reaching binding decides now — an import of a framework class under any alias is a constructor, a spelling bound to a local `def`/`class` is not.

Worth flagging: **my first implementation over-corrected.** Honouring any alias of a name in `AGENT_NAME_CLASSES` also promoted `from mypkg.agents import Agent as Wrapped`, widening recognition instead of sharpening it. I caught it because I had written the negative case into the same test. Aliases are now honoured only from framework modules; unaliased spellings still fall through to the terminal name, so third-party re-exports and dotted `adk.LlmAgent` are unaffected.

### P1 · Enclosing-scope ordering (`signals.py:842`)

Reproduced: `StaleGlobalRoot`. A function body executes when called, not where written, so the line filter is only valid inside the reference's own scope. In an enclosing or module scope the only provable case is a single unconditional binding.

### P1 · Conditional inline roots (`signals.py:580`)

Reproduced: `BranchOne`. `conditional` was computed and then dropped at the `(call, scope)` boundary. It is carried through to root evidence now.

### P1 · Binding forms without a `Store` name (`signals.py:613`)

All five reproduced. `del`, `class`, `except … as`, and `case` are recorded as bindings that hold no agent, which is what retires the previous root; `from x import *` marks the file unprovable outright.

### P1 · `global` / `nonlocal` (`signals.py:619`)

Reproduced: `OldRoot`. They are declarations now, not writes, and stores are routed to the scope they actually bind. The redirected write is marked **conditional** rather than simply moved: whether the function runs, and when relative to the module body, is not something the file says — so the answer is `CHANGE_ME`, not `NewRoot`.

### P1 · Symlink logical path (`shipgate-detect.py:232`)

Reproduced exactly, and it is the worst of the six precisely because it is not about ranking: `is_agent_project` flipped to `false` and the script reported zero Python files. Resolution proves containment and nothing else now; `_rel()` no longer re-resolves.

### P2s

`DetectResult(agent_name_candidates=[NameCandidate(...)])` works again — instance and legacy dict are both upgraded at the model boundary using the rule that decided selection before ranking existed, so the silent `selectable: false` path is gone too. `tests.py`/`test.py` count as test code. The fallback walk has a documented ceiling (`MAX_WALK_FILES`) that **raises** rather than truncating, matching the existing `DiscoveryError` path — a partial inventory is a partial scope verdict. It counts per entry, not per directory; my first version checked once per directory and a single flat tree sailed past it, which the test caught.

### Two behaviour changes I want your explicit sign-off on

Both fall out of resolving provenance instead of spellings, and both touch things that were previously documented as correct:

1. `samples/_anti_patterns/dynamic_toolset_factory` defines a local `class Agent`. Its literal is no longer offered as a name candidate. The fixture comment and README stated the old behaviour and are updated.
2. `test_an_unrelated_agent_class_does_not_define_a_project` (yours, from #370) asserted that same literal stayed a "low-confidence name suggestion". I inverted that assertion: offering it under `source: Agent_name_literal` claims a provenance it never had, and `init` could write it as the reviewed identity — structurally the same fabrication as the `def Agent` repro in your P1. The test's actual subject, that the module does not define a *project*, is untouched. If you'd rather keep unattributed literals as suggestions under a distinct source value, that's a small change and I'll make it.

### Verification

Full suite green, `ruff` clean, `compileall` clean, `llms-full.txt` regenerated. Byte parity between the CLI and the standalone detector holds across all 14 reproducers for `agent_name_candidates`, `is_agent_project`, and `agent_scope`. Sample selections unchanged across all 19 fixtures; `Strix` and `SmartCloserAgent` still resolve.

One incidental: the new docs wording tripped `test_trust_claims_carry_meta_cli_qualifier` on the phrase "does not run" (about Python execution order, not about Shipgate running user code). I reworded the doc rather than touch the guard.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -366,30 +602,619 @@ def _parse_python_facts(path: Path, workspace: Path) -> _PyFacts | None:
...
+    :meth:`_AgentHierarchy._reaching`.
+    """
+    seen: set[int] = set()
+    current = scope
+    while current not in seen:
+        seen.add(current)
+        if current not in facts.class_scopes or current == scope:
+            found = [w for w in facts.writes.get(name, []) if w.scope == current]
+            if found:
+                return max(found, key=lambda write: write.lineno)
```
pengfei-threemoonslab: **[P1] Resolve constructor bindings at the call site, not at the end of the scope**

`_reaching_binding()` has no call location and always returns the highest-line binding. For example, `def Agent(*, name): return object(); root_agent = Agent(name="FabricatedRoot"); from google.adk.agents import Agent` calls the local fake at runtime, but detection retroactively uses the later framework import and publishes `FabricatedRoot` as a selectable `root_agent`. A conditional import is also treated as certain: `if os.getenv("USE"): from google.adk.agents import Agent as A; root_agent = A(name="MaybeRoot")` selects `MaybeRoot` even when the import never ran. Pass the call location, use the latest prior binding in the current straight-line scope, reject conditional bindings, and fail closed where an enclosing binding depends on invocation order. Mirror the rule in `tools/shipgate-detect.py` and add late/conditional-import regressions.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -366,30 +602,619 @@ def _parse_python_facts(path: Path, workspace: Path) -> _PyFacts | None:
...
+
+    An unbound spelling keeps the terminal-name reading. It is how a
+    third-party re-export (``from mypkg.agents import Agent``) and a
+    dotted ``adk.LlmAgent`` are still recognised, and there is nothing in
+    the file to contradict it.
+    """
+    parts = ctor.split(".")
+    tail = parts[-1]
+    binding = _reaching_binding(facts, parts[0], scope)
+    if binding is not None and len(parts) == 1:
```
pengfei-threemoonslab: **[P1] Validate the binding behind dotted constructors too**

The provenance check only runs when `len(parts) == 1`; every dotted call falls through to the terminal spelling. With a genuine framework import elsewhere, `class fake: class Agent: ...; root_agent = fake.Agent(name="FabricatedRoot")` is therefore accepted and selected even though `fake.Agent` is an unrelated local class. Rebinding a known module alias, such as `sdk = object(); root_agent = sdk.Agent(...)`, has the same problem. When a dotted head has a binding, require that binding to prove the expected framework module/alias and reject definitions, assignments, unrelated imports, and uncertainty. Keep any terminal-name heuristic only for genuinely unbound heads, and mirror this in the standalone detector.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -366,30 +602,619 @@ def _parse_python_facts(path: Path, workspace: Path) -> _PyFacts | None:
...
+                    )
+        elif isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef, ast.ClassDef)):
+            # A `class` statement binds its name in the enclosing scope just
+            # as `def` does, and either one retires an agent bound there.
+            _record_write(facts, node.name, scope, node.lineno, conditional, "definition")
             for decorator in node.decorator_list:
                 name = _decorator_name(decorator)
                 if name:
                     facts.decorators.add(name)
+        elif isinstance(node, ast.Name) and isinstance(node.ctx, ast.Store):
```
pengfei-threemoonslab, line 757: **[P1] Attribute writes can invalidate constructor and environment provenance**

The binding table records `Name` stores/deletes but not `Attribute` stores/deletes. Thus `import google.adk.agents as adk; adk.Agent = fake; root_agent = adk.Agent(name="FabricatedRoot")` still selects `FabricatedRoot`, although runtime constructs the fake object. The same gap makes environment defaults unsafe: after `import os; os.getenv = fake`, `NAME = os.getenv("NAME", "FabricatedRoot")` is trusted even when runtime returns another name. Record relevant attribute mutations (including deletes) and make `_constructor_role()` / `_is_stdlib_env_lookup()` decline after a possible rebind. The zero-install implementation needs the same model and regressions.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -341,19 +424,172 @@ def _collect_python_files(
...
+        """
+        seen: list[str] = []
+        for evidence in self.agent_names:
+            if evidence.literal is not None and evidence.literal not in seen:
+                seen.append(evidence.literal)
+        return seen
+
+
+_MODULE_SCOPE = 0
+_SCOPE_NODES = (ast.FunctionDef, ast.AsyncFunctionDef, ast.Lambda, ast.ClassDef)
```
pengfei-threemoonslab, line 523: **[P1] Model comprehension scopes before using their targets as bindings**

Python 3 comprehensions have an implicit child scope, but all four forms are omitted here. Repro: import real `Agent`/`App`; create `worker = Agent(name="WorkerAgent")`; evaluate `[App for App in ()]`; then `app = App(root_agent=Agent(name="ActualRoot"))`. Runtime keeps the imported `App` and the root is `ActualRoot`; the detector records the comprehension target as a module write, misses the App/root edge, and selects `WorkerAgent`. Add list/set/dict/generator-comprehension scopes (with the first iterable evaluated in the enclosing scope), mirror the standalone walker, and cover this wrong-identity case.

pengfei-threemoonslab (commented):
Re-reviewed `f10d0d31` against current `main`. All nine findings from the previous pass are fixed.

The focused detector/init/schema/scope suite passes (one skip), Ruff and compilation checks are clean, hosted CI is green, and the PR-scoped Agents Shipgate control is `complete`. I nevertheless reproduced seven new P1 identity-correctness defects and one P2 compatibility/validation regression on this exact head; each detector-semantics issue is present in both the canonical and zero-install implementations.

`git diff --check` also reports an extra blank line at EOF in `src/agents_shipgate/schemas/detect.py`.

These are change-requesting findings. Because this review is being submitted by the PR author's account, GitHub only permits a `COMMENT` review.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -341,19 +424,172 @@ def _collect_python_files(
...
+            parents[inner] = scope
+            if isinstance(node, ast.ClassDef):
+                class_scopes.add(inner)
+            # A nested definition's own body is straight-line relative to
+            # itself, whatever branch the definition sits in.
+            inner_conditional = False
+        else:
+            inner = scope
+            inner_conditional = conditional or isinstance(node, _BRANCH_NODES)
+        for child in ast.iter_child_nodes(node):
```
pengfei-threemoonslab, line 625: **[P1] Definition header expressions run in the enclosing scope**

Every child of a function/class node is currently assigned to its new inner scope, but defaults, decorators, annotations, class bases, and class keywords are evaluated outside that body. A valid repro is `worker = Agent(name="WorkerAgent"); def configure(App=App(root_agent=Agent(name="ActualRoot"))): return App; app = configure()`. Runtime constructs the App/root while evaluating the default, before parameter `App` exists; detection lets that parameter shadow the constructor, misses the root, and selects `WorkerAgent`. Walk header expressions in the enclosing scope and only parameters/body in the function scope (with the corresponding class rules), then mirror the standalone implementation.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -341,19 +424,172 @@ def _collect_python_files(
...
+        node, scope, conditional = stack.pop()
+        out.append((node, scope, conditional))
+        if isinstance(node, _SCOPE_NODES):
+            inner: int = id(node)
+            parents[inner] = scope
+            if isinstance(node, ast.ClassDef):
+                class_scopes.add(inner)
+            # A nested definition's own body is straight-line relative to
+            # itself, whatever branch the definition sits in.
+            inner_conditional = False
```
pengfei-threemoonslab: **[P1] Preserve conditional provenance of a definition's existence**

Resetting `inner_conditional` makes roots inside branch-defined functions unconditional evidence. For `if USE: def build(): return App(root_agent=Agent(name="BranchOne")) else: def build(): return App(root_agent=Agent(name="BranchTwo")); app = build()`, runtime identity depends on the branch, while detection marks both names selectable roots and deterministically chooses `BranchOne`. The body may be straight-line relative to itself, but the scope's existence/callability is conditional. Carry that declaration provenance into root evidence and fail closed when competing roots arise through conditional definitions; mirror it in `tools/shipgate-detect.py`.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -366,30 +602,619 @@ def _parse_python_facts(path: Path, workspace: Path) -> _PyFacts | None:
...
+                    agent_construction_targets[id(target)] = node.value
+
+    hierarchy.star_import = facts.star_import
+    _apply_scope_declarations(facts)
+
+    # Constructor provenance needs the completed binding table: a bare
+    # ``Agent`` is only a framework constructor when nothing in this file
+    # rebound the spelling, and an aliased ``RealAgent`` is one when the
+    # import says so. Deciding during the walk would read a half-built table.
+    roles = [
```
pengfei-threemoonslab, line 810: **[P1] Apply wildcard-import uncertainty to constructor provenance**

`star_import` is copied into the hierarchy, but constructor roles are still assigned without consulting it. In `from google.adk.agents import Agent; from google.adk.apps import App; from replacement import *; app = App(root_agent=Agent(name="FabricatedRoot"))`, `replacement` may legally replace both constructors; runtime can build plain fakes while detection selects `FabricatedRoot` as a proven inline root. Treat a preceding wildcard import as an unknown binding for constructor spellings until a later explicit binding re-establishes provenance, and apply the same source-ordered rule in the standalone detector.

src/agents_shipgate/schemas/detect.py
```diff
@@ -95,3 +159,31 @@ class DetectResult(BaseModel):
...
+        """
+        if not isinstance(value, list):
+            return value
+        upgraded: list[object] = []
+        for entry in value:
+            if isinstance(entry, AgentNameCandidate):
+                upgraded.append(entry)
+            elif isinstance(entry, NameCandidate):
+                upgraded.append(_upgrade_legacy_candidate(entry.model_dump()))
+            elif isinstance(entry, dict) and not _RANKED_FIELDS & set(entry):
```
pengfei-threemoonslab, line 196: **[P2] Preserve legacy sequence inputs without bypassing schema validation**

The old `list[NameCandidate]` field accepted tuples, but this shim upgrades only an actual `list`: a tuple of `NameCandidate` now raises, while a tuple of legacy dicts silently parses with `selectable=False`. For lists, `_upgrade_legacy_candidate()` also bypasses `NameCandidate` validation—missing values, wrong types, and extra keys are stringified or dropped despite `extra="forbid"` (for example a legacy dict with `unexpected: 1` is accepted). Normalize every sequence form Pydantic previously accepted, validate each legacy item through `NameCandidate.model_validate()` first, and only then enrich it. Add tuple and malformed-input regressions.

pengfei-threemoonslab:
All seven P1s, the P2, and the `git diff --check` EOF whitespace are addressed in `7b73aa30`. Each reproduced against `f10d0d31` first, and each fix has a test that fails without it (13 CLI tests, plus 10 new detector parity cases).

The seven P1s share a premise I had wrong. My last pass asked *"is this spelling bound to a framework import somewhere in this module?"* — a question that can be true in places where it does not apply. **Provenance is a question about a location, not about a file.**

### P1 · Binding must reach the call site (`signals.py:857`)

Both reproduced. `_reaching_binding()` took the highest-line binding, so an import at the bottom retroactively validated a decoy above it, and a conditional import counted as certain. It takes a line number now — and I collapsed it onto the same `_reaching_write` that agent-symbol resolution uses. That rule existed in two copies, which is how they had drifted into disagreeing about certainty.

### P1 · Dotted heads (`signals.py:807`)

Reproduced. Fixing it surfaced a latent bug in the bookkeeping it depends on: `import a.b.c` binds `a` denoting `a`, but I was storing the full path under the head, so `a.b.c.Agent` resolved to `a.b.c.b.c` and would never have matched. `import google.adk.agents` + `google.adk.agents.LlmAgent(...)` now has its own test.

### P1 · Attribute rebinding (`signals.py:652`)

Both reproduced. `adk.Agent = fake` and `os.getenv = fake` replace a constructor and a stdlib lookup while binding no name at all, so the import that "proved" them was still sitting there. Recorded now, and both `_constructor_role` and `_is_stdlib_env_lookup` decline after one.

### P1 · Comprehensions, headers, conditional definitions (`signals.py:517`, `:568`, `:564`)

All three reproduced, and all three were the walker treating Python's scoping as simpler than it is. Rewritten as a recursive visitor: comprehensions get their own scope with the outermost iterable outside it; definition headers are walked in the enclosing scope; and `declaration_conditional` carries a branch-defined scope's contingency into everything it declares.

### P1 · Wildcard imports and constructor roles (`signals.py:699`)

Reproduced. `star_import` was consulted for symbol lookups but not for constructor roles. It is source-ordered as you asked, so a later explicit binding restores what a spelling means — that case has its own test, since a permanent verdict would be easy to write and wrong.

### P2 · Legacy sequence inputs (`detect.py:184`)

All four behaviours reproduced (tuple of instances raising; tuple of dicts silently `selectable=False`; `unexpected: 1` accepted; missing `value` stringified to `''`). Any sequence is normalised now, and each legacy entry goes through `NameCandidate.model_validate()` before enrichment. `git diff --check` is clean.

### One process note

Mid-way through mirroring into the detector, a scripted splice matched an anchor *before* its start index and duplicated a large region of the file. It compiled, so `compileall` said nothing — the parity run caught it. I reverted `tools/shipgate-detect.py` to `HEAD` and redid the port with anchored, count-asserted replacements rather than patching the damaged file. Worth stating plainly because a mangled file that compiles is exactly the kind of thing a green lint run hides.

### Verification

Full suite green, `ruff` clean, `compileall` clean, `git diff --check` clean, `llms-full.txt` regenerated. Byte parity between the CLI and the standalone detector holds across all 19 reproducers for `agent_name_candidates`, `is_agent_project`, and `agent_scope`. Sample selections unchanged across all 19 fixtures; `Strix` and `SmartCloserAgent` still resolve.

Still open from last round, in case it got lost: the two behaviour changes I asked you to sign off on — the anti-pattern sample's literal no longer being a candidate, and the inverted assertion in `test_an_unrelated_agent_class_does_not_define_a_project`.

PR #376, opened 2026-08-14
fix(verify): rank release blockers above the trust-root notice and split the no-base fail-safe (#365)

### LINKED ISSUE #365: verifier headline ranks a medium trust-root notice above critical release blockers
## Summary

When a PR both touches the trust root and contains critical release blockers, `verifier.json.headline` reports the trust-root fact and the blockers are not mentioned. On first adoption — where touching the trust root is unavoidable — the headline is guaranteed to describe governance plumbing instead of the safety finding that actually blocks the merge.

## Evidence (reproduced 2026-08-10, CLI 0.16.0b7, contract 21)

`verify --base <merge-base> --head sg-head` on `google/adk-samples` PR #1917, with `shipgate.yaml` introduced by the same diff (first adoption):

```
merge_verdict:           blocked
control.state:           human_review_required   (must_stop: true)
base_status:             missing_manifest
headline: "This PR weakens the release policy that evaluates it; a coding agent cannot
           self-approve that change — a human must review it."
```

Meanwhile, in the same artifact:

```
release_decision.decision: blocked
release_decision.reason:   4 active findings block release
blockers[0..3]: SHIP-ACTION-FINANCIAL-WRITE-CONTROL-MISSING on spraay_batch_eth,
                spraay_batch_eth_variable, spraay_batch_token, spraay_batch_token_variable
```

The two findings driving the headline are both `medium`:

- `SHIP-VERIFY-TRUST-ROOT-TOUCHED` — *Release trust root touched: …/shipgate.yaml*
- `SHIP-VERIFY-POLICY-WEAKENED` — *Policy change cannot be proven safe (no base snapshot)*

`SHIP-VERIFY-POLICY-WEAKENED` here does not mean the policy was weakened; it means there was no base manifest to compare against, which is definitionally true for every first adoption (`base_status: missing_manifest`). So on first adoption the headline is *always* the governance message, regardless of what the PR does.

The subject of this PR is an agent that batch-pays ETH and ERC-20 on Base mainnet whose only confirmation is prompt text (`"Reply 'confirm' to execute"`). That does not appear in the headline.

## Impact

`headline` is the one line that reaches a PR comment, a chat reply, or a triage list. Ranking a medium governance notice above four critical financial-write blockers understates severity at the exact moment a human is deciding how much attention to spend — and it does so most reliably for new adopters, whose first impression is that the tool is about its own configuration rather than about their agent.

## Proposal

1. Rank `headline` by the strongest signal: critical/high release blockers outrank trust-root and policy notices. When both are present, lead with the blocker and append the review requirement.
2. Distinguish "policy weakened relative to a base" from "no base snapshot exists." The latter is a first-adoption condition, not a weakening claim; give it its own reason code and a message that says so.

## Before / After

**Before**
> This PR weakens the release policy that evaluates it; a coding agent cannot self-approve that change — a human must review it.

**After**
> Release blocked: 4 tools can move funds with no enforced controls. This PR also introduces the Shipgate policy that evaluates it, so a human — not the coding agent — must review both.

## Acceptance criteria

- [ ] Repro's headline names the financial-write blockers first and retains the human-review requirement.
- [ ] `control.state`, `must_stop`, `merge_verdict`, `can_merge_without_human`, and `permissions` are unchanged — ordering only.
- [ ] First adoption (`base_status: missing_manifest`) emits a distinct reason code from a real base-relative weakening, with distinct copy.
- [ ] A trust-root-only PR with no blockers still leads with the trust-root message.

## Scope / non-goals

No change to gating, control state, or human-review routing. This is headline selection and one reason-code split. It does not relax the first-adoption human-review requirement, which stays correct.

## Headline metric

**Merge-decision comprehension on first adoption** — the fraction of blocked verdicts whose headline names the blocking cause. Currently 0% for first-adoption PRs that also carry blockers.

pengfei-threemoonslab:
Closes #365.

Two facts were being reported in the wrong order, and one of them under a name that claimed something it could not know.

## 1. The headline ranks by severity, not by category

`verifier.json.headline` is the one line that reaches a PR comment, a chat reply, or a triage list. When a PR both touched the release trust root and blocked release on critical findings, it reported the trust-root fact and never mentioned the blockers. The two findings driving it are `medium`; the blockers they outranked were `critical`.

Reproduced on this branch before the change, on a repository that has already adopted Shipgate (`samples/support_refund_agent`, manifest edited by the diff):

```
merge_verdict: blocked        release_decision.blockers: 5 (3 critical, 2 high)
headline: "This PR edits a release trust root (the manifest, CI gate, agent
           instructions, or trigger catalog used to evaluate it); a coding agent
           cannot self-approve that change — a human must review it."
```

After:

```
headline: "5 active finding(s) block release; 11 review item(s) accepted as debt.
           Most severe: stripe.create_refund has financial write capability without
           required controls. This PR edits a release trust root (…); a coding agent
           cannot self-approve that change — a human must review it."
```

The rule: a run carrying a **critical or high** release blocker leads with the scan's own verdict line and **appends** the self-approval prohibition. The notice is never dropped, so the human-review requirement survives in the same string. A trust-root or policy change with **no** blockers still leads with the governance notice — it is then the whole story (issue AC4).

The leading clause also now names the worst blocker instead of only counting it. `"4 active finding(s) block release"` reads the same whether the agent is missing a docstring or can move funds with no enforced control; the blocker title already carries the tool and the capability, so naming it costs one clause. The pick is deterministic — severity, then check id, then title — so two runs of the same tree name the same row and output stays byte-stable. The same clause now serves the first-adoption headline shipped in #297, which previously also stopped at the count.

`control.human_review.why` follows the headline rather than the bare notice, so the control envelope can never name less than the headline does.

**Nothing else moves.** `control.state`, `must_stop`, `merge_verdict`, `can_merge_without_human`, `permissions`, `fix_task` routing, and `release_decision` are untouched — ordering only, asserted end-to-end in the new tests.

## 2. A first adoption no longer reports a weakening that could not have happened

The fail-safe that fires when there is no base policy to compare against shared a reason code with a proven base-relative weakening. So `verifier.json` reported `SHIP-VERIFY-POLICY-WEAKENED` against a base that carried no gate at all — the condition every first adopter meets by definition. In the issue's repro that is the finding titled *"Policy change cannot be proven safe (no base snapshot)"*, sitting under a reason code that says the policy was weakened.

That fail-safe is now its own reason code: **`SHIP-VERIFY-POLICY-BASE-ABSENT`** (medium, floor medium, category `verify`), carrying both evidence kinds:

| kind | meaning | `policy_weakened` | `policy_weakening_proven` |
|---|---|---|---|
| `manifest_introduced` | git proves the base carries no manifest under any name — a first adoption | `false` | `false` |
| `base_snapshot_unavailable` | no base report was obtainable; the direction is unprovable | `true` (fail-closed) | `false` |
| `ci_mode_weakened` / `fail_on_loosened` / `severity_override_lowered` (under `SHIP-VERIFY-POLICY-WEAKENED`) | a base-vs-head comparison ran and found the head weaker | `true` | `true` |

`SHIP-VERIFY-POLICY-WEAKENED` keeps firing, unchanged, for every proven base-relative weakening (`ci_mode_weakened`, `fail_on_loosened`, `severity_override_lowered`). It is **narrowed, not deprecated** — permitted by [STABILITY.md § Check IDs](STABILITY.md#check-ids) ("the conditions under which it fires may *narrow*"), and new check IDs may be added in any minor release.

Nothing about the gate moves with the reason code:

- same severity (`medium`) and same suppression immunity (category `verify`),
- same `human_ack` requirement on the `policy` surface (`_ACK_REQUIRING_CHECKS`),
- same `protected_surface_changes` rows (`_VERIFY_PATH_ROWS`),
- same release decision — including for configuration already written against the pre-split id: `checks.severity_overrides: {SHIP-VERIFY-POLICY-WEAKENED: critical}` still reaches the new check and still blocks, via a new `SPLIT_CHECK_ID_ALIASES` umbrella (kept out of the legacy/deprecated map, since the umbrella still emits for its own half). An override on the new id wins; floor validation is unchanged.
- same projection for artifacts written **before** the split: every read path accepts either id with the same evidence kinds, so a stored `report.json` carrying the old id with `manifest_introduced` is still an adoption. Nothing re-emits the old id.

In particular `verifier_summary.policy_weakened` keeps its **fail-safe** meaning: only a git-proven adoption clears it. A rename-and-loosen diff — which also finds nothing at the configured path on the base — still cannot clear the gate-bypass alarm by breaking the base scan. That is pinned by `test_an_unprovable_direction_still_counts_as_weakened`.

## Surface discipline

Per [`CONTRIBUTING.md` § Surface discipline](CONTRIBUTING.md#surface-discipline): the new check ID is not one of the listed new-surface classes (no CLI command, no schema version, no new report/`verifier.json` block, no discovery surface, no adapter). It moves the issue's headline metric — **merge-decision comprehension on first adoption**, the fraction of blocked verdicts whose headline names the blocking cause, currently 0% for first-adoption PRs that also carry blockers — and it extends the one decision engine rather than adding a parallel one: both reason codes are ordinary `Finding`s routed through `release_decision`.

## Review cycle (commit `15ce3b6`)

A ChatGPT review recorded a technical **REQUEST_CHANGES** with six defects. All six were reproduced and fixed; per-finding evidence is in the inline replies and [this comment](https://github.com/ThreeMoonsLab/agents-shipgate/pull/376#issuecomment-5298995307).

1. **Configured gating moved with the id** — a pre-split `severity_overrides` entry stopped applying, turning a configured `critical`/`blocked` into `medium`/`review_required`. Fixed with the `SPLIT_CHECK_ID_ALIASES` umbrella above.
2. **Fail-closed routing was speaking for the copy** — `policy_weakened` stays raised for an unprovable direction, but every renderer read it as a fact and claimed a weakening. `capability_review` gains the additive `policy_weakening_proven`; routing is unchanged, only the claim differs.
3. **The named blocker title was untrusted, unbounded, and multiline-capable** — newlines survived into a one-sentence field, and length alone pushed the appended `cannot self-approve` clause past the compact control envelope's 400-byte prose budget. Control characters are collapsed, the title is capped, and the governance suffix now gets a reserved byte budget.
4. **Pre-split artifacts stopped reprojecting** — old id + `manifest_introduced` read as a weakening and no longer as an adoption. One shared vocabulary (`core/policy_reason_codes`) now serves all four read paths.
5. **Stale public rationale** — `docs/triggers.json` and its five propagated copies corrected. `AGENTS.md:421` is a protected trust root that preflight routes to a human; the exact edit is written out in the comment above. Its claim is incomplete, not false — the safety property it asserts still holds.
6. **Reversed catalog `fires_when`** — corrected and regenerated.

## Second review cycle (commit `1a39c13`)

A second ChatGPT review recorded a technical **REQUEST_CHANGES** with five further defects. All five reproduced and fixed; per-finding evidence is in the inline replies and [this comment](https://github.com/ThreeMoonsLab/agents-shipgate/pull/376#issuecomment-5300720746).

1. **[P1] No-base PR copy claimed a proven weakening** — `pr-comment.md` printed `Policy weakened: true` off the fail-closed routing flag. It now prints `Policy changed, weakening unproven: true` with the reason; the route it reports is unchanged.
2. **[P1] A frozen schema identifier gained an emitted field** — an artifact declaring `verifier_schema_version: "0.8"` failed validation against the published v0.8 schema. The verifier schema advances to **`0.9`** (following the `0.7 → 0.8` precedent in #346); `docs/verifier-schema.v0.8.json` is restored byte-for-byte from `main` and frozen, `"0.8"` joins the legacy read set, and the model now enforces `policy_weakening_proven=true ⇒ policy_weakened=true`.
3. **[P1] Split-ID compatibility was incomplete** — the Tier B comparator read literal override keys (false negatives *and* false positives), and pre-split accepted-debt fingerprints stopped matching (a `critical` item moved from `review_required` to `blocked`). Both read paths now use exact-new-then-umbrella semantics; the legacy fingerprint candidate is scoped to declared split targets and still gated on `support_hash`.
4. **[P1] The reserved budget was spent after it was reserved** — the evidence-gap provenance note was appended after composition, so a long multibyte title plus one gap note made 443 bytes and the 400-byte compact projection dropped `a human must review it.` Every later addition goes through one composition; the configured-manifest path is bounded; when room runs out the gap note yields first and the requirement never does.
5. **[P2] Unicode format controls survived C0/C1 filtering** — U+202E and U+2066 reorder rendered text without changing a byte, and a lone surrogate raised inside the byte budgeting. Filtering is now by Unicode general category (`Cc`, `Cf`, `Cs`, `Co`, `Cn`, `Zl`, `Zp`).

**Wire contract note for consumers:** `verifier.json` now declares `verifier_schema_version: "0.9"`. Pin to [`docs/verifier-schema.v0.9.json`](docs/verifier-schema.v0.9.json); v0.8 and earlier stay frozen and still read (the new field defaults to `false`).

## Tests

New: [`tests/test_headline_ranking.py`](tests/test_headline_ranking.py) — 41 tests covering the ranking predicate, the named cause and its determinism, headline composition (blockers lead / governance-only still leads / failed scan still wins), the control envelope staying ordering-only, the reason-code split, the preserved `policy_weakened` fail-safe, the preserved `human_ack` and `protected_surface_changes` accounting, and two end-to-end `run_verify` runs for the repro shape and the trust-root-only shape.

Also new: [`tests/test_policy_reason_code_split.py`](tests/test_policy_reason_code_split.py) — 22 tests for the review findings: override compatibility (unit **and** an end-to-end gating run asserting the finding lands in `release_decision.blockers`), fail-closed-vs-honest copy across headline / control / fix task, legacy-artifact reprojection across all five read paths, the Action's findings fallback, the catalog direction, the frozen-v0.8 / current-v0.9 schema split and its forward normalization, the proven⇒weakened invariant, comparator false-negative and false-positive cases, and pre-split accepted debt keeping its `review_required` verdict. The headline suite adds the final-budget, priority-order, Unicode-category, and PR-comment-claim regressions.

Updated: `tests/test_first_adoption.py` (the fail-safe now carries its own reason code; the adoption records *no* `SHIP-VERIFY-POLICY-WEAKENED`; the adoption-precedence test is parametrized over proven/unprovable), `tests/test_fix_task_contract.py` (adoption fixture check id), `tests/test_cross_block_consistency.py` (the `policy_weakened` mirror reads both reason codes).

Full suite: `python -m pytest` — green except `test_codex_boundary_check.py::test_codex_check_boundary_json_golden_outputs`, which reproduces **byte-identically on a clean tree** (Python 3.14's `tomllib` attaches `line`/`column` to `TOMLDecodeError`, changing the evidence dict and therefore the fingerprint; the golden predates it). Unrelated to this PR and left alone. `python -m ruff check .` — clean. Generated surfaces (`scripts/generate_schemas.py`, `scripts/build-llms-full.py`) run twice produce no diff.

## Agents Shipgate local control

- `preflight --plan` was run before editing. It routed `human_review_required` on the first plan solely because it listed `AGENTS.md` (a protected trust root). **`AGENTS.md` was dropped from the change** and preflight then returned `agent_action_required` with `next_action: verify`.
- One follow-up is therefore left for a human: `AGENTS.md` (line 421) and the `rationale` text in `docs/triggers.json` both name `SHIP-VERIFY-POLICY-WEAKENED` as the check whose fail-safe treats `**/policies/**` as a trust root. That is now `SHIP-VERIFY-POLICY-BASE-ABSENT`. The surface selection itself (`_POLICY_SURFACES`) is unchanged, so the routing parity those sentences describe still holds — only the reason code named in the prose is stale. A coding agent must not self-approve an `AGENTS.md` edit, so this is routed rather than done.
- `verify` control state for this branch is reported at the bottom of this description.

## Not in scope

The issue's "After" copy is matched in substance, not verbatim: the leading clause is the scan's own verdict line plus the worst blocker's own title, rather than newly-authored prose. Nothing in the headline is generated text that the report does not already state.

## Agents Shipgate control state for this branch

Unchanged from the previous round: editing `docs/triggers.json` to fix the stale rationale is itself a trust-root change, so this branch's own verdict is review-required. That is the correct outcome, and it doubles as a live demonstration of the issue's AC4.

```
shipgate check --agent claude-code --workspace . --format agent-boundary-json
  decision: allow · input_coverage: complete · violations: 0
  control.state: agent_action_required  → next_action: verify

agents-shipgate verify --workspace . --config shipgate.yaml --base main --head HEAD \
  --ci-mode advisory --format control
  control_state:      review_publishable
  decision:           review_required      (decision_source: release_decision)
  execution:          succeeded            exit_code: 0
  human_review:       required=true
  permissions:        commit/push/update_pr=true · merge=false · report_complete=false
  current_control_id: sha256:f54fa33ec43624c328d53a94945e093260346b70e5468df559f5968478663036
  contract_version:   24

report.json → review_items: 1 · blockers: 0
  SHIP-VERIFY-TRIGGER-CATALOG-DRIFT (medium) · changed_trigger_files: ["docs/triggers.json"]
verifier.json → verifier_schema_version: 0.9 · headline: 194 bytes
                capability_review.policy_weakened: false · policy_weakening_proven: false
```

**A human must approve the merge.** `review_publishable` permits commit, push, and PR updates so review can happen; `permissions.merge` and `permissions.report_complete` are `false`. Only the current `next_action` was followed at each step, and nothing was merged.

## Human-owned follow-ups

Preflight routes any `AGENTS.md` edit to `human_review_required` / `must_stop: true`, so a coding agent must not make these. Two edits, both prose:

1. **Line 421** — in `the Tier B checks (\`SHIP-VERIFY-POLICY-WEAKENED\`, \`SHIP-VERIFY-CI-GATE-REMOVED\`, agent-instruction weakening, trigger-catalog drift)`, insert `` `SHIP-VERIFY-POLICY-BASE-ABSENT` `` after `` `SHIP-VERIFY-POLICY-WEAKENED` ``. The sentence is incomplete, not false: it claims the Tier B checks select their changed files the same way, and the new check selects through the same `_POLICY_SURFACES` / `touched()` path.
2. **Schemas table** — the `Verifier schema (current)` row still points at `docs/verifier-schema.v0.5.json` / `0.5`. It has been stale since v0.6 and this PR takes the current version to `0.9`; the row should read [`docs/verifier-schema.v0.9.json`](docs/verifier-schema.v0.9.json) / `0.9`.

After either edit, re-run `python scripts/build-llms-full.py` (`llms-full.txt` carries a copy).


pengfei-threemoonslab (commented):
Technical verdict: **REQUEST_CHANGES**.

GitHub identity limitation: the connected account (`pengfei-threemoonslab`) is also the PR author, so GitHub cannot accept a self-request-changes review; I am submitting this as COMMENT while recording the technical verdict explicitly.

The critical/high blocker-first ordering is deterministic, the governance-only case still leads with its notice, and the default-config tests preserve the control/permission/merge/fix-task/release-decision fields. However, six actionable defects remain:

1. The reason-code split changes configured gating: an existing critical severity override for `SHIP-VERIFY-POLICY-WEAKENED` no longer applies to the renamed no-base finding, changing `blocked` to `review_required`.
2. `base_snapshot_unavailable` gets the honest new finding ID/title, but the headline, control reason, fix task, and PR prose still claim a proven policy weakening.
3. The selected blocker title is unbounded and multiline-capable. A 200-character title is enough for the compact control projection to truncate away the appended self-approval clause; larger values also defeat the 6,000-character PR-comment budget.
4. Historical reports using the old ID plus `kind=manifest_introduced` are reprojected as weakened and are no longer recognized as pure adoption.
5. `AGENTS.md:421` and `docs/triggers.json:253` still name the old ID as the no-base fail-safe; that stale public rationale propagates to `llms-full.txt`, examples, and protocol goldens.
6. The machine catalog's `fires_when` comparison says base is weaker than head, the opposite of the implementation.

Evidence/checks: exact base/head inspected; 654 focused tests passed; Ruff, schema generation, llms-full consistency, and diff checks passed. The full local run had 5,163 pass / 5 skip; its one golden failure reproduces on `main` under Python 3.14, and eight packaging setup errors were sandbox-network build-isolation failures. All current GitHub workflows are green. Fresh Shipgate control is `complete` at `sha256:90fa930989591a3166a5214d30a3f2941f5e07061ca3ec44f49b7360ce9eca3b`; the defects above are contract/compatibility edges not covered by that run.

Claude needs another implementation cycle before approval.

src/agents_shipgate/checks/verify_policy.py
```diff
@@ -219,7 +239,7 @@ def _fail_safe(context: ScanContext) -> list[Finding]:
         return [
             verify_finding(
                 context,
-                check_id=CHECK_ID,
+                check_id=BASE_ABSENT_CHECK_ID,
```
pengfei-threemoonslab, line 293: [P1] Preserve configured severity semantics when splitting this ID. An existing `checks.severity_overrides.SHIP-VERIFY-POLICY-WEAKENED: critical` upgraded this no-base finding before the split. With the new emitted ID and no compatibility mapping, that override is silently ignored; the same fixture changes from `critical / blocked` to `medium / review_required`. That violates the ordering-only acceptance criterion and the STABILITY claim that the resulting release decision is unchanged. Please add a deterministic compatibility/migration mapping and a regression test for existing overrides.

src/agents_shipgate/cli/verify/capability_review.py
```diff
@@ -164,11 +165,16 @@ def _trust_root_touched(report: ReadinessReport) -> bool:
...
-    # without weakening anything, and this flag is read as a fact downstream.
+    # Same rule as ``verifier_blocks``: the no-base fail-safe keeps this flag
+    # raised (an unprovable direction is treated as weakened for routing), but
+    # a proven first adoption weakens nothing, and this flag is read as a fact
+    # downstream.
     return any(
         f.check_id == POLICY_WEAKENING_CHECK_ID
-        and (f.evidence or {}).get("kind") != "manifest_introduced"
+        or (
+            f.check_id == POLICY_BASE_ABSENT_CHECK_ID
```
pengfei-threemoonslab: [P1] Keep fail-closed routing separate from proven-weakening copy. For `SHIP-VERIFY-POLICY-BASE-ABSENT` with `kind=base_snapshot_unavailable`, this deliberately leaves `policy_weakened=True`; existing `_self_approval_note` and fix/PR renderers interpret that boolean as a fact and say “This PR weakens the release policy.” I reproduced a new-ID/no-base finding whose headline and `control.human_review.why` still use the old false claim. Preserve the conservative route/boolean if required, but select human-facing copy from the distinct reason/evidence so an unprovable direction is not reported as proven weakening.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -1824,6 +1829,80 @@ def _gap_provenance_note(
...
+    if report.agent_summary is not None:
+        summary = report.agent_summary.headline
+    elif report.release_decision is not None:
+        summary = report.release_decision.reason
+    else:
+        return "Shipgate requires human review."
+    worst = _worst_blocker(report.release_decision)
+    if worst is None:
+        return summary
+    title = worst.title.rstrip(".")
```
pengfei-threemoonslab: [P1] Normalize and bound this untrusted title before composing the headline. `ReleaseDecisionItem.title` can include an unconstrained OpenAPI/MCP tool name. Newlines survive, violating the single-sentence contract; a 200-character title already makes the 400-byte compact control projection truncate away the appended `cannot self-approve` clause, and a 7k title expands the PR comment to ~30k despite its 6k cap. Reserve a deterministic budget for the governance suffix, collapse control whitespace, and add long/control-character tests across verifier, compact control, and PR-comment output.

src/agents_shipgate/cli/verify/fix_task.py
```diff
@@ -32,7 +32,7 @@
 
 _MAX_INSTRUCTIONS = 5
 _MAX_REPAIRS = 10
-_ADOPTION_CHECK_ID = "SHIP-VERIFY-POLICY-WEAKENED"
+_ADOPTION_CHECK_ID = "SHIP-VERIFY-POLICY-BASE-ABSENT"
```
pengfei-threemoonslab: [P2] Emit the new ID, but continue accepting the historical adoption shape on read/reprojection paths. Existing artifacts carry `SHIP-VERIFY-POLICY-WEAKENED` plus `kind=manifest_introduced`; after this replacement `is_pure_adoption_review` no longer recognizes them, while the companion capability/verifier-summary fallbacks now return `policy_weakened=True` (main returned false). Supporting both IDs for legacy `manifest_introduced` evidence preserves adoption wording, protected-surface projection, and repair routing without re-emitting the old ID.

docs/checks/verify.yaml
```diff
@@ -43,11 +43,11 @@ checks:
...
     agent that makes CI pass by weakening the gate rather than fixing the finding.
-    When no base snapshot is available but a policy trust root was touched, the
-    check fails safe to a review-required finding instead of silently passing.
-  fires_when: A VerificationContext is present and the base report's
-    effective_policy is weaker than the head manifest's; or, with no base
-    snapshot, a policy/manifest trust root was changed.
+    The claim is base-relative, so it fires only when a base snapshot exists to
+    compare against; the no-base fail-safe is SHIP-VERIFY-POLICY-BASE-ABSENT.
+  fires_when: A VerificationContext and a base effective-policy snapshot are both
+    present and the base report's effective_policy is weaker than the head
```
pengfei-threemoonslab: [P2] This comparison direction is backwards. The implementation fires when the **head** effective policy is weaker than the base, not when the base is weaker than head. Because this source regenerates `docs/checks.json` and other public catalog surfaces, please correct the machine-readable `fires_when` text and regenerate.

src/agents_shipgate/checks/verify_policy.py
pengfei-threemoonslab: Fixed in 15ce3b6. Reproduced first — the override stopped reaching the finding, exactly as described:

```
before: emitted SHIP-VERIFY-POLICY-BASE-ABSENT · severity after
        {SHIP-VERIFY-POLICY-WEAKENED: critical} = medium
after:  emitted SHIP-VERIFY-POLICY-BASE-ABSENT · severity after
        {SHIP-VERIFY-POLICY-WEAKENED: critical} = critical
```

The pre-split id is now an umbrella over both halves via a new `SPLIT_CHECK_ID_ALIASES` map in `core/check_ids.py`, consulted by `expands_to_check_id` alongside the legacy map. It is deliberately **not** added to `LEGACY_CHECK_ID_ALIASES`: that map also drives `_verify_deprecated_check_ids`, and `SHIP-VERIFY-POLICY-WEAKENED` is not deprecated — it still emits for its own half — so a baseline entry naming it must not be reported as a stale alias.

Two properties preserved on purpose:

- An override written against `SHIP-VERIFY-POLICY-BASE-ABSENT` still wins, because `_severity_override_for_check` prefers the exact id before falling back to expansion.
- Floor validation is unchanged: `_resolve_metadata` resolves the umbrella id directly in the catalog, so an override against `SHIP-VERIFY-POLICY-WEAKENED` is still checked against its own `high` floor rather than the new check's `medium` one.

Regression tests in `tests/test_policy_reason_code_split.py`:

- `test_the_pre_split_id_is_an_umbrella_for_both_halves` — expansion is one-directional, and the umbrella is not in the legacy/deprecated map.
- `test_an_existing_override_still_reaches_the_no_base_finding` — parametrized over both evidence kinds.
- `test_an_override_on_the_new_id_wins_over_the_umbrella`.
- `test_a_configured_override_still_blocks_the_release_end_to_end` — your gating reproduction: a real `run_verify` over a repo whose manifest carries the pre-split override, asserting the finding is `critical`, the decision is `blocked`, **and** that `SHIP-VERIFY-POLICY-BASE-ABSENT` appears in `release_decision.blockers`. That last assertion is the one that fails without the mapping (the sample carries other blockers, so `decision == "blocked"` alone would not have caught it).

src/agents_shipgate/cli/verify/capability_review.py
pengfei-threemoonslab: Fixed in 15ce3b6. Reproduced first:

```
before: policy_weakened=True  note="This PR weakens the release policy that evaluates it; …"
after:  policy_weakened=True  policy_weakening_proven=False
        note="This PR changes the release policy that evaluates it and no base policy was
              available to prove the change does not weaken the gate; a coding agent cannot
              self-approve that change — a human must review it."
```

Routing and claim are now separate facts. `capability_review.policy_weakened` keeps its fail-closed meaning — raised whenever the direction may have moved, including when nothing compared — and everything that gates still reads it. A new additive field `capability_review.policy_weakening_proven` (default `false`, never `true` without `policy_weakened`) carries the narrower fact that a base-vs-head comparison actually ran and found the head weaker. Copy is selected from the narrower one; the route is byte-identical either way.

Three renderers were saying it, and all three now branch on the narrow fact:

- `_self_approval_note` → the verifier headline and `control.human_review.why` / `stop_reason`.
- `fix_task`'s `review_policy_weakening` repair reason ("A human must approve release-policy weakening" → "…the release-policy change, whose direction could not be proven without a base policy to compare against").
- The PR comment, which quotes the headline.

The shared predicate lives in a new `core/policy_reason_codes` module (`counts_as_weakened` for routing, `weakening_is_proven` for copy) so the four consumers cannot drift apart.

Regression tests in `tests/test_policy_reason_code_split.py`:

- `test_an_unprovable_direction_routes_closed_but_claims_nothing` — asserts `policy_weakened is True` **and** `policy_weakening_proven is False`, no "weakens the release policy", and that "cannot self-approve" / "a human must review it" survive.
- `test_a_proven_weakening_still_says_it_weakens_the_policy` — the `ci_mode_weakened` case is unchanged.
- `test_the_repair_reason_does_not_assert_an_unproven_weakening`.
- `test_a_no_base_run_reports_the_honest_copy_end_to_end` — a real `run_verify` over an adoption that also touches an existing policy pack (so the kind is `base_snapshot_unavailable`, not `manifest_introduced`), asserting the string is absent from the headline, the whole serialized `control`, and the whole serialized `fix_task`, while `can_merge_without_human` stays `False` and `human_review.required` stays `True`.

`tests/test_first_adoption.py::test_a_policy_finding_beats_the_adoption_wording` is parametrized over proven/unprovable now: the adoption wording loses to a policy finding either way, only the claim differs.

src/agents_shipgate/cli/verify/orchestrator.py
pengfei-threemoonslab: Fixed in 15ce3b6. Reproduced first, with a title carrying newlines plus 600 characters of padding:

```
before: headline contains newline: True   length: 845
        truncate_prose(headline) keeps "self-approve": False
        compact tail: " X X X X X X X X X X X X X X X X X X X X X X X […]"
after:  headline contains newline: False  length: 328
        truncate_prose(headline) keeps "self-approve": True
        compact tail: "coding agent cannot self-approve that change — a human must review it."
```

Three changes, all in `cli/verify/orchestrator`:

1. **Normalize.** `_one_clause` collapses `[\x00-\x1f\x7f-\x9f]` and whitespace runs to single spaces. Collapsed rather than escaped — the headline is prose, not a transcript, and a `\x0a` escape mid-sentence is no more readable than the newline it replaces. (`single_line_text` stays the right tool for the terminal renderer, where showing the byte matters.)
2. **Bound the untrusted value on its own account.** `_HEADLINE_TITLE_MAX_CHARS = 120`, marked with `…`, independent of the composition budget — that is what keeps a 7 k title from expanding any downstream renderer that quotes the headline.
3. **Reserve the suffix budget.** `_compose_with_reserved_suffix` computes `MAX_ENVELOPE_PROSE_BYTES - len(suffix) - 1` and shortens the *lead* to fit, so `truncate_prose` never has anything to cut. If the requirement alone fills the budget, it is published on its own — which is exactly what the headline said before blockers ever led it, so the degenerate case degrades to the pre-change behaviour rather than losing the requirement. Both composing branches use it (blocker-leads and mixed-adoption), and `configured_manifest` is normalized on the way into the adoption suffix.

Regression tests in `tests/test_headline_ranking.py`:

- `test_a_control_character_title_stays_one_clause_of_one_sentence` — parametrized over `\n`-forged `Control: complete` / `Agent must stop: false` lines, `\r\n`, `\t`, NUL, and an ANSI escape; asserts none survive, that nothing is escaped into `\x`, and that the self-approval clause is still there.
- `test_a_long_title_is_bounded_and_marked`.
- `test_the_human_review_requirement_survives_the_compact_control_budget` — parametrized at 200 / 2 000 / 7 000 characters (200 is your exact reproduction); asserts the composed headline fits `MAX_ENVELOPE_PROSE_BYTES`, that `truncate_prose(headline) == headline`, and that it still ends with "a human must review it."
- `test_a_requirement_that_fills_the_budget_is_published_on_its_own`.
- `test_the_adoption_headline_reserves_the_same_room`.
- `test_a_hostile_title_does_not_expand_the_pr_comment` — renders both PR-comment styles from a real `VerifierArtifact` with a 7 k multiline title and asserts `len(comment) <= 6000` and that no forged `\nControl: complete` line appears.

src/agents_shipgate/cli/verify/fix_task.py
pengfei-threemoonslab: Fixed in 15ce3b6. Reproduced first, on a stored finding carrying the old id plus `kind=manifest_introduced`:

```
                                    before   after   main
capability_review.policy_weakened    True     False   False
verifier_summary.policy_weakened     True     False   False
protected_surface_changes            []       [('policy','shipgate.yaml')]  same as main
is_pure_adoption_review              False    True    True
```

The read side now goes through one shared vocabulary — a new `core/policy_reason_codes` module — rather than four hand-written id comparisons:

- `counts_as_weakened(check_id, evidence)` — fail-closed routing: either reason code counts as weakened unless the evidence kind is `manifest_introduced`.
- `weakening_is_proven(check_id, evidence)` — the honest-copy fact: either reason code, but only for kinds that came from an actual comparison (so a *legacy* `base_snapshot_unavailable` is not retroactively promoted to a proven weakening either).
- `is_adoption_evidence(check_id, evidence)` — the adoption route, which is what `_ADOPTION_CHECK_ID` was pinning here.

Consumers updated: `capability_review`, `verifier_blocks.build_verifier_summary`, `verifier_blocks._VERIFY_PATH_ROWS` (the pre-split id is mapped again, so legacy artifacts keep their `policy` rows and therefore their `human_ack` requirement), `fix_task` (both the pure-adoption predicate and `_is_same_manifest_adoption_finding`), and `scripts/github_action_outputs._verifier_flags`.

The emitter is untouched: a no-base run still emits only `SHIP-VERIFY-POLICY-BASE-ABSENT`, pinned by `test_the_split_never_re_emits_the_pre_split_id_for_a_no_base_run`.

Regression tests in `tests/test_policy_reason_code_split.py`:

- `test_a_legacy_adoption_artifact_still_reprojects_as_an_adoption` — all five projections at once (capability review, verifier summary, protected-surface rows, `human_ack.required`, `is_pure_adoption_review`).
- `test_a_legacy_no_base_artifact_still_reprojects_as_weakened` — the other half: still `policy_weakened=True`, still not a proven weakening, still not an adoption.
- `test_github_action_outputs_read_both_reason_codes` — the Action's findings fallback, both ids × both kinds.

docs/checks/verify.yaml
pengfei-threemoonslab: Fixed in 15ce3b6. You are right, and it was wrong before this PR too — I copied the direction from the line I was editing instead of checking it against `_compare`, which fires on `head_rank < base_rank` / `base_fail - head_fail` / `head_sev < base_sev`.

```
before: "…and the base report's effective_policy is weaker than the head manifest's."
after:  "…and the head manifest's effective policy is weaker than the base report's."
```

`description` was already correct ("The PR's effective release policy is weaker than the base report's"), so the two no longer contradict each other. Regenerated with `scripts/generate_schemas.py`; `docs/checks.json` carries the corrected text, and `scripts/build-llms-full.py` was re-run for the propagated copy.

`test_the_catalog_states_the_comparison_in_the_direction_it_runs` in `tests/test_policy_reason_code_split.py` asserts the corrected sentence is present, that the reversed one is absent, and that the pair is described as mutually exclusive on the base snapshot (`SHIP-VERIFY-POLICY-BASE-ABSENT`'s `fires_when` requires *no* base snapshot). Re-running the generator twice produces no diff, so the checked-in catalog matches this source.

pengfei-threemoonslab:
All six defects reproduced and addressed in **15ce3b6**. Inline replies carry the per-finding evidence; this covers **(5)**, which had no inline thread, and the control state.

## (5) Stale public rationale

Preflight first, as required.

`docs/triggers.json` — preflight on a plan naming it and its five propagated copies returned `agent_action_required` (`next_action: verify`, `must_stop: false`), so I made the edit. The sentence

> `SHIP-VERIFY-POLICY-WEAKENED` additionally treats `**/policies/**` as a fail-safe trust root

became

> the policy check's no-base fail-safe, `SHIP-VERIFY-POLICY-BASE-ABSENT`, additionally treats `**/policies/**` as a fail-safe trust root

and the identical string was corrected in all five surfaces that copy it: `tests/golden/agent_protocol/{codex,claude-code,cursor}-block-stop.json` and `examples/agent-protocol/expected/{block-stop,policy-bypass}.json`. `llms-full.txt` regenerated.

`AGENTS.md:421` — **genuine human-only blocker.** Preflight routes that edit to `human_review_required` / `must_stop: true`:

```
reason: AGENTS.md matches protected surface **/AGENTS.md; a coding agent must not
        self-approve trust-root edits. A coding agent must stop and route this to a human.
permissions: {edit: false, commit: false, push: false, update_pr: false, merge: false,
              report_complete: false}
```

so I did not touch it. On the redesign-or-report choice you offered: the AGENTS.md sentence is now **incomplete, not false**, so it does not need a redesign to stay true. It claims the Tier B checks *select their changed files the same way*, and that safety property is intact — `SHIP-VERIFY-POLICY-BASE-ABSENT` selects through the same `_POLICY_SURFACES` / `touched()` path, case-insensitively and at any depth, so a case variant still cannot be a trust root in Tier A and invisible to the specialized check that carries the severity. What is stale is only the enumeration.

The exact human-owned edit, for whoever picks it up:

> **File** `AGENTS.md`, line 421 · in `the Tier B checks (\`SHIP-VERIFY-POLICY-WEAKENED\`, \`SHIP-VERIFY-CI-GATE-REMOVED\`, agent-instruction weakening, trigger-catalog drift)`, insert `` `SHIP-VERIFY-POLICY-BASE-ABSENT` `` after `` `SHIP-VERIFY-POLICY-WEAKENED` ``. Then re-run `python scripts/build-llms-full.py` (`llms-full.txt` carries a copy of this paragraph).

## Control state

Editing `docs/triggers.json` is itself a trust-root change, so this branch's own verdict moved — honestly — from `complete` to review-required. That is the correct outcome, not a regression, and it doubles as a live demonstration of the issue's AC4 (a trust-root change with no blockers still leads with the trust-root message):

```
check  --agent claude-code --format agent-boundary-json
  decision: allow · input_coverage: complete · violations: 0 · issues: 0
  control.state: agent_action_required → next_action: verify

verify --base main --head HEAD --ci-mode advisory --format control
  control_state:      review_publishable
  decision:           review_required   (decision_source: release_decision)
  execution:          succeeded         exit_code: 0
  human_review:       required=true
  permissions:        commit/push/update_pr=true · merge=false · report_complete=false
  current_control_id: sha256:32011dba8d57c9357417deb40038e4c2095b4cabb5f8b058218e505ba824247e
  contract_version:   24

report.json → review_items: 1 · blockers: 0
  SHIP-VERIFY-TRIGGER-CATALOG-DRIFT (medium) · changed_trigger_files: ["docs/triggers.json"]
  verifier_summary.policy_weakened: false
verifier.json.headline
  "This PR edits a release trust root (…); a coding agent cannot self-approve that
   change — a human must review it."
```

I followed only the current `next_action` at each step: `check` → `verify`, then `review_publishable`, which permits commit/push/PR update and denies merge. Not merged.

## Verification

- Focused: `tests/test_headline_ranking.py` (29), `tests/test_policy_reason_code_split.py` (14), plus `test_first_adoption.py`, `test_self_approval_signal.py`, `test_fix_task_contract.py`, `test_cross_block_consistency.py` — all pass.
- Full: `python -m pytest` — one failure, `test_codex_boundary_check.py::test_codex_check_boundary_json_golden_outputs`, which I confirmed reproduces **byte-identically on a clean tree** (`git stash` → same `audit_id`, same fingerprints, same `line`/`column` keys). Python 3.14's `tomllib` attaches `line`/`column` to `TOMLDecodeError`, which changes the evidence dict and therefore the fingerprint; the golden predates it. Unrelated to this PR, and I left it alone rather than regenerating an unrelated golden.
- `python -m ruff check .` — clean.
- Generated surfaces: `scripts/generate_schemas.py` and `scripts/build-llms-full.py` run twice produce no diff.

pengfei-threemoonslab (commented):
Technical verdict: **REQUEST_CHANGES**. GitHub will not allow a blocking self-review because the connected identity (`pengfei-threemoonslab`) authored this PR, so this is submitted as a COMMENT review; it should be treated as REQUEST_CHANGES.

I re-reviewed exact head `15ce3b67d41d075c2f629b23712a2366a075420e`, including the updated description, evidence comment, every reply to review #4941576705, and the `ca4d21f..15ce3b6` diff. The fix does close the direct pre-split severity-override/explicit-new precedence case, honest headline/control/fix-task routing, old-ID adoption reprojection, rationale copies, and catalog direction. However, actionable defects remain:

1. **[P1] No-base PR copy still claims a proven weakening.** In an end-to-end `base_snapshot_unavailable` run, `policy_weakened=true` and `policy_weakening_proven=false`; headline/control/fix-task copy is honest, but generated `pr-comment.md` still prints `- Policy weakened: true` because `report/pr_comment.py:151-155` reads the broad fail-closed routing flag. Human-facing PR copy must read the proven fact.
2. **[P1] The frozen strict verifier v0.8 wire contract changed in place.** A fresh artifact still declares `verifier_schema_version: "0.8"` while emitting `policy_weakening_proven`; validating it against main/ca4's published v0.8 schema fails with `Additional properties are not allowed ('policy_weakening_proven' was unexpected)`. Advance/freeze the schema version or avoid changing the v0.8 wire shape.
3. **[P1] Split-ID compatibility is incomplete for policy comparison and accepted-debt baselines.** Runtime overrides honor exact-new precedence plus old-ID umbrella fallback, but the Tier-B comparator does not, producing both false negatives and false positives for `severity_override_lowered`. Separately, an old-ID accepted-debt fingerprint no longer matches the new no-base finding: the same critical item changes from matched/`review_required` to new/`blocked`.
4. **[P1] The final headline can still exceed the 400-byte control budget and lose the exact review suffix.** A real `_build_verifier` repro with a long multibyte critical blocker, trust-root notice, and one new evidence gap emitted 443 bytes; compact `reason` and `human_review.why` truncated at 400 and no longer contained the full `a human must review it.` requirement. The post-compose provenance append bypasses the reservation; an unbounded configured-manifest suffix has the same failure.
5. **[P2] Unicode format controls remain unsanitized.** U+202E RTL OVERRIDE and U+2066 LRI survive into headline/control/PR copy and can visually reorder the reserved governance suffix; a lone surrogate raises during UTF-8 budgeting.

Focused tests (836) and lint/schema/catalog generation checks pass, and current GitHub CI is green, but those suites do not exercise these reproductions. Please add regressions for each compatibility/read path and rerun the final-output projections.

src/agents_shipgate/core/check_ids.py
```diff
@@ -31,11 +31,35 @@
...
+# where it previously got `critical` and `blocked` — a gate that quietly
+# loosened because an id moved. Configuration written against the umbrella must
+# keep reaching both halves; an override written against a half still wins,
+# because ``_severity_override_for_check`` prefers the exact id.
+#
+# Floor enforcement is unaffected: ``_resolve_metadata`` resolves the umbrella
+# id directly in the catalog, so an override against it is still validated
+# against the umbrella's own (stricter) floor.
+SPLIT_CHECK_ID_ALIASES: dict[str, tuple[str, ...]] = {
+    "SHIP-VERIFY-POLICY-WEAKENED": ("SHIP-VERIFY-POLICY-BASE-ABSENT",),
```
pengfei-threemoonslab, line 53: **[P1] Apply split compatibility to comparator and baseline consumers.** This map reaches severity mutation lookup, but `verify_policy._effective_severity` still does exact lookup only and baseline matching still hashes only the emitted ID. I reproduced (a) runtime `OLD: critical` + explicit `NEW: medium` lowering critical→medium while the comparator emits nothing, (b) removing explicit `NEW: critical` leaving runtime critical through the umbrella while the comparator falsely reports critical→medium, and (c) an old-ID accepted-debt baseline becoming new/resolved and changing a critical release decision from `review_required` to `blocked`. Both read paths need exact-new-then-umbrella semantics/legacy fingerprint candidates without re-emitting the old ID.

src/agents_shipgate/schemas/verifier.py
```diff
@@ -309,7 +309,15 @@ class VerifierCapabilityReview(BaseModel):
...
     trust_root_touched: bool = False
+    # Fail-closed routing: true whenever the release policy may have gotten
+    # weaker, including when the direction could not be established at all
+    # (no base snapshot). Consumers gate on this.
     policy_weakened: bool = False
+    # The narrower, honest fact: a base-vs-head comparison actually ran and
+    # found the head weaker. Never true without ``policy_weakened``. Copy is
+    # selected from this one, so an unprovable direction is not reported to a
+    # human as a proven weakening while the conservative route is preserved.
+    policy_weakening_proven: bool = False
```
pengfei-threemoonslab, line 320: **[P1] Do not add this property under the frozen `0.8` schema identifier.** A current artifact declares `verifier_schema_version: "0.8"` but fails validation against main/ca4's strict v0.8 schema because this is an additional property. Please advance the verifier schema and retain the prior v0.8 file, or keep the existing wire shape. Also enforce the documented implication: the model and generated schema currently accept `policy_weakening_proven=true` with `policy_weakened=false`.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -1824,6 +1851,162 @@ def _gap_provenance_note(
...
+    reads. The lead is shortened instead — and if the requirement alone fills
+    the budget, it is published on its own, which is what the headline said
+    before blockers ever led it.
+    """
+
+    if not suffix:
+        return lead
+    if not lead:
+        return suffix
+    room = MAX_ENVELOPE_PROSE_BYTES - len(suffix.encode("utf-8")) - 1
```
pengfei-threemoonslab, line 1982: **[P1] Reserve the suffix against the final composed headline, not this intermediate value.** `_build_verifier` later appends `_gap_provenance_note` without rebounding. With a long multibyte blocker + trust-root change + one new evidence gap, the raw headline is 443 UTF-8 bytes; the 400-byte compact projection drops the complete `a human must review it.` suffix from both `reason` and `human_review.why`. A long `configured_manifest` can also make `suffix` itself exceed the budget. Normalize/bound dynamic suffix inputs and run all later additions through one final reserved-budget composition.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -1824,6 +1851,162 @@ def _gap_provenance_note(
...
+            SEVERITY_ORDER.get(item.severity, 99),
+            item.check_id,
+            item.title,
+        ),
+    )
+
+
+# Every character class that would let an interpolated value stop being one
+# clause of one sentence: newlines, tabs, and the C0/C1 control range.
+_HEADLINE_CONTROL_CHARACTERS = re.compile(r"[\x00-\x1f\x7f-\x9f]")
```
pengfei-threemoonslab: **[P2] C0/C1 filtering does not cover arbitrary Unicode formatting controls.** U+202E RTL OVERRIDE and U+2066 LRI survive this regex into headline/control/PR output and can visually reorder or obscure the governance suffix; a lone surrogate also raises in `_bounded_bytes().encode()`. Sanitize unsafe Unicode categories before byte accounting and add format-control/surrogate regressions.

src/agents_shipgate/core/check_ids.py
pengfei-threemoonslab: Fixed in 1a39c13, both halves.

**(3a/3b) The comparator now resolves exactly as the runtime applier does.** Reproduced first:

```
                                                     before   after   expected
(a) base {OLD: critical}; head {OLD: critical, NEW: medium}
    runtime applies medium — comparator findings        0       1        1
(b) base {OLD: critical, NEW: critical}; head {OLD: critical}
    runtime still applies critical — comparator         1       0        0
```

`_effective_severity` was `overrides.get(check_id) or defaults.get(check_id)` — a literal-key read. It now mirrors `core.findings.mutations._severity_override_for_check` exactly: the exact id wins, then `expands_to_check_id` walks the umbrella, then the catalog default. The iteration set is widened the same way (`_comparable_check_ids`), because a configured umbrella key that no longer *names* the check it governs would otherwise drop out of the comparison entirely.

**(3c) Accepted debt survives the split.** New `legacy_split_check_id_fingerprints` in `core/findings/identity.py`, consumed by both `apply_baseline` and the resolved-not-pruned scan in `core/baseline.py`. It is deliberately narrow, per your "without weakening trust or silently accepting unrelated debt":

- offered only for a finding whose own check id is a declared split **target** (an unrelated finding, and the umbrella id itself, get an empty set);
- only for the umbrella ids that declare that target;
- still subject to the existing `support_hash` agreement and the `legacy_identity_unambiguous` guard before it can match;
- and it never re-emits or revives the old id — `test_the_split_never_re_emits_the_pre_split_id_for_a_no_base_run` still holds.

Regressions in `tests/test_policy_reason_code_split.py`:

- `test_the_comparator_sees_a_lowering_written_against_the_new_id` — your (a), asserting the target id and both severities.
- `test_the_comparator_does_not_invent_a_lowering_that_never_happened` — your (b).
- `test_a_real_lowering_on_an_unrelated_check_is_untouched` — the ordinary path is unchanged.
- `test_pre_split_accepted_debt_still_matches_and_keeps_the_verdict` — your (c), end to end: asserts the two fingerprints genuinely differ, that `baseline_status == "matched"`, that the summary is `(matched=1, new=0, resolved=0)`, and that `build_release_decision` returns **`review_required` with no blockers** rather than `blocked`.
- `test_the_legacy_fingerprint_candidate_is_scoped_to_declared_splits` — the false-acceptance guard.

src/agents_shipgate/schemas/verifier.py
pengfei-threemoonslab: Fixed in 1a39c13. You were right on both counts; reproduced first:

```
before: declared verifier_schema_version "0.8"
        vs published v0.8 → "Additional properties are not allowed
                             ('policy_weakening_proven' was unexpected)"
        VerifierCapabilityReview(policy_weakened=False,
                                 policy_weakening_proven=True) → accepted
after:  declared verifier_schema_version "0.9"
        docs/verifier-schema.v0.8.json restored to its published bytes
        same construction → ValidationError:
          "policy_weakening_proven=True requires policy_weakened=True"
```

I advanced the schema rather than dropping the field, following the repository's own precedent for this exact move (`0.7 → 0.8` in #346):

- `verifier_schema_version: Literal["0.9"]`, `"0.8"` added to the legacy read set, `_normalize_legacy_control` normalizing forward to `"0.9"`.
- `docs/verifier-schema.v0.9.json` generated; `docs/verifier-schema.v0.8.json` restored byte-for-byte from `main` and left frozen.
- Propagated: `.well-known/agents-shipgate.json` (version + schema URL), `docs/INDEX.md` (0.9 current, 0.8 listed as frozen), `docs/agent-contract-current.md`, `STABILITY.md`.
- `tests/fixtures/local-agent-contract.v9.json` deliberately **keeps** `"0.8"` — it is a real shipped v9 render, and rewriting it would make `test_local_contract_upgrades_a_real_shipped_v9_render` see a user-modified file. The live renderer emits `0.9` and `tests/test_local_contract.py` pins that.

Reading a v0.8 artifact stays lossless: the field defaults to `false`, which is exactly what "this artifact recorded no base-vs-head comparison" means, so an old artifact can never be reprojected as a proven weakening.

The implication is now enforced in the model (`_proven_implies_weakened`), not only documented — a payload cannot route as safe on `policy_weakened` while telling a human the policy was weakened.

Regressions in `tests/test_policy_reason_code_split.py`:

- `test_the_frozen_v0_8_schema_still_describes_v0_8_artifacts` — asserts the property is absent from v0.8's `$defs`, present in v0.9's, and that each schema's `verifier_schema_version` const matches its filename.
- `test_a_v0_8_artifact_still_reads_and_normalizes_forward` — takes a real `run_verify` artifact, relabels it `0.8`, strips the field, and asserts it validates, normalizes to `0.9`, and reads `policy_weakening_proven is False`.
- `test_proven_weakening_cannot_contradict_the_routing_flag` — the rejected combination plus the three supported ones.

src/agents_shipgate/cli/verify/orchestrator.py
pengfei-threemoonslab: Fixed in 1a39c13. Reproduced your case exactly — a multibyte critical title, a trust-root notice, and one gap note:

```
before: composed 398 B, then " " + provenance appended → 444 B
        truncate_prose(final) contains "a human must review it." → False
        adoption headline with a long configured_manifest → 547 B
after:  final 398 B · truncate_prose(headline) == headline
        ends with "a human must review it." → True
        adoption headline with the same manifest → 397 B, ends with the
        complete "…a separate human-review decision."
```

Three changes:

1. **One composition.** `_verifier_headline` takes a `context_note` and folds it into the *lead*; `_build_verifier` passes `_gap_provenance_note(...)` in instead of appending to the result. There is no longer any path that adds to the headline after the reservation.
2. **The suffix's one dynamic input is bounded.** `_manifest_label` normalizes and caps `configured_manifest` at 80 chars, so the adoption suffix is bounded by construction and the reservation stays satisfiable.
3. **Priority order when room runs out.** `_report_lead` gives the verdict and the named blocking cause their full room and shrinks — then drops, rather than reducing to a stub — the gap note. The reserved suffix is never touched. That ordering is deliberate: the verdict and the cause are why the headline was reordered for #365 in the first place; the provenance note is context about where the gaps came from.

Regressions in `tests/test_headline_ranking.py`:

- `test_the_final_composed_headline_fits_the_control_budget` — parametrized over an ordinary ASCII title, your multibyte one, and 7 000 chars; asserts `len(headline.encode()) <= MAX_ENVELOPE_PROSE_BYTES`, `truncate_prose(headline) == headline`, and that it still ends with `a human must review it.`
- `test_an_unbounded_configured_manifest_cannot_eat_the_adoption_suffix` — an 80-segment path, asserting the complete adoption suffix survives.
- `test_the_blocking_cause_outranks_the_gap_note_when_room_runs_out` — pins the priority order in both directions.

src/agents_shipgate/cli/verify/orchestrator.py
pengfei-threemoonslab: Fixed in 1a39c13. Reproduced first — `pay‮funds⁦now​` survived the C0/C1 regex intact; a lone surrogate raised inside `_bounded_bytes`.

`_one_clause` now filters by Unicode general category rather than by codepoint range, replacing any of `Cc`, `Cf`, `Cs`, `Co`, `Cn`, `Zl`, `Zp` with a space before whitespace collapsing:

```
before: categories surviving: {'Cf', 'Ll', 'Zs'}   U+202E present: True
        _one_clause("pay\ud800funds").encode("utf-8") → UnicodeEncodeError
after:  categories surviving: {'Ll', 'Zs'}          U+202E present: False
        result "pay funds now" · encodes cleanly
```

Categories, not a blocklist, because the property being protected is positional: `Cf` is the class that reorders *rendered* text without changing a byte, which is exactly how a tool name could visually move the reserved governance suffix out of the position the composition guarantees. `Cs` also removes the lone-surrogate crash from the byte budgeting rather than papering over it — every value reaching `_bounded_bytes` is now UTF-8 encodable by construction.

`_one_clause` is applied to every dynamic input in the line: the blocker title, the report headline, the configured-manifest label, and the context note.

Regression in `tests/test_headline_ranking.py`: `test_unicode_format_controls_never_reach_the_headline`, parametrized over U+202E RTL override, U+2066 LRI, U+200B zero-width space, a bidi pop/isolate pair, a lone surrogate, and a private-use codepoint. Each asserts no surviving character is in the unsafe categories, that the headline encodes to UTF-8, and that it still ends with the review requirement.

pengfei-threemoonslab:
All five defects reproduced and fixed in **1a39c13**. Inline replies carry the per-finding evidence; this covers **(1)**, which had no inline thread, and the verification/control state.

## (1) No-base PR copy claimed a proven weakening

Reproduced at `report/pr_comment.py:151-155`, exactly as you described:

```
policy_weakened=True · policy_weakening_proven=False
before: "- Policy weakened: `true`"
after:  "- Policy changed, weakening unproven: `true` (no base policy was
         available to compare against; routed to human review as if weakened)"
```

The line now branches on `policy_weakening_proven`, the narrower fact, via a new `_policy_change_lines` helper. The fail-closed route is untouched and still reported on the same surface — the same comment carries `Merge verdict: blocked`, `Agent may merge: false`, and the unchanged control block. I also corrected the same claim in `_trust_root_warning_lines`; note that function is currently unreferenced, so it is consistency, not a live path.

Regression: `test_the_pr_comment_reports_the_proven_fact_not_the_routing_flag` in `tests/test_headline_ranking.py` renders the comment both ways and asserts the unproven case never prints `Policy weakened: true`, the proven case still does, and the merge/route fields are identical in both.

## Verification

- **Focused:** `tests/test_headline_ranking.py` (41) and `tests/test_policy_reason_code_split.py` (22) pass, plus every suite the schema bump touched (`test_local_contract`, `test_setup_control`, `test_human_authorization`, `test_verifier_control_contract`, `test_diff_input_status`, `test_agent_mode`, `test_agent_instructions_renderers`, `test_agent_instructions_apply`).
- **Full:** `python -m pytest` — one failure, `test_codex_boundary_check.py::test_codex_check_boundary_json_golden_outputs`, which reproduces byte-identically on a clean tree (Python 3.14's `tomllib` attaches `line`/`column` to `TOMLDecodeError`, changing the evidence dict and the fingerprint; the golden predates it). Unrelated; CI's `test` job passes it.
- **Lint:** `python -m ruff check .` clean.
- **Generated surfaces:** `scripts/generate_schemas.py` and `scripts/build-llms-full.py` run twice produce no diff. `docs/verifier-schema.v0.8.json` verified byte-identical to `main`.

## Preflight and control

Preflight ran before the edits, on the full changed-file plan; it returned `agent_action_required` with `next_action: verify` and no protected-surface signal — `AGENTS.md` is not in the plan and was not touched.

```
check  --agent claude-code --format agent-boundary-json
  decision: allow · input_coverage: complete · violations: 0
  control.state: agent_action_required → next_action: verify

verify --base main --head HEAD --ci-mode advisory --format control
  control_state:      review_publishable
  decision:           review_required   (decision_source: release_decision)
  execution:          succeeded         exit_code: 0
  human_review:       required=true
  permissions:        commit/push/update_pr=true · merge=false · report_complete=false
  current_control_id: sha256:f54fa33ec43624c328d53a94945e093260346b70e5468df559f5968478663036
  contract_version:   24

verifier.json → verifier_schema_version: 0.9
                capability_review.policy_weakened: false · proven: false
                headline: 194 bytes
```

`review_publishable` because the branch edits `docs/triggers.json`, a trigger-catalog trust root — unchanged from the previous round, and still the honest outcome. I followed only the current `next_action` at each step. Not merged.

## Still human-owned

`AGENTS.md:421` — the Tier B enumeration, and now also the schema table row naming `docs/verifier-schema.v0.5.json` (stale since v0.6; this PR makes it two versions further behind). Preflight routes any `AGENTS.md` edit to `human_review_required` / `must_stop: true`, so both are left for a human. Exact edits are in the PR description.

PR #383, opened 2026-08-17
feat(cli): give the repository one launcher and doctor an environment block (#334)

### LINKED ISSUE #334: Provide a repository-local CLI launcher and environment-aware recovery
> Parent epic: #338
>
> Dependency split: #322 owns invocation capture, structured executable/arguments, and the `python -m` / `__main__.py` regression. This issue owns the repository-local launcher, environment diagnosis, and recovery UX.

## Problem

A stale or broken global console script can fail before Agents Shipgate starts, while a source-checkout invocation requires the user to discover interpreter and `PYTHONPATH` details. Contributors and coding agents need one repository-local command and a deterministic diagnosis when the selected launcher cannot import the checkout.

## Scope

- provide and document one canonical contributor launcher;
- deterministically select the checkout code and a supported interpreter;
- diagnose launcher/import/version mismatches;
- emit an exact recovery using the shared invocation policy from #322.

## Acceptance criteria

- [ ] One canonical contributor command works from a clean source checkout without a global installation.
- [ ] The launcher deterministically selects the checkout’s code and supported interpreter.
- [ ] A stale global console script produces actionable guidance rather than an unexplained traceback where feasible.
- [ ] `doctor --json` reports interpreter, launcher, import source, installed version, source-tree version, and mismatches.
- [ ] #322 is complete and launcher recovery uses its structured invocation policy.
- [ ] CI covers clean checkout, editable install, released install, stale launcher, and mismatched interpreter cases.
- [ ] Contributor and generated agent documentation use the same canonical command.

## Success measure

A user should never need to understand `PYTHONPATH` to let a coding agent run Agents Shipgate in this repository.

pengfei-threemoonslab:
Closes #334. Part of #338.

## Surface discipline

**Headline metric: activation rate / time-to-first-verdict.** An agent that cannot launch the CLI never produces a verdict; the epic's own reproduction stalled at exactly that step and a human took over the terminal. No new CLI command or sub-app, no new schema version, no new report block, no new adapter. The one new field extends a command that already exists.

## The problem

Running Shipgate from a source checkout meant either a bare `agents-shipgate`, which resolves through `PATH` and can silently execute a pipx or base-conda copy — a `0.8.0` shadowing a worktree makes new subcommands look "missing" — or `PYTHONPATH=src python -m agents_shipgate`, correct but undiscoverable. A console script promoted from an environment that no longer exists is worse than either: it dies with `ModuleNotFoundError` before a line of Shipgate runs, so nothing in Shipgate's output can explain it.

## What this adds

**`./shipgate`** — the canonical contributor command, documented in `CONTRIBUTING.md`, `AGENTS.md`, and `CLAUDE.md`. No installation, no activated virtualenv, no `PYTHONPATH`. It puts this tree's `src/` ahead of every installed copy (in child processes too) and selects an interpreter — `AGENTS_SHIPGATE_PYTHON`, else the project virtualenv, looked up in the main checkout as well so a `git worktree` shares it — re-executing exactly once under a loop guard. Standard library only, one file, not part of the wheel.

**Recovery through the #322 policy.** The launcher announces itself via `AGENTS_SHIPGATE_CLI`, so every command it prints is runnable as printed. This is not decoration: its `argv[0]` is named `shipgate`, so without the announcement the invocation policy would read it as a console script and emit `agents-shipgate …`, which a clean checkout has no way to run — #322's defect, one layer down. An operator's own override still wins.

**`environment` on `doctor --json`** — and on every `doctor` agent-mode error line, including the discovery failure where `--json` prints no payload at all. Interpreter, launcher, every Shipgate console script on `PATH` with the interpreter each shebang names, import source, installed/imported/source-tree versions, and `mismatches[]` with severities and runnable recovery commands.

Nothing runs an interpreter or executes a console script to find out. That is the trust-model invariant (`subprocess` and `os.exec*` are banned under `src/`), and it is also the only thing that could work: the environments worth diagnosing are the ones where running something is what fails. A stale wrapper is identified from its shebang.

## Two orderings that are load-bearing

- **The checkout search is caller-first.** A `git worktree` borrows the main checkout's editable install, so searching the import root first would find *that* tree, agree with itself, and report nothing — while every edit in the worktree went unexecuted. Caller-first names the worktree, sees the import outside it, and says so.
- **A source checkout out-voting an install is not a mismatch.** An editable install's metadata is captured once and lags every version bump; running a worktree shadows the install on purpose. Reporting either would put a permanent warning in front of every contributor for a state that is correct. Two *installed* copies disagreeing still is a mismatch — nothing chose that.

## Also

- `invocation.render_cli_override`, the host-rules inverse of how `AGENTS_SHIPGATE_CLI` is parsed. Needed now that something writes the variable: `join_argv` is POSIX everywhere by design, so on Windows it would round-trip a checkout path back with literal quotes in it. One renderer per parser.
- `environment_error` in `docs/errors.json` (exit 4 — the existing "other error" code, not a new one). The published catalog's contract test now scans the launcher too, since it is a real emit site and the only one that can report a Shipgate that never started.

No schema or contract version changes. `environment` is deliberately excluded from `control.input_id`: it carries machine-specific absolute paths, and `input_id` is the identity of what `doctor` decided about a *manifest*.

## Verification

All five acceptance environments are covered twice — judged from a stated `EnvironmentProbe`, and driven as real subprocesses against a synthetic checkout:

- **clean checkout** — new CI job `clean-checkout-launcher` installs the dependency closure and *not* the project, then asserts `installed_version` is null, this tree answered, and every emitted command names the launcher. That environment is one the `test` job can never have.
- **released install** — new CI step installs the built wheel into a fresh venv and runs it from outside the checkout, asserting `import_source.kind == "installed"` and no mismatches.
- **editable install, stale console script, unsupported interpreter, missing dependencies, incomplete checkout, worktree shadow** — `tests/test_environment.py` (35 tests).

Full suite green (5340 passed, 5 skipped), `ruff` clean, `generate_schemas.py --check` clean, dependency locks verified. The stale-console-script story was also reproduced by hand: the wrapper's own failure is an opaque `bad interpreter` shell error; `./shipgate doctor --json` names it as `console_script_interpreter_missing` and hands back a working entry point.

## Boundary

Everything written *into another repo* — adoption kits, `.cursorrules`, `skills/`, `.claude/commands/`, and every block `init --write --agent-instructions=…` renders — keeps saying `agents-shipgate`. Those run where the package is installed and there is no launcher, and their bytes are pinned by render hashes. `AGENTS.md` now states that boundary so a future sweep does not "helpfully" rewrite them.


shipgate
```diff
@@ -0,0 +1,424 @@
...
+        ],
+        environment=report,
+    )
+    return ENVIRONMENT_EXIT_CODE
+
+
+def _install_command() -> str:
+    from agents_shipgate.invocation import join_argv
+
+    return join_argv([sys.executable, "-m", "pip", "install", "-e", str(CHECKOUT)])
```
pengfei-threemoonslab: [P1] Preserve quoted Python recovery commands

`_install_command()` returns a POSIX-quoted non-Shipgate command, but `emit_agent_mode_error()` normalizes it through `retarget_command()`, whose `_program_token_span()` scans only to raw whitespace. For a plausible clone such as `/tmp/agents-shipgate worktree/.venv/bin/python`, the first raw span is `'/tmp/agents-shipgate`; `_program_name()` sees `agents-shipgate` and replaces it with the launcher, producing a broken command like `'/tmp/agents-shipgate worktree/shipgate' worktree/.venv/bin/python' -m pip ...`. I reproduced this through `AGENTS_SHIPGATE_PYTHON`; both `next_action` and `next_actions[0].command` were corrupted. Since runnable recovery commands are the core contract here, parse the quoted program token before deciding it is a Shipgate command (or explicitly bypass Shipgate retargeting for this action), and add a regression test with an `agents-shipgate …` path containing spaces.

tests/test_environment.py
```diff
@@ -0,0 +1,745 @@
...
+    environment = _bare_environment(tmp_path)
+    environment["AGENTS_SHIPGATE_PYTHON"] = str(isolated_interpreter)
+
+    result = _run_launcher(root, "doctor", "--json", env=environment)
+
+    assert result.returncode == 4
+    assert "Traceback" not in result.stderr
+    payload = _agent_mode_line(result.stderr)
+    assert payload["error"] == "environment_error"
+    assert payload["next_actions"][0]["executable"] == [str(isolated_interpreter)]
```
pengfei-threemoonslab: [P2] Exercise the recovery in the no-pip environment

This fixture deliberately creates the selected interpreter with `venv --without-pip`, but the test only checks the emitted tokens. Running that exact `next_actions[0]` fails immediately with `No module named pip`, so the promised recovery is not runnable in the environment being tested. Please either detect/route the missing-pip case (for example via an `ensurepip` or fresh-venv step) or explicitly narrow the supported contract, and make the test execute the proposed recovery far enough to prove it can start.

src/agents_shipgate/environment.py
```diff
@@ -0,0 +1,549 @@
...
+    line = head.split(b"\n", 1)[0]
+    try:
+        text = line.decode("utf-8").strip()
+    except UnicodeDecodeError:
+        return None
+    tokens = text[2:].strip().split()
+    if not tokens:
+        return None
+    interpreter = tokens[0]
+    if Path(interpreter).name == "env":
```
pengfei-threemoonslab: [P2] Recognize pip’s POSIX shell trampoline

When an interpreter path contains spaces, pip generates console scripts beginning with `#!/bin/sh` and then `exec "/path with space/bin/python" "$0" "$@"`. This code records `/bin/sh` as the interpreter, emits `console_script_runs_other_interpreter` for a valid current install (twice, for the aliases), and cannot detect when the underlying Python in the `exec` line has gone stale. I reproduced it by installing the built wheel into a venv whose path contains spaces. Parse that standard trampoline or treat shell shebangs as unknown, and cover the generated wrapper shape in the environment tests.

pengfei-threemoonslab (commented):
Blocking findings: I found three reproducible correctness gaps in the launcher’s core promises. One recovery command is corrupted by invocation retargeting for a plausible checkout path containing spaces, the missing-dependency recovery assumes `pip` even in the no-pip environment used by its own test, and pip’s POSIX shell trampoline is misclassified as the actual Python interpreter. All GitHub checks are green, the focused launcher/public-contract tests and packaging tests pass locally, and the Shipgate verifier reports `review_required` only for the intentional trust-root edits (no blockers). The overall direction is solid, but these failure-path issues should be fixed before merge because this PR specifically promises accurate diagnosis and runnable recovery when the environment is broken.

GitHub does not permit the PR author’s account to submit a formal `REQUEST_CHANGES` review, so this is posted as a comment review; please treat the P1 finding as blocking.

shipgate
pengfei-threemoonslab: Fixed in a4f05a3d, and you were right that the corruption is the serious half — I reproduced it and it is slightly worse than described: the resulting string carries a dangling quote, so `split_invocation` returns `None` and the action also loses its `executable`/`args` pair. A runnable command that runs the wrong program, with the structured form silently withdrawn.

I fixed it in `retarget_command` rather than by bypassing retargeting for this action. Bypassing would have made the launcher a call site that opts out of the wire-level policy, which is the pattern #322 spent a round removing — and it would leave the same trap armed for the next emitter that renders a quoted path.

`_program_token` now finds the span with quoting honoured and takes the token's **value** from `shlex`, the same grammar the string was rendered with, then cross-checks the two: if `shlex.split(command[start:end])` is not exactly the token it thought it had, it returns `None` and the command is left alone. Locating a token and reading it are different jobs, and only the second decides whether the command is ours to rewrite.

The old scan's defence — that our console-script names contain no whitespace or quotes — was true of the *names* and irrelevant to the strings they appear in. That is now written down where the assumption used to be.

Regression tests at both levels:

- `tests/test_invocation_policy.py::test_a_quoted_path_is_read_before_it_is_judged_to_be_ours` — your exact `/tmp/agents-shipgate worktree/.venv/bin/python` path.
- `::test_a_quoted_shipgate_path_is_still_retargeted` — the converse, since reading the token properly must not mean leaving quoted tokens alone; `'/opt/my tools/agents-shipgate'` is now retargeted where it previously was not.
- `::test_the_program_token_span_agrees_with_the_grammar` — parametrized, asserts the span and the grammar agree on the value *and* on the remainder.
- `tests/test_environment.py` — the isolated-interpreter fixture now lives at `<...>/agents-shipgate worktree/.venv/`, so the missing-dependency test is also the end-to-end regression: every emitted action must name the interpreter, and `next_action` must not start with the launcher path.

tests/test_environment.py
pengfei-threemoonslab: Fixed in a4f05a3d — I took the "detect and route" option rather than narrowing the contract, because a fresh `venv --without-pip` is not an exotic environment and the ranked-actions contract already exists for exactly this shape.

`_recovery_actions()` now returns a ranked list: when `pip` is absent and `ensurepip` is present, rank 1 is `<interpreter> -m ensurepip --upgrade` and rank 2 is the editable install. Both prerequisites are answered by asking the *selected* interpreter about itself with `importlib.util.find_spec` — this code is already running inside it after the re-exec, so nothing has to be spawned. An interpreter with neither (some distribution packages strip `ensurepip`) now gets the diagnosis and **no command at all**, plus a message saying why; emitting a command that cannot run is the failure mode this PR is about, so silence is the honest answer there.

On making the test run it: `test_the_emitted_recovery_runs_in_the_environment_it_is_emitted_for` asserts the fixture really has no pip, executes rank 1 for real (~1.3s, offline), asserts it exits 0, then asserts rank 2's program resolves via `python -m pip --version`. I deliberately stopped short of running the editable install itself — that resolves this project's dependencies from the network, which would make the test slow and flaky, and the claim being verified is that the emitted sequence *can start* in the environment under test.

Also added `::test_the_recovery_is_ranked_by_what_the_interpreter_can_actually_do`, which drives `_recovery_actions()` directly over all three availability combinations, and `::test_the_launcher_asks_this_interpreter_about_its_own_modules`.

src/agents_shipgate/environment.py
pengfei-threemoonslab: Fixed in a4f05a3d. Reproduced exactly as you describe — installing the built wheel into a venv whose path contains a space produced two `console_script_runs_other_interpreter` warnings against a perfectly healthy install, one per alias.

I took the "parse the trampoline" option rather than "treat shell shebangs as unknown". Both remove the false positive, but only parsing preserves the detection that matters most here: with `/bin/sh` recorded, a *stale* underlying interpreter is invisible, which would have left `console_script_interpreter_missing` unable to fire for exactly the installs most likely to hit it.

`_shebang_interpreter` now checks whether the shebang program is a shell and, if so, reads the `exec` target from the head of the file — quoted or bare, tolerating the trailing quote from the polyglot heredoc. A handoff it does not recognise (`exec "$0.real"`, or no `exec` line) reports `null`, your fallback, rather than the shell: `/bin/sh` is never the answer to "which Python runs this", and reporting it would be a fact about the wrapper dressed up as a fact about the install. The read budget went 512 → 1024 bytes, since the `exec` line carries a full absolute path rather than a short shebang.

Three tests, all built on `_trampoline()`, which reproduces `pip`'s generated shape verbatim rather than approximating it:

- `::test_a_shell_trampoline_names_the_interpreter_it_execs` — the false positive, now clean.
- `::test_a_stale_shell_trampoline_is_caught_like_any_other` — the detection that was previously impossible.
- `::test_an_unrecognised_shell_wrapper_is_unknown_rather_than_wrong`.

Verified end to end against the same spaced-path wheel install: `mismatches: []`, and the wrapper's `interpreter` is the venv's real `python`.

shipgate
```diff
@@ -0,0 +1,475 @@
+#!/usr/bin/env python3
```
pengfei-threemoonslab, line 1: [P1] Make the repository launcher invocable on Windows

This extensionless shebang script is directly runnable on POSIX, but Windows PowerShell/CreateProcess does not use this shebang to execute `./shipgate` (or `.\shipgate`) through Python. That conflicts with making this the canonical command in `CONTRIBUTING.md`, `AGENTS.md`, and `CLAUDE.md` for an OS-independent package, and it also leaves `_announce_launcher()` publishing the same non-runnable path in recovery commands. The new integration job is Ubuntu-only, so the `os.name == "nt"` re-exec branch never proves the entry point can be started. Please provide a Windows-invocable shim/spelling (or use `python shipgate` there), make the announced argv match it, and cover it on Windows.

src/agents_shipgate/environment.py
```diff
@@ -0,0 +1,588 @@
...
+        return None
+    target = next(group for group in match.groups() if group is not None)
+    # ``exec "$0"`` and friends are the wrapper re-entering itself, not an
+    # interpreter path.
+    return target if target.startswith(("/", "\\")) or ":" in target else None
+
+
+def _same_file(left: str, right: str) -> bool:
+    try:
+        return Path(left).resolve() == Path(right).resolve()
```
pengfei-threemoonslab: [P1] Keep virtual environments distinct when comparing interpreters

Resolving both paths collapses a POSIX virtualenv's `venv/bin/python` symlink onto the base interpreter binary. If `doctor` is running under the base Python while the `agents-shipgate` wrapper on PATH names that venv Python, this reports `runs_this_interpreter: true` and emits no mismatch, even though launching the wrapper selects a different `sys.prefix` and site-packages. I reproduced this with a real `venv --without-pip`; both paths resolve to the same binary and the report is clean. Compare normalized path spellings without dereferencing (as the launcher's `_same_interpreter()` already deliberately does) and add a real-venv regression.

src/agents_shipgate/environment.py
```diff
@@ -0,0 +1,588 @@
...
+    would make the five environment cases untestable without mutating the
+    process running the tests.
+    """
+
+    suffixes = ("", ".exe", ".bat", ".cmd") if os.name == "nt" else ("",)
+    for entry in path_entries:
+        for suffix in suffixes:
+            candidate = Path(entry) / f"{name}{suffix}"
+            try:
+                if candidate.is_file():
```
pengfei-threemoonslab: [P2] Match executable PATH lookup semantics

`is_file()` accepts a regular file that lacks execute permission. POSIX command lookup skips such an entry and can run an executable with the same name from a later PATH directory, but this helper stops at the non-executable file, describes the wrong wrapper, and can hide a stale/different-interpreter diagnostic for the command the shell actually runs. I reproduced `command -v agents-shipgate` selecting the second entry while this function selected the first. Filter by executability on POSIX (with appropriate Windows/PATHEXT handling) and cover a non-executable first hit followed by an executable one.

pengfei-threemoonslab (commented):
Second-pass result: the three findings from my first review are fixed and independently verified at `a4f05a3d` (quoted recovery commands, no-`pip` recovery ordering, and pip's spaced-path shell trampoline). This pass found four remaining correctness gaps in the environment contract, including two common-path P1s: the canonical launcher has no directly invocable Windows spelling, and interpreter comparison collapses POSIX virtualenvs into their base interpreter. The other two findings can misidentify which PATH entry or trampoline target actually runs.

All current GitHub checks are green. Locally, the focused environment/invocation/public-contract suite, Ruff, schema generation, dependency-lock validation, static-only checks, and packaging tests pass; the aggregate suite's only sandboxed errors were the same packaging build attempting a network fetch, and that packaging module passed when rerun with dependency access. Agents Shipgate reports `review_publishable` / `review_required` solely because this PR changes protected CI and agent-instruction trust roots, with no Shipgate blockers.

GitHub does not allow the PR author's connected account to submit a formal `REQUEST_CHANGES` review, so this is posted as a comment review; please treat the P1 findings as blocking.

src/agents_shipgate/environment.py
```diff
@@ -0,0 +1,588 @@
...
+        return None
+    if name in _SHELLS:
+        return _trampoline_target(text)
+    return interpreter
+
+
+def _trampoline_target(text: str) -> str | None:
+    """The interpreter a shell wrapper ``exec``s, or ``None`` if unrecognised."""
+
+    match = _EXEC_TARGET.search(text)
```
pengfei-threemoonslab: [P2] Restrict trampoline matching to executable shell lines

Searching the entire wrapper head lets an `exec` mention in a comment or printed string win over the real handoff. For example, a valid wrapper containing `# old target: exec "/deleted/python"` before `exec "/working/python" "$0" "$@"` is reported as `console_script_interpreter_missing`, even though it runs the working interpreter. I reproduced that through `describe_environment()`. Parse candidate command lines (at minimum excluding comments/non-command contexts) and add a regression so diagnostic prose cannot become the interpreter target.

shipgate
pengfei-threemoonslab: Fixed in 7b3ef0b4. You are right that this is the launcher's own defect relocated: announcing a path the OS will not start publishes exactly the unrunnable recovery command the whole change exists to remove.

`launcher_argv()` now announces `<interpreter> <launcher>` wherever the file cannot be started on its own, and `python shipgate ...` is documented as the Windows spelling in `CONTRIBUTING.md`, `AGENTS.md`, and `CLAUDE.md` — so the emitted command and the documented command are the same command. The condition is not `os.name == "nt"` alone; it is "the OS will not run this file", which also covers an archive extracted without modes or a checkout on a `noexec` mount. One rule, two causes.

I deliberately did **not** add a `.cmd` shim. `CreateProcess` runs a batch file by handing it to `cmd.exe`, which re-parses the arguments under a different quoting grammar — and this project publishes structured argv that callers execute without a shell (`docs/diagnostics.md` tells agents to prefer `[*executable, *args]`). Two tokens are cheaper than a second grammar, and safer given the batch-argument-escaping class of issue. If you would rather have the shim for ergonomics, say so and I will add it as a convenience that does not participate in the announcement.

New `windows-launcher` CI job on `windows-latest`, in three parts:

1. `python shipgate --version` — the entry point starts.
2. `python shipgate doctor --json` — asserts `environment.launcher.executable == [sys.executable, <launcher>]`, that `import_source.root` is this checkout's `src`, that `mismatches` is empty, and that `next_actions[0].executable` carries the same two tokens.
3. The `os.name == "nt"` re-execution branch, via a `--system-site-packages` venv and `AGENTS_SHIPGATE_PYTHON`: asserts the selected interpreter is what answered, **and** that `doctor`'s exit 2 on a missing manifest survives the hop — status forwarding is precisely what a branch that spawns instead of replacing the process can lose.

It installs the same hash-locked `constraints/dev.txt` as the other jobs; the lock is universal (it carries the Windows wheels and the `os_name == 'nt'` markers), so this is the environment CI approves everywhere rather than a Windows-only resolve.

Unit coverage without Windows: `test_a_launcher_the_os_will_not_start_is_announced_with_its_interpreter`, parametrized over the faked platform and a non-executable copy, plus `test_the_launcher_announces_what_it_says_it_announces`, which asserts the value survives the round trip through the environment variable rather than trusting `launcher_argv()` alone.

src/agents_shipgate/environment.py
pengfei-threemoonslab: Fixed in 7b3ef0b4, and this one is worth naming precisely because I had already written the correct rule — in the launcher's `_same_interpreter()`, with a docstring explaining exactly this trap — and then wrote the opposite one in `environment.py`. A second implementation of a rule drifts; that is the bug class this PR has now hit twice.

Reproduced on this machine: the project's `.venv/bin/python` and a fresh `venv --without-pip` both resolve to the same uv-managed CPython binary, so `resolve() == resolve()` was `True` for two environments with different `sys.prefix` and different `site-packages`.

`_same_file` is gone, replaced by a public `same_interpreter()` that compares `os.path.normcase(os.path.abspath(...))` and never dereferences. Its docstring states the consequence in both directions, since the same rule has opposite failure modes in the two callers: resolving here reports a wrapper for another environment as clean, and resolving in the launcher skips the interpreter switch that would have supplied the dependencies.

On sharing rather than duplicating: the launcher genuinely cannot import this function. Interpreter selection runs before the checkout is on `sys.path` and before the version gate — and it has to work on an interpreter too old to parse `environment.py` at all, since rejecting that interpreter is its job. So the copy stays, and `test_the_launcher_and_the_package_compare_interpreters_the_same_way` pins the two against a shared table. That is the same duplicate-plus-parity pattern already used for the generated Claude Code hook's command-line splitter.

Regressions:

- `test_two_virtualenvs_over_one_base_are_not_the_same_interpreter` — the real-venv one you asked for. It first asserts the trap is present on this platform (`resolve()` collapses the two) and skips otherwise, so it can never pass vacuously; Windows copies the interpreter into a venv rather than linking it.
- `test_a_wrapper_for_another_environment_is_reported_as_one` — the same trap through `describe_environment`, asserting `runs_this_interpreter: False` and the resulting warning. It uses a symlink rather than the venv fixture, because that fixture's path contains a space and a shebang cannot carry one — the kernel splits the interpreter at the first space, which is why `pip` writes a trampoline for those installs in the first place.

Verified end to end too: a wrapper naming one venv, `doctor` running under another, now reports `console_script_runs_other_interpreter`.

src/agents_shipgate/environment.py
pengfei-threemoonslab: Fixed in 7b3ef0b4. The helper was answering "does a file exist there", which is not the question — the question is which entry the caller's shell would run, and a file without the execute bit is one POSIX lookup walks past.

`_which` now takes the probe and requires executability: `os.access(X_OK)` on POSIX, and on Windows the extension instead, since there is no execute bit there and `PATHEXT` is the whole rule. The extensions come from the probed environment (new `EnvironmentProbe.path_extensions`, filled by `probe_environment` and falling back to the stock `.COM;.EXE;.BAT;.CMD` when the variable is absent), so the lookup stays a pure function of stated facts rather than reading the live environment — which is why `shutil.which` was avoided in the first place.

Tests:

- `test_a_non_executable_entry_does_not_shadow_the_one_the_shell_runs` — your case, and it asserts the consequence rather than just the path: the second entry is stale, so the previously-hidden `console_script_interpreter_missing` now fires.
- `test_the_lookup_agrees_with_the_shell` — cross-checks the selection against `command -v` in `/bin/sh` on that same layout, so the claim is verified against the shell rather than against my reading of it.
- `test_windows_lookup_follows_pathext` and `test_pathext_comes_from_the_environment_that_was_probed`.

One knock-on worth flagging: several existing fixtures wrote console scripts without `chmod`, so they were exactly the non-executable files the lookup is now required to skip, and they failed. That is the fix working — they were not console scripts. There is now a `_wrapper()` helper that writes and chmods, and it says why in its docstring so the mode does not get dropped again.

src/agents_shipgate/environment.py
pengfei-threemoonslab: Fixed in 7b3ef0b4. Agreed on the principle, and it is the sharper statement of the bug: a diagnostic may never be derived from a string the shell does not execute. Deriving `console_script_interpreter_missing` from a comment is worse than missing the detection, because it accuses a wrapper that runs fine.

`_trampoline_target` now reads the head line by line, skipping blanks and comments, and requires `exec` in **command position** — the pattern is anchored with `re.match` against the stripped line. That is not the same as "the line starts with `exec`": `pip`'s trampoline writes `'''exec' "/path" "$0" "$@"`, where the leading quotes are what make the same bytes valid Python and the shell concatenates them away. So quotes around the word are allowed and nothing else before it is. The first matching line wins; an unrecognised handoff still reports `null` rather than the shell.

Tests:

- `test_a_commented_out_target_does_not_become_the_interpreter` — your exact shape, `# old target: exec "/deleted/python"` above a working `exec`, asserting the working interpreter is reported and `mismatches` is empty.
- `test_a_mention_of_exec_inside_the_script_body_is_not_a_handoff` — a `printf` whose *argument* contains `exec "/not/an/interpreter"`, which the old whole-file search would also have taken.

Both build on a `_trampoline()` helper that reproduces `pip`'s generated bytes verbatim rather than approximating them, since the whole point of the parsing is that it matches what is actually on disk.

PR #390, opened 2026-08-19
fix(google-adk): reach sub-agent tools and name every unreached one (#385)

### LINKED ISSUE #385: ADK sub-agent tools are excluded from the analyzed surface with no evidence gap, and agent_bindings.declarations cannot resolve them
## Impact

On a multi-agent Google ADK app, **every tool owned by a sub-agent is excluded from the
analyzed surface, and no evidence gap names it.** There is currently no working way for a
user to bring that surface into scope.

Discovered while dry-running shipgate on a real third-party PR:
[google/adk-samples#1745](https://github.com/google/adk-samples/pull/1745) — a Salesforce +
SAP S/4HANA "Quote-to-Cash" agent. Its shape is the common ADK pattern: a root coordinator
with `sub_agents=[salesforce_agent, sap_agent]`.

The excluded half is the half a release gate exists to judge:

| Reached (root agent's own tools) | **Not reached (sub-agent tools)** |
|---|---|
| `list_messages`, `get_user_identity` | `get_salesforce_opportunities`, `get_opportunity_line_items` |
| `send_email`, `get_manager_email` | **`create_salesforce_quote`** |
| `map_salesforce_account_to_sap_bp` | **`update_opportunity_status`** (sets deals to `Closed Won`) |
| `map_salesforce_product_to_sap_material` | **`create_sap_sales_order`** |

All three write tools are on the right-hand side. A user who closes every gap the tool
names still never learns these tools exist.

## Severity: P1

This is a **silent** coverage hole, not a reported one. Of 25 entries in
`release_decision.evidence_coverage.evidence_gaps`, **zero** name a sub-agent tool. They are
also absent from `suggested-inventory.json` and `suggested-declarations.yaml`. The only
trace is the ratio `6/12 catalog tools reachable`.

For a product positioned as a deterministic verifier, "we didn't look at the write tools and
didn't say so" is a credibility risk, not just a coverage metric.

## Reproduction

CLI `0.16.0b7`, contract `24`. No git checkout required.

```bash
git clone https://github.com/google/adk-samples.git && cd adk-samples
git fetch origin pull/1745/head:pr-1745 && git checkout pr-1745
cd python/agents/smart_closer
agents-shipgate init --workspace . --write
# declare the root so the graph resolves at all:
cat >> shipgate.yaml <<'YAML'
agent_bindings:
  root:
    source_id: adk_agent
    object: root_agent
YAML
```

Observed: `6/12 catalog tools reachable`, and no evidence gap referencing any sub-agent tool.

## The documented lever does not work

`agent_bindings.declarations` is the apparent remedy. It fails:

```yaml
agent_bindings:
  declarations:
    - agent: SalesforceAgent
      complete: true
      tools:
        - tool: create_salesforce_quote
      reason: Declared from agent.py salesforce_agent LlmAgent tools list.
```

```
Insufficient evidence: an agent binding target does not resolve (agent_v1:010e7b61...)
why: "Declaration references unresolved agent 'SalesforceAgent'."
accepted_values: ["exact_agent_object", "exact_handoff_target"]
```

Both spellings were tried — the declared name (`SalesforceAgent`) and the Python object
(`salesforce_agent`). **Both are rejected.**

**This is the core defect:** the scan's own `binding_surface_facts.agents` already contains
those names. It lists **7 agent nodes for 3 actual agents**:

```
root_agent | SalesforceAgent | SapAgent | SalesforceAgent | SapAgent | salesforce_agent | sap_agent
```

`SalesforceAgent` and `SapAgent` each appear **twice**, and the variable spellings appear as
separate nodes. A plausible root cause is that the duplicate nodes make every sub-agent name
ambiguous, so no declaration can ever resolve. The resolver is rejecting names that the
scanner itself emitted.

## Expected behavior

1. Reachability traverses `sub_agents=[...]` from the declared root, **or**
2. If traversal must stay conservative, an evidence gap is raised **per unreached tool**, so
   the surface is visibly excluded rather than silently absent, **and**
3. `agent_bindings.declarations[].agent` resolves any agent name present in
   `binding_surface_facts.agents`.

## Acceptance criteria

- [ ] Scanning PR #1745's `smart_closer` reports either `12/12` reachable, or `6/12` **plus 6
      evidence gaps naming the unreached tools**.
- [ ] `binding_surface_facts.agents` contains one node per agent (3 here, not 7).
- [ ] A declaration naming any agent in `binding_surface_facts.agents` resolves.
- [ ] Regression test: a fixture with `sub_agents=[...]` asserts no tool is both
      in-catalog and gap-free while unreachable.

## Notes

The "no silent exclusion" property in criterion 2 is the one worth holding even if full
traversal is deferred. Being told the gate did not look is materially different from being
told nothing.

pengfei-threemoonslab:
## Summary

Closes #385.

On the canonical Google ADK multi-agent shape — a coordinator with `sub_agents=[salesforce_agent, sap_agent]` — every tool the sub-agents owned fell out of the root-reachable graph, and none of the 25 evidence gaps named one. On [google/adk-samples#1745](https://github.com/google/adk-samples/pull/1745) the excluded half was the half a release gate exists to judge: three financial writes, including one that sets opportunities to `Closed Won` and one that creates an SAP sales order.

Three independent defects produced the one symptom. Each was reproduced against that PR before being fixed.

**1. `sub_agents` was read by the wrong name.** ADK routes a handoff by the sub-agent's `name=`, but `sub_agents=[…]` spells the Python variable the agent was assigned to. Reading the variable as an agent name produced one phantom node per sub-agent, owning no tools, so the handoff landed on a node with nothing behind it and the real agent stayed unreachable — this is also where the duplicate nodes the issue reported came from. The two spellings are now reconciled from the module's own `Agent(...)` assignments. An element that cannot be named — built inline, imported from elsewhere, or behind a non-literal `sub_agents` value — now fails closed as partial evidence; naming two of three sub-agents previously reported the two as the whole handoff set.

**2. `agent_bindings.declarations` invalidated its own remedy.** Declaring an agent seeded a synthetic node with no `source_id`. For an agent the scanner had *already observed*, that second node landed under a different dedupe key and made the name ambiguous, so `_resolve_agent` rejected names its own scan had emitted. Declarations now reuse the observed node, and a genuinely ambiguous name reports how many agents share it and which sources they came from instead of reading as a spelling mistake.

**3. Unreached tools were unmentioned.** A tool bound to an agent the configured root cannot reach now gets an evidence gap naming it. Everything downstream of the binding graph — findings, the action surface, semantic coverage, `tool_inventory` — is narrowed to root-reachable tools, so such a tool is never judged; before this it was not mentioned either, leaving the ratio `6/12 catalog tools reachable` as its only trace.

### Result on the reported repository

| | before | after |
|---|---|---|
| catalog tools reachable | 6/12 | **12/12** |
| agent nodes for 3 agents | 5 (7 with the issue's config) | **3** |
| `create_salesforce_quote` / `update_opportunity_status` / `create_sap_sales_order` | absent from findings and gaps | judged, findings emitted |

The issue's example declaration now resolves. It then reports `conflicting_binding_evidence` — a true positive, because `SalesforceAgent` really holds 6 tools, not the 1 the example declared.

## Type

- [x] Input adapter change
- [x] Report, schema, or SARIF output — new `evidence_gaps` rows only; no schema change (see below)
- [x] Documentation only — `docs/manifest-v0.1.md` gains the agent-name vocabulary for `agent_bindings`

## Reviewer notes

**The judgment call worth your review.** The new gap fires **only for a tool carrying a structural edge** to an agent the root cannot reach — a hole in the graph *we* built: the scan proved some agent owns the tool and then failed to connect that agent to the root. A catalog tool with **no** edge at all is deliberately left ungated, because catalog membership is not evidence of capability: `samples/large_multi_framework_agent` has **58 of 63** declared spec operations correctly out of scope, and gating them would make declaring an OpenAPI spec or MCP server self-blocking. Both existing samples with `unbound_tools > 0` turned out to be the no-edge kind — that is how the boundary was chosen rather than guessed, and the rationale is in the `_unreached_tool_gaps` docstring.

**No schema bump.** The new rows reuse the existing `missing_binding_evidence` kind, whose meaning ("the agent's tool bindings are unproven") already matches exactly. `EvidenceGap.kind` is a frozen `Literal`, so a new member would have forced a report-schema bump for no added meaning.

**Verified and deliberately left out of scope.** Both were reproduced; both now fail *closed* and are visible rather than silent, so neither blocks this fix:

- **Cross-module `sub_agents`.** `from subs.salesforce import salesforce_agent` resolves to a dotted name no per-file map holds, so the phantom node persists and the sub-agent's tool stays unbound — but it is now named in a gap (`create_salesforce_quote … is bound to SalesforceAgent, which the configured root agent does not reach`). Closing it needs cross-source variable→name matching, which is its own piece of work. This is the branch #385 explicitly allows: "`6/12` **plus** evidence gaps naming the unreached tools".
- **ADK Agent Config (YAML) binds nothing.** A `root.yaml` with `sub_agents[].config_path` reports **0/2 reachable** — including the root's own tool — because that path never sets `adk_agent_name`. It falls back to the graph-level `missing_binding_evidence`. Separately, config agent records carry no `source_id`, so a `root: {source_id: …}` selector can never match a YAML-config agent. Worth its own issue.

## Verification

Additional local checks run:

- Full suite green, **no golden churn** — an earlier, broader version of the gap rule flipped `samples/large_multi_framework_agent` and the `support_refund_agent` packet golden; that is what surfaced the boundary above.
- `ruff check .` and `compileall -q src tests` clean.
- End-to-end against a real checkout of google/adk-samples#1745 for the before/after table, plus purpose-built multi-file and Agent-Config-YAML workspaces for the two out-of-scope findings.
- Four regression tests in `tests/test_google_adk.py`, including issue #385's acceptance criterion 4: a `sub_agents=[…]` fixture asserting no tool is both in-catalog and gap-free while unreachable.

## Release-readiness notes

- [x] No user-code import added to default scan paths — the fixtures keep the module-level `raise RuntimeError("this file must never be imported")` guard
- [x] Report/schema changes are additive or documented in `STABILITY.md` — no schema change; new rows reuse the existing `missing_binding_evidence` gap kind

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,11 +17,17 @@
   agent name produced one phantom node per sub-agent, owning no tools, so the
   handoff landed on a node with nothing behind it and the real agent stayed
   unreachable. The two spellings are now reconciled from the module's own
-  assignments, which also collapses the duplicate nodes the graph used to
-  report. An element that cannot be named — built inline, imported from
-  elsewhere, or behind a non-literal `sub_agents` value — now fails closed as
-  partial evidence; naming two of three sub-agents previously reported the two
-  as the whole handoff set.
+  assignments, resolved innermost-out through the enclosing scopes so that two
+  factories reusing one local name cannot cross their sub-agents. This also
+  collapses the duplicate nodes the graph used to report.
+
+  An element that cannot be resolved to an agent definition — built inline,
+  imported from another module, behind a non-literal `sub_agents` value, or
+  rebound ambiguously within one scope — now fails closed as partial evidence
+  naming the spelling. It no longer becomes a node of its own: an empty tool
+  set on a node named after an import reads as proof the sub-agent has no
+  capability, which is the opposite of what is known about it. Naming two of
+  three sub-agents likewise no longer reports the two as the whole handoff set.
 
   `agent_bindings.declarations` was the documented remedy and could not work.
   Declaring an agent seeded a synthetic node for it, and for an agent the
@@ -35,10 +41,10 @@
   to root-reachable tools, so such a tool is never judged; before this it was
   not mentioned either, leaving the ratio `6/12 catalog tools reachable` as its
   only trace. This covers tools whose owning agent the scan identified — a hole
-  in the graph shipgate built. A catalog entry that no agent binds at all is a
-  different claim and is unchanged: catalog membership is deliberately not
-  evidence of capability, so declaring an OpenAPI spec or MCP server does not
-  become self-blocking.
+  in the graph Agents Shipgate built. A catalog entry that no agent binds at
+  all is a different claim and is unchanged: catalog membership is deliberately
+  not evidence of capability, so declaring an OpenAPI spec or MCP server does
+  not become self-blocking.
 
 - **One command runs Agents Shipgate from this checkout, and `doctor` now says
   which Shipgate answered.** Running the CLI from a source tree meant either a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,44 @@
 
 ## Unreleased
 
+- **A Google ADK sub-agent's tools are part of the analyzed surface, and a tool
+  the gate did not look at can no longer go unmentioned.** On the canonical ADK
+  multi-agent shape — a coordinator with `sub_agents=[salesforce_agent,
+  sap_agent]` — every tool the sub-agents owned fell out of the root-reachable
+  graph, and none of the 25 evidence gaps named one. On the reported repository
+  the excluded half was the half a release gate exists to judge: three financial
+  writes, including one that sets opportunities to `Closed Won` and one that
+  creates an SAP sales order
+  ([#385](https://github.com/ThreeMoonsLab/agents-shipgate/issues/385)).
+
+  ADK routes a handoff by the sub-agent's `name=`, but `sub_agents=[…]` spells
+  the Python variable the agent was assigned to. Reading the variable as an
+  agent name produced one phantom node per sub-agent, owning no tools, so the
+  handoff landed on a node with nothing behind it and the real agent stayed
+  unreachable. The two spellings are now reconciled from the module's own
+  assignments, which also collapses the duplicate nodes the graph used to
+  report. An element that cannot be named — built inline, imported from
+  elsewhere, or behind a non-literal `sub_agents` value — now fails closed as
+  partial evidence; naming two of three sub-agents previously reported the two
+  as the whole handoff set.
+
+  `agent_bindings.declarations` was the documented remedy and could not work.
+  Declaring an agent seeded a synthetic node for it, and for an agent the
+  scanner had already observed that second node made the name ambiguous, so the
+  resolver rejected names its own scan had emitted. Declarations now reuse the
+  observed node, and a genuinely ambiguous name says how many agents share it
+  and which sources they came from instead of reporting the name as unknown.
+
+  Finally, a tool bound to an agent the configured root cannot reach now gets an
+  evidence gap naming it. Everything downstream of the binding graph is narrowed
+  to root-reachable tools, so such a tool is never judged; before this it was
+  not mentioned either, leaving the ratio `6/12 catalog tools reachable` as its
+  only trace. This covers tools whose owning agent the scan identified — a hole
+  in the graph shipgate built. A catalog entry that no agent binds at all is a
+  different claim and is unchanged: catalog membership is deliberately not
+  evidence of capability, so declaring an OpenAPI spec or MCP server does not
+  become self-blocking.
+
 - **One command runs Agents Shipgate from this checkout, and `doctor` now says
   which Shipgate answered.** Running the CLI from a source tree meant either a
   bare `agents-shipgate`, which resolves through `PATH` and can silently execute
--- a/src/agents_shipgate/inputs/google_adk.py
+++ b/src/agents_shipgate/inputs/google_adk.py
@@ -73,6 +73,16 @@ CALLBACK_KEYS = {
 OPENAPI_PATH_KEYS = {"spec_path", "path", "spec_file", "openapi_path", "openapi_spec"}
 MCP_INVENTORY_KEYS = {"inventory_path", "tool_inventory_path", "mcp_tools_path", "mcp_inventory"}
 EVAL_PATH_KEYS = {"eval_set", "eval_sets", "eval_file", "eval_files", "eval_path", "eval_paths"}
+# Python constructs that own a name binding. A ``variable = Agent(...)`` is
+# reachable only from its own scope outward, so resolving a ``sub_agents``
+# element has to respect them.
+_SCOPE_NODES = (
+    ast.Module,
+    ast.FunctionDef,
+    ast.AsyncFunctionDef,
+    ast.Lambda,
+    ast.ClassDef,
+)
 
 
 def load_google_adk_artifacts(
@@ -589,17 +599,12 @@ class _PythonAdkExtractor:
         # reused: ``extract`` iterates it, and the sub-agent spelling map
         # below is built from it.
         self.agent_call_list = self._agent_calls()
-        # ADK routes a handoff to the sub-agent's ``name=``, but
-        # ``sub_agents=[salesforce_agent]`` spells the Python variable the
-        # agent was assigned to. Agent nodes are keyed by the name, so the
-        # two spellings must be reconciled here — otherwise the handoff
-        # lands on a phantom node that owns no tools and every tool the
-        # sub-agent holds drops out of the root-reachable surface (#385).
-        self.agent_names_by_variable = {
-            target_name: _kwarg_string(call, "name") or target_name
-            for target_name, call in self.agent_call_list
-            if target_name
+        self.parents = {
+            child: node
+            for node in ast.walk(tree)
+            for child in ast.iter_child_nodes(node)
         }
+        self.agent_names_by_variable = self._agent_names_by_variable()
 
     def extract(self) -> list[LoadedToolSource]:
         tools: list[Tool] = []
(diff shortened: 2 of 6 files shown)


src/agents_shipgate/inputs/google_adk.py
```diff
@@ -1032,6 +1056,16 @@ def _record_agent_callbacks_plugins_subagents(self, call: ast.Call, agent_name:
                         "source_id": self.source_id,
                         "sub_agent_count": sub_agent_count,
                         "sub_agents": sub_agent_names,
+                        # False when ``sub_agents`` is not a literal sequence,
+                        # or when any element could not be named. The binding
+                        # graph turns that into a partial-evidence gap: a
+                        # handoff we cannot name is a branch of the capability
+                        # surface we did not follow, and it must say so rather
+                        # than report the named subset as the whole.
+                        "sub_agents_complete": (
```
pengfei-threemoonslab: [P1] Do not mark unresolved imported sub-agents complete

This completeness check only counts `_qualified_name` successes. An imported name qualifies (`from sub import worker` becomes `sub.worker`) even though no local `Agent(...)` assignment was resolved. I reproduced a manifest that scans only `root.py`: `root.py` imports `worker` from `sub.py`, where `worker` owns `delete_record`; `delete_record` never enters the catalog, yet the graph reports `structural` and `pass_eligible=True` with no partial-evidence gap. That recreates the silent exclusion this PR is intended to close and contradicts the stated fail-closed behavior for imports. Track whether every element resolves to an observed local/cross-source agent; otherwise mark the handoff incomplete, and add the single-entrypoint imported-sub-agent regression.

src/agents_shipgate/inputs/google_adk.py
```diff
@@ -585,12 +585,27 @@ def __init__(
...
+        # reused: ``extract`` iterates it, and the sub-agent spelling map
+        # below is built from it.
+        self.agent_call_list = self._agent_calls()
+        # ADK routes a handoff to the sub-agent's ``name=``, but
+        # ``sub_agents=[salesforce_agent]`` spells the Python variable the
+        # agent was assigned to. Agent nodes are keyed by the name, so the
+        # two spellings must be reconciled here — otherwise the handoff
+        # lands on a phantom node that owns no tools and every tool the
+        # sub-agent holds drops out of the root-reachable surface (#385).
+        self.agent_names_by_variable = {
```
pengfei-threemoonslab: [P1] Preserve lexical scope in the variable-to-agent map

`_agent_calls()` walks the entire AST, including nested functions, while this dictionary keys assignments only by the bare target name. If `build_a()` and `build_b()` both use a local `worker`, one entry silently overwrites the other. I reproduced `RootA(sub_agents=[worker])` reaching `WorkerB`'s `write_b` tool and excluding `WorkerA`'s actual `read_a` tool. Because this can make the gate judge the wrong capability surface, key the mapping by lexical scope/call site or fail closed when the identifier is ambiguous, and add a two-factory regression.

CHANGELOG.md
```diff
@@ -2,6 +2,44 @@
...
+  resolver rejected names its own scan had emitted. Declarations now reuse the
+  observed node, and a genuinely ambiguous name says how many agents share it
+  and which sources they came from instead of reporting the name as unknown.
+
+  Finally, a tool bound to an agent the configured root cannot reach now gets an
+  evidence gap naming it. Everything downstream of the binding graph is narrowed
+  to root-reachable tools, so such a tool is never judged; before this it was
+  not mentioned either, leaving the ratio `6/12 catalog tools reachable` as its
+  only trace. This covers tools whose owning agent the scan identified — a hole
+  in the graph shipgate built. A catalog entry that no agent binds at all is a
```
pengfei-threemoonslab: [P3] Use the canonical display name in prose

Repository naming rules reserve `shipgate` for the CLI alias. This user-visible sentence should say “a hole in the graph Agents Shipgate built.”

pengfei-threemoonslab (commented):
Engineer review — changes requested

The canonical #385 path is materially improved: same-module variable/name reconciliation, declaration reuse, and per-unreached-tool evidence gaps all behave as intended in the added fixtures. I reviewed the full six-file diff and validated the exact head with focused binding/release tests, Ruff, Compileall, the full pytest suite (the packaging fixture passed on a network-enabled rerun), and the local Agents Shipgate verifier. GitHub CI, Agents Shipgate, and Self Dogfood are all green; the verifier reports `complete` / `passed`.

Two P1 semantic gaps remain, both in the new ADK name reconciliation:

1. Imported sub-agents are marked complete without being resolved or scanned, so their entire tool surface can still disappear without a gap.
2. The variable→agent-name map drops lexical scope, so repeated local names in factory functions can route a root to the wrong agent and analyze the wrong tools.

Both can produce a false-negative capability surface, so I do not think this should merge until they are fixed with regression coverage. I found no other blocking issues. The changelog also has one nonblocking canonical-naming correction.

src/agents_shipgate/inputs/google_adk.py
pengfei-threemoonslab: Confirmed and fixed in 8162e27 — you were right, and it was worse than the completeness flag suggested: because `sub.py` is not a declared source, `delete_record` never enters the catalog at all, so the per-tool gap this PR adds could not have covered it either. Reproduced exactly as you described (`structural` / `pass_eligible=True`, no partial evidence, no trace of the tool).

Elements are now sorted into three outcomes rather than two. Resolved to an agent this module defines → a real handoff target. Named but matching no agent definition → recorded in a new `unresolved_sub_agents` list and reported as `partial_binding_evidence` naming the spelling. Not nameable at all → left to the count comparison, as before.

The unresolved element also no longer becomes a node of its own. That was the second half of the problem: an empty tool set on a node named `sub.worker` reads as proof the sub-agent has no capability, which is the opposite of what is known about it.

Your fixture now gives `status: partial`, `pass_eligible: False`, `agents: ["RootAgent"]`, and `Google ADK agent 'RootAgent' hands off to sub.worker, which this scan could not match to an agent definition; that sub-agent's tools were not analyzed.`

The message deliberately says "could not match" rather than "no such agent": in the two-entrypoint layout the agent *is* in the report under its own name, reached through a source this entrypoint cannot resolve, so the claim is about the spelling and must not read as a claim the agent does not exist.

Regression added: `test_google_adk_imported_sub_agent_is_not_reported_as_a_proven_surface`, single-entrypoint, with the import deliberately unscanned.

src/agents_shipgate/inputs/google_adk.py
pengfei-threemoonslab: Confirmed and fixed in 8162e27. Your reproduction is exact — I got `RootA -> WorkerB`, reachable `["write_b"]`, `read_a` unbound, and `status: structural` / `pass_eligible: True`. Analyzing a tool the root cannot call while excluding one it can, and calling that proven, is a worse failure than the one this PR set out to fix.

I took the scope-preserving option rather than the fail-closed one, because the two-factory shape is a legitimate ADK pattern and failing closed on it would be a false negative on real code. The map now keys on `(enclosing scope, name)` — scope resolved through a parent map over `Module` / `FunctionDef` / `AsyncFunctionDef` / `Lambda` / `ClassDef` — and `_sub_agent_name` walks scopes innermost-out from the referencing call. Your fixture now yields `{("RootA", "WorkerA"), ("RootB", "WorkerB")}` and reachable `["read_a"]`, with `write_b` named in a gap as out of `RootA`'s scope.

Ambiguity still fails closed where it genuinely cannot be settled: one name rebound to differently named agents *within a single scope* maps to `None` and resolves to nothing. Straight-line AST order is not control flow, so choosing either binding would be a guess.

Two regressions added: `test_google_adk_factory_locals_do_not_leak_between_scopes` (your two-factory case, asserting both handoff pairs and the reachable set) and `test_google_adk_ambiguous_agent_variable_resolves_to_nothing` (same-scope rebinding → partial, no handoff edges).

CHANGELOG.md
pengfei-threemoonslab: Fixed in 8162e27 — reads "a hole in the graph Agents Shipgate built" now. Checked the rest of the entry for the same slip; that was the only bare `shipgate` in prose. Per AGENTS.md the alias is for shell snippets only.

pengfei-threemoonslab:
Both P1s fixed in 8162e27, plus the P3. Each was reproduced first — thank you, they were real and the second was worse than a miss.

**P1.2 (scope)** — repro gave `RootA -> WorkerB`: the gate analyzed `write_b`, a tool that root cannot call, excluded `read_a`, which it can, and reported `pass_eligible=True`. I took the scope-preserving fix rather than fail-closed-on-ambiguity, since two factories with a local `worker` is a legitimate ADK shape and failing closed there would be a false negative on real code. The map keys on `(enclosing scope, name)` and resolves innermost-out from the referencing call. Genuine ambiguity — one name rebound to differently named agents *within one scope* — still resolves to nothing, because straight-line AST order is not control flow.

**P1.1 (imports)** — worse than the flag suggested: with `sub.py` undeclared, `delete_record` never enters the catalog, so the per-tool gap this PR adds could not have covered it either. Elements now sort into resolved / named-but-unmatched / unnameable, and an unmatched element no longer becomes a node — an empty tool set on a node called `sub.worker` reads as proof the sub-agent has no capability. The message says "could not match" rather than "no such agent", because in a two-entrypoint layout the agent *is* in the report under its own name.

**Regressions added** (3): single-entrypoint imported sub-agent, two factories sharing a local name, same-scope ambiguous rebinding.

**Verification** — `ruff check .`, `compileall -q src tests`, and the full `pytest` suite all clean, with no golden churn. Self-dogfood scan `passed`. Re-confirmed the canonical #385 path is unaffected: adk-samples#1745 still reports `12/12 catalog tools reachable` and 3 agent nodes.

One behavior change worth flagging: the two-entrypoint multi-file layout I called out as out-of-scope in the PR description now reports `partial` / `pass_eligible=False` instead of `structural`, since the imported handoff is exactly the case P1.1 asks to fail closed. That is the correct verdict — we can see both agents but cannot prove the handoff connects them — and `agent_bindings.declarations[].handoffs` is the lever, which this PR also makes resolvable.

PR #391, opened 2026-08-19
fix(inputs): refuse an absent input instead of reporting it as a malformed one (#389, #387, #384)

### LINKED ISSUE #384: Missing manifest reports as malformed; control.reason contradicts control.next_action
## Summary

When the manifest **does not exist**, `doctor` reports it as **malformed**:

```
Config error: Config file must contain a YAML object: <path>
```

An absent file, an empty file, and a valid-YAML-but-not-a-map file all produce this
identical string. The message asserts the file exists and has the wrong shape, which
for the absent case is simply false — it sends you looking for a broken file that
isn't there.

`config/loader.py` already has the correct message for this case. The path `doctor`
takes never reaches it.

## Reproduction

Three cases, one message:

```bash
# A — file absent
agents-shipgate doctor --config /tmp/absent-dir/shipgate.yaml --json

# B — file exists, empty
: > /tmp/empty.yaml && agents-shipgate doctor --config /tmp/empty.yaml --json

# C — file exists, valid YAML, but a list
printf -- "- a\n- b\n" > /tmp/list.yaml && agents-shipgate doctor --config /tmp/list.yaml --json
```

All three:

```
Config error: Config file must contain a YAML object: <path>
```

The *routing* does tell them apart, so the distinction is known internally:

| Case | `control.next_action.kind` | `verify_required` |
|---|---|---|
| A absent | `verify` → `verify --preview` | `true` |
| B empty | `edit` → "Fix the manifest in place" | `false` |
| C list  | `edit` → "Fix the manifest in place" | `false` |

## Root cause

Two sites collapse a failed read into empty bytes:

- `src/agents_shipgate/cli/_register_doctor.py:245` — `snapshots[path] = b""`
- `src/agents_shipgate/cli/scan/inspect.py:82` — `except OSError: manifest_bytes = b""`

Then `decode_manifest(b"")` → `""` → `yaml.safe_load("")` → `None` → fails the
`isinstance(data, dict)` test at `src/agents_shipgate/config/loader.py:53`, which
raises the "must contain a YAML object" ConfigError.

Meanwhile the other reader, `load_yaml_file()` at `src/agents_shipgate/config/loader.py:20-24`,
has the branch that should have fired:

```python
if not path.exists():
    hint = ""
    if path.name == "shipgate.yaml":
        hint = " Run `agents-shipgate init --workspace . --write` to create one."
    raise ConfigError(f"Config file not found: {path} in {Path.cwd()}.{hint}")
```

This is the overlay-reader drift class from #348: two readers of one input with
divergent error semantics. The `OSError → b""` is deliberate — it keeps a failure
reported against the same bytes the identity hashed — but it conflates *absent*
with *empty*.

## Why it matters

**Human-facing.** This is a first-run papercut, hit before reaching any real work.
It was found while pointing `doctor` at a relative `--config` from the wrong CWD;
the message claimed a malformed file, and the file did not exist at all.

**Agent-facing — the worse half.** The control envelope contradicts itself:

- `control.reason` = `"Config file must contain a YAML object: <path>"` → implies *fix the file*
- `control.next_action.kind` = `"verify"` → implies *bootstrap from scratch*

A coding agent reasoning from `reason` will try to edit a file that does not exist.
An agent following `next_action` behaves correctly. Two fields of one envelope
disagree about what the workspace looks like.

## Suggested fix

Distinguish absent from unreadable at the snapshot read, and let the not-found
ConfigError carry the existing `init --write` hint. Keeping `b""` for the
identity binding is fine as long as the absent case is classified before the
YAML shape check, so that `reason` and `next_action` agree.

Worth a negative-control test asserting the three cases produce three distinct
`message`/`reason` strings — otherwise a future refactor re-collapses them silently.

## Environment

- CLI `0.16.0b7` (source checkout), contract `24`
- Python 3.12.13, macOS

### LINKED ISSUE #387: Manifest type mismatch surfaces as internal_error and tells the user to file a bug (TypeError escapes pydantic validation)
## Impact

A routine manifest mistake — a YAML mapping where a list is expected — is reported as a
**shipgate crash**, and the user is told to file a bug for their own typo:

```
Internal error: artifact entries must be a list
next_action: "Re-run with --verbose for a stack trace; this is a bug — please file an issue."
```

No field, no file, no line number. The user cannot tell which key is wrong, and the guidance
sends them to the issue tracker instead of to their editor.

## Severity: P2 — small fix, disproportionate cost

Two-line fix. It fires on the adoption path (`google_adk.tool_inventories` is the prescribed
remedy for the first gap most ADK users hit), and it converts a five-second correction into a
bug report. It also pollutes the issue tracker with reports that are not defects.

## Reproduction

CLI `0.16.0b7`, contract `24`.

```yaml
# shipgate.yaml — mapping instead of a list
google_adk:
  tool_inventories:
    adk_agent: tool-inventory.json
```

```bash
```

## The revealing contrast

The same field with a **list** containing a bad key behaves correctly:

```yaml
google_adk:
  tool_inventories:
    - source_id: adk_agent      # not a valid key
      path: tool-inventory.json
```

```
Config error: Invalid shipgate.yaml:
- google_adk.tool_inventories.source_id: Extra inputs are not permitted
next_action: "Edit shipgate.yaml"
```

Exact path, correct error class, actionable next step. Only the mapping shape crashes.

## Root cause

`src/agents_shipgate/schemas/manifest/_artifacts.py:26` and `:44`:

```python
if not isinstance(value, list):
    raise TypeError("artifact entries must be a list")
```

Pydantic converts `ValueError` and `AssertionError` raised inside a validator into a
`ValidationError`, but lets **`TypeError` propagate**. So this escapes the config-loading
boundary and is caught by the generic internal-error handler. When the value *is* a list, the
bad item reaches `ArtifactPathConfig.model_validate(item)`, which raises a proper
`ValidationError` — hence the clean message above.

The same file raises `TypeError("artifact entries must be strings or objects")` at the
element level, which will have the identical problem.

## Expected behavior

A manifest type mismatch produces a `config_error` naming the offending path, consistent with
every other manifest validation failure.

## Acceptance criteria

- [ ] `tool_inventories` given a mapping produces `config_error`, not `internal_error`.
- [ ] The message names the manifest path (e.g. `google_adk.tool_inventories`).
- [ ] `next_action.kind` is `edit` against `shipgate.yaml`, not "file an issue".
- [ ] Same for the element-level `TypeError` in the same module.
- [ ] Sweep for other `raise TypeError` inside pydantic validators; add a test asserting no
      manifest input shape yields `internal_error`.

## Notes

Worth treating the last criterion as the real fix. Any `TypeError` inside a validator is a
latent "tell the user to file a bug" path; a targeted sweep closes the class rather than this
instance.

Related: #384 (a different config-loading path that also misreports user input state).

### LINKED ISSUE #389: verify --preview creates a nonexistent --workspace path and exits 0 instead of reporting a bad argument
## Impact

`verify --preview` is documented as the safe probe:

> "Lightweight relevance check: evaluate triggers and report whether Shipgate is relevant +
> what to run next, **WITHOUT** running a scan, requiring a manifest, or writing any files
> beyond the verifier artifacts. Always exits 0."

Given a `--workspace` path that does not exist, it **creates the entire path** and writes
artifacts into it, then exits 0. A typo in `--workspace` silently succeeds and leaves a new
directory tree on disk instead of reporting a bad argument.

## Severity: P2

No data is lost, but the safe probe violates its own contract, and the failure is silent.
Two ways this bites:

- **Interactively** a mistyped path produces a confident-looking JSON result about a workspace
  that was never there — and litters the filesystem.
- **In CI** a misconfigured `--workspace` yields exit 0 plus a complete artifact set in a
  bogus location. Anything gating on exit code or on "did artifacts appear" reads that as
  healthy.

Found by a first-time user who ran the preview step before the clone step. The tool created
`~/tmp/adk-1745/adk-samples/` on their machine, which then **blocked the subsequent
`git clone`** (non-empty destination) — turning one skipped step into a second, unrelated
failure.

## Reproduction

CLI `0.16.0b7`, contract `24`.

```bash
T=/tmp/does-not-exist-$$/nested/workspace
[ -e "$T" ] && echo EXISTS || echo ABSENT     # ABSENT
agents-shipgate verify --preview --json --workspace "$T" --base main --head HEAD
echo "exit=$?"                                 # exit=0
find /tmp/does-not-exist-$$
```

Observed — **four** directory levels created, none of which existed:

```
/tmp/does-not-exist-NNN
/tmp/does-not-exist-NNN/nested
/tmp/does-not-exist-NNN/nested/workspace
/tmp/does-not-exist-NNN/nested/workspace/agents-shipgate-reports
   agent-handoff.json
   current-control.json
   pr-comment.md
   verifier.json
```

## Secondary: the absent path is reported as a shape problem

```json
"reason":      "git_failed",
"detail":      "Workspace is not inside a git checkout: <path>",
"remediation": "Inspect the reported Git failure before rerunning; fetching refs cannot repair a deterministic input failure.",
"fetch_repairable": false
```

The path did not exist. "Not inside a git checkout" implies it exists and lacks git, which
sends the user to diagnose the wrong thing. This is the same absent-reported-as-malformed
shape as #384, in an unrelated subsystem — see the comment added there.

## What preview got right

Worth recording, because the safety property held: the unreadable input did **not** become a
skip. `trigger.evaluation_status: not_evaluated`, `should_run: null` (not `false`),
`merge_verdict: unknown` (not `pass`), `can_merge_without_human: false`,
`control.must_stop: true`. The rationale is explicit:

> "That is not evidence the PR is unrelated to agent capabilities — repair the diff input and
> re-evaluate."

This issue is about the side effect and the diagnosis, not the verdict.

## Expected behavior

A `--workspace` that does not exist is an **invocation error**, reported before any directory
is created.

## Acceptance criteria

- [ ] `verify --preview` with a nonexistent `--workspace` creates **no** directories.
- [ ] The message names the missing path and says it does not exist, distinct from
      "exists but is not a git checkout".
- [ ] Same guard on non-preview `verify`, `scan`, and any other command that resolves
      `--workspace` before writing an output directory.
- [ ] Regression test asserting the filesystem is unchanged after the command.

## Open question for maintainers

`--preview` is contractually "always exits 0", so this deliberately does not ask for a
nonzero exit. But an absent `--workspace` is arguably an argument error rather than a preview
outcome, and belongs on a different path from "preview ran and could not read the diff".
Your call whether the guard lands in argument validation (before the preview contract
applies) or inside preview.

## Notes

Found while running the adoption path on
[google/adk-samples#1745](https://github.com/google/adk-samples/pull/1745). Related: #384,
#385, #386, #387, #388.

pengfei-threemoonslab:
## Summary

- Three reports, one class: **an input that is not there was reported as an input with the wrong shape** — and in one case the command created the input it had been asked to inspect.
- Closes #389, closes #387, closes #384.

### #389 — an absent `--workspace` is an invocation error, decided before anything is created

`verify --preview` given a `--workspace` that did not exist created the entire four-level path, wrote a full artifact set into it, and **exited 0**. A typo produced a confident-looking result about a workspace that was never there; in CI both signals a caller can gate on — exit code and "did artifacts appear" — read healthy. The leftover directory then blocked the `git clone` the reporter had skipped.

New pure classifier `core/workspace_input.py` (`present` / `absent` / `not_a_directory` / `unreadable`) plus the CLI guard `cli/workspace_guard.py`, called as the **first statement** of all 19 commands that expose `--workspace`, and from `ensure_git_workspace` so the diff-input path stops reporting an absent path as "not inside a git checkout".

Probing the whole surface turned up four more live instances of the same class:

| Command | Before | After |
|---|---|---|
| `verify --preview` | created 4 dirs + 4 artifacts, exit 0 | exit 2, nothing created |
| `init --write`, `audit --host`, `verification prepare` / `worker` | bare `FileNotFoundError` traceback, exit 1 | exit 2, `config_error` |
| `install-hooks --write` | **wrote hooks** into the mistyped tree | exit 2, nothing created |
| `mcp audit` | `decision: allow` about a nonexistent directory | exit 2, `config_error` |
| `detect`, `check`, `trigger` | exit 0 with a payload describing nothing | exit 2, `config_error` |

**The open question in #389, answered:** the guard lands in argument validation, and preview exits 2. Its documented "always exits 0" is a promise about workspaces it *evaluates*; there is nothing here to evaluate, and the exit code is exactly the signal the silent 0 was misleading. Recorded as a STABILITY.md unreleased migration note — no contract or schema version moves, and the refusal reuses the existing `config_error` / exit 2 slot rather than adding an error kind.

### #384 — absent, empty, and malformed manifests read differently

Classified at the read boundary (`config/loader.py::manifest_read_error`), so the deliberate `OSError → b""` collapse still binds a failure to the bytes the identity hashed, while absent no longer reaches the YAML shape check. `control.reason` and `control.next_action` now tell one story instead of two:

| State | Message | `next_action.kind` |
|---|---|---|
| absent | `Config file not found: <path> in <cwd>. Run agents-shipgate init …` | `verify` |
| empty | `Config file is empty: <path>.` | `edit` |
| not a mapping | `Config file must contain a YAML object: <path>` | `edit` |

Routing is unchanged — it always distinguished the cases; only the message did not.

The cross-cutting check asked for in the issue comment is `tests/test_absent_input_messages.py`, a table over all six inputs Shipgate reads. Audit result: **only the manifest and the workspace conflated the states**. Tool source, policy pack, baseline, and diff refs already told them apart and are now pinned so they stay that way.

### #387 — a manifest type mismatch is an edit, not a bug report

Pydantic converts `ValueError` and `AssertionError` raised inside a validator into a `ValidationError`, but lets `TypeError` propagate past the config-loading boundary into the generic internal-error handler — so `google_adk.tool_inventories:` written as a mapping told the user their own typo was a Shipgate bug and asked them to file an issue. Every `raise TypeError` under `schemas/` is now `ValueError` (12 sites, 8 modules):

```
Config error: Invalid shipgate.yaml:
- google_adk.tool_inventories: Value error, must be a list of artifact paths, but is a mapping
```

`config_error`, exit 2, `next_action.kind: edit` against the manifest path. Messages also name the shape that *was* written.

### What keeps the class closed

Two sweeps, not two patches:

- `test_every_workspace_command_is_swept` enumerates `--workspace` commands from the **live Typer app** and fails when a new one is not in the coverage table.
- `test_no_schema_module_raises_typeerror` bans `raise TypeError` outright under `schemas/`.

## Type

- [ ] Input adapter change
- [ ] Documentation only

## Verification

Additional local checks run:

- Full suite green (`pytest -n auto`, exit 0) and `ruff check .` clean.
- `./shipgate self-check --json` → `ready: true`, all fixtures `ok`.
- All three issue reproductions re-run verbatim against the built CLI; the `#389` one asserted on the filesystem (`find` returns nothing).
- All 19 `--workspace` commands probed by hand with an absent path: uniform exit 2, zero filesystem creation.

## Release-readiness notes

- [x] Report/schema changes are additive or documented in `STABILITY.md` — no schema change; the behavior narrowing is documented in a new unreleased migration note

## Notes for the reviewer

- **The guard is side-effect only.** `require_workspace(workspace)` is deliberately not `workspace = require_workspace(workspace)`: `check` and the invocation policy (#322) render `--workspace .` relative to cwd, and resolving would rewrite the commands Shipgate emits.
- **Behavior narrowing to weigh.** Commands that previously exited 0 with a payload about a nonexistent workspace (`detect`, `check`, `trigger`, `mcp audit`, `install-hooks`) now exit 2. Every documented invocation passes an existing `--workspace` and is unaffected; commands whose `--workspace` is optional still accept its absence, which means "discover instead", not "a workspace that is missing".
- **Message-text change.** A consumer that pattern-matched `must contain a YAML object` to detect a *missing* manifest must switch to `SHIP-DIAG-MISSING-MANIFEST` or `control.next_action` — which is what `docs/errors.json` already directs. Same for the `#387` field messages: the field-path prefix is the stable part.
- **Incidental fix.** `tests/test_adapter_static_only.py` pins allowlisted `subprocess` call sites by line number; the new imports in `cli/bootstrap.py` and `cli/verify/git.py` shifted three of them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,57 @@
 
 ## Unreleased
 
+- **An input that is not there is no longer reported as an input with the
+  wrong shape, and no command creates the workspace it was asked to inspect.**
+  Three reports, one class. `verify --preview` given a `--workspace` that did
+  not exist created the entire four-level path, wrote a full artifact set into
+  it, and exited 0 — so a typo produced a confident result about a workspace
+  that was never there, and in CI both signals a caller can gate on read
+  healthy. The leftover directory then blocked the `git clone` the reporter had
+  skipped, turning one missed step into a second, unrelated failure
+  ([#389](https://github.com/ThreeMoonsLab/agents-shipgate/issues/389)).
+
+  An absent `--workspace` is now an invocation error — `config_error`, exit 2,
+  decided before any directory is resolved or created — on **every** command
+  that takes the option, `--preview` included. Preview's documented "always
+  exits 0" is a promise about workspaces it evaluated; there is nothing here to
+  evaluate. The sweep that enforces this is generated from the live command
+  table, so a new `--workspace` command cannot quietly reopen the hole. It also
+  closes four defects found while closing the first: `init --write`,
+  `audit --host`, and `verification prepare`/`worker` raised bare
+  `FileNotFoundError` tracebacks, `install-hooks --write` wrote hooks into the
+  mistyped tree, and `mcp audit` answered `decision: allow` about a directory
+  that did not exist.
+
+  `doctor` reported an **absent** manifest as a malformed one — "Config file
+  must contain a YAML object" — for a file that had never been created. The
+  routing knew the difference all along, so the control envelope contradicted
+  itself: `control.reason` said *fix the file* while `control.next_action` said
+  *bootstrap from scratch*, and an agent reasoning from the reason edited a file
+  that was not there. Absent, empty, and present-but-not-a-mapping now produce
+  three distinct messages, classified where the read failed rather than after
+  the bytes have been flattened to `b""`
+  ([#384](https://github.com/ThreeMoonsLab/agents-shipgate/issues/384)). The
+  same conversion in the diff-input path — "Workspace is not inside a git
+  checkout" for a path that did not exist — is gone with it.
+
+- **A manifest type mismatch is an edit, not a bug report.** A YAML mapping
+  where a list belongs — `google_adk.tool_inventories:` with keys under it, the
+  prescribed remedy for the first gap most ADK adopters hit — was reported as
+  `internal_error` with "this is a bug — please file an issue", naming no field,
+  no file, and no line. Pydantic converts `ValueError` and `AssertionError`
+  raised inside a validator into a `ValidationError`, but lets `TypeError`
+  propagate past the config-loading boundary
+  ([#387](https://github.com/ThreeMoonsLab/agents-shipgate/issues/387)).
+
+  Every validator in the manifest schema now raises `ValueError`, so the same
+  mistake produces `config_error` (exit 2) naming the manifest path and routing
+  to `edit`, exactly as a bad key inside a correctly shaped list already did.
+  Messages also name the shape that was written — "must be a list of artifact
+  paths, but is a mapping" — and a build-time sweep fails if `raise TypeError`
+  reappears anywhere under `schemas/`, which closes the class rather than the
+  instance.
+
 - **A Google ADK sub-agent's tools are part of the analyzed surface, and a tool
   the gate did not look at can no longer go unmentioned.** On the canonical ADK
   multi-agent shape — a coordinator with `sub_agents=[salesforce_agent,
--- a/STABILITY.md
+++ b/STABILITY.md
@@ -13,6 +13,72 @@ for reproducible CI.
 
 ---
 
+<a id="migration-note-unreleased-absent-input"></a>
+
+## Migration Note: unreleased — an absent input is refused, not misreported
+
+No version moves: `contract_version`, `report_schema_version`,
+`minimum_control_contract_version`, and every published schema document are
+unchanged. No new error kind and no new exit code — the refusals below all use
+the existing `config_error` / exit `2` slot published in
+[`docs/errors.json`](docs/errors.json).
+
+**`--workspace` must name an existing directory, on every command that takes
+it.** A path that does not exist is an invocation error, decided before any
+output directory is resolved and before anything is written. This changes
+observable behavior in three ways, all of them narrowing a case that
+previously "succeeded":
+
+- `verify --preview` no longer exits 0 for a `--workspace` that does not
+  exist. Its documented "always exits 0" holds for every workspace it
+  *evaluates*; an absent path is not one, and preview previously created the
+  entire path and wrote its artifact set into it before answering
+  ([#389](https://github.com/ThreeMoonsLab/agents-shipgate/issues/389)).
+- `detect`, `check`, `trigger`, `mcp audit`, and `install-hooks` previously
+  exited 0 with a payload describing a workspace that was not there; they now
+  exit 2 with `config_error`.
+- `init --write`, `audit --host`, and `verification prepare`/`worker`
+  previously raised an unhandled `FileNotFoundError` (exit 1); they now exit 2
+  with `config_error`.
+
+A caller that passes an existing `--workspace` — every documented invocation —
+is unaffected. Commands whose `--workspace` is optional still accept its
+absence, which means "discover instead", not "a workspace that is missing".
+
+**Three manifest states, three messages.** An absent manifest, an empty one,
+and a present-but-not-a-mapping one previously produced one identical
+`Config file must contain a YAML object: <path>` string while routing to two
+different actions, so `control.reason` and `control.next_action` could
+disagree about whether the file existed
+([#384](https://github.com/ThreeMoonsLab/agents-shipgate/issues/384)). The
+messages are now:
+
+| State | Message | `next_action.kind` |
+|---|---|---|
+| absent | `Config file not found: <path> in <cwd>.` (plus the `init --write` hint for `shipgate.yaml`) | `verify` |
+| empty | `Config file is empty: <path>.` | `edit` |
+| not a mapping | `Config file must contain a YAML object: <path>` | `edit` |
+
+Routing is unchanged — it always distinguished the cases. `ConfigError` and
+exit `2` are unchanged for all three. Consumers that pattern-matched on the
+`must contain a YAML object` text to detect a *missing* manifest must switch
+to the diagnostic id (`SHIP-DIAG-MISSING-MANIFEST`) or to
+`control.next_action`, which is what the recovery hint in `docs/errors.json`
+already directs.
+
+**A manifest type mismatch is a `config_error`, not an `internal_error`.**
+`raise TypeError` inside a Pydantic validator propagates past the
+config-loading boundary; the manifest schema now raises `ValueError`
+throughout, so a YAML mapping where a list belongs produces
+`Invalid shipgate.yaml:` with the offending field path and an `edit` route,
+the same as any other manifest validation failure
+([#387](https://github.com/ThreeMoonsLab/agents-shipgate/issues/387)). Error
+text for these fields changed shape (it now names what was written, e.g.
+"must be a list of artifact paths, but is a mapping"); the field path prefix
+is the stable part.
+
+---
+
 <a id="migration-note-unreleased-doctor-environment"></a>
 
 ## Migration Note: unreleased — `doctor --json` reports the environment that answered
--- a/docs/agent-contract-current.md
+++ b/docs/agent-contract-current.md
@@ -532,7 +532,11 @@ skips a head-only scan; `verifier.json.merge_verdict` is `unknown` and the
 command exits 2.
 
 `agents-shipgate verify --preview --json` is a lightweight relevance check — no
-scan, no manifest required, exits 0. It emits a `verifier.json` with
+scan, no manifest required, exits 0 for every workspace it evaluates. A
+`--workspace` that does not exist is not a workspace it evaluates: it is
+refused as an invocation error (`config_error`, exit 2) before any directory
+is created, on preview and on every other command that takes the option. It
+emits a `verifier.json` with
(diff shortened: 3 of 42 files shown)


pengfei-threemoonslab (commented):
Engineering review result: **no blocking findings**.

Reviewed the full 42-file diff at `e5ffd0108e357a5c22842b66515dcdeb69d07ccf`.

The implementation closes the three reported failure classes coherently:

- the workspace guard runs before command-side resolution/writes across the live Typer `--workspace` surface, while deliberately leaving caller-visible relative paths unchanged;
- snapshot readers retain the bytes used for identity while preserving the original read failure for diagnosis, so absent/empty/non-mapping manifests no longer collapse into one message;
- validator shape errors now remain inside Pydantic's `ValidationError` boundary, and the AST sweep prevents direct `raise TypeError` regressions under `schemas/`.

Validation performed:

- focused workspace/config/setup/scan regression tests passed;
- Ruff and `compileall` passed;
- the full parallel suite produced no assertion failures; its isolated wheel fixture was initially blocked by sandboxed dependency access, then `tests/test_packaging.py` passed 8/8 with network access;
- GitHub CI, Agents Shipgate, and Agents Shipgate Self Dogfood are all green;
- local deterministic verification reports `control.state=complete`, merge verdict `mergeable`, and release decision `passed`.

Non-blocking follow-up: `load_yaml_file()` still checks `Path.exists()` before `read_text()`. For a path whose parent component is a regular file, that short-circuits to the generic not-found message, so the new `NotADirectoryError` branch in `manifest_read_error()` is only reached by the snapshot readers. Letting the read classify `FileNotFoundError`/`NotADirectoryError` directly would make that edge consistent, but it does not block the acceptance criteria for #384/#387/#389.

I would approve this revision; GitHub does not allow the connected account to approve its own PR, so this is recorded as a comment review.

PR #392, opened 2026-08-20
fix(inventory): bind a tool inventory to the source it completes (#386)

### LINKED ISSUE #386: google_adk.tool_inventories duplicates catalog entries and does not close the incomplete_surface gap it is prescribed for
## Impact

The remediation the tool prescribes for `incomplete_surface` **does not close the gap, and
makes the headline coverage metric worse.** A user who follows the instruction exactly is
left further from a pass than before they started.

This is on the critical path of every ADK adoption: `incomplete_surface` fires for all six
statically-extracted tools on first scan, and `tool_inventories` is the only remedy offered.

## Severity: P1

Not a cosmetic issue — the loop is unclosable. The instruction is emitted by the tool, the
user complies, and the same verdict comes back with a degraded number. There is no third
step offered.

## Reproduction

CLI `0.16.0b7`, contract `24`. Subject: [google/adk-samples#1745](https://github.com/google/adk-samples/pull/1745),
`python/agents/smart_closer`.

The verdict instructs:

> "Review the skeleton written next to report.json, save it in your repo, reference it from
> `google_adk.tool_inventories` in shipgate.yaml, then rerun verification."

Doing exactly that:

```bash
cp agents-shipgate-reports/suggested-inventory.json ./tool-inventory.json
cat >> shipgate.yaml <<'YAML'
google_adk:
  tool_inventories:
    - path: tool-inventory.json
YAML
```

## Observed

| | before | after |
|---|---|---|
| catalog tools | 12 | **18** |
| reachable | 6 | 6 |
| ratio | 6/12 | **6/18** |
| `incomplete_surface: get_manager_email` | present | **still present** |

The catalog holds **18 entries but only 12 unique names**. Every one of the six inventory
tools is double-counted:

```
list_messages: 2   get_user_identity: 2   get_manager_email: 2
send_email: 2      map_salesforce_account_to_sap_bp: 2
map_salesforce_product_to_sap_material: 2
```

Removing the `tool_inventories` block returns the catalog to 12, confirming the inventory is
the cause.

## Root cause

`google_adk.tool_inventories` is `list[ArtifactPathConfig]`, and `ArtifactPathConfig` has
only two fields:

```
path: str      (required)
optional: bool (optional)
```

There is **no field binding an inventory to a `tool_source`.** Confirmed by probing
`source_id`, `id`, `source`, `tool_source`, `tool_source_id`, `for_source`, and `source_ref`
— all rejected with `Extra inputs are not permitted`.

So an inventory cannot *complete* the tools of the source that raised the gap; its entries
are appended to the catalog as new tools that happen to share names. Since the gap is keyed
to the original source's tool (`get_manager_email [adk_agent]`), it is never satisfied.

## Expected behavior

An inventory referenced to resolve `incomplete_surface` for source `adk_agent` should raise
**that source's** tools to high confidence, closing the gap, with the catalog remaining at 12.

## Acceptance criteria

- [ ] After the documented remediation, `incomplete_surface` for `get_manager_email [adk_agent]`
      is **closed**.
- [ ] Catalog size does not grow when an inventory names tools already extracted from a source.
- [ ] Either a source-binding field exists on inventory entries, or entries are merged by
      `(source_id, tool_name)`.
- [ ] Coverage ratio is monotonic: following a prescribed remediation never lowers it.
- [ ] Regression test asserting the ratio does not decrease across a prescribed fix.

## Notes

The monotonicity property in criterion 4 is worth a general guard. A remediation that
degrades the metric it was issued to improve will erode trust in every other number the
report prints.

pengfei-threemoonslab:
## Summary

- `<framework>.tool_inventories[]` entries now take an optional **`source_id`** naming the tool source whose surface the file enumerates. Entries matching a name that source already exposes are joined to that observation instead of added beside it — the catalog keeps its size, the merged tool inherits the inventory's high extraction confidence, and the `incomplete_surface` gap that asked for the file closes.
- Fixes [#386](https://github.com/ThreeMoonsLab/agents-shipgate/issues/386): the prescribed remediation did not close the gap it was issued for, and made the headline coverage metric worse.

### The reported defect, reproduced

`incomplete_surface` fires for every statically-extracted tool on a first ADK/LangChain/CrewAI/n8n scan, and the only remedy offered was "save the skeleton, reference it from `<framework>.tool_inventories`". Doing exactly that loaded the inventory as an **independent source**. On a minimal ADK repro the result was strictly worse than before the user started:

| | before | after (old behavior) |
|---|---|---|
| catalog tools | 4 | **6** (2 distinct names doubled) |
| reachable | 4/4 | **4/6** |
| pass-eligible actions | 2/4 | 2/4 |
| semantic gap count | 2 | **8** |
| `incomplete_surface` | present | **still present** |

The four extra gaps included `ambiguous_tool_selector` on the `action_surface.actions` rows that previously resolved cleanly — the remediation broke declarations that were already correct. With `source_id`, the same repro reaches catalog 4, reachable 4/4, pass-eligible **4/4**, gaps **0**.

### Design: one merge engine, not two

`build_tool_identity_catalog` opens with *"Build canonical tools without ever joining observations by name"* — a deliberate trust-root invariant. This PR keeps it. `source_id` is a **manifest declaration desugared into the existing `tool_identity.bindings` engine**: one synthesized binding per matched name, with the inventory as `primary` (which is what carries `extraction_confidence` through `_merge_bound_observations`). The manifest — the trust root — is what licenses the join; the inventory file never joins itself to anything.

The issue's acceptance criterion 3 also offered implicit `(source_id, tool_name)` merging as an alternative. I did **not** take that option: silently reversing the never-join-by-name rule would trade the invariant for one line of user convenience. Instead the explicit path is made cheap (see below).

Three cases deliberately do not merge:
- an entry the completed source does not expose — a tool static extraction missed, which is why inventories exist; it stays standalone and is honestly reported as unbound
- a name the source exposes twice — the inventory alone does not say which observation it describes; asks for an explicit `tool_identity.bindings` entry
- an observation a reviewed binding already claims — the human declaration outranks the desugared one

### Closing the loop end to end

- Both emitters of the prescribed text (`incomplete_surface`, `low_confidence_tool`) now route through one `_inventory_remediation` builder that names the field and the concrete source id. They had drifted into two near-identical copies.
- `suggested-inventory.json`'s note carries the exact YAML entry to write, and tells a multi-source skeleton to split.
- The note and the fix-task rollup name the route **only for source types that actually have a `tool_inventories` key** — my first pass told `samples/conductor_agent` to add `source_id` to a key that does not exist for its source type. `fix_task.py` already had this guard; `writing.py` did not.
- An inventory declared without `source_id` that shadows a low-confidence source now emits a `source_warning` naming the exact `source_id` to add. That is the "third step" the issue says was missing — the legacy spelling still loads (an inventory can legitimately describe a separate surface), but the degrading case is no longer silent.

### Scope note

Fixed for **all four frameworks** with a `tool_inventories` key — ADK, LangChain, CrewAI, n8n. The schema, the loaders, and the prescribed-text table are shared code, so an ADK-only fix would have left three identical P1s.

`samples/google_adk_agent/shipgate.yaml` drops from 20 lines of hand-written `tool_identity.bindings` to one entry, verified byte-identical in coverage and findings against the pre-change manifest. It was the ADK adoption exemplar teaching the workaround. The LangChain and CrewAI samples are untouched and remain the demonstration of hand-written bindings. **Happy to drop the sample change** if you'd rather keep the diff strictly to the reported surface.

## Type

- [x] Input adapter change
- [ ] Documentation only

## Verification

Additional local checks run:

- Full suite green (`pytest -q -n auto -m "not perf"`, exit 0), plus `ruff check`, `compileall`, and `scripts/generate_schemas.py --check`.
- **Every new test was proved to fail without the fix** by stubbing `_inventory_completion_bindings` to return `([], [])`: 8 failures, all passing again on restore.
- End-to-end scans exercising all four states (no inventory / bound / unbound / unknown `source_id`) on ADK and LangChain projects.
- Acceptance criteria 1, 2 and 4 are asserted in one end-to-end test that **reads the `expects` string the report emits** and applies it mechanically, so the instruction text and the behavior cannot drift apart. Monotonicity is checked as `after_reachable × before_catalog ≥ before_reachable × after_catalog`.
- 12 new tests across `tests/test_tool_identity.py`, `tests/test_google_adk.py`, `tests/test_config.py` — including that the entry model still rejects unknown keys, since #386 was reported after probing seven spellings that all had to keep failing loudly.

## Release-readiness notes

The manifest change is **additive and backward compatible**: `source_id` is optional and the bare-path form (`- inv.json`) parses unchanged. `STABILITY.md`'s degraded-evidence recovery section is updated to state that the prescribed route includes `source_id`. `docs/manifest-v0.1.json` regenerated; new "Tool Inventories" section in `docs/manifest-v0.1.md`, plus `troubleshooting.md`, `agent-action-guide.md`, and the full example manifest.

Per [`CONTRIBUTING.md` § Surface discipline](../blob/main/CONTRIBUTING.md#surface-discipline): this is an optional field on an existing config model, not a new surface (no new command, schema version, report block, discovery surface, or adapter). It moves **activation rate / time-to-first-verdict** — it is on the critical path of every ADK adoption, and today that path is unclosable. Per (2), it extends the one decision engine rather than adding a parallel one: the join is desugared into the existing binding machinery.


src/agents_shipgate/core/tool_identity.py
```diff
@@ -321,6 +334,168 @@ def resolve_selectors_by_tool_id(
...
+            )
+            synthesized.append(
+                ToolIdentityBindingConfig(
+                    id=binding_id,
+                    provider=target,
+                    reason=(
+                        f"reviewed tool inventory {inventory_id!r} completes "
+                        f"source {target!r} (tool_inventories[].source_id)"
+                    ),
+                    primary=inventory_selector,
```
pengfei-threemoonslab, line 587: [P1] Preserve completed-source selectors after the join

Making the inventory observation primary rekeys the canonical tool's `source_id` to `google_adk_inventory:...`. A selector such as `{tool: lookup, source_id: adk_agent}` resolves before the documented remediation but matches zero tools afterward (the merged tool has `provider=adk_agent`, but selector resolution compares `source_id` only). This is especially important when same-name providers require a source-qualified selector, and Shipgate itself emits source-qualified action-declaration scaffolds. Preserve the completed source as a valid selector identity (or retain/resolve member-source aliases) and add a regression test that applies a pre-existing `source_id`-qualified action row across inventory completion.

src/agents_shipgate/core/tool_identity.py
```diff
@@ -321,6 +334,168 @@ def resolve_selectors_by_tool_id(
...
+            if binding_id in claimed_ids:
+                continue
+            claimed_ids.add(binding_id)
+            inventory_selector = ToolObservationSelectorConfig(
+                source_id=inventory_id, tool=name
+            )
+            synthesized.append(
+                ToolIdentityBindingConfig(
+                    id=binding_id,
+                    provider=target,
```
pengfei-threemoonslab, line 582: [P2] Do not discard source-owned semantics when promoting confidence

The automatic inventory-primary choice relies on `_merge_bound_observations`, but that merge starts from the primary and does not backfill fields such as `output_schema`, `owner`, `function_signature`, or auth `type`/`mode`/`credential_mode`/`source`; it only unions auth scopes/alternatives. In an n8n-style repro, completing a medium-confidence tool with `apiKey`/`unscoped` auth, an output schema, and an owner produces a high-confidence tool with unknown auth, `{}` output, and no owner, replacing the closed `incomplete_surface` gap with `partial_authority_evidence`. Completion should raise extraction fidelity without erasing stronger source evidence. Please merge missing semantic fields (with conflict checks) or keep the target observation primary and promote the reviewed inventory fields explicitly, with a regression test.

src/agents_shipgate/cli/scan/writing.py
```diff
@@ -181,6 +182,55 @@ def _write_suggested_declarations(
...
+        return (
+            "declare it for this source — as a `tool_sources[]` entry of type "
+            "`mcp`, or from the matching tool_inventories key where the "
+            "framework block has one"
+        )
+    if len(bindable) == 1:
+        key, source_id = bindable[0]
+        return (
+            f"reference it from `{key}` in shipgate.yaml as "
+            f"`- {{path: <saved file>, source_id: {source_id}}}` — without "
```
pengfei-threemoonslab: [P2] Quote source IDs in emitted YAML

`source_id` is an unconstrained string, and generated framework source IDs include the configured path. Interpolating it unquoted into a flow mapping makes legal values such as `google_adk:agent,prod.py` parse as `source_id: google_adk:agent` plus an extra `prod.py` key, so copying the exact remediation fails manifest validation. Serialize the value as a YAML/JSON-safe scalar here and in `_inventory_remediation`, and cover a path/source ID containing a comma (ideally also `#`/braces).

pengfei-threemoonslab (commented):
Changes are needed on the source-completion path before merge. The happy-path tests and CI are green, but the findings below show that the prescribed remediation is not monotonic for qualified selectors or semantic metadata, and its copy-paste YAML is not valid for every schema-legal source ID. I reproduced each issue against commit 16f8d2e. The P1 finding is blocking. Local Agents Shipgate verification returned `review_publishable` / `review_required` for the touched sample manifest trust root, and the verification receipt validated successfully. (Posted as a comment review because GitHub does not permit an author account to formally request changes on its own PR.)

src/agents_shipgate/core/tool_identity.py
pengfei-threemoonslab: Confirmed and fixed in 07117998.

Reproduced first — `{tool: lookup, source_id: adk_agent}` went `resolved: True` → `unresolved_tool_selector` across completion. Your point about Shipgate emitting source-qualified scaffolds itself is the sharpest part: `_action_selector` qualifies every action row by `source_id`, so a user who pasted our scaffold and then applied our inventory instruction got unresolved rows on declarations that worked a minute earlier — the same self-defeating-remediation class as #386 itself.

Took the alias branch of your suggestion: a canonical tool now answers to the `source_type`/`source_id` of **any** observation bound into it, read from `identity_assessment` claims (one entry per bound member; tools built outside the catalog fall back to their own pair). This repairs hand-written `tool_identity.bindings` too, not just the desugared ones.

One thing I tightened beyond the ask: given together, both qualifiers must be satisfied by the **same** observation. Checking them independently would resolve a selector pairing one member's type with another member's id — a tool neither observation describes. `test_source_qualifiers_must_be_satisfied_by_one_observation` pins that negative.

Regression tests: `test_completion_keeps_the_completed_source_as_a_selector_identity` (unit) and `test_source_qualified_action_rows_survive_inventory_completion` (end-to-end — a real scan with pre-existing `source_id`-qualified action rows, asserting no `unresolved_tool_selector`/`ambiguous_tool_selector` before or after, and that the rows still pass-eligible). Both fail without the fix.

src/agents_shipgate/core/tool_identity.py
pengfei-threemoonslab: Confirmed and fixed in 07117998. Your n8n-style repro reproduced exactly: a medium-confidence tool with `apiKey`/`unscoped` auth, an output schema, and an owner came back `extraction_confidence=high, owner=None, output_schema={}, auth.type=None, auth.mode=unknown, function_signature=None`.

Went with the backfill option rather than "keep the target observation primary". `_surface_is_complete` keys completeness off `source_type` (`_AST_ONLY_SOURCE_TYPES`), so keeping the AST observation primary would also have required redefining that predicate — a much wider change for the same outcome.

`_merge_bound_observations` now backfills `description`, `input_schema`, `output_schema`, `parameters`, `function_signature`, `owner`, and auth `type`/`credential_mode`/`source`/`mode`/`explicit` — **into empty slots only**, and it runs after the existing conflict checks, so two populated disagreeing values remain `conflicting_tool_identity` rather than a silent pick (`test_backfill_never_overwrites_what_the_reviewed_primary_states` pins both halves). `auth.mode` is special-cased because `unknown` is the absence of a claim, not a claim of `none`, so a plain falsiness test would have skipped it. Source identity is deliberately excluded from the backfill list — the row keeps the primary's provenance, and other members stay selectable via the alias path in your P1 thread.

Worth flagging: **the erasure was suppressing a real finding, not only degrading evidence.** `samples/support_refund_agent` binds `send_email_preview(...) -> str` to an inventory that declares only `inputSchema`. The merge dropped both things `SHIP-SCHEMA-FREEFORM-OUTPUT` looks at — the AST-derived `{"type": "string"}` schema *and* the `sdk_function` source type its `-> str` fallback branch requires — so the shipped golden recorded no free-form-output finding for a tool that plainly returns free-form text. That golden has been wrong since the binding was added.

I regenerated the goldens rather than narrowing the backfill to keep them quiet. Delta is exactly one restored MEDIUM review item; decision stays `blocked`, blockers 5 → 5, review items 9 → 10, added check IDs `['SHIP-SCHEMA-FREEFORM-OUTPUT']`, removed none.

src/agents_shipgate/cli/scan/writing.py
pengfei-threemoonslab: Confirmed and fixed in 07117998. Reproduced verbatim — `source_id: google_adk:agent,prod.py` parsed as `{"path": "<saved file>", "source_id": "google_adk:agent", "prod.py": None}`, and `extra="forbid"` then rejects the stray key, so copying the exact prescribed text fails manifest validation.

Added `yaml_scalar()` in `core/evidence_actions.py` and applied it at both emitters (`_inventory_remediation` and the `suggested-inventory.json` note, single- and multi-source forms). JSON strings are a subset of YAML double-quoted scalars, escape sequences included, so `json.dumps(..., ensure_ascii=False)` is a total encoder here — no new dependency, and non-ASCII ids stay readable rather than escaped.

Covered the cases you named plus two more: `test_prescribed_entry_parses_back_to_the_source_it_names` is parametrized over `plain`, `comma`, `hash`, `braces`, and `colon-space`, and asserts against **both** emitters. Only `comma`, `braces`, and `colon-space` actually failed unquoted — `#` needs preceding whitespace to start a comment — but I kept the `hash` case since it documents that boundary rather than leaving the next reader to rediscover it.

Your review also changed how these tests assert. Every check on the prescribed text now parses it with `yaml.safe_load` via a `_inventory_entry_from` helper instead of substring-matching: my original `assert "source_id: adk_agent" in instruction` passes on output that silently splits into two keys, which is precisely the bug. The instruction is only worth emitting if it can be pasted, so the tests now read it the way a user would.

tests/test_google_adk.py
```diff
@@ -1197,3 +1198,327 @@ def write_b(value: str) -> str:
...
+    that resolved a minute earlier.
+    """
+
+    project = tmp_path / "project"
+    project.mkdir()
+    (project / "agent.py").write_text(_ADK_AGENT_SOURCE, encoding="utf-8")
+    declarations = """
+action_surface:
+  actions:
+    - tool: get_manager_email
```
pengfei-threemoonslab, line 1410: [P1] Exercise the actual generated selector, including `tool_id`

This regression test hand-writes only `tool` and `source_id`, but `_action_selector` also emits `tool_id`. Inventory completion changes the canonical ID from the observation-derived `tool_v2_*` to a binding-derived ID, and `ToolSelectorIndex.resolve` prioritizes `tool_id`; therefore the exact pre-completion scaffold still becomes `unresolved_tool_selector` after completion. I reproduced this with the before/after ADK catalogs: the selector with the old `tool_id` resolves before and fails after, while removing only `tool_id` makes the new alias test pass. Please test the actual emitted `declaration_template` and either preserve/alias the prior ID or stop emitting an ID that this remediation invalidates.

src/agents_shipgate/core/tool_identity.py
```diff
@@ -65,6 +102,24 @@ def build(cls, tools: Sequence[Tool]) -> ToolSelectorIndex:
                 name: tuple(sorted(matches, key=lambda tool: tool.id))
                 for name, matches in by_name.items()
             },
+            source_identities={tool.id: _source_identities(tool) for tool in tools},
```
pengfei-threemoonslab: [P1] Apply member-source aliases to every selector consumer

The aliases live only in `ToolSelectorIndex`, while two established selector paths still compare the canonical primary fields directly: `_action_has_policy_control` in `action_surface.py` and `_matching_suppression` in `findings/mutations.py`. In an end-to-end ADK case, the action row continues to resolve after completion, but `{tool: send_email, source_id: adk_agent}` in `require_confirmation_for_tools` stops matching because the `ActionFact` now carries the inventory source; the scan then emits missing `confirmation.required` and moves to `blocked`. A source-qualified `checks.ignore` row similarly becomes inert. The documented claim that a canonical tool answers to every bound observation identity needs to hold for these consumers too—centralize matching on the alias-aware index or carry the member identity pairs into `ActionFact`, and add policy/suppression regressions.

src/agents_shipgate/core/tool_identity.py
```diff
@@ -455,9 +680,61 @@ def _merge_bound_observations(primary: Tool, members: list[Tool]) -> tuple[Tool,
...
+    which is the same non-monotonic trade #386 is about (#386 review).
+
+    Only empty slots are filled, so the reviewed observation still wins wherever
+    it actually says something. Disagreements between two populated values are
+    not silently resolved here — ``input_schema``, ``auth.type`` and
+    ``auth.mode`` already raise ``conflicting_tool_identity`` above, and this
+    runs after those checks.
+    """
+
+    for field in _BACKFILL_FIELDS:
```
pengfei-threemoonslab: [P1] Reject disagreements for the newly preserved semantic fields

Only `input_schema`, annotations, `auth.type`, and `auth.mode` are conflict-checked above. Populated disagreements in `output_schema`, `function_signature`, `owner`, `auth.credential_mode`, and `auth.source` are silently resolved in favor of whichever value is already present. With a silent primary and two bound members reporting `string`/`team-a` versus `integer`/`team-b`, the catalog deterministically picked one by observation-ID order, emitted no issues, and left the identity `pass_eligible=True`; a populated primary versus member also silently wins. This contradicts the new manifest contract that populated disagreements produce `conflicting_tool_identity` and can suppress downstream checks such as `SHIP-SCHEMA-FREEFORM-OUTPUT`. Compare populated member values before backfilling (with explicit compatibility rules where appropriate) and make conflicting identities non-pass-eligible.

pengfei-threemoonslab (commented):
Follow-up review of `07117998`. I confirmed that the covered cases for the three earlier findings are improved, but found five remaining regressions. The three P1 items mean applying the prescribed inventory can still invalidate the exact generated action row, drop source-qualified confirmation/suppression selectors, or accept contradictory identity evidence as pass-eligible; please treat those as blocking. The P2 items cover evidence traceability and total YAML scalar encoding.

Validation performed: focused identity/ADK/config tests, Ruff, compileall, the full non-performance suite, packaging tests with isolated-build network access, targeted standalone reproductions, and `agents-shipgate verify` from the source checkout. All existing tests and CI pass, but they do not cover these cases.

src/agents_shipgate/core/tool_identity.py
```diff
@@ -455,9 +680,61 @@ def _merge_bound_observations(primary: Tool, members: list[Tool]) -> tuple[Tool,
...
+        if not getattr(merged, field):
+            value = getattr(member, field)
+            if value:
+                setattr(merged, field, deepcopy(value))
```
pengfei-threemoonslab: [P2] Preserve the donor's provenance when backfilling evidence

`setattr` copies the member's value into the primary `Tool` but leaves the primary inventory's source locator/provenance on the canonical row. The restored support-sample `SHIP-SCHEMA-FREEFORM-OUTPUT` finding consequently points to `inventories/sdk-tools.json#/tools/0`, although that object has no output schema; the `-> str` evidence exists only in `agents/refund_agent.py`. Backfilled n8n auth has the same problem: values supplied by workflow credentials are reported through the silent inventory/MCP source. Please retain per-field donor provenance (or otherwise make findings resolve the supplying observation) so review evidence points to the artifact that actually supports it.

src/agents_shipgate/core/evidence_actions.py
```diff
@@ -111,6 +112,25 @@ def _needs_escape(char: str, *, injective: bool) -> bool:
...
+    path (``google_adk:agents/agent.py``). Written bare into a flow mapping, a
+    value containing ``,`` splits into two keys — ``source_id: google_adk:agent``
+    plus a stray ``prod.py`` — so the exact text the tool prescribed fails
+    manifest validation under ``extra="forbid"`` (#386 review). ``#``, ``{``,
+    ``}``, ``:`` and leading indicators are the same class of hazard.
+
+    JSON strings are a subset of both YAML 1.1 and 1.2 double-quoted scalars,
+    escape sequences included, so ``json.dumps`` is a total encoder here. ``ensure_ascii=False`` keeps non-ASCII identifiers readable rather
+    than escaping them.
+    """
```
pengfei-threemoonslab, line 139: [P2] Escape C1 characters so the YAML round trip is actually total

`ensure_ascii=False` emits C1 characters literally. PyYAML rejects values containing U+0080/U+009F, and normalizes U+0085 NEL to a space, so `safe_load(yaml_scalar("adk\u0085agent")) != value`. These source IDs are still representable in a manifest via an escaped double-quoted scalar and the schema does not forbid them, which means the newly emitted remediation can be invalid or name the wrong source. Default `json.dumps` escaping (or explicit escaping of YAML-forbidden/normalized code points and lone surrogates) fixes this; please add C1 round-trip cases beyond the printable punctuation tests.

tests/test_google_adk.py
pengfei-threemoonslab: Confirmed and fixed in d0e68e90. You are right that my regression only passed because it hand-wrote `tool` + `source_id` — the field the tool actually emits was the one I left out, which made the test look like it covered the case while testing a subset.

Reproduced: `tool_id` before `tool_v2_bcb2f4…`, after `tool_v2_5fb072…`; the scaffold with the old id resolved before and returned `unresolved_tool_selector` after.

Took the alias branch rather than dropping `tool_id`: it is what disambiguates same-name providers, so removing it would trade this bug for the one it exists to prevent. A canonical tool now answers to the id each of its observations carried while unbound, computed through `_observation_tool_id` — the same expression `build_tool_identity_catalog` uses to issue one, so the two cannot drift.

One thing worth flagging that fell out of the fix: putting alias ids into `ToolSelectorIndex.by_id` broke `agent_bindings`, which reads `set(by_id)` as the entire catalog when it partitions reachable/possible/unbound. The `tool_catalog does not match binding graph partitions` invariant caught it on the first run. Aliases now live in a separate `by_selectable_id` used only by `resolve`, and `test_alias_ids_never_enter_the_catalog_partition` pins that.

Tests: `test_the_id_an_observation_had_unbound_still_selects_the_merged_tool` (unit, asserts the id really does change and the old selector still lands on the merged tool) and `test_the_generated_action_scaffold_still_resolves_after_completion` — which reads the real `declaration_template` off the emitted gap, asserts it carries `tool_id`, and writes *that* into the manifest, so the selector under test is the generated one rather than my paraphrase of it.

src/agents_shipgate/core/tool_identity.py
pengfei-threemoonslab: Confirmed and fixed in d0e68e90. Both consumers were comparing the canonical primary fields directly, exactly as described — I documented the claim in `STABILITY.md` without making it true everywhere, which is the worse half of the mistake.

Centralized rather than duplicated: `IdentityAliases.matches(tool_id=, source_type=, source_id=)` in `core/tool_identity.py` is now the single predicate, and `_action_has_policy_control` (via `action_identity_aliases`, reading `ActionFact.semantic_assessment.identity` — the wire projection carries `observation_ids` and `claims`, so no new field on `ActionFact` was needed) and `_matching_suppression` (via `tool_identity_aliases`) both go through it. That also picked up `tool_id` aliasing in both places for free, which they needed for the same reason as the P1 above.

`ToolSelectorIndex` keeps the alias map out of `by_id` — see my reply on the `tool_id` thread; `agent_bindings` reads `set(by_id)` as the catalog partition and the consistency invariant caught it.

Regression: `test_source_qualified_policy_and_suppression_survive_completion` runs a real scan before and after completion with a source-qualified `require_confirmation_for_tools` entry and a source-qualified `checks.ignore`, and asserts the confirmation control does not go missing, the suppression still bites, and the decision does not get worse across the remediation. It fails on all three counts without the fix.

Note I had to swap the suppressed check in that test: `SHIP-SCHEMA-BROAD-FREE-TEXT` does not fire on the ADK fixture, so the assertion would have passed vacuously. It uses `SHIP-DOC-MISSING-DESCRIPTION`, which does.

src/agents_shipgate/core/tool_identity.py
pengfei-threemoonslab: Confirmed and fixed in d0e68e90 — with one deliberate departure I want to put in front of you rather than bury.

Reproduced both halves: with a silent primary, members reporting `team-a`/`string` and `team-b`/`integer` gave `owner='team-a'`, `output_schema={'type': 'string'}`, zero issues, `pass_eligible=True`. My own test that claimed to cover this asserted `conflicting_tool_identity` was present — but that came from `auth.type`/`auth.mode`, which were already checked, so it gave me false confidence about the fields it did not cover.

`_backfill_preserved_evidence` now collects every contributor per field, primary included, and more than one distinct populated value raises `conflicting_tool_identity` naming the field and the donors. `_assessment` already makes any issue non-pass-eligible, so that half needed no change. Schemas compare through `_schema_signature` — the compatibility rule bound `input_schema` already used — so an inventory adding formats or `additionalProperties: false` still reads as a refinement.

**The departure: `auth.source` is preserved but not compared.** Implementing your list as written made *every* completed ADK tool `conflicting_tool_identity` and took `pass_eligible_actions` to 0. The reason is that `auth.source` names the extractor that produced the auth record, not the credential — `'google_adk_static'` on the AST observation, `'mcp'` on the inventory reading of the same tool — so two observations of one capability disagree by construction. `credential_mode` is a real claim and is checked. If you meant `auth.source` as a semantic claim rather than a provenance label I will reconsider, but as the adapters populate it today, comparing it is a guaranteed false positive.

I also excluded `description` and `parameters` from comparison (they backfill): prose differs benignly whenever an inventory rewords a docstring, and `parameters` restates `input_schema`, which is already compared — reporting either would double-count or manufacture noise. Reasoning is in the `_BACKFILL_FIELDS` comment.

Tests: disagreement-is-a-conflict, populated-primary-vs-member-is-also-a-conflict, agreeing-members-stay-quiet (so the check is not just noise), and an explicit assertion that `auth.source` does *not* appear in the conflict messages.

src/agents_shipgate/core/tool_identity.py
pengfei-threemoonslab: Confirmed and fixed in d0e68e90. Reproduced: the restored finding cited `inventories/sdk-tools.json` `#/tools/0`, an object with no `outputSchema` key in it — so the one artifact a reviewer would open to check the claim is the one artifact that cannot support it.

Did both halves of your "or". Each preserved value now records the observation that supplied it in `Tool.evidence_provenance` (keyed by Tool field name, `auth.*` included; in-memory like the other assessments, so no wire surface moves). And `tool_finding` takes an optional `evidence_field`: a check names the field it judged, and the finding's `SourceReference` resolves to the supplying observation instead of the primary, with an `evidence_supplied_by` entry recording the redirect. `SHIP-SCHEMA-FREEFORM-OUTPUT` now reports:

```
source.ref      : agents/refund_agent.py
source.location : agents/refund_agent.py:5
evidence_supplied_by: {field: output_schema, source_type: sdk_function, source_ref: agents/refund_agent.py}
```

which is where the `-> str` actually is.

**It is opt-in per check, deliberately.** I first tried deriving the donor generically by intersecting the finding's evidence keys with the provenance map — it needs no per-check edits, but it misfires whenever an evidence key happens to share a Tool field name, and `source` and `owner` both collide with keys other checks already use. An explicit `evidence_field` cannot silently re-point the wrong finding. Only `SHIP-SCHEMA-FREEFORM-OUTPUT` opts in so far, since that is the one with a demonstrated mismatch; the map is on the tool for any other check you want wired, and I am happy to do the auth ones in this PR if you would rather not leave them.

Goldens regenerated — the finding's source and fingerprint move; `report.md` is unchanged. Regression: `test_preserved_evidence_findings_cite_the_artifact_that_declares_it`, plus a unit test that the donor map records the right observation for `output_schema`, `owner`, and `auth.source`.

src/agents_shipgate/core/evidence_actions.py
pengfei-threemoonslab: Confirmed and fixed in d0e68e90. Reproduced all of it:

```
NEL U+0085     round_trips=False  got='adk agent'     <- names a DIFFERENT source
PAD U+0080     PyYAML REJECTED: ReaderError
APC U+009F     PyYAML REJECTED: ReaderError
lone surrogate PyYAML REJECTED: ReaderError
```

The NEL case is the one that bothers me most: it parses cleanly and silently renames the source, so the failure is invisible at the point of use.

Switched to default `json.dumps` escaping. I had reached for `ensure_ascii=False` for readability on accented identifiers and called the function "total" in its own docstring while it was not — the docstring now says why the escaping is load-bearing rather than incidental, so the next person does not re-optimize it back for prettiness.

Your list was also incomplete in one direction: **U+007F DEL** fails the same way and is ASCII, so it survives any "escape non-ASCII" rule that reasons only about the C1 block. `json.dumps` escapes it because it is a control character, which is another argument for the default encoder over hand-rolled code-point rules — I would have missed DEL writing those by hand.

Regression: `test_yaml_scalar_round_trips_every_schema_legal_source_id`, parametrized over `plain`, `comma`, `hash`, `braces`, `colon-space`, `non-ascii`, `c1-pad`, `c1-apc`, `del`, `c1-nel`, `lone-surrogate`. Five of the eleven fail with `ensure_ascii=False`.

Issue #394, opened 2026-08-20
verify --preview routes monorepo adoption to a root `init` that always refuses (`resolve_change_scope` called without `evidence_dirs`)


pengfei-threemoonslab:
## Summary

`verify --preview` on a monorepo PR emits `init --workspace <repo root> --write`, which `init` then refuses deterministically. The change-scope resolver had everything it needed to route to the correct sub-project and returned nothing, because the preview call site does not pass `evidence_dirs`.

## Evidence

`src/agents_shipgate/cli/verify/orchestrator.py:4129`:

```python
resolution = resolve_change_scope(
    root=root,
    changed_files=changed_files,
    limit=requested_root,
)
```

`evidence_dirs` is omitted, so it defaults to `frozenset()`. That parameter is what unlocks `WEAK_PROJECT_MARKERS` (`src/agents_shipgate/cli/discovery/scope.py:62`), and `find_project_root` documents the exact case it exists for (`scope.py:200`):

> `evidence_dirs` unlocks `WEAK_PROJECT_MARKERS` for exactly the directories the caller has already found agent evidence in. A bare `requirements.txt` is not a project boundary … but two sibling agents whose only marker is `requirements.txt` beside `agent.py` are two projects

`python/agents/smart_closer/` has **no `pyproject.toml`**; `requirements.txt` beside `agent.py` is its entire boundary. Without `evidence_dirs`, `find_project_root` walks past it, finds no strong marker, reaches the workspace root and returns `None`.

`detect`'s own candidate builder (`src/agents_shipgate/cli/discovery/signals.py:2085`) *does* construct and pass `evidence_dirs`, and correctly reports `smart_closer` with `marker: "requirements.txt"`. The same question is answered by two code paths with different inputs.

## Reproduction

[google/adk-samples#1745](https://github.com/google/adk-samples/pull/1745). All 17 changed files are under `python/agents/smart_closer/`. CLI `0.16.0b7`.

```
agents-shipgate verify --preview --json --workspace <repo> \
  --base 765b4d1fbdcdd397fc48c622c22d757d8c87bafe --head pr-1745
```

Actual:

```json
"next_action": {
  "kind": "initialize",
  "command": "shipgate init --workspace <repo root> --write --json",
  "why": "This unconfigured workspace looks agent-related; initialize the local Shipgate contract and advisory agent workflow."
}
```

Running that exact command:

```
Refusing to write shipgate.yaml: <repo root> holds 53 self-contained
[...]

PR #400, opened 2026-08-20
fix(google-adk): measure extraction confidence instead of assuming it (#393)

### LINKED ISSUE #393: `insufficient_evidence` is structural for every code-derived framework: two independent ceilings deny surface completeness by constant
## Summary

A tool surface derived from parsed code can never be treated as complete, regardless of how statically analysable it is. Two independent ceilings enforce this, and both are constants keyed on the code path rather than measurements of what was observed. The consequence is that `insufficient_evidence` is the default first-run verdict for **every code-derived framework** — OpenAI Agents SDK, LangChain, CrewAI and Google ADK — not a property of any particular repository.

## The two ceilings

**Ceiling 1 — `extraction_confidence` never exceeds `medium`.**

- `src/agents_shipgate/inputs/google_adk.py:988` hardcodes `extraction_confidence="medium"` for `google_adk_python_ast`.
- `src/agents_shipgate/inputs/_python_framework.py:246` is the only path that sets `"high"`, and it applies exclusively to tools loaded from a `tool_inventory` artifact.

**Ceiling 2 — completeness is denied by `source_type` alone.**

`src/agents_shipgate/core/semantic_assessment.py:888`:

```python
def _surface_is_complete(tool: Tool) -> bool:
    return not (
        tool.source_type in _AST_ONLY_SOURCE_TYPES
        or tool.annotations.get("wildcard_tools") is True
        ...
    )
```

`_AST_ONLY_SOURCE_TYPES` (`semantic_assessment.py:45`) contains `sdk_function`, `langchain_function`, `langchain_structured_tool`, `crewai_function`, `crewai_class_tool`, `crewai_prebuilt_tool`, `google_adk`, `google_adk_config`, `google_adk_function`.

The gate at `semantic_assessment.py:567` requires **both**:

```python
if tool.extraction_confidence != "high" or not _surface_is_complete(tool):
```

so raising confidence alone changes nothing — the source type still denies completeness. Any fix must address both sites together.

Downstream consumers of ceiling 1: `ci/release_decision.py:255` (`low_confidence_tool_count`), `core/semantic_assessment.py:114` (`extraction_complete`).

## Reproduction

[google/adk-samples#1745](https://github.com/google/adk-samples/pull/1745), `python/agents/smart_closer` — a single 247-line file, 12 module-level functions wrapped in `FunctionTool`, no factories, no dynamic toolsets, no imported tool symbols. CLI `0.16.0b7`, contract 24.

```
Summary: Insufficient evidence: the tool surface could not be fully enumerated
         (create_salesforce_quote [adk_agent]).
Evidence coverage: mixed (12 low-confidence tool(s); 12/12 catalog tools reachable;
                   36 semantic evidence gap(s); 0/12 actions pass-eligible)
```

```json
"binding_coverage":  {"total_catalog_tools": 12, "reachable_tools": 12,
                      "unbound_tools": 0, "pass_eligible": true, "gap_count": 0},
"identity_coverage": {"canonical_tools": 12, "pass_eligible_tools": 12, "gap_count": 0},
"source_warning_count": 0,
"low_confidence_tool_count": 12
```

Every one of the 12 `low_confidence_tool` gaps reads `extraction_confidence=medium; static extraction could not prove the full tool surface.`

## Why this matters

**A condition that holds for every input carries no information.** `low_confidence_tool` currently cannot distinguish a repository that assembles its tools in a factory loop from one that declares twelve module-level functions. Both are `medium`; both abstain. Ceiling 2 is the same statement made a second time, even more coarsely — an entire source type is denied completeness without reference to anything observed in the source.

**The real boundary is declarative-artifact vs. parsed-code, and it is drawn by constant rather than by evidence.** MCP, OpenAPI, Anthropic Messages API and OpenAI API artifacts can reach a complete surface; every framework whose tools are read out of source cannot.

**It makes the README inaccurate.** The README frames this as a dynamic-code edge case:

> when a change builds its tool surface dynamically — a toolkit factory, a config-bound allowlist, tools assembled at runtime — static extraction often can't enumerate the result, and Shipgate returns `insufficient_evidence`

The observed behaviour is not "often, on dynamic code". It is "always, on first run, for every code-derived framework" — including the most statically trivial case we could find.

**The prescribed remedy is transcription.** The user is told to write `suggested-inventory.json` listing the same 12 tools shipgate just extracted correctly. No new fact enters the system. The gap's own fields — `accepted_values: ["reviewed_explicit_inventory"]`, `requires_human_review: true` — show that what is actually being requested is a **human attestation**, not additional evidence.

## Suggested direction

Replace both ceilings with properties measured during the parse:

- Were all tool names statically resolvable?
- Was the tool list built by any dynamic construct — factory call, loop/comprehension append, config-bound allowlist, `*Toolset` construction, imported symbols?
- Were there unparsed or unreachable regions in the analysed module?
- Did every reference bind?

A source exhibiting none of those, with all catalog tools reachable and zero source warnings, is complete — whatever language it is written in. A source exhibiting any of them stays `medium`/`low` as today, and now the value actually distinguishes the two.

If the intended rule is instead that *a human must attest any tool surface before it may gate a release*, that is a defensible design — but it should be stated in the verdict as an attestation requirement rather than reported as an extraction failure. See #396.

Because this changes verdicts (`insufficient_evidence` → real answers), the adversarial-shape suite has to land with it; fail-open regressions will hide there.

---

Found during a first-time-user adoption walk of adk-samples#1745 on 2026-08-20, following only documented commands and CLI output. **Amended 2026-08-21** — originally filed as an ADK-only, single-ceiling issue; the second ceiling and the wider framework scope were found while designing the fix.

pengfei-threemoonslab:
## Summary

- **Google ADK extraction confidence is measured on the parsed module instead of hardcoded.** `inputs/google_adk.py` set `extraction_confidence="medium"` on every tool the AST path produced, and the only code that ever set `"high"` applied to tools loaded from a `tool_inventory` artifact. Every gate tests `!= "high"`, so no ADK repository could reach a pass from source however statically analysable it was — `insufficient_evidence` was not a property of a repository, it was the framework's default first-run verdict (#393).
- **`_surface_is_complete` was the same constant one layer down.** `core/semantic_assessment.py` disqualified any `source_type in _AST_ONLY_SOURCE_TYPES` outright. Fixing only `extraction_confidence` would have changed nothing observable: `incomplete_surface` would still have fired on every tool and still forced the same verdict. Membership now poses the question; the adapter's own attestation answers it.
- **The `low_confidence_tool` evidence gap names the construct responsible** instead of repeating one sentence on every AST-extracted tool in every repository.
- Docs (`upstream-integrations.md`, `manifest-v0.1.md`), CHANGELOG, and the `samples/google_adk_agent` manifest comment updated — that sample's inventory is no longer what earns it high confidence.

### Why this is worth the churn

The reported subject was the most statically trivial case available: one file, twelve annotated module-level functions, `12/12` catalog tools reachable, `0` unbound, `0` source warnings. It still reported twelve `low_confidence_tool` gaps and abstained. A condition that holds for every input carries no information — it could not tell a toolkit factory from twelve plain functions — and the remedy it prescribed was transcription: copy the tools Shipgate had just extracted correctly into `suggested-inventory.json`, adding no fact to the system.

Measured on the reproduction fixture:

| | before | after |
|---|---|---|
| clean module, no declarations | `insufficient_evidence`, 36 semantic gaps, "Fix at `suggested-inventory.json`" | `insufficient_evidence`, 24 gaps, "Fix at `shipgate.yaml#action_surface`" — no skeleton written |
| clean module, actions declared | `insufficient_evidence`, 0/12 pass-eligible | **`passed`**, 12/12 pass-eligible |

The second row is the point. Before this change, declaring effect and authority for every tool — the one thing the gate genuinely needs a human for — still left `incomplete_surface` on all twelve, because the extraction ceiling was unreachable from the manifest. Effect and authority remain human assertions; what changed is that they are now the *last* step rather than one behind a step no repository could take.

## How it works

The adapter writes `Tool.extraction["surface"]` (`enumerated` / `partial`) plus `surface_gaps[]` reason codes. `extraction` is written only by adapter code — no loader copies it out of a parsed input file — so unlike `annotations` the claim cannot be self-declared by an untrusted artifact.

The proof is scoped to the **file**, not the agent: one unresolved construct anywhere holds every tool the file produced, because a fully-resolved agent in a half-resolved module can still reach tools nobody enumerated. Per-function reasons (`untyped_parameter`, `variadic_parameters`, `decorated_tool_function`) are separate and do not affect siblings.

Absence of the attestation reads as incomplete everywhere downstream, so LangChain, CrewAI, and the OpenAI Agents SDK static path keep their previous verdict untouched, and a wildcard exposure still outranks any completeness claim.

## Reviewer notes

**Promoting anything to `high` is where the risk lives — please read this section first.** I wrote 17 adversarial constructs as probes against `_load_python_path`. **Eight were fail-open in the first draft** and would have shipped a soundness bug:

- `root_agent.tools.append(x)`, the same through an alias (`bucket = root_agent.tools`), `setattr(agent, "tools", ...)`, and `agent.tools[0] = x` → `mutable_tool_binding`
- `Agent(**config)`, which hides `tools` entirely and silently records `tool_count: 0` → `dynamic_agent_kwargs`
- four name-resolution shapes → `shadowed_tool_definition`. `self.functions` is a flat `ast.walk` name map (it has to be, to name a tool at all), so `tools=[helper]` happily resolves to a factory's inner def, a method lifted out of a class body, whichever of two conditional defs the walk saw last, or a def a later import/assignment rebinds. Each reports a signature the running agent will not use.

All are closed, each with a parametrized regression test.

**The unaccounted-warning backstop is deliberate.** Every extractor warning routes through `_surface_warning` (records a reason) or `_note_warning` (surface-neutral — eval artifacts only). `_resolve_extraction_evidence` compares the accounted count against the real growth of `artifacts.warnings`; a mismatch records `unclassified_extractor_warning`. A future contributor adding an ambiguity with a plain `warnings.append` therefore demotes the module automatically rather than failing open. `test_an_unclassified_extractor_warning_fails_closed` proves it load-bearing by simulating exactly that slip.

**Four fixtures were repaired rather than deleted.** `_ADK_AGENT_SOURCE` (the #386 inventory tests), `_scan_google_adk_insufficient_evidence_project` in `test_ci.py`, and the ADK repo in `test_verify.py` all existed only because everything used to be `medium`. Each now drops a parameter annotation — a warning-free, per-tool gap that keeps their `source_warnings == []` assertions valid and gives the inventory a genuine job (the real parameter types).

**Scope held deliberately.** The mechanism is reusable, but `crewai`, `langchain`, `openai_sdk_static`, `conductor`, and `n8n` still hardcode `medium` and are untouched. Each needs its own probe list; doing five at once is how the eight fail-opens above would have slipped through. Worth a follow-up issue.

**One pre-existing gap found, not filed and not fixed here.** `import google.adk.agents` followed by `google.adk.agents.LlmAgent(...)` is not recognised as an agent call at all — `_qualified_name` mangles the dotted alias — so that spelling reports zero tools. It fails closed (`SHIP-INVENTORY-NOT-ENUMERABLE`), so it is out of scope for #393. `test_a_qualified_module_path_is_not_mistaken_for_an_agent_attribute` documents it.

## Type

- [x] Input adapter change
- [ ] Documentation only

`Tool.extraction` is in-memory only and is not serialized into `report.json`, so no report schema version bump is required. The only user-visible output change is the `evidence_gaps[].why` free-text string on `low_confidence_tool` rows, which gains a trailing `Unresolved: <reasons>.` sentence for adapters that measure completeness.

## Verification

Additional local checks run:

- `ruff check .` — clean
- `compileall -q src tests` — clean
- Full suite, `pytest tests -n auto` — `PYTEST_EXIT=0`, zero failures
- End-to-end reproduction of the issue's `smart_closer` shape before and after, plus the fully-declared A/B that isolates the extraction ceiling as the terminal blocker
- 17-construct adversarial probe against `_load_python_path`; `samples/google_adk_agent` rescanned (unchanged: all four tools still `high`, no new source warnings)

One unrelated flake, `test_authorization_verify_integration.py::test_invalid_authorization_keeps_human_stop_and_receipt_valid[missing_trust]`, appeared in one intermediate `-n auto` run and did not recur; it passes standalone and on a clean tree, and nothing in this change touches authorization or trust policy.

## Release-readiness notes

- [x] New or changed check IDs are documented in `docs/checks.md` — no check IDs added or changed
- [x] Report/schema changes are additive or documented in `STABILITY.md` — no schema change; see Type note above

Closes #393

### DIFF
--- a/src/agents_shipgate/inputs/google_adk.py
+++ b/src/agents_shipgate/inputs/google_adk.py
@@ -10,6 +10,8 @@ from agents_shipgate.core.artifact_models import (
     GoogleAdkToolset,
 )
 from agents_shipgate.core.domain import (
+    SURFACE_ENUMERATED,
+    SURFACE_PARTIAL,
     AgentBindingObservation,
     AuthInfo,
     LoadedToolSource,
@@ -84,6 +86,62 @@ _SCOPE_NODES = (
     ast.Lambda,
     ast.ClassDef,
 )
+#: ``Tool.extraction["surface"]`` — whether this adapter *proved* the tool
+#: surface it reports, rather than merely produced it.
+#:
+#: Before #393 the Python AST path hardcoded ``confidence="medium"``, so no ADK
+#: repository could reach ``high`` from source however statically analysable it
+#: was, and ``insufficient_evidence`` was the framework's default first-run
+#: verdict rather than a property of any repository. A condition that holds for
+#: every input carries no information: it could not tell a toolkit factory from
+#: twelve module-level functions, and the remedy it prescribed — transcribe the
+#: twelve tools Shipgate had just extracted correctly into an inventory — added
+#: no fact to the system.
+#:
+#: ``SURFACE_ENUMERATED`` is therefore a claim the extractor has to earn on the
+#: parsed module. Every construct that leaves any part of the surface
+#: unresolved records a reason code below and holds the whole module at
+#: ``medium``; absence of the attestation reads as incomplete everywhere
+#: downstream, so a new ambiguity nobody classified fails closed.
+
+#: Module-scoped reasons: one unresolved construct anywhere in the file means
+#: this file's tool surface was not proven, so it holds every tool the file
+#: produced. Scoping them per agent instead would let a fully-resolved agent in
+#: a half-resolved module claim a proof the module cannot support.
+SURFACE_GAP_DYNAMIC_TOOLS = "dynamic_tools_expression"
+SURFACE_GAP_UNRESOLVED_REFERENCE = "unresolved_tool_reference"
+SURFACE_GAP_UNRESOLVED_EXPRESSION = "unresolved_tool_expression"
+SURFACE_GAP_UNRESOLVED_WRAPPER = "unresolved_tool_wrapper"
+SURFACE_GAP_DYNAMIC_TOOLSET = "dynamic_toolset"
+SURFACE_GAP_CONFLICTING_CONTRACT = "conflicting_tool_contract"
+SURFACE_GAP_UNRESOLVED_SUB_AGENT = "unresolved_sub_agent"
+#: The module reaches an agent's ``tools`` attribute after construction, or
+#: builds an agent from unpacked keyword arguments. Reading the ``tools=``
+#: literal proves the surface only if that literal is the whole story;
+#: ``agent.tools.append(imported)`` and ``Agent(**config)`` both make it not be.
+SURFACE_GAP_MUTABLE_TOOL_BINDING = "mutable_tool_binding"
+SURFACE_GAP_DYNAMIC_AGENT_KWARGS = "dynamic_agent_kwargs"
+#: A ``tools=`` element resolved to a definition the name may not actually
+#: refer to. ``self.functions`` is a flat, scope-blind name map, so it happily
+#: answers with a function defined inside a factory, a method lifted out of a
+#: class body, one of two conditional definitions, or a definition whose name
+#: was later rebound. Naming the tool is still useful; claiming its signature
+#: was proven is not.
+SURFACE_GAP_SHADOWED_DEFINITION = "shadowed_tool_definition"
+#: Emitted only by the fail-closed backstop in
+#: ``_PythonAdkExtractor._resolve_extraction_evidence``:
+#: a warning this module raised through neither surface helper. It means a new
+#: ambiguity was added without deciding what it says about the surface, so the
+#: module declines to claim one.
+SURFACE_GAP_UNCLASSIFIED = "unclassified_extractor_warning"
+#: Per-tool reasons: the module may be fully enumerated while one function's
+#: own callable interface still is not.
+SURFACE_GAP_UNTYPED_PARAMETER = "untyped_parameter"
+SURFACE_GAP_VARIADIC_PARAMETERS = "variadic_parameters"
+SURFACE_GAP_DECORATED_FUNCTION = "decorated_tool_function"
+#: An Agent Config lists tool *names*; there is no signature to read, so this
+#: path never claims an enumerated surface.
+SURFACE_GAP_TOOL_REFERENCE_ONLY = "tool_reference_only"
 
 
 def load_google_adk_artifacts(
@@ -422,7 +480,12 @@ def _tool_from_config_entry(
         annotations={"adk_tool_reference": name, "agent_name": agent_name},
         auth=AuthInfo(source="google_adk_config"),
         extraction_confidence="low",
-        extraction={"method": "google_adk_agent_config", "confidence": "low"},
+        extraction={
+            "method": "google_adk_agent_config",
+            "confidence": "low",
+            "surface": SURFACE_PARTIAL,
+            "surface_gaps": [SURFACE_GAP_TOOL_REFERENCE_ONLY],
+        },
     )
     tools.append(tool)
     artifacts.function_tools.append(
@@ -588,6 +651,11 @@ class _PythonAdkExtractor:
             for node in ast.walk(tree)
             if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef))
         }
+        # Names ``self.functions`` can answer for without being right. The map
+        # above is flat and scope-blind by design — it has to be, to name a
+        # tool at all — but the answer is only a *proof* of the signature when
+        # the name is defined once, at module scope, and never rebound.
+        self.unproven_function_names = _unproven_function_names(tree)
         self.wrappers = self._wrapper_assignments()
         self.toolset_assignments = self._toolset_assignments()
         # One canonical Tool per function definition, keyed by the def name.
@@ -598,6 +666,13 @@ class _PythonAdkExtractor:
         # agents is loaded once, not once per agent.
         self.toolset_tool_names: dict[int, list[str]] = {}
         self.agent_bindings: dict[str, _AdkAgentBinding] = {}
+        # Reasons this module's tool surface was not proven complete (#393).
+        # Empty at the end of ``extract`` is what earns ``SURFACE_ENUMERATED``.
+        self.surface_gaps: list[str] = []
+        # Warnings this module raised through ``_surface_warning`` or
+        # ``_note_warning``. Compared against the real growth of
+        # ``artifacts.warnings`` so an unclassified append fails closed.
+        self._accounted_warnings = 0
         # Every ``Agent(...)`` assignment in this module, walked once and
         # reused: ``extract`` iterates it, and the sub-agent spelling map
         # below is built from it.
@@ -612,9 +687,16 @@ class _PythonAdkExtractor:
     def extract(self) -> list[LoadedToolSource]:
         tools: list[Tool] = []
         loaded_sources: list[LoadedToolSource] = []
+        warnings_before = len(self.artifacts.warnings)
         self._record_eval_references()
+        self._record_mutable_tool_bindings()
         for target_name, call in self.agent_call_list:
             agent_name = _kwarg_string(call, "name") or target_name or "adk_agent"
+            if any(keyword.arg is None for keyword in call.keywords):
+                # ``Agent(**config)`` hides every keyword, ``tools`` included.
+                # Without ``tools=`` the agent silently records tool_count 0,
+                # which would otherwise read as a proven empty surface.
+                self._note_surface_gap(SURFACE_GAP_DYNAMIC_AGENT_KWARGS)
             tools_expr = _kwarg(call, "tools")
             tool_count = len(tools_expr.elts) if isinstance(tools_expr, (ast.List, ast.Tuple)) else 0
             self.artifacts.agents.append(
@@ -630,8 +712,9 @@ class _PythonAdkExtractor:
             self._record_agent_callbacks_plugins_subagents(call, agent_name)
             if not isinstance(tools_expr, (ast.List, ast.Tuple)):
                 if tools_expr is not None:
-                    self.artifacts.warnings.append(
-                        f"Google ADK agent {agent_name!r} uses a dynamic tools expression."
+                    self._surface_warning(
+                        f"Google ADK agent {agent_name!r} uses a dynamic tools expression.",
+                        SURFACE_GAP_DYNAMIC_TOOLS,
                     )
                     self.artifacts.toolsets.append(
                         GoogleAdkToolset(
@@ -648,6 +731,7 @@ class _PythonAdkExtractor:
                 loaded_sources.extend(
                     self._extract_tool_expr(item, tools, agent_name, binding)
                 )
+        self._resolve_extraction_evidence(warnings_before)
         return [
             LoadedToolSource(
                 source_id=self.source_id,
@@ -658,6 +742,102 @@ class _PythonAdkExtractor:
             ),
             *loaded_sources,
         ]
+
+    def _surface_warning(self, message: str, reason: str) -> None:
+        """Report a construct that leaves part of this module's surface unknown."""
+
+        self.artifacts.warnings.append(message)
+        self._accounted_warnings += 1
+        self._note_surface_gap(reason)
+
+    def _note_warning(self, message: str) -> None:
+        """Report a warning that says nothing about the tool surface.
+
+        Eval-artifact references are about test collateral, not about which
+        tools an agent can call, so they must not cost the module its
+        completeness claim. Everything else goes through ``_surface_warning``.
+        """
+
+        self.artifacts.warnings.append(message)
+        self._accounted_warnings += 1
+
+    def _note_surface_gap(self, reason: str) -> None:
+        if reason not in self.surface_gaps:
+            self.surface_gaps.append(reason)
+
+    def _record_mutable_tool_bindings(self) -> None:
+        """Notice any reach for an agent's ``tools`` after it is constructed.
+
+        The ``tools=`` literal is only a proof of the surface while nothing
+        else touches it. ``root_agent.tools.append(imported_tool)``,
+        ``root_agent.tools = [...]``, ``setattr(root_agent, "tools", ...)``,
+        and an alias bound with ``bucket = root_agent.tools`` all add tools the
+        walk above never sees, and every one of them was silently promoted to
+        ``high`` before this guard.
+
+        Any ``.tools`` access at all counts, not only the mutating spellings:
+        a read can be aliased and a subscript store hides behind a load. The
+        cost of over-reporting is one module that stays at ``medium``, which is
+        where it already was; the cost of under-reporting is a proven-surface
+        claim over tools nobody enumerated. Dotted module paths such as
+        ``google.adk.tools.FunctionTool`` are excluded by their imported root —
+        those are packages, not agents.
+        """
+
+        for node in ast.walk(self.tree):
+            if isinstance(node, ast.Attribute) and node.attr == "tools":
+                if not self._is_imported_module_path(node.value):
+                    self._note_surface_gap(SURFACE_GAP_MUTABLE_TOOL_BINDING)
+                    return
+            if (
+                isinstance(node, ast.Call)
+                and isinstance(node.func, ast.Name)
+                and node.func.id == "setattr"
+                and len(node.args) >= 2
+                and isinstance(node.args[1], ast.Constant)
+                and node.args[1].value == "tools"
+            ):
+                self._note_surface_gap(SURFACE_GAP_MUTABLE_TOOL_BINDING)
+                return
+
+    def _is_imported_module_path(self, node: ast.AST) -> bool:
+        """Whether ``node`` roots in a name this module imported."""
+
+        current = node
+        while isinstance(current, ast.Attribute):
+            current = current.value
+        return isinstance(current, ast.Name) and current.id in self.aliases
+
+    def _resolve_extraction_evidence(self, warnings_before: int) -> None:
+        """Settle each tool's extraction confidence on what this module proved.
+
+        Runs once, after the whole module is walked, because completeness is a
+        property of the file rather than of the agent that happened to be
+        visited first: a dynamic tools expression on the last agent invalidates
+        the proof for tools bound by the first.
+
+        The unaccounted-warning backstop is the load-bearing part. A future
+        ambiguity added with a plain ``artifacts.warnings.append`` would
+        otherwise leave ``surface_gaps`` empty and silently promote an
+        unresolved module to ``high`` — the fail-open shape a "safe" block-level
+        signal clearing a path-wide guard produces. Counting warnings makes the
+        default answer "not proven".
+        """
+
+        emitted = len(self.artifacts.warnings) - warnings_before
+        if emitted != self._accounted_warnings:
+            self._note_surface_gap(SURFACE_GAP_UNCLASSIFIED)
+        for tool in self.canonical_function_tools.values():
+            raw_gaps = tool.extraction.get("surface_gaps")
+            local_gaps = raw_gaps if isinstance(raw_gaps, list) else []
+            gaps = sorted({*self.surface_gaps, *local_gaps})
+            confidence = "medium" if gaps else "high"
+            tool.extraction["surface"] = (
+                SURFACE_PARTIAL if gaps else SURFACE_ENUMERATED
+            )
+            tool.extraction["surface_gaps"] = gaps
+            tool.extraction["confidence"] = confidence
+            tool.extraction_confidence = confidence
 
     def _agent_names_by_variable(self) -> dict[tuple[ast.AST | None, str], str | None]:
         """Map each ``variable = Agent(...)`` to the agent's declared ``name=``.
@@ -813,8 +993,9 @@ class _PythonAdkExtractor:
                     self.functions[expr.id], tools, agent_name, binding, False
                 )
             else:
-                self.artifacts.warnings.append(
-                    adk_unresolved_tool_warning(agent_name, expr.id)
+                self._surface_warning(
+                    adk_unresolved_tool_warning(agent_name, expr.id),
+                    SURFACE_GAP_UNRESOLVED_REFERENCE,
                 )
             return []
         if isinstance(expr, ast.Call):
@@ -832,8 +1013,9 @@ class _PythonAdkExtractor:
                 return []
             if call_name in OPENAPI_TOOLSET_NAMES | MCP_TOOLSET_NAMES:
                 return self._extract_toolset_call(expr, agent_name, binding)
-        self.artifacts.warnings.append(
-            f"Google ADK agent {agent_name!r} has a tool expression that could not be statically resolved."
+        self._surface_warning(
+            f"Google ADK agent {agent_name!r} has a tool expression that could not be statically resolved.",
+            SURFACE_GAP_UNRESOLVED_EXPRESSION,
         )
         return []
 
@@ -855,8 +1037,9 @@ class _PythonAdkExtractor:
                 bool(wrapper.get("long_running")),
             )
             return
-        self.artifacts.warnings.append(
-            f"Google ADK tool wrapper {wrapper_name!r} has no statically resolvable function."
+        self._surface_warning(
+            f"Google ADK tool wrapper {wrapper_name!r} has no statically resolvable function.",
+            SURFACE_GAP_UNRESOLVED_WRAPPER,
         )
 
     def _bind_function_tool(
@@ -875,6 +1058,8 @@ class _PythonAdkExtractor:
         (and collide on tool observation identity).
         """
 
+        if node.name in self.unproven_function_names:
+            self._note_surface_gap(SURFACE_GAP_SHADOWED_DEFINITION)
         tool = self.canonical_function_tools.get(node.name)
         if tool is None:
             tool = self._function_to_tool(node, agent_name, long_running)
@@ -885,9 +1070,10 @@ class _PythonAdkExtractor:
             # LongRunningFunctionTool is a contradictory declaration about one
             # action. Keep the stricter contract and route it to review rather
             # than letting binding order decide.
-            self.artifacts.warnings.append(
+            self._surface_warning(
                 f"Google ADK function {node.name!r} is bound as both a long-running "
-                "and a standard function tool; review its operation contract."
+                "and a standard function tool; review its operation contract.",
+                SURFACE_GAP_CONFLICTING_CONTRACT,
             )
             if long_running:
                 tool.annotations["long_running"] = True
@@ -985,8 +1171,15 @@ class _PythonAdkExtractor:
                 "long_running": long_running,
             },
             auth=AuthInfo(source="google_adk_static"),
+            # Provisional: ``_resolve_extraction_evidence`` settles both fields
+            # once the whole module has been walked. ``medium`` here so a tool
+            # is never high-confidence in flight.
             extraction_confidence="medium",
-            extraction={"method": "google_adk_python_ast", "confidence": "medium"},
+            extraction={
+                "method": "google_adk_python_ast",
+                "confidence": "medium",
+                "surface_gaps": _function_surface_gaps(node, parameters),
+            },
         )
         payload = self._function_tool_payload(tool, agent_name)
         self.artifacts.function_tools.append(payload)
@@ -1021,9 +1214,10 @@ class _PythonAdkExtractor:
         )
         self.artifacts.toolsets.append(toolset)
         if not spec_path:
-            self.artifacts.warnings.append(
+            self._surface_warning(
                 f"Google ADK OpenAPIToolset at {self.source_ref}:{call.lineno} "
-                "has no static local spec path."
+                "has no static local spec path.",
+                SURFACE_GAP_DYNAMIC_TOOLSET,
             )
             return []
         loaded = load_openapi_tools(
@@ -1059,9 +1253,10 @@ class _PythonAdkExtractor:
         )
         self.artifacts.toolsets.append(toolset)
         if not inventory_path:
-            self.artifacts.warnings.append(
+            self._surface_warning(
                 f"Google ADK McpToolset at {self.source_ref}:{call.lineno} "
-                "has no static MCP tool inventory path."
+                "has no static MCP tool inventory path.",
+                SURFACE_GAP_DYNAMIC_TOOLSET,
             )
             return []
         loaded = load_mcp_tools(
@@ -1121,6 +1316,14 @@ class _PythonAdkExtractor:
                     resolved = self._sub_agent_name(variable, call)
                     if resolved is None:
                         unresolved_sub_agents.append(variable)
+                        # A handoff target this module does not define owns
+                        # tools this module never saw, so the file cannot
+                        # claim it enumerated the surface reachable from
+                        # here. Deliberately no warning: #385 left an
+                        # unreached branch ungated, and adding one now would
+                        # move repositories between verdicts for a reason
+                        # this change is not about.
+                        self._note_surface_gap(SURFACE_GAP_UNRESOLVED_SUB_AGENT)
                     else:
                         sub_agent_names.append(resolved)
                 self.artifacts.sub_agents.append(
@@ -1160,12 +1363,12 @@ class _PythonAdkExtractor:
             try:
                 path = resolve_input_path(self.entrypoint_dir, value)
             except InputParseError:
-                self.artifacts.warnings.append(
+                self._note_warning(
                     f"Google ADK eval reference {value!r} resolves outside the entrypoint directory."
                 )
                 continue
             if not path.exists():
-                self.artifacts.warnings.append(
+                self._note_warning(
                     f"Google ADK eval reference {value!r} was detected but not found."
                 )
                 continue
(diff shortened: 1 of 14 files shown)


src/agents_shipgate/inputs/google_adk.py
```diff
@@ -832,8 +1013,9 @@ def _extract_tool_expr(
                 return []
             if call_name in OPENAPI_TOOLSET_NAMES | MCP_TOOLSET_NAMES:
                 return self._extract_toolset_call(expr, agent_name, binding)
-        self.artifacts.warnings.append(
-            f"Google ADK agent {agent_name!r} has a tool expression that could not be statically resolved."
+        self._surface_warning(
```
pengfei-threemoonslab, line 1224: [P1] Record unresolved inline wrappers before promoting the module. This fallback is never reached for a recognized `FunctionTool`/`LongRunningFunctionTool`: lines 1003–1013 return even when `_call_func_name()` is absent or names an imported function. With `tools=[known, FunctionTool(func=imported_tool)]`, only `known` entered the catalog, it became high/enumerated, and a fully declared scan returned `passed` with zero warnings, gaps, or findings while omitting the imported capability. Add `SURFACE_GAP_UNRESOLVED_WRAPPER` on the negative inline-wrapper branch and cover imported, attribute, lambda, and missing-`func` forms.

src/agents_shipgate/inputs/google_adk.py
```diff
@@ -1431,6 +1634,75 @@ def _parameters(node: ast.FunctionDef | ast.AsyncFunctionDef) -> list[ToolParame
...
+        if isinstance(node, ast.FunctionDef | ast.AsyncFunctionDef):
+            definitions[node.name] = definitions.get(node.name, 0) + 1
+    module_level = {
+        node.name
+        for node in tree.body
+        if isinstance(node, ast.FunctionDef | ast.AsyncFunctionDef)
+    }
+    rebound: set[str] = set()
+    for node in ast.walk(tree):
+        if isinstance(node, ast.Name) and isinstance(node.ctx, ast.Store | ast.Del):
```
pengfei-threemoonslab: [P1] This file-global rebound set does not prove lexical name resolution. Function parameters are `ast.arg`, class bindings live on `ClassDef.name`, and pattern/exception targets have other shapes. Repro: a module-level `def known(...)` plus `def build(known): return LlmAgent(tools=[known])`, invoked with `external_tool`, was resolved as the module function, marked high, and returned `passed`. A same-named class and wrapper-variable rebinding fail similarly. Resolve each tool/wrapper/toolset name scope- and order-aware, and add regressions for parameters, classes, pattern captures, exception targets, and later wrapper rebindings.

src/agents_shipgate/inputs/google_adk.py
```diff
@@ -659,6 +743,102 @@ def extract(self) -> list[LoadedToolSource]:
...
+        otherwise leave ``surface_gaps`` empty and silently promote an
+        unresolved module to ``high`` — the fail-open shape a "safe" block-level
+        signal clearing a path-wide guard produces. Counting warnings makes the
+        default answer "not proven".
+        """
+
+        emitted = len(self.artifacts.warnings) - warnings_before
+        if emitted != self._accounted_warnings:
+            self._note_surface_gap(SURFACE_GAP_UNCLASSIFIED)
+        for tool in self.canonical_function_tools.values():
```
pengfei-threemoonslab, line 986: [P1] Module gaps disappear when the module exposes only OpenAPI/MCP toolset tools, because this loop updates only `canonical_function_tools`. A resolved toolset combined with `Agent(**config)` or a later `.tools` mutation records `dynamic_agent_kwargs`/`mutable_tool_binding` internally, but the loop can be empty; the toolset operation stays high and the scan returned `passed` with a pass-eligible binding graph and no gaps. Propagate module completeness to every returned/bound tool (or publish a source/binding-level incomplete signal) and add toolset-only kwargs/mutation/sub-agent cases.

src/agents_shipgate/inputs/google_adk.py
```diff
@@ -659,6 +743,102 @@ def extract(self) -> list[LoadedToolSource]:
...
+        those are packages, not agents.
+        """
+
+        for node in ast.walk(self.tree):
+            if isinstance(node, ast.Attribute) and node.attr == "tools":
+                if not self._is_imported_module_path(node.value):
+                    self._note_surface_gap(SURFACE_GAP_MUTABLE_TOOL_BINDING)
+                    return
+            if (
+                isinstance(node, ast.Call)
```
pengfei-threemoonslab: [P1] Reflective access bypasses this mutation guard. `getattr(root_agent, "tools").append(imported_tool)` contains neither an `Attribute(attr="tools")` nor a bare `setattr`, so the known local tool remained high/enumerated and the fully declared scan returned `passed` while omitting `imported_tool`. Also, `_is_imported_module_path` exempts any root in `self.aliases` even after that name is rebound to an agent. Detect constant-string reflective access and make the imported-module exclusion binding-aware; add both regressions.

src/agents_shipgate/inputs/google_adk.py
```diff
@@ -1431,6 +1634,75 @@ def _parameters(node: ast.FunctionDef | ast.AsyncFunctionDef) -> list[ToolParame
...
+      :func:`_json_schema_type`'s fallback. That is a guess presented as a
+      schema, which is exactly what a high-confidence extraction may not do.
+    """
+
+    gaps: list[str] = []
+    if node.decorator_list:
+        gaps.append(SURFACE_GAP_DECORATED_FUNCTION)
+    if node.args.vararg is not None or node.args.kwarg is not None:
+        gaps.append(SURFACE_GAP_VARIADIC_PARAMETERS)
+    if any(parameter.type is None for parameter in parameters):
```
pengfei-threemoonslab: [P1] Annotation presence is not proof that the emitted schema is faithful. `_json_schema_type` falls back to `"string"` for `set[str]`, `int | None`, tuples, custom/Pydantic models, and other unsupported forms, yet this code records no gap and promotes the tool to high/enumerated. End-to-end probes produced an incorrect string schema and `release_decision="passed"`. Make conversion report whether it was lossless (including return types), or record an unsupported/partial-annotation surface gap; then test these forms.

pengfei-threemoonslab (commented):
Blocking expert review: the direction is sound and the happy-path tests/CI are green, but the new high-confidence promotion still has five independent fail-open paths. In each reproduced case, the report reached `release_decision="passed"` with no evidence gaps while a reachable tool was omitted or its interface was represented by a guessed schema.

This is a COMMENT review because the authenticated GitHub account owns this PR; GitHub does not allow an author to submit REQUEST_CHANGES on their own PR. Please treat the inline P1 findings as merge-blocking.

Validation performed:
- exact-head GitHub Actions: CI and both Agents Shipgate workflows succeeded;
- focused changed-area pytest suite: passed;
- `git diff --check`: clean;
- adversarial end-to-end `run_scan` probes: reproduced the five false-pass cases below;
- verifier: `review_publishable` / `review_required` for the touched trust root (merge remains human-owned).

Non-blocking docs follow-up: `docs/examples.md` still says AST-only binding graphs cannot qualify for an evidence-backed pass, which this PR intentionally changes.

src/agents_shipgate/inputs/google_adk.py
pengfei-threemoonslab: Confirmed and fixed. Reproduced exactly as described: `tools=[known, FunctionTool(func=imported_tool)]` gave `catalog=[('known','high')]`, zero warnings, zero gaps, `passed`.

The negative branch now records `SURFACE_GAP_UNRESOLVED_WRAPPER` with a warning naming the agent and the unresolved `func`. Parametrized regressions cover all four spellings you listed plus the long-running variant: imported function, attribute (`mod.thing`), lambda, missing `func`, and `LongRunningFunctionTool(func=imported)`. All five now report `medium` / `unresolved_tool_wrapper` and block.

src/agents_shipgate/inputs/google_adk.py
pengfei-threemoonslab: Confirmed and fixed. Your repro reproduced verbatim — the parameter resolved as the module function, high, `passed`. Same for a same-named class and a rebound wrapper variable.

Replaced the name/count/rebound heuristic with `_name_binding_occurrences`, which collects every binding form: `FunctionDef`/`AsyncFunctionDef`/`ClassDef.name`, `Name` Store/Del, `ast.arg`, `ast.alias`, `ExceptHandler.name`, `MatchAs`/`MatchStar.name`, `MatchMapping.rest`, and `Global`/`Nonlocal`. `_name_is_proven` requires exactly one occurrence, `_scope_of(...) is self.tree`, and a top-level statement that is one of `Assign`/`AnnAssign`/`FunctionDef`/`AsyncFunctionDef`/`ClassDef`/`Import`/`ImportFrom` — so a module-level `if X: helper = ...` is also refused. It is applied to the wrapper and toolset *variable* names too, since both maps are last-write-wins.

Regressions added for parameters, classes, `except ... as`, `case ... as`, `global`, and later wrapper rebinding.

One scoping note worth your call: I made mis-resolution *refuse to prove* rather than resolve correctly. A genuinely scope- and order-aware resolver would change which tools land in the catalog for every existing ADK repo, which is a larger change than #393 and not one I wanted to smuggle in behind a confidence fix. The fail-open you found is closed either way — the residual is that the catalog may still *name* the wrong definition, now at `medium` with `shadowed_tool_definition` rather than as a proven surface. Happy to open a follow-up for the real resolver if you'd rather have it here.

src/agents_shipgate/inputs/google_adk.py
pengfei-threemoonslab: Confirmed and fixed — this was the one I had not considered at all. A resolved `McpToolset` beside `Agent(**config)` recorded `dynamic_agent_kwargs` into a loop over `canonical_function_tools` that was empty, and the gap evaporated: `catalog=[('remote_lookup','high')]`, `gaps={}`.

`_resolve_extraction_evidence` now also takes `loaded_sources` and, when `self.surface_gaps` is non-empty, stamps the module reasons onto every tool a toolset in this module contributed. It only ever lowers — `high` becomes `medium`, `low` stays `low` — so this step can never promote a tool the adapter did not extract, and a toolset in a clean module keeps its `high` (covered by its own test).

Regressions added for the toolset-only module with `**kwargs`, with a `.tools` mutation, and with an unresolved sub-agent, plus the clean control.

src/agents_shipgate/inputs/google_adk.py
pengfei-threemoonslab: Confirmed and fixed, both halves.

Reflective access: added `_is_reflective_tools_access`, matching `getattr`/`setattr`/`delattr` with a constant `"tools"` second argument. It matches on the *trailing* callee name, so `builtins.setattr(...)` counts as well as the bare builtin — regressions for `getattr(...).append`, `builtins.setattr`, and `delattr`.

Imported-root exemption: `_is_imported_module_path` now requires the root name to have exactly one binding occurrence *and* for that binding to be an `ast.alias`. `from x import agents` followed by `agents = LlmAgent(...)` gives two occurrences, so it is no longer treated as a package path. The `google.adk.tools.FunctionTool` case that motivated the exemption still passes.

src/agents_shipgate/inputs/google_adk.py
pengfei-threemoonslab: Confirmed and fixed. All of `set[str]`, `int | None`, `tuple[int, str]`, and a custom class shipped `{"type": "string"}` on a high/enumerated tool and returned `passed`.

Rather than maintain a second type vocabulary alongside the emitter's, `_annotation_is_faithful` asks `_json_schema_type` what it *would* emit and compares that to what `_annotation_json_type` says the annotation denotes (read from the AST, not the unparsed string). Anything the emitter mishandles is unfaithful by construction, including spellings added later — which is how `typing.List[str]` is caught: the type is representable, but the emitter's string match misses the module prefix and emits a scalar. Return annotations go through the same check since `output_schema` uses the same fallback; an *absent* return annotation stays an honest omission (`output_schema` is `{}`) and is not a gap.

New reason code `unrepresentable_annotation`. `_bound_args` was extracted so the schema and the faithfulness check cannot disagree about which arguments they are describing. Regressions cover all your listed forms plus `Optional[int]`, `Literal[...]`, `list[SomeModel]`, `dict[int, str]`, `typing.List[str]`, and unrepresentable returns — with a positive-control table (`str`, `int`, `float`, `bool`, bare `list`/`dict`, `list[str]`, `dict[str, int]`, `list[dict[str, str]]`) asserting the check does not demote what the emitter gets right.

pengfei-threemoonslab:
All five P1 findings confirmed and fixed in 68650f73. Each is replied to inline with the specific change; the short version:

| # | Fail-open | Reason code now recorded |
|---|---|---|
| 1 | `FunctionTool(func=<not defined here>)` returned silently | `unresolved_tool_wrapper` |
| 2 | Binding model missed `ast.arg`, `ClassDef.name`, `except/case/global` | `shadowed_tool_definition` |
| 3 | Module gaps only reached `canonical_function_tools` | module reasons now reach toolset tools |
| 4 | `getattr(agent, "tools")`; alias-map exemption after rebinding | `mutable_tool_binding` |
| 5 | Annotation presence read as schema fidelity | `unrepresentable_annotation` (new) |

**I reproduced all five before changing anything**, and the adversarial probe set now stands at 20 constructs — every one blocks, against a 6-case control group that still reaches `passed`. Thank you for these; #3 in particular is a class I had not considered at all (I had reasoned about module completeness entirely in terms of function tools), and #5's `typing.List[str]` case showed that a hand-maintained type vocabulary would have drifted from the emitter, so the faithfulness check now asks the emitter directly rather than keeping a second list.

**One scoping decision I would like your call on.** For #2 I made mis-resolution *refuse to prove* rather than resolve correctly. A genuinely scope- and order-aware resolver would change which tools land in the catalog for every existing ADK repository, which is a larger change than #393 and not one I wanted to smuggle in behind a confidence fix. The fail-open is closed either way; the residual is that the catalog may still *name* the wrong definition — now at `medium` with `shadowed_tool_definition` attached instead of as a proven surface. Say the word if you would rather have the real resolver in this PR, and I will open a follow-up otherwise.

Also picked up the non-blocking docs item: `docs/examples.md` no longer claims an AST-only binding graph cannot qualify for an evidence-backed pass, and now points at the ADK section for exactly what "fully resolved" requires.

Verification on the new head: `ruff check .` clean, `compileall -q src tests` clean, full `pytest -n auto` suite `PYTEST_EXIT=0` with zero failures. The #393 headline scenarios are unchanged — a fully static ADK project with actions declared still reaches `passed` at 12/12 pass-eligible, and `samples/google_adk_agent` still scans identically with all four tools `high` and no new source warnings.

src/agents_shipgate/inputs/google_adk.py
```diff
@@ -659,6 +747,167 @@ def extract(self) -> list[LoadedToolSource]:
...
+            tool.extraction["confidence"] = confidence
+            tool.extraction_confidence = confidence
+        if not self.surface_gaps:
+            return
+        module_gaps = sorted(self.surface_gaps)
+        for loaded in loaded_sources:
+            for tool in loaded.tools:
+                raw_gaps = tool.extraction.get("surface_gaps")
+                local_gaps = raw_gaps if isinstance(raw_gaps, list) else []
+                tool.extraction["surface"] = SURFACE_PARTIAL
```
pengfei-threemoonslab: [P1] Preserve member incompleteness through tool-identity merging. This correctly lowers an ADK-loaded OpenAPI observation, but a reviewed identity binding to the same operation from a direct OpenAPI source later deep-copies the high-confidence primary and does not combine `extraction.surface`, `surface_gaps`, or confidence from members. Repro: `OpenAPIToolset(api.yaml)` on `LlmAgent(..., **overrides)`, the same `api.yaml` configured directly, and an identity binding between the two `lookup_record` observations. The ADK member has `dynamic_agent_kwargs`/medium here, yet the canonical catalog and agent binding become high/pass-eligible and the final decision is `passed` with no gaps. An identity assertion proves operation sameness, not that the ADK module exposes no additional tools; conservatively carry the worst member completeness through canonicalization and add this end-to-end case.

src/agents_shipgate/inputs/google_adk.py
```diff
@@ -803,8 +1052,13 @@ def _extract_tool_expr(
     ) -> list[LoadedToolSource]:
         if isinstance(expr, ast.Name):
             if expr.id in self.wrappers:
+                # The variable's own name has to hold up too, not just the
+                # function it wraps: a wrapper reassigned later resolves
+                # through a last-write-wins map.
+                self._require_proven_name(expr.id)
```
pengfei-threemoonslab, line 1178: [P1] Apply the same binding proof to every recognized framework constructor, not only the wrapper/toolset variable or wrapped function. With `from google.adk.tools import FunctionTool; FunctionTool = imported_replacement; tools=[FunctionTool(func=known)]`, `_qualified_name` still resolves the stale import alias, treats the replacement factory as Google's wrapper, catalogs `known` high, and returns `passed` even when the factory exposes a different capability. Rebinding `LlmAgent` itself has the same result. Require proven provenance for Agent, wrapper, and toolset callees before trusting their semantics; otherwise record a module gap.

src/agents_shipgate/inputs/google_adk.py
```diff
@@ -1423,12 +1722,214 @@ def _parameters(node: ast.FunctionDef | ast.AsyncFunctionDef) -> list[ToolParame
     for arg, default in zip(positional_args, positional_defaults, strict=True):
         if arg.arg in {"self", "ctx", "context", "tool_context"}:
             continue
-        parameters.append(_parameter(arg, required=default is None))
+        bound.append((arg, default is None))
```
pengfei-threemoonslab, line 1879: [P1] Do not discard a parameter solely because its name is `ctx` or `context`. `def known(context: str, record_id: str) -> dict` produces a schema containing only `record_id`, is promoted high/enumerated, and returns `passed` with no gap even though `context: str` is an ordinary model-visible input. Google ADK identifies injected context by its type annotation (with `tool_context` as the name fallback), not by arbitrary `ctx`/`context` names; see the official [FunctionTool implementation](https://github.com/google/adk-python/blob/main/src/google/adk/tools/function_tool.py). Only omit a statically verified injected parameter; otherwise retain it or demote the interface.

src/agents_shipgate/inputs/google_adk.py
```diff
@@ -1423,12 +1722,214 @@ def _parameters(node: ast.FunctionDef | ast.AsyncFunctionDef) -> list[ToolParame
...
+    """The JSON type this annotation really denotes, or None if it denotes none.
+
+    Reads the annotation as a tree rather than as the unparsed string, so
+    ``set[str]``, ``tuple[int, str]``, ``int | None``, ``Optional[int]``,
+    ``Literal[...]``, a string forward reference, and any custom or Pydantic
+    class all answer None instead of silently becoming a scalar.
+    """
+
+    if isinstance(node, ast.Name):
+        return _SCALAR_ANNOTATION_TYPES.get(node.id)
```
pengfei-threemoonslab: [P1] Resolve supported annotation spellings against binding provenance before trusting them. `from domain import Account as str; def process(account: str) -> dict: ...` makes ADK see `Account` at runtime, but this spelling-only lookup emits a string schema, marks the tool high/enumerated, and allows `passed` with no gap. `list`, `dict`, `List`, and `Dict` can be shadowed the same way. Only unshadowed builtins and canonical typing aliases should qualify as faithful; rebound or ambiguous names need `unrepresentable_annotation` (or an equivalent gap).

src/agents_shipgate/inputs/google_adk.py
```diff
@@ -1423,12 +1722,214 @@ def _parameters(node: ast.FunctionDef | ast.AsyncFunctionDef) -> list[ToolParame
...
+        return False
+    func = node.func
+    called = (
+        func.id
+        if isinstance(func, ast.Name)
+        else func.attr
+        if isinstance(func, ast.Attribute)
+        else None
+    )
+    if called not in _REFLECTIVE_ATTRIBUTE_BUILTINS:
```
pengfei-threemoonslab: [P1] Resolve aliases and equivalent constant-name access when detecting reflective tool mutations. `from builtins import getattr as read_attr; read_attr(root_agent, "tools").append(imported_tool)` calls the real builtin, but this raw spelling check sees `read_attr`, omits the added runtime tool, and still returns `passed` at high confidence with no gap. `vars(root_agent)["tools"]` and `root_agent.__dict__["tools"]` also bypass the guard. Resolve builtin aliases through import provenance and conservatively recognize these constant `"tools"` access paths.

src/agents_shipgate/inputs/google_adk.py
```diff
@@ -1423,12 +1722,214 @@ def _parameters(node: ast.FunctionDef | ast.AsyncFunctionDef) -> list[ToolParame
...
+
+    for node in ast.walk(tree):
+        if isinstance(node, ast.FunctionDef | ast.AsyncFunctionDef | ast.ClassDef):
+            record(node.name, node)
+        elif isinstance(node, ast.Name) and isinstance(node.ctx, ast.Store | ast.Del):
+            record(node.id, node)
+        elif isinstance(node, ast.arg):
+            record(node.arg, node)
+        elif isinstance(node, ast.alias):
+            record(node.asname or node.name.split(".", 1)[0], node)
```
pengfei-threemoonslab, line 1989: [P1] Treat a star import as an unknown binding for every name it may introduce. In `def known(...): ...; from external import *; root = LlmAgent(..., tools=[known])`, the alias is recorded only under `"*"`, so `_name_is_proven("known")` succeeds even though the import can replace `known` at runtime. The extractor catalogs the stale local definition high and returns `passed`. A reachable `import *` should invalidate otherwise-proven names (or add a module-wide shadowing gap), with a regression asserting the release cannot pass.

pengfei-threemoonslab (commented):
Blocking second-pass review of `68650f7`: the follow-up commit closes all five reproductions from my first review, and those probes now fail closed with their expected named gaps. The new proof machinery still has six independent false-pass paths, though. For each inline P1 below, an end-to-end scan with otherwise complete declarations returned `release_decision="passed"`, high/enumerated evidence, and no warnings, evidence gaps, or findings while the runtime capability surface or schema was incomplete or wrong.

This is a COMMENT review because the authenticated account owns the PR; GitHub does not allow an author to request changes on their own PR. Please treat the six inline P1s as merge-blocking.

Validation:
- exact-head GitHub Actions: CI, Agents Shipgate, and Self Dogfood all succeeded;
- focused changed-area pytest suite and all 8 packaging tests passed;
- `git diff --check` and the worktree are clean;
- the exact base/head verifier succeeded, with `control_state=review_publishable` / `decision=review_required` for the touched trust root.

Two non-blocking follow-ups:
- a propagated ADK module gap on an OpenAPI/MCP-loaded tool is still routed by that tool's source type, so remediation asks for an inventory/spec that already exists instead of naming the ADK `dynamic_agent_kwargs` or mutation that must be fixed;
- `docs/upstream-integrations.md` says every listed reason holds the whole file at medium, but per-function signature/annotation gaps only lower the affected tool (as the same page later states).

src/agents_shipgate/inputs/google_adk.py
pengfei-threemoonslab: Confirmed and fixed. Reproduced with your exact setup — ADK `OpenAPIToolset(spec_path="api.yaml")` under `LlmAgent(..., **overrides)`, the same `api.yaml` as a direct source, and an identity binding across the two `lookup_record` observations: canonical `high`, `gaps={}`, `passed`.

Your framing is the fix. "An identity assertion proves operation sameness, not that the ADK module exposes no additional tools" is exactly the line the code was missing, so reasons are now split into two kinds:

- **Interface-scoped** (`untyped_parameter`, `variadic_parameters`, `decorated_tool_function`, `unrepresentable_annotation`, `tool_reference_only`) — claims about one tool's own schema. A reviewed inventory or identity binding legitimately closes these; that is what #386 is for, and the primary still wins.
- **Set-scoped** (everything else) — claims about *which tools exist*. Nothing a human can say about one tool establishes that a module exposes no others.

The adapter records the distinction as `extraction["tool_set_proven"]`, and `_carry_unproven_tool_set` in `core/tool_identity.py` carries a `False` from any member onto the canonical tool, capping it at `medium` and merging the member's reasons. Your case now returns `insufficient_evidence` with `dynamic_agent_kwargs` named.

I deliberately did *not* take the simpler "min confidence over members", because that would have undone #386 — an inventory completing a medium AST observation would drop straight back to medium. Its tests confirm that still works.

Also took your first non-blocking note here, since it is the same row: an unproven set no longer routes remediation by source type. It was asking for a spec that was already present and complete; it now names the construct to fix and says plainly that an inventory cannot close it.

src/agents_shipgate/inputs/google_adk.py
pengfei-threemoonslab: Confirmed and fixed. Both spellings reproduced as `passed` with `known` catalogued high.

Added `_require_proven_framework_symbol`, applied wherever a call is classified as a framework constructor rather than only where its arguments are read: the `Agent`/`LlmAgent` call in the extract loop, the inline `FunctionTool`/`LongRunningFunctionTool` branch, `_extract_toolset_call`, and the wrapper-variable path through `_append_wrapper_tool`. It walks the callee to its root `Name` and requires that name to have exactly one binding occurrence which is an `ast.alias` — so the resolution through the alias table means something. Otherwise the module records `shadowed_framework_symbol`.

This shares the binding table with the `shadowed_tool_definition` check, so `from google.adk.agents import LlmAgent` in a module that also does `from x import *` fails for that reason too.

src/agents_shipgate/inputs/google_adk.py
pengfei-threemoonslab: Confirmed and fixed, and thank you for the link to the ADK source — the name list was inherited from the OpenAI SDK adapter's conventions and was simply wrong for ADK.

I took the "retain" option rather than "demote", because the schema was the thing that was wrong: demoting would have left an incorrect `input_schema` in place that a reviewed inventory could then have "closed" over. `_is_injected_context_arg` now omits a parameter only when it is verifiably injected:

- its annotation resolves through the alias table to `ToolContext` / `CallbackContext` / `ReadonlyContext`, or
- it is named `tool_context` (ADK's documented name fallback), or
- it is `self` **at position 0** — previously `self` was dropped from any position.

`ctx` and `context` are gone from the name list. Verified end to end:

```
context: str, record_id: str    -> properties ['context', 'record_id'], high
ctx: ToolContext, record_id     -> properties ['record_id'],            high
tool_context, record_id: str    -> properties ['record_id'],            high
ctx, record_id: str             -> properties ['ctx', 'record_id'],     medium (untyped_parameter)
```

The last row is the one I want to draw attention to: a parameter *named* `ctx` with no annotation is no longer silently deleted, it is a real parameter with an unreadable type, which is honest and blocks. Regressions cover all four.

src/agents_shipgate/inputs/google_adk.py
pengfei-threemoonslab: Confirmed and fixed. `from domain import Account as str` reproduced as `passed` with a string schema over a model, and so did the `list` variant.

`_annotation_json_type` now takes a `name_is_canonical` predicate, supplied by the extractor from the same binding table the other checks use:

- a builtin (`str`, `int`, `float`, `bool`, `list`, `dict`) is canonical only while the module binds **nothing** of that name;
- `List` / `Dict` are canonical only when their single binding is an import resolving into `typing` (or `typing_extensions`).

Anything else answers `None`, which makes the annotation unfaithful and records `unrepresentable_annotation`. The check recurses, so `dict[str, int]` also requires `str` itself to be canonical.

One thing this surfaced that I fixed rather than papered over: `from typing import List` with `List[str]` was *already* emitting `{"type": "string"}`, because `_json_schema_type`'s string match only looked for a lowercase `list[` prefix. The new faithfulness check correctly refused to certify that — but the right fix is the emitter, not holding the user's tool at medium for our gap, so `_json_schema_type` now recognises `List[...]`/`Dict[...]` too. There is a positive-control test asserting a genuine `from typing import List` stays proven.

src/agents_shipgate/inputs/google_adk.py
pengfei-threemoonslab: Confirmed and fixed — all three reproduced as `passed`.

`_is_reflective_tools_access` now takes the alias table and resolves the callee through `_qualified_name` before matching on the trailing segment, so `from builtins import getattr as read_attr` resolves to `builtins.getattr` and matches. That also covers `import builtins` + `builtins.getattr(...)` and any other aliasing of the same builtin.

The two mapping forms are handled as a separate branch: a `Subscript` with a constant `"tools"` slice, where the subscripted expression is either a `vars(...)` call or a `__dict__` attribute.

I kept that branch narrow rather than flagging every `["tools"]` subscript. A module holding `CONFIG = {"tools": [...]}` is ordinary and would otherwise be demoted for a dictionary key, so there is a positive-control test asserting that case stays proven. If you would rather have the broader rule and eat the false positives, say so — the trade is one line.

src/agents_shipgate/inputs/google_adk.py
pengfei-threemoonslab: Confirmed and fixed. Reproduced as `passed`, and the mechanism is exactly as you describe — the alias is recorded under `"*"`, so `known` had one binding occurrence and read as proven.

`_record_star_imports` runs once per module in `extract()` and records `star_import_shadowing` if any `ImportFrom` carries a `*` alias. I took the module-wide route rather than trying to invalidate individual names: a star import can replace anything the module defines, so no name in the file is provable, and one clearly-named reason is more useful to a reader than `shadowed_tool_definition` appearing on every tool with no explanation of why.

That also covers the framework symbols — `LlmAgent` and `FunctionTool` are no more provable than `known` under a star import — since the module cannot be proven at all once the reason is recorded. Regression asserts the release cannot pass and that the reason is named in the gap.

pengfei-threemoonslab:
All six P1 findings confirmed and fixed in `fe5803bd`, each replied to inline. Short version:

| # | Fail-open | Fix |
|---|---|---|
| A | Identity merge discarded member incompleteness | reasons split interface- vs set-scoped; set-scoped survive the merge |
| B | Framework symbols trusted by spelling | `Agent`/wrapper/toolset callees must root in a singly-bound import |
| C | Injected context identified by name | identified by type, `tool_context` name fallback, `self` at position 0 only |
| D | Annotation spellings trusted without provenance | builtins canonical only while unbound; `List`/`Dict` only from `typing` |
| E | Reflective mutation had three more spellings | builtin resolved through aliases; `vars()`/`__dict__` handled |
| F | `import *` recorded only under `"*"` | module-wide `star_import_shadowing` |

**The one that changed my model of the problem is A.** I had been treating "confidence" as a single scalar that a reviewed assertion could always raise. Your sentence — *an identity assertion proves operation sameness, not that the ADK module exposes no additional tools* — is the distinction the code was missing, and it turns out to be the thing that makes #386 and #393 coexist. Reasons are now two kinds: **interface-scoped** ones a reviewed inventory legitimately closes (that is what #386 is for, and the primary still wins), and **set-scoped** ones nothing said about a single tool can close, which now travel with the observation as `extraction["tool_set_proven"]` and cap the canonical tool wherever it is merged. I deliberately avoided the simpler "min confidence over members", which would have silently undone #386.

**C was a plain correctness bug, and I took your "retain" option rather than "demote".** Demoting would have left an incorrect `input_schema` in place that a reviewed inventory could then close over. So `def known(context: str, record_id: str)` now emits both properties and passes honestly, while `def known(ctx, record_id: str)` keeps `ctx` as a real parameter with an unreadable type and blocks on `untyped_parameter`. Thank you for the link to `function_tool.py` — the name list was inherited from the OpenAI SDK adapter's conventions and was simply wrong for ADK.

**One emitter bug surfaced on the way.** `from typing import List` with `List[str]` was already emitting `{"type": "string"}` — `_json_schema_type`'s string match only looked for a lowercase `list[` prefix. The new faithfulness check correctly refused to certify it, but the right fix is the emitter rather than holding a user's tool at medium for our gap, so `List[...]`/`Dict[...]` are now recognised. Positive-control test included.

Both non-blocking follow-ups are done: an unproven tool set no longer routes remediation by source type (it was asking for a spec already present and complete — it now names the construct to fix and says an inventory cannot close it), and `docs/upstream-integrations.md` now distinguishes reasons that hold the whole file from per-function ones that lower only the affected tool.

Adversarial probe set is now 26 constructs across both reviews, all blocking, with an 8-case control group that still reaches `passed`.

Verification on `fe5803bd`: `ruff check .` clean, `compileall -q src tests` clean, full `pytest -n auto` suite `PYTEST_EXIT=0` with zero failures. Headline #393 scenarios unchanged — declared static ADK project still `passed` at 12/12 pass-eligible — and `samples/google_adk_agent` still scans with all four tools `high` and no source warnings.

The scoping question from my last summary still stands if you want it answered: for `shadowed_tool_definition` I made mis-resolution refuse to prove rather than resolve correctly, leaving a real scope- and order-aware resolver as a follow-up.

pengfei-threemoonslab:
Rebased onto `main` to clear the merge conflict — `52e88c3f` (#394/#395) landed in the meantime. Commit SHAs changed: `a2d7a8c3 → 016643bd`, `68650f73 → 59aa9661`, `fe5803bd → 50b3257d`.

The only overlap was `CHANGELOG.md`, where both branches added an entry under `## Unreleased`. Both are kept verbatim, with the #394/#395 entry leading since it landed first. No code conflicted — that change is in `cli/discovery/` and `cli/verify/`, this one in `inputs/google_adk.py`, `core/`, and `ci/release_decision.py`.

Re-verified on the rebased tree rather than assuming the merge was semantically clean: `ruff check .` clean, `compileall -q src tests` clean, full `pytest -n auto` suite `PYTEST_EXIT=0` with zero failures, `git diff --check` clean. The #393 behaviour is unchanged — declared static ADK project still `passed` at 12/12 pass-eligible, `samples/google_adk_agent` still all four tools `high` with no source warnings.

Nothing else changed; the review threads above still describe the current code.

Issue #362, opened 2026-08-11
insufficient_evidence reason ranks symptom over action, and first_recommended_action contradicts it


pengfei-threemoonslab:
## Summary

On an `insufficient_evidence` verdict, the headline `reason` counts source warnings (the symptom) while the one actionable gap is demoted to a secondary line — and the `Next action:` line printed immediately below actively contradicts it, telling the user no machine-actionable fix exists and to "gather deeper evidence."

## Evidence (reproduced 2026-08-10, CLI 0.16.0b7)

Stage-2 stdout, verbatim, three consecutive lines:

```
Reason: Evidence coverage below threshold (6 source warning(s)); scan results are not
        trustworthy enough to gate release.
Improve evidence: Add a reviewed closed-world binding declaration and rerun verification.
        Target: shipgate.yaml#agent_bindings.declarations.
Next action: Evidence coverage below threshold (6 source warning(s)); ... Surface this to
        the user and gather deeper evidence (e.g. MCP/OpenAPI inputs, eval traces,
        additional source files) before re-running the scan; applying patches does not
        clear an evidence verdict, so no machine-applicable fix is available.
```

Line 2 is specific and correct. Line 3 is generic, contradicts line 2, and is the field a coding agent is contractually told to read (`agent_summary.first_recommended_action`). The six warnings named in lines 1 and 3 are *not* independently fixable — they are the same fact ("ADK static analysis cannot resolve imported tool symbols") restated six times, and they persist unchanged after the binding gap is closed (still 6 at the final `blocked` verdict).

Secondary evidence — the wrong lever gets no correction signal. Declaring `tool_identity.bindings` against a source that produced zero observations raised `source_warning_count` from 6 to 12 with six copies of:

```
Invalid tool binding 'bind_spraay_batch_eth': member source_id='adk_crypto_payroll_agent',
tool='spraay_batch_eth' matched 0 observations
```

No message states the actual rule (a source with no observations cannot be a binding member; `agent_bindings` is the surface for this).

## Impact

The gating engine is right; the ranking is wrong. A consumer that reads `reason` and `first_recommended_action` — which is exactly what `docs/agent-contract-current.md` instructs — gets "no fix available, go gather more evidence" when a precise one-block fix is available and already computed. For a coding agent this is a dead end, and the cheap escape routes from a dead end are the ones enumerated in `forbidden_actions`.

## Proposal

[...]

PR #268, opened 2026-07-11
P0: prove actual agent-to-tool bindings

pengfei-threemoonslab:
## What changed

This implements the P0 binding milestone: Agents Shipgate no longer assumes that every extracted tool declaration belongs to the configured agent.

- Adds one deterministic `AgentBindingGraphAssessment` rooted at the configured agent.
- Separates the complete `tool_catalog` from proven-reachable `tool_inventory` capabilities.
- Normalizes direct tools, tool nodes, toolsets, workflows, subagents, and handoffs into canonical binding edges with provenance and completeness.
- Adds reviewed, closed-world `agent_bindings` declarations with exact canonical selectors.
- Makes semantic assessments, controls, action facts, capability facts, locks, diffs, and release decisions consume only the root-reachable surface.
- Keeps unbound catalog tools visible for audit without allowing them to create capability findings or affect the verdict.
- Routes missing, partial, dynamic, ambiguous, malformed, and conflicting binding evidence to `insufficient_evidence` outside the suppressible Finding model.

Framework normalization covers OpenAI Agents SDK, LangChain/LangGraph, CrewAI, Google ADK, n8n, and exact Codex plugin MCP inventories. MCP, OpenAPI, OpenAI tool JSON, and Anthropic artifacts remain catalogs unless structural wiring or a reviewed declaration binds them. Conductor remains non-pass-eligible until it emits the same normalized graph contract.

## Why

The previous pipeline conflated extraction with reachability: every catalog declaration was assigned to one synthetic agent. That could both attribute unrelated provider tools to an agent and evaluate controls against capabilities the agent could not actually call. Conversely, it could not defend a `passed` result with evidence of the deployed agent's complete reachable tool and handoff graph.

This change makes binding a first-class static evidence dimension. A pass now requires an unambiguous root, a complete reachable graph, and pass-eligible identity, binding, effect, and authority evidence for every reachable capability. Static analysis still does not prove runtime behavior.

## Public contract and migration

- Package: `0.16.0b2`
- Runtime contract: `13`
- Report schema: `0.31`
- Packet schema: `0.10`
- Capability standard: `0.4`
- Capability lock / diff: `0.5` / `0.6`
- Action snapshot: `0.3`
- Safety corpus, receipt index, and qualification formats: v2

Pre-0.31 binding data and mixed lock versions fail closed with regeneration guidance. There is no legacy catalog-as-agent runtime switch; users requiring that behavior must pin `0.16.0b1`.

## Validation

- Full pytest suite passed.
[...]

PR #401, opened 2026-08-21
fix(scaffold): carry the vocabulary, and reach the scans that need it (#361, #388)

### LINKED ISSUE #361: Emit the declaration scaffold for binding gaps, and from the first scan
## Summary

`suggested-declarations.yaml` — the single highest-value onboarding artifact we ship — is only emitted once the binding layer is already closed. During the two scans where the user is most stuck, it is not written at all, and the binding gap that *is* reported carries `declaration_template: null`.

## Evidence (reproduced 2026-08-10, CLI 0.16.0b7, contract 21)

Target: `google/adk-samples` PR #1917 (Google ADK agent; 6 tools, all imported symbols — 4 from an external package).

| Stage | Manifest | Decision | `suggested-declarations.yaml` |
|---|---|---|---|
| 1 | bare `init` output | `insufficient_evidence` | **not written** |
| 2 | + `google_adk.tool_inventories` | `insufficient_evidence` | **not written** |
| 3 | + `agent_bindings` | `insufficient_evidence` (12 semantic gaps) | ✅ written |
| 4 | + `action_surface.actions` | `blocked` (4 CRITICAL) | ✅ written |

At stage 2 the binding gap is correctly reported and correctly routed:

```json
{"kind": "declare_agent_bindings",
 "path": "shipgate.yaml#agent_bindings.declarations",
 "accepted_values": ["agent", "complete:true", "tools", "handoffs", "reason"],
 "declaration_template": null}
```

`accepted_values` names the *keys*; the user still has to hand-author the block. By contrast every semantic gap at stage 3 carries a full `declaration_template` **and** a merge-ready YAML block with a `# merge into: shipgate.yaml#…` pointer.

Cost of the asymmetry: hand-authoring `inventories/function-tools.json` (98 lines) plus the `agent_bindings` block, with one round wasted on an invalid enum (`auth.mode: api_key`; accepted set is `none|scoped|unscoped|ambient`) that the scaffold would have pre-filled correctly.

## Impact

Binding is the *most* mechanical layer — the six tool names are already extracted from the `tools=[...]` list and sit in `tool_catalog[]` at stage 2. Nothing about scaffolding them requires human judgment; only `complete: true` does. We ask the user to retype data we already hold, at exactly the point where they have the least context.

## Proposal

1. Emit `suggested-declarations.yaml` whenever **any** gap has a next action, including the first scan — not only when semantic gaps exist.
2. Populate `declaration_template` for `missing_binding_evidence`, pre-filling `agent`, `tools[]` (from `tool_catalog[]`), and `handoffs: []`, leaving `complete` and `reason` as `<REVIEW_REQUIRED>`.
3. When a `google_adk` source yields zero observations but the agent's `tools=[...]` names are extractable, also scaffold a `tool_inventories` JSON stub with names pre-filled and schemas/enums left as `<REVIEW_REQUIRED>`.

## Before / After

**Before** (stage 2, stdout): one `Improve evidence:` line pointing at a YAML path the user must author from the docs.

**After** (stage 2, stdout):
```
Improve evidence: Add a reviewed closed-world binding declaration.
  Scaffold written: agents-shipgate-reports/suggested-declarations.yaml
  6 tool names pre-filled; set `complete:` and `reason:` and merge into shipgate.yaml.
```

## Acceptance criteria

- [ ] Stage-1 and stage-2 scans in the repro emit `suggested-declarations.yaml`.
- [ ] The stage-2 scaffold contains an `agent_bindings` block listing all 6 tools with correct `source_id`s, and no invented `complete`/`reason` values.
- [ ] Merging the scaffold verbatim after replacing `<REVIEW_REQUIRED>` closes `binding_coverage.gap_count` in one iteration.
- [ ] Existing stage-3/4 scaffold output is unchanged (golden test).

## Scope / non-goals

Scaffolds stay inert: `<REVIEW_REQUIRED>` placeholders only, never auto-applied, never auto-merged. Effect, authority, and `complete: true` remain human assertions — this issue only removes retyping, not judgment.

## Headline metric

**Cold-start iterations to first gating verdict: 5 → 2.** No new CLI command, no schema version bump — this populates an existing optional field and changes when an existing artifact is written.

### LINKED ISSUE #388: Declaration scaffolds hand users blanks for values shipgate already computed (root binding candidate, accepted_values)
## Impact

`suggested-declarations.yaml` is the file the tool tells users to edit. It hands them
`<REVIEW_REQUIRED>` blanks for values **shipgate has already computed and stored elsewhere**.
The user is asked to guess, or to go spelunking in a 242 KB `report.json`, for answers the
tool already has.

Two concrete instances, both on the first-adoption path.

## Severity: P2 — the dominant cost of first adoption

Neither is a correctness bug; together they are why a first scan lands in
`insufficient_evidence` and stays there. Adoption friction is the constraint on this product
far more than engine accuracy.

## Instance 1 — `init` computes the root agent, then writes a manifest without it

```bash
agents-shipgate init --workspace . --write --json
```

Reports, with high confidence:

```json
"agent_name": "SmartCloserAgent",
"agent_name_candidates": [{
  "value": "SmartCloserAgent", "role": "root_agent",
  "rank_score": 4.0, "selectable": true,
  "rationale": ["declared as Agent(name=…) in agent.py",
                "bound as App(root_agent=root_agent)"]
}]
```

The manifest it writes contains `agent.name` and **no `agent_bindings.root`**. The very next
`scan` returns:

```
insufficient_evidence: the root agent is ambiguous (agent_binding_graph).
Fix at shipgate.yaml#agent_bindings.root
```

The scaffold then asks for two `<REVIEW_REQUIRED>` values — `source_id` and `object` — whose
correct answers (`adk_agent`, `root_agent`) init derived one command earlier and printed in
its own rationale.

**This is not a request to auto-declare the trust root.** Inferring `agent_bindings.root`
from AST evidence is the self-declaration attack surface from #268, and refusing to guess is
correct. The ask is that the scaffold *offer* the computed candidate for a human to confirm:

```yaml
# closes: ambiguous_root_agent
# merge into: shipgate.yaml#agent_bindings.root
# shipgate detected: source_id: adk_agent, object: root_agent
#   (declared as Agent(name=…) in agent.py; bound as App(root_agent=root_agent))
#   Confirm these are correct, then replace the placeholders below.
agent_bindings:
  root:
    source_id: <REVIEW_REQUIRED>
    object: <REVIEW_REQUIRED>
```

The human declaration property is preserved. The guessing game is removed.

## Instance 2 — the scaffold omits `accepted_values` that `report.json` already carries

`suggested-declarations.yaml` emits:

```yaml
effect: <REVIEW_REQUIRED>
authority:
  mode: <REVIEW_REQUIRED>
```

with no vocabulary. `agents-shipgate explain` takes only a `check_id`, so there is no CLI
route to the answer.

Meanwhile `report.json` carries it exactly, per gap, at
`release_decision.evidence_coverage.evidence_gaps[].next_action.accepted_values`:

```json
"accepted_values": ["read", "write", "destructive",
                    "external_communication", "financial_write", "production_operation"]
"accepted_values": ["none", "scoped", "unscoped", "ambient"]
```

The data exists and is per-field accurate. It is simply not written into the file users are
directed to fill in. (The scaffold header does document the `authority.mode` co-requirement
rules, so the omission is inconsistent rather than total.)

## Expected behavior

The scaffold is self-sufficient: for every `<REVIEW_REQUIRED>`, it carries the allowed values
and any candidate shipgate computed, as comments.

## Acceptance criteria

- [ ] Every `<REVIEW_REQUIRED>` in `suggested-declarations.yaml` is preceded by a comment
      listing that field's `accepted_values`.
- [ ] Where a candidate was computed (root agent object/source), it appears as a commented
      suggestion with its rationale, still requiring human confirmation.
- [ ] A user can complete the scaffold without opening `report.json` or the JSON schema.
- [ ] Test asserting scaffold `accepted_values` comments stay in sync with the gap's
      `accepted_values` (so the vocabulary cannot drift between the two artifacts).

## Notes

Found while running the full adoption path on
[google/adk-samples#1745](https://github.com/google/adk-samples/pull/1745) as a first-time
user, following only CLI output. These two items plus #385 and #386 were the reasons that run
never reached a gate-able verdict.

pengfei-threemoonslab:
## Summary

- **Closes #361** — `suggested-declarations.yaml` was written only once the binding layer was already closed, so during the two scans where an adopter is most stuck it did not exist, and the binding gap that *was* reported carried `declaration_template: null`. Two new templates cover those stages.
- **Closes #388** — every `<REVIEW_REQUIRED>` in the scaffold now carries the values it accepts (or the shape its answer takes) as a comment, and the `agent_bindings.root` block lists the agent objects the scan already observed for a human to confirm.

Both were filed from the same first-time-adoption walk on `google/adk-samples`, and both are adoption friction rather than engine accuracy.

### #361 — the scaffold reaches the scans that need it

Reproduced the reported walk (ADK agent, six imported tool symbols): stage 1 emitted no scaffold and no skeleton; stage 2's `missing_binding_evidence` gap carried a null template.

**Stage 1 — the unenumerable source.** The symbol names exist nowhere structured: the loader records the *reason* as a `surface_gap` and the *names* only in its own warning prose. `core.source_warnings` owns both halves of that prose, so they are read back through the same exact, `repr`-delimited decoder `group_source_warnings` already relies on — never a regex over loader text. That feeds one `incomplete_surface` row **per source** carrying the exact `tool_inventories` entry with `source_id` bound to the source it completes (#386), and the inventory skeleton is now written with those names pre-filled.

Per source, not per symbol, and that redirection came from a test failure worth calling out: a `source_warning` gap's `subject` **is** the raw warning text, so routing the repair onto those rows made one of six restatements addressable and put loader prose straight back into the CLI headline that the #362 grouping work removed (`test_cli_summary_prints_both_the_raw_count_and_the_mechanism_count` caught it). The six `source_warning` rows stay inert `review_warning`s with no path and no command.

**Stage 2 — the unbound catalog.** A resolved root, a populated catalog, and not one static edge between them now scaffolds the closed-world `agent_bindings.declarations` row. The 2026-07 defect this block was cut for was not the block, it was the *assertion* inside it: what is pre-filled here is only what was read off the surface (which agent, which catalog tools exist, which handoffs were observed), and `complete` and `reason` — the two values that are a judgement — stay `<REVIEW_REQUIRED>`.

Past **50 tools the template is withheld, not truncated**. `complete: true` claims the listed tools are *all* the agent can reach, so a silently cut list is false exactly where a reviewer cannot see it.

Both instructions are also withdrawn once followed: the inventory skeleton is deleted when the source it was written for has been completed, the way the declaration scaffold already was.

### #388 — the scaffold says what a legal answer is

Each blank is preceded by a comment carrying either the field's `accepted_values` or, where the answer is not from a closed set, the shape it takes and which modes make it required.

The vocabulary renders **from the gap's own `accepted_values`**, never a second copy, so the two artifacts cannot drift. That is possible for exactly two fields and the reason is worth a reviewer's attention: `accepted_values` is **overloaded** on the wire. For most gap kinds it lists the manifest *keys* a repair must set (`["agent", "complete:true", "tools", …]`); only for `declare_action_effect` and `declare_action_authority` does it list the legal *values* of one field. Printing the first reading above a blank would tell a reviewer that `agent` is a legal `effect`, so only the second is routed, and the routing table is pinned to the action-kind enum by a test.

The root block now lists the observed agent objects. Note what that also fixes: `agent_bindings.root.object` matches the agent's **declared** name (`_select_root` filters on `agent.name`), not the Python variable it was assigned to — the issue's own worked example guessed the variable and would not have matched. Nothing is filled in; inferring the trust root from AST evidence stays the self-declaration surface #268 closed.

### Two things found while self-reviewing

- PyYAML cannot carry comments, so the block is emitted by hand — and a hand-rolled emitter is where a second opinion about YAML style creeps in. A test asserts that stripping the comment lines back out yields **exactly** what `safe_dump` writes for the same template. A separate fallback covers a string holding a newline, which `safe_dump` renders as a multi-line scalar whose continuation indent this emitter cannot compute for a leaf on its own.
- A *name* is repository data. An agent name holding a newline would close the `#` of a candidate comment and write a **filled-in root selector** for a reader to paste. Names in comments are escaped with `display_literal` (injective), with an adversarial test holding both vectors shut.

### One layer earlier

`agent_bindings.declarations[].complete` is `Literal[True]`, so pasting the scaffold unfinished failed with pydantic's "Input should be True" — true, and silent about the scaffold the reader pasted. The placeholder is now rejected from the raw input before field validation, so one wording covers every field whatever its type.

## Type

- [ ] Input adapter change
- [x] Documentation only

## Verification

Additional local checks run:

- `ruff check .` — clean; full `pytest -n auto` — green (unpiped, exit code checked).
- The three-stage cold-start walk driven end to end through `./shipgate scan`, and encoded as `test_cold_start_walk_scaffolds_both_layers_in_two_iterations`: stage 1 and 2 both emit `suggested-declarations.yaml`; merging the stage-2 block verbatim after replacing the two placeholders takes `binding_coverage.gap_count` 1 → 0 in one iteration and reaches 6/6 catalog tools.
- #388's own measurement on the generated file, which the issue reported as `0`: the effect vocabulary and the four authority modes now appear once per block.
- Sample goldens checked rather than assumed: regenerated `conductor_agent` into a temp dir and diffed — **no** `evidence_gaps` differences, so no golden needed regenerating (its `incomplete_surface` rows have no `tool_inventories` key to scaffold). `report.md` and `packet.*` are byte-compared by the suite and unchanged.

**Not verified:** #361's headline metric (cold-start iterations 5 → 2) is measured against the walk reconstructed in this repo, not against live `google/adk-samples#1917` — I did not clone that PR.

## Release-readiness notes

- [x] New or changed check IDs are documented in `docs/checks.md` — no check IDs added or changed.
- [x] Report/schema changes are additive or documented in `STABILITY.md` — no `report_schema_version` bump. Existing optional fields (`declaration_template`, `accepted_values`) are populated where they were null, and `build_release_decision` gains a keyword-only `tool_catalog` that defaults to `tools`. **Nothing gates on any of it**: `evidence_below_ie_threshold` and every coverage count read the same inputs as before.

### Surface discipline

No new CLI command, no schema version, no adapter. The headline metric is cold-start iterations to a first gating verdict, and the change is confined to the one decision engine plus the artifact that projects it. `docs/engineering/insufficient-evidence-cold-start.md` carries the design record, including the earlier round this reverses part of and why.

### DIFF
--- a/src/agents_shipgate/ci/release_decision.py
+++ b/src/agents_shipgate/ci/release_decision.py
@@ -1,7 +1,10 @@
 from __future__ import annotations
 
 import math
+from collections.abc import Sequence
 from copy import deepcopy
+from dataclasses import dataclass
+from typing import Any
 
 from agents_shipgate.ci.exit_policy import (
     effective_fail_on,
@@ -14,6 +17,12 @@ from agents_shipgate.core.evidence_actions import (
     evidence_gap_target,
     primary_evidence_gap,
     yaml_scalar,
+)
+from agents_shipgate.core.source_warnings import unresolved_adk_tool_symbols
+from agents_shipgate.schemas.bindings import (
+    AgentBindingGraphAssessment,
+    AgentBindingIssue,
+    AgentBindingNode,
 )
 from agents_shipgate.schemas.common import Severity
 from agents_shipgate.schemas.report import (
@@ -111,8 +120,20 @@ def build_release_decision(
     ci_mode: str,
     fail_on: list[Severity] | None,
     new_findings_only: bool,
+    tool_catalog: Sequence[Tool] | None = None,
 ) -> ReleaseDecision:
+    """Compute the release decision.
+
+    ``tools`` is the root-reachable surface — the population every count and
+    every gate reads. ``tool_catalog`` is everything extraction found, reachable
+    or not, and is used for exactly one thing: scaffolding the binding
+    declaration for a repository whose catalog is populated and whose reachable
+    set is therefore empty (#361). Nothing gates on it. It defaults to ``tools``
+    so a caller that never had the distinction keeps its behaviour.
+    """
+
     fail_on_resolved = effective_fail_on(ci_mode, fail_on)
+    catalog = tools if tool_catalog is None else tool_catalog
 
     # blockers/review_items consider the full findings set, NOT
     # new_findings_only: baseline-matched criticals must remain visible
@@ -255,7 +276,7 @@ def build_release_decision(
     low_confidence_tool_count = sum(1 for tool in tools if tool.extraction_confidence != "high")
     semantic_coverage, semantic_gaps = _semantic_coverage(tools)
     identity_coverage = _identity_coverage(tools)
-    binding_coverage, binding_gaps = _binding_coverage(report)
+    binding_coverage, binding_gaps = _binding_coverage(report, catalog)
     evidence = EvidenceCoverageDecision(
         level=report.summary.evidence_coverage,
         human_review_recommended=report.summary.human_review_recommended,
@@ -444,6 +465,120 @@ AGENT_BINDINGS_ROOT_TEMPLATE: dict[str, object] = {
         },
     }
 }
+
+# Ceiling on the tool rows a closed-world binding declaration is scaffolded
+# with. Above it the template is withheld rather than truncated: a
+# ``declarations`` row is a claim that the listed tools are ALL the agent can
+# reach, so a list silently cut at N would be false exactly where the reviewer
+# is least able to notice — and a repository with hundreds of unbound catalog
+# entries is telling us to wire the binding in source, not to retype it.
+_MAX_SCAFFOLDED_BINDING_TOOLS = 50
+
+
+def _inventory_declaration_template(
+    manifest_key: str, source_id: str | None
+) -> dict[str, object] | None:
+    """The manifest wiring that joins a reviewed inventory to its source.
+
+    ``_inventory_remediation`` already prescribes this entry as prose. Emitting
+    it as a template puts it in the file the reader was actually told to edit,
+    which is the whole point of the scaffold (#388): the vocabulary and the
+    shape stop being something to reconstruct from a sentence.
+
+    Withheld when the source has no id: an inventory referenced without
+    ``source_id`` is an *independent* source added beside the extracted tools
+    instead of completing them, so the gap that asked for the file stays open
+    (#386). A template that quietly does that is worse than none.
+    """
+
+    if not source_id:
+        return None
+    block, _, key = manifest_key.partition(".")
+    if not block or not key:
+        return None
+    return {
+        block: {key: [{"path": REVIEW_REQUIRED_SENTINEL, "source_id": source_id}]}
+    }
+
+
+def _binding_declaration_template(
+    graph: AgentBindingGraphAssessment,
+    issue: AgentBindingIssue,
+    tool_catalog: Sequence[Tool],
+) -> dict[str, object] | None:
+    """The manifest block one binding issue is repaired in, or ``None``.
+
+    Two binding issues are scaffoldable, and they want different blocks.
+
+    ``ambiguous_root_agent`` wants root *selection* only. In a decorator-only
+    repository there is no agent object for a filled root selector to match, so
+    offering one would send the reader after a value that cannot exist.
+
+    ``missing_binding_evidence`` — a resolved root, a populated catalog, and
+    not one static edge between them — wants a closed-world ``declarations``
+    row. Everything in it that a human owns stays a sentinel: ``complete`` is
+    the closed-world assertion and ``reason`` is how they verified it. What is
+    pre-filled is only what was *read off the surface*: which agent, which
+    catalog tools exist, and which handoffs were observed. That is the
+    retyping #361 measured — six tool names already extracted, hand-copied at
+    the point the user has the least context — not the judgement.
+
+    The other binding kinds are repaired in source wiring or in an existing
+    declaration, so a block aimed at ``agent_bindings.declarations`` would not
+    fit the path their action names.
+    """
+
+    if issue.kind == "ambiguous_root_agent":
+        return deepcopy(AGENT_BINDINGS_ROOT_TEMPLATE) if graph.agents else None
+    if issue.kind != "missing_binding_evidence":
+        return None
+    # Root-scoped only. The same kind is also raised per tool for capabilities
+    # bound to an agent the root does not reach (``_unbound_tool_gaps``); those
+    # are repaired by wiring the handoff, not by declaring the root's tool set.
+    if issue.tool_id is not None or issue.agent_id != graph.root_agent_id:
+        return None
+    catalog = {tool.id: tool for tool in tool_catalog}
+    unbound = [catalog[tool_id] for tool_id in graph.unbound_tool_ids if tool_id in catalog]
+    if not unbound or len(unbound) > _MAX_SCAFFOLDED_BINDING_TOOLS:
+        return None
+    names = {agent.agent_id: agent.name for agent in graph.agents}
+    root_name = names.get(issue.agent_id or "")
+    # ``root`` is the documented alias for the configured root agent, and it is
+    # what a duplicated agent name has to fall back to: a declaration naming a
+    # name two agents share resolves to neither.
+    agent = (
+        root_name
+        if root_name and sum(1 for value in names.values() if value == root_name) == 1
+        else "root"
+    )
+    handoffs = sorted(
+        {
+            names.get(edge.target_agent_id, edge.target_agent_id)
+            for edge in graph.handoff_edges
+            if edge.source_agent_id == issue.agent_id
+        }
+    )
+    return {
+        "agent_bindings": {
+            "declarations": [
+                {
+                    "agent": agent,
+                    "complete": REVIEW_REQUIRED_SENTINEL,
+                    # The same selector shape the action templates use: both
+                    # resolve through ToolSelectorIndex, where ``tool_id`` is
+                    # exact and the source qualifier keeps the row readable.
+                    "tools": [
+                        _action_selector(tool)
+                        for tool in sorted(unbound, key=lambda item: (item.name, item.id))
+                    ],
+                    "handoffs": handoffs,
+                    "reason": REVIEW_REQUIRED_SENTINEL,
+                }
+            ]
+        }
+    }
+
+
 _SEMANTIC_RERUN_COMMAND = (
     "agents-shipgate verify --workspace . --config shipgate.yaml --ci-mode advisory --format json"
 )
@@ -471,6 +606,7 @@ _MANDATORY_CURRENT_CONTROL_CHECKS = frozenset(
 
 def _binding_coverage(
     report: ReadinessReport,
+    tool_catalog: Sequence[Tool] = (),
 ) -> tuple[BindingCoverageDecision, list[EvidenceGap]]:
     graph = report.binding_surface_facts
     reason_counts: dict[str, int] = {}
@@ -507,17 +643,7 @@ def _binding_coverage(
             path = "shipgate.yaml#agent_bindings"
             accepted_values = ["literal_binding", "reviewed_declaration"]
             expects = "Correct the binding annotation or provide an exact reviewed declaration."
-        # Only root SELECTION is scaffoldable. The other binding issues are
-        # repaired under ``agent_bindings.declarations`` or in the source
-        # wiring, so a root-only block would not fit the path they name — and
-        # in a decorator-only repository there is no agent object for a filled
-        # root selector to match, so offering one would send the reader after a
-        # value that cannot exist.
-        root_template = (
-            AGENT_BINDINGS_ROOT_TEMPLATE
-            if issue.kind == "ambiguous_root_agent" and graph.agents
-            else None
-        )
+        template = _binding_declaration_template(graph, issue, tool_catalog)
         gaps.append(
             EvidenceGap(
                 kind=issue.kind,
@@ -529,9 +655,9 @@ def _binding_coverage(
                     command=_SEMANTIC_RERUN_COMMAND,
                     path=path,
                     why="A complete root-reachable static binding graph is required for passed.",
-                    expects=_with_scaffold_pointer(expects, root_template),
+                    expects=_with_scaffold_pointer(expects, template),
                     accepted_values=accepted_values,
-                    declaration_template=deepcopy(root_template) if root_template else None,
+                    declaration_template=template,
                 ),
             )
         )
@@ -816,6 +942,12 @@ def _semantic_gap(
     action_kind: str
     accepted_values: list[str]
     declaration_template: dict[str, object] | None = None
+    # Whether ``expects`` should also name the on-disk scaffold. Off for the
+    # inventory repair: that row's own ``path`` is the skeleton to open and its
+    # remediation spells the manifest entry inline, so naming a second file for
+    # the same one-line edit splits one instruction in two. The block is still
+    # written to the scaffold for a reader who works from there.
+    name_scaffold = True
     if kind in {"incomplete_tool_identity"}:
         action_kind = "declare_source_identity"
         accepted_values = ["unique_source_id", "stable_native_locator"]
@@ -886,6 +1018,10 @@ def _semantic_gap(
             expects = _inventory_remediation(
                 manifest_key, tool.source_id, rerun="rerun verification."
             )
+            declaration_template = _inventory_declaration_template(
+                manifest_key, tool.source_id
+            )
+            name_scaffold = False
         else:
             action_kind = "provide_complete_inventory"
             accepted_values = [
@@ -979,7 +1115,9 @@ def _semantic_gap(
             command=_SEMANTIC_RERUN_COMMAND,
             path=_semantic_gap_path(kind, tool),
             why=action_why,
-            expects=_with_scaffold_pointer(expects, declaration_template),
+            expects=_with_scaffold_pointer(
+                expects, declaration_template if name_scaffold else None
+            ),
             accepted_values=accepted_values,
             declaration_template=declaration_template,
         ),
@@ -1104,9 +1242,9 @@ def _evidence_gaps(report: ReadinessReport, tools: list[Tool]) -> list[EvidenceG
     """v0.26: one actionable row per measurable evidence gap.
 
     Deterministic projection of the same inputs the counts use:
-    low-confidence tools (sorted by name) first, then source warnings in
-    report order. Never gates — `build_release_decision` keeps deciding
-    on the counts alone.
+    low-confidence tools (sorted by name) first, then unenumerable sources by
+    id, then source warnings in report order. Never gates —
+    `build_release_decision` keeps deciding on the counts alone.
     """
     gaps: list[EvidenceGap] = []
     low_confidence = sorted(
@@ -1139,6 +1277,9 @@ def _evidence_gaps(report: ReadinessReport, tools: list[Tool]) -> list[EvidenceG
                 ),
             )
         elif manifest_key is not None:
+            inventory_template = _inventory_declaration_template(
+                manifest_key, tool.source_id
+            )
             action = EvidenceGapAction(
                 kind="declare_tool_inventory",
                 path=SUGGESTED_INVENTORY_FILENAME,
@@ -1150,6 +1291,12 @@ def _evidence_gaps(report: ReadinessReport, tools: list[Tool]) -> list[EvidenceG
                 expects=_inventory_remediation(
                     manifest_key, tool.source_id, rerun="rerun the scan."
                 ),
+                # No scaffold pointer: this row's own ``path`` is the skeleton
+                # to open, and its remediation already spells the manifest
+                # entry inline. Naming a second file for the same one-line edit
+                # splits one instruction across two artifacts. The block is
+                # still in the scaffold for a reader who works from there.
+                declaration_template=inventory_template,
             )
         else:
             action = EvidenceGapAction(
@@ -1173,6 +1320,55 @@ def _evidence_gaps(report: ReadinessReport, tools: list[Tool]) -> list[EvidenceG
                     f"{_surface_gap_note(tool)}"
                 ),
                 next_action=action,
+            )
+        )
+    covered_sources = {
+        tool.source_id for tool in low_confidence if tool.source_id
+    }
+    for source in unresolved_symbol_sources(report):
+        if source.source_id in covered_sources:
+            # A low-confidence row for this source already prescribes the same
+            # inventory; two rows would be one repair asked for twice.
+            continue
+        inventory_template = _inventory_declaration_template(
+            source.manifest_key, source.source_id
+        )
+        listed = ", ".join(source.symbols[:_MAX_LISTED_SYMBOLS])
+        more = (
+            f" (+{len(source.symbols) - _MAX_LISTED_SYMBOLS} more)"
+            if len(source.symbols) > _MAX_LISTED_SYMBOLS
+            else ""
+        )
+        agents = ", ".join(source.agents)
+        gaps.append(
+            EvidenceGap(
+                kind="incomplete_surface",
+                subject=f"{source.source_id} [{source.manifest_key.split('.')[0]}]",
+                source_type=source.manifest_key.split(".")[0],
+                source_ref=source.source_ref,
+                why=(
+                    f"{len(source.symbols)} tool symbol(s) {agents} lists are "
+                    f"not defined in this entrypoint, so the source produced no "
+                    f"observation for them: {listed}{more}."
+                ),
+                next_action=EvidenceGapAction(
+                    kind="declare_tool_inventory",
+                    path=f"shipgate.yaml#{source.manifest_key}",
+                    why=(
+                        "The complete statically-bound tool surface must be "
+                        "enumerable before any tool of it can be judged."
+                    ),
+                    expects=_with_scaffold_pointer(
+                        _inventory_remediation(
+                            source.manifest_key,
+                            source.source_id,
+                            rerun="rerun the scan.",
+                        ),
+                        inventory_template,
+                    ),
+                    accepted_values=["reviewed_explicit_inventory"],
+                    declaration_template=inventory_template,
+                ),
             )
         )
     for warning in report.source_warnings:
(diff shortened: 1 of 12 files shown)


src/agents_shipgate/ci/release_decision.py
```diff
@@ -1214,6 +1410,110 @@ def _evidence_gaps(report: ReadinessReport, tools: list[Tool]) -> list[EvidenceG
...
+    agents: tuple[str, ...]
+    symbols: tuple[str, ...]
+
+
+def unresolved_symbol_names(report: ReadinessReport) -> list[str]:
+    """Tool symbols an agent names that the catalog holds no observation for.
+
+    Once a reviewed inventory has been joined to the source, the symbol has an
+    observation and the repair is *made* — the entrypoint still cannot resolve
+    the import, so the warning stays, but re-prescribing the inventory would
```
pengfei-threemoonslab: [P1] Stop completed import warnings from gating forever

This explicitly leaves the unresolved-import warnings in place after the inventory repair is considered made, but `evidence_below_ie_threshold` still gates on the raw warning count. I followed the generated flow for the six-symbol fixture—source-bound inventory, closed-world binding declaration, then `effect: read` / `authority.mode: none` for all six tools—and the final result is still `decision=insufficient_evidence`, `source_warning_count=6`, `binding_gap_count=0`, `semantic_gap_count=0`, with no actionable non-warning gap. Thus the advertised cold-start route cannot reach `passed`. Once source completion is accepted, these resolved warnings need to be removed, reclassified, or excluded from gating, and the integration test should assert the final verdict after all declaration layers.

src/agents_shipgate/ci/release_decision.py
```diff
@@ -1214,6 +1410,110 @@ def _evidence_gaps(report: ReadinessReport, tools: list[Tool]) -> list[EvidenceG
...
+    tell the reader to do again what they just did.
+    """
+
+    known = {
```
pengfei-threemoonslab: [P1] Determine inventory completion per source, not by global name

`known` discards source identity and also assumes the Python symbol equals the exposed tool name. Both directions fail end to end: an unresolved `search` in source A is falsely treated as repaired when unrelated source B exposes a tool named `search`; conversely, a source-bound inventory that follows the generated instruction to replace `search_toolset` with its real tools (`web_search`, `document_search`) is prescribed again forever. Track the reviewed `tool_inventories[].source_id` completion relationship (and its provenance) rather than subtracting a catalog-wide exact-name set. Please cover renamed/split toolsets and same-name tools from distinct sources.

src/agents_shipgate/cli/scan/declarations.py
```diff
@@ -13,33 +13,120 @@
...
+    "root.object": (
+        "the agent's declared name — what `Agent(name=…)` was given, not the "
+        "Python variable it was assigned to."
+    ),
+    "root.source_id": "the shipgate.yaml#tool_sources[].id that defines it.",
+    "declarations.complete": (
+        "`true`, and only once you have checked that the tools listed below "
+        "are the COMPLETE set this agent can reach — delete any it cannot. "
+        "This is the closed-world assertion: the list was read off the "
+        "catalog, the claim is yours."
```
pengfei-threemoonslab: [P1] Require review of handoffs before accepting `complete: true`

`AgentBindingDeclarationConfig.complete` is the closed-world assertion for both `tools` and `handoffs`, but this hint—and the `reason` hint below—only tells the reviewer to check the tools list. The scaffold pre-fills observed handoffs, and after completion those become high-confidence declaration edges. A reviewer following this instruction can therefore ratify the tools while omitting an unobserved/dynamic downstream agent surface. Require the reviewer to verify the complete tool **and handoff** sets, including adding missing and deleting invalid entries, and make the reason cite both checks.

src/agents_shipgate/ci/release_decision.py
```diff
@@ -444,6 +465,120 @@ def _inventory_remediation(
...
+    # what a duplicated agent name has to fall back to: a declaration naming a
+    # name two agents share resolves to neither.
+    agent = (
+        root_name
+        if root_name and sum(1 for value in names.values() if value == root_name) == 1
+        else "root"
+    )
+    handoffs = sorted(
+        {
+            names.get(edge.target_agent_id, edge.target_agent_id)
```
pengfei-threemoonslab, line 578: [P2] Withhold handoff scaffolds whose names are ambiguous

The target is serialized by bare name even though declaration handoffs have no source qualifier. With `Root@A -> Twin@A` plus an unrelated `Twin@B`, this emits `handoffs: [Twin]`; after filling only `complete` and `reason`, resolution reports that `Twin` names two agents and then reports a closed-world handoff conflict, so the advertised scaffold cannot close its own gap. Check global target-name uniqueness before emitting this template; when a target is ambiguous, withhold it and route the reviewer to source wiring (the current schema cannot express the exact target).

src/agents_shipgate/schemas/manifest/root.py
```diff
@@ -92,6 +107,25 @@ class AgentsShipgateManifest(BaseModel):
...
+        The after-validator below can only see values that already parsed, so a
+        sentinel in a *typed* field never reached it: pasting the binding
+        scaffold with ``complete: <REVIEW_REQUIRED>`` — where the schema accepts
+        only ``true`` — failed with "Input should be True", which does not tell
+        the reader they pasted an unfinished scaffold. Reading the raw input
+        first makes one wording cover every field, whatever its type.
+        """
+
+        if isinstance(data, dict):
+            unfilled = sorted(_sentinel_paths(data))
```
pengfei-threemoonslab, line 154: [P2] Make the raw sentinel traversal cycle-safe

`yaml.safe_load` preserves recursive aliases, while `_sentinel_paths` recursively walks the raw dict/list graph without tracking visited container identities. A syntactically valid YAML manifest such as `bogus: &loop {x: *loop}` now reaches this new before-validator and raises an uncaught `RecursionError`, instead of the prior structured `extra_forbidden` / `ConfigError` (and agent-mode recovery payload). Add cycle detection or use an iterative bounded traversal before inspecting raw values.

src/agents_shipgate/cli/scan/declarations.py
```diff
@@ -104,13 +197,257 @@ def build_declaration_scaffold(gaps: Sequence[EvidenceGap]) -> str | None:
...
+    # between names, and an id-hash ordering looks arbitrary to them.
+    ordered = sorted(agents, key=lambda agent: (agent.name, agent.source_id or ""))
+    for agent in ordered[:_MAX_RENDERED_CANDIDATES]:
+        # ``display_literal`` because these are repository-controlled identity
+        # strings, and the caller writes them into a YAML `#` comment. An
+        # `Agent(name="...\nroot:\n  object: X")` would otherwise close the
+        # comment and render a *filled-in* root block for a reader to paste —
+        # a self-declaration the scaffold exists to refuse (#268). Injective,
+        # so the escape still names exactly one object.
+        name = display_literal(agent.name)
```
pengfei-threemoonslab, line 310: [P2] Escape YAML-forbidden noncharacters in candidate comments

`display_literal` escapes line/control hazards but leaves U+FFFE and U+FFFF literal. A repository-controlled agent name such as `"bad\uFFFE"` therefore writes that raw code point into this comment, and `yaml.safe_load_all` rejects the generated scaffold with `ReaderError: unacceptable character #xfffe`. These values can enter a Python string via an escape, so extend the identity renderer (or this sink) to escape YAML-forbidden noncharacters and add a parse/round-trip regression.

pengfei-threemoonslab (commented):
Reviewing `7bd027f70ac96d6fa2c98c99f7ed24b57fd2c5b5`, I found six reproducible issues, led by a cold-start path that remains permanently `insufficient_evidence`, source completion inferred without source identity, and a closed-world instruction that omits handoff review. The remaining inline findings cover an invalid duplicate-name handoff scaffold and two deterministic YAML/manifest robustness regressions.

Validation: GitHub CI, Agents Shipgate, and self-dogfood are green; local Shipgate control is `complete` with `mergeable / passed`; the focused 185-test scaffold/ranking suite, Ruff, and compileall pass. The broad local run completed all non-packaging tests, but eight packaging cases could not execute because their shared isolated build fixture could not download `hatchling` in the restricted environment. The green static gate does not exercise the end-to-end/adversarial cases called out inline.

GitHub does not permit me to request changes on my own PR, so treat the P1 findings below as merge-blocking review feedback.

pengfei-threemoonslab (commented):
All six addressed in `d4415abb`. Every one reproduced first; the two P1s turned out to be the same mistake seen from two sides, and are fixed together.

**[P1] The prescribed route did not terminate.** Confirmed by walking it to the end — inventory, binding declaration, then `effect`/`authority` for all six tools — and landing on `insufficient_evidence` with `source_warning_count=6` and no non-warning gap left. A reviewed `tool_inventories` entry naming a source in `source_id` *is* the answer shipgate prescribes for that source's warnings, so those warnings are now withdrawn once the manifest declares the source. The same walk now reaches a real gating verdict.

Only the **warning** is withdrawn. The loader's `surface_gaps` entry stays, so nothing claims static analysis resolved what it could not and extraction confidence is unmoved. I also checked the cheapest abuse — declare an *empty* source-bound inventory to silence the warnings — and it still stops at `SHIP-INVENTORY-NOT-ENUMERABLE` → `review_required`. That independence is now pinned by a test rather than left to argument.

**[P1] Completion per source, not by global name.** Agreed, and both directions you named are exactly right. The relation is now the reviewed `tool_inventories[].source_id`, already desugared into `LoadedToolSource.completes_source_id`, applied in the one place holding both halves of it: the loaded sources know which inventory completes which source, the ADK artifacts know which source published each agent. The catalog-wide name subtraction is gone. Both cases are end-to-end tests: `test_a_split_toolset_inventory_is_not_prescribed_forever` (the inventory follows the skeleton's own instruction to split `search_toolset` into `web_search`/`document_search`, and the source is still complete) and `test_a_same_named_tool_elsewhere_does_not_complete_this_source`. An agent name two sources both publish is dropped rather than guessed, so a warning is never withdrawn against the wrong source.

**[P1] Handoffs in the `complete:` instruction.** Correct — the scaffold pre-fills observed handoffs and the hint spoke only about tools, so a reviewer could ratify the tool set while silently asserting a downstream agent surface. Both hints now name both lists and ask for additions *and* deletions on each, and `handoffs:` got its own hint saying it is covered by `complete:`.

**[P2] Ambiguous handoff targets.** Reproduced. The whole template is withheld now, not just the handoff — dropping the target would understate a closed world the reviewer is about to assert, and as you say the schema cannot express the exact target, so source wiring is the honest route.

**[P2] Cycle-safe raw traversal.** Reproduced (`RecursionError` on `bogus: &loop {x: *loop}`). Fixed with identity-based cycle detection; the cyclic manifest gets its structured `ValidationError` back. A placeholder *inside* a cycle is still found and named — cycle-safety must not blind the check, so that is asserted too.

**[P2] YAML-forbidden noncharacters.** Reproduced (`ReaderError: unacceptable character #xfffe`). I took the identity-renderer option rather than patching the one sink: `display_literal`/`one_line` now escape Unicode noncharacters alongside the invisible code points they already covered — same hazard, and two of them additionally make the document unloadable. The encoding stays injective, verified through `undisplay_literal`.

**On the validation gap you flagged.** Taken: the integration test now walks every declaration layer and asserts the final verdict, so an unreachable remedy fails the suite instead of a review. `ruff` and the full local suite are green.

src/agents_shipgate/cli/scan/tools_agent.py
```diff
@@ -43,6 +84,15 @@ def _build_tools_and_agent(
     # Some adapters expose the same warnings through both LoadedToolSource
     # and the artifact bag; keep report warning output stable and unique.
     warnings = list(dict.fromkeys(warnings))
+    # The one place holding both halves of the completion relationship: the
+    # loaded sources know which inventory completes which source, and the ADK
+    # artifacts know which source each agent came from. A warning asking for
+    # evidence the manifest now declares is answered, and must stop gating.
+    warnings = withdraw_completed_adk_tool_warnings(
```
pengfei-threemoonslab, line 95: [P1] Keep unresolved membership through binding resolution

For a mixed ADK surface `tools=[local_tool, remote_tool]`, where `local_tool` is defined in the entrypoint and `remote_tool` is imported, I followed the generated skeleton exactly and bound it to the source. This withdrawal clears the warning, but the graph then has `issues=[]`, only `local_tool` reachable, and the reviewed `remote_tool` inventory entry unbound; the one structural edge suppresses the `missing_binding_evidence` scaffold. The decision remains `insufficient_evidence`, and the surviving remediation tells the user to bind another inventory to `source_id: "google_adk_inventory:inventory.json"` (a self-reference), while the adjacent low-confidence action says an inventory cannot close it. Preserve structured `(source, agent, symbol)` membership into binding assessment and mark this graph partial so the missing edge has a valid declaration/source-wiring route; add a mixed local/imported end-to-end test.

tests/test_declaration_scaffold.py
```diff
@@ -415,3 +482,947 @@ def _gaps(agents: list, kind: str):
...
+                }
+            },
+            sort_keys=False,
+        ),
+        encoding="utf-8",
+    )
+    final = _scan_cold_start(project, reports)
+    assert final.release_decision is not None
+    coverage = final.release_decision.evidence_coverage
+    assert coverage.semantic_coverage.gap_count == 0
```
pengfei-threemoonslab: [P1] Assert that the cold-start walk actually leaves insufficient evidence

The new final stage still returns `decision=insufficient_evidence`: exact counts are binding `0`, semantic `0`, source warnings `0`, but policy gaps `1`, with `mixed_policy_evidence` for `send_quote_email`. The fixture uniformly declares every effect as `read`, so these assertions prove warning removal and actionability of the remaining gap, not the comment's claim that the route is no longer an abstention. Consume the emitted action declarations (or at least review `send_quote_email` as `external_communication`) and assert the substantive final verdict plus `policy_gap_count == 0`; with the reviewed effect the fixture currently reaches `blocked` with all four evidence-gap counts at zero.

src/agents_shipgate/schemas/manifest/root.py
```diff
@@ -39,26 +41,58 @@
...
+    with it — with a stack overflow (PR #401 review). A container already open
+    on this branch is not descended again: it can hold no placeholder its first
+    visit did not already report.
+    """
 
+    if isinstance(node, (dict, list, tuple)):
+        marker = id(node)
+        if marker in _open:
+            return []
+        _open = _open | {marker}
```
pengfei-threemoonslab: [P1] Bound shared-alias traversal, not only ancestor cycles

`_open` contains only the containers on the current branch. After one sibling returns, the same aliased child is not remembered, so an acyclic YAML DAG such as repeated `{left: *previous, right: *previous}` is traversed once per logical path. An actual `safe_load`/manifest-validation repro with 20 levels was only about 1.5 KB / 50 lines yet took 1.14 seconds here, growing roughly 4x every two levels; putting a sentinel at the shared leaf also materializes `2**n` path strings. This can stall manifest loading before the structured error is produced. Use a traversal-wide visited set, a bounded traversal, or memoize sentinel presence; reporting one representative path for an aliased placeholder is sufficient to reject it.

src/agents_shipgate/cli/scan/tools_agent.py
```diff
@@ -15,6 +18,44 @@
...
+
+    if adk is None:
+        return {}
+    by_name: dict[str, set[str]] = {}
+    for agent in adk.agents:
+        name, source_id = agent.get("name"), agent.get("source_id")
+        if isinstance(name, str) and isinstance(source_id, str):
+            by_name.setdefault(name, set()).add(source_id)
+    return {
+        name: next(iter(ids)) for name, ids in by_name.items() if len(ids) == 1
```
pengfei-threemoonslab: [P2] Recognize completion when every colliding source is reviewed

Dropping a duplicated agent name entirely is conservative only while at least one candidate source is incomplete. With `adk_a` and `adk_b` both publishing `LlmAgent(name="Closer")`, each referencing two distinct imported symbols, and each having a nonempty reviewed inventory bound to its own source ID, all four warnings remain and the decision stays `insufficient_evidence`. The separate root ambiguity is actionable, but these warnings are not; the release-decision name join skips them too. Keep `name -> set[source_id]` and withdraw when every candidate ID is completed, or carry source provenance with each warning so display-name uniqueness is unnecessary.

src/agents_shipgate/cli/scan/tools_agent.py
```diff
@@ -15,6 +18,44 @@
...
+    source would clear a diagnostic nobody answered.
+    """
+
+    if adk is None:
+        return {}
+    by_name: dict[str, set[str]] = {}
+    for agent in adk.agents:
+        name, source_id = agent.get("name"), agent.get("source_id")
+        if isinstance(name, str) and isinstance(source_id, str):
+            by_name.setdefault(name, set()).add(source_id)
```
pengfei-threemoonslab: [P2] Normalize both sides of the source-ID comparison

`_completed_source_ids` strips completion IDs at lines 31–35, but this map stores the artifact `source_id` verbatim. The manifest schema currently permits surrounding whitespace, so a source and inventory both using the exact quoted ID `' adk '` load successfully yet compare as `' adk '` versus `'adk'`; the answered unresolved-symbol warning remains. Strip here as well, or reject/normalize surrounding whitespace consistently at the schema boundary.

pengfei-threemoonslab (commented):
Re-reviewing `d4415abb5bc729e7e9166b9d2a5dc1c2e48d8812`: the six original isolated reproductions are addressed (single-source warning completion, split/renamed inventories, handoff review guidance, ambiguous handoff scaffolding, direct YAML cycles, and forbidden noncharacters). The expanded review found five current-head issues: three P1s in mixed-surface termination, the purported terminal regression test, and shared-alias traversal, plus two P2 source-identity edge cases.

Validation: GitHub CI, Agents Shipgate, and self-dogfood are green. The focused 193-test scaffold/ranking suite passes; Ruff, compileall, and diff-check pass. The broad suite completed all non-packaging tests, while the same eight packaging cases could not run because their isolated build fixture could not download `hatchling` in the restricted environment. PR-mode Shipgate reports `mergeable / passed` and local control is `complete`, but those static checks do not exercise the reproduced cases below.

GitHub does not permit me to request changes on my own PR, so treat the P1 findings as merge-blocking review feedback.

pengfei-threemoonslab (commented):
Four of five fixed in `c04b4dc6`. **The mixed-surface P1 is not fixed, deliberately** — I reproduced it, tried three versions, and each made something else worse. Analysis below; the call on whether that blocks merge is yours.

**[P1] Shared-alias traversal.** Right, and my guard was ancestor-only, which handles cycles and not DAGs. The visited set is traversal-wide now, so the walk is bounded by the document rather than by path count. `test_shared_aliases_do_not_blow_up_the_sentinel_walk` builds your 20-level case and asserts one path comes back, not 2**20. Your point that one representative path suffices is what makes this safe: a container skipped was already fully walked where it was first reached, so no placeholder is missed.

**[P2] Completion when every colliding source is reviewed.** Agreed — dropping the name was conservative only while some candidate was still incomplete. `agent_source_ids` is `name -> set[source_id]` now and withdrawal needs `completed ⊇ candidates`. Both directions are pinned.

**[P2] Normalizing both sides.** Fixed, with a test on the `' adk '` case.

**[P1] The regression test didn't prove its own comment.** Correct, and thank you for the exact counts. It now consumes the scaffold's emitted action blocks, answers them the way the scaffold's comments instruct (`mode: none`, delete what it doesn't take), reviews `send_quote_email` as `external_communication`, and asserts the substantive result: all four counts zero, `evidence_gaps == []`, `blocked` with blockers. As a side effect it also proves the scaffold's action rows are pasteable.

---

**[P1] Mixed local/imported surface — reproduced, not fixed.**

Three defects sit behind your repro, and I confirmed each:

1. `incomplete_surface` prescribes an inventory for an unproven tool *set*, contradicting the `low_confidence_tool` row about the same tool — and after the merge the `source_id` it prints is the inventory's own id, which the loader rejects as self-referential.
2. `structural_complete` is derived from edge completeness. The one resolved edge is genuinely complete, so a reviewed closed-world declaration adding the imported tool — the repair the gap asks for — comes back `conflicting`.
3. `partials` is a bare `set[str]` with no agent attribution, so even an accepted declaration leaves the row standing.

What I tried, and why each was withdrawn:

- **Publishing `tools_complete=False` from the ADK adapter** (matching what `openai_sdk_static` already does for an unresolved tool name) marks each emitted edge uncertain, which drops the tools that *were* resolved out of `reachable_tool_ids`. They stop being judged, and the per-tool `low_confidence_tool` rows #393 added disappear — eight tests in `test_google_adk.py` catch exactly that. The uncertainty is about the members that produced no edge, not the ones that did.
- **Publishing the membership via `AgentBindingObservation.issues` instead**, plus seeding `structural_complete` from agent-level completeness, keeps the tools reachable and clears the false conflict — but (3) then leaves `binding_coverage.gap_count == 1` after the declaration, so the route still doesn't terminate.
- **The routing half of (1) alone** removes a demonstrably wrong instruction, but names `agent_bindings.declarations` as the target — which (2) then rejects. That is swapping one broken instruction for another while looking like progress, so I reverted it too rather than ship it.

A correct fix has to separate "this edge is uncertain" from "this list is not exhaustive" in the resolver, and give `partials` an agent so a reviewed declaration can supersede one. Both are shared semantics the OpenAI Agents SDK adapter also depends on, and getting them wrong is precisely where a quiet fail-open enters. I'd rather hand you a reproduced defect chain than a half-landed resolver change made under review pressure — the full analysis is in `docs/engineering/insufficient-evidence-cold-start.md` under "Known gap: the mixed local/imported surface", and I have the repro ready if you want it as a follow-up issue.

One thing I did keep: the closed-world `declarations` row is now scaffolded for a partly-read tool list too, listing the agent's existing edges alongside the unbound catalog. That fires where edges are genuinely incomplete (dynamic tools, toolsets) — where `structural_complete` is already `False`, so the declaration route works — and a row omitting a tool the repository wires to the agent would be false.

`ruff`, `compileall` and the full local suite are green.

PR #402, opened 2026-08-21
fix(control): ask for the checkout instead of dead-ending the loop (#397)

### LINKED ISSUE #397: Control loop dead-ends in two moves when the evaluated head is not the checked-out worktree
## Summary

When the evaluated head is not the commit checked out in the worktree, `verify --preview` routes to `detect` against that worktree. Because the worktree does not contain the changed files, `detect` cannot name the project, returns `control_state: human_review_required` with `next_action.command: null`, and the loop ends. Two commands, both authorized by `allowed_next_commands`, and there is no forward path for either actor.

The remedy — check out the head — is derivable from what preview already knows and is never stated.

## Reproduction

[google/adk-samples#1745](https://github.com/google/adk-samples/pull/1745), worktree on `main`, CLI `0.16.0b7`, contract 24.

**Step 1**

```
agents-shipgate verify --preview --json --workspace <repo> \
  --base 765b4d1fbdcdd397fc48c622c22d757d8c87bafe --head pr-1745
```

```json
"control": {
  "state": "agent_action_required",
  "next_action": {
    "kind": "discover",
    "command": "shipgate detect --workspace <repo> --json",
    "why": "The project this change belongs to could not be established (the evaluated head 'pr-1745' is not the commit this worktree has checked out, and project markers are read from the worktree that init would run against) …"
  },
  "allowed_next_commands": ["shipgate detect --workspace <repo> --json"]
}
```

**Step 2** — the only allowed command:

```json
"control_state": "human_review_required",
"next_actor": "human",
"next_action": {"actor": "human", "kind": "review", "command": null},
"verify_required": false
```

The diagnosis in step 1 is correct and sharp. The routing that follows from it is not: `detect` is pointed at a tree that structurally cannot contain `python/agents/smart_closer/`, because that directory exists only on the PR branch.

Checking out `pr-1745` and re-running the identical step-1 command clears the state immediately.

## Secondary defect in the same path

`detect`'s escalation hands the human a JSON selector inside a shell command, and no runnable command:

```json
"next_action": "Review: Agents were found in 25 separate projects; this workspace is not one manifest's scope. Run `init --workspace <agent_project_candidates[].path> --write` for the project you are changing.",
"next_actions": [{"kind": "review", "command": null, …}]
```

`init` already does this correctly — its `next_actions[]` carry concrete per-candidate commands with `executable` and `args` arrays. `detect` should adopt the same shape.

## Why this matters

`allowed_next_commands` promises a deterministic loop. A loop that terminates in two moves on a legitimate PR — with `command: null` for the human it escalates to — is worse than no loop, because the contract implied determinism.

## Suggested direction

1. When `_head_is_current_worktree` is false, say so as the actionable instruction ("check out `<head>`, or re-run with `--head` omitted"), rather than routing to a discovery command that reads the wrong tree.
2. Give `detect`'s unresolved-scope escalation the same concrete per-candidate commands `init` already emits.

Related: #327, #369 (structured argv on control contracts).

---

pengfei-threemoonslab:
Closes #397.

## The loop, before

`verify --preview --head <ref>` while some other commit is checked out reads
project markers from the working tree — deliberately, because that is the tree
the `init` it recommends would write to. On [google/adk-samples#1745] the
changed directory exists only on the PR branch, so no project could be named.

That state routed to a human: `must_stop: true`, `next_action.command: null`,
`allowed_next_commands: []`. One move, no forward path for either actor — and
the remedy, derivable from the very ref preview was handed, went unsaid. The
`detect` the same walk reaches on a monorepo was the second dead end: its
escalation published a JSON selector inside a shell command
(`init --workspace <agent_project_candidates[].path> --write`) and no runnable
command anywhere in `next_actions[]`.

## The loop, after

**1 · The head-mismatch route asks for the checkout.** Nothing about the change
is in doubt there. What is missing is an *input* — a working tree holding the
commit under review — and producing it is one mechanical step the caller owns.
That is what `fetch_base` already exists for, so no new surface.

```json
"next_action": {
  "actor": "coding_agent",
  "kind": "fetch_base",
  "command": null,
  "expects": "commit 66f837355087 checked out in this worktree",
  "why": "The project this change belongs to could not be established (the evaluated head 'pr-1745' (66f837355087) is not the commit this worktree has checked out, and project markers are read from the worktree that init would run against). Check 66f837355087 out in this worktree, then re-run this preview with --base 20f865664ada --head 66f837355087: a revision expression re-resolves against the new HEAD. Discovery of the worktree as it stands answers about a different tree."
}
```

`expects` names a **commit id**, and the rerun is **pinned on both refs**,
because the step being requested moves `HEAD`. Spelled with the caller's own
`--head HEAD~1` the route named one commit before the checkout and its parent
after, and following it walked history backwards one commit per iteration; a
`HEAD`-relative `--base` re-ranges the same way, which is quieter and worse —
the rerun succeeds against a diff nobody asked for.

The requested checkout also makes the preview's control pointer stale, so
`agent control` — the one refresh entry point — stops reporting the request as
outstanding once it has been performed.

The two causes no checkout can repair — evidence the change deleted, an
unreadable inventory — keep their human route with no command. Plain `verify`
is unaffected: it reads `--head` from the object database, not the worktree.

**2 · `detect` publishes a runnable command for every candidate.** Rank 1 stays
the decision (`kind: "review"`, `command: null` — naming one candidate would
make the arbitrary pick `init --write` refuses to make), and below it goes one
exact command per candidate with `executable`/`args`. `init` and `detect` build
that list from one helper, so the two an adopter runs in sequence cannot publish
different recoveries for one workspace.

- **Every** candidate, not ten: the display cap that keeps a human refusal
  readable had been applied to the routing, leaving candidate 11 onward
  selectable and unrunnable. `detect --workspace samples --json` finds 22
  projects; the reported repository has 25.
- A candidate that **already carries a manifest** routes to
  `doctor --config <that manifest> --json`. A nested `shipgate.yaml` is itself
  evidence of a project, so adopted directories are candidates too — on this
  repository's own `samples/`, 21 of 22 are, and every `init --write` emitted
  for them exited 2. The exception is `--agent-instructions`, which makes
  `init --write` the advertised refresh and exits 0.
- Both commands' *printed* summaries mark those candidates and name the doctor
  route, from the one formatter they now share.
- `detect`'s `control.input_id` covers the route it publishes. Every command is
  spelled for the entry point the process came in through, so the same
  workspace read two ways published different commands under one identity —
  and that identity is the documented cache boundary for the answer.

**3 · A regression found while doing 2.** `setup_control_envelope` kept
`advance_alternatives` even when a *diagnostic* outranked the advance they carry
out, so a workspace whose only agent evidence is two nested manifests published
`stop` — "not a Shipgate target" — followed by an `init --write` for each of the
two. They now ride only with the decision that selected them.

## Deviations from the issue's suggested wording

The issue suggests saying "check out `<head>`, **or re-run with `--head`
omitted**". The second half is left out on purpose: this route only fires when
the worktree holds a *different* commit, so omitting `--head` would evaluate a
different subject and answer confidently about the wrong tree. The `why` says so
explicitly rather than leaving it implied.

`detect`'s ambiguous-scope control state stays `human_review_required`. Choosing
the project is the pick Shipgate refuses to make; the issue asked for the
runnable commands beneath that decision, not for the decision to be automated.

## Review

Nine passes over the branch, each fix its own commit. The two worth naming:

- **A defect the fix introduced.** Splitting `head_mismatch` into its own branch
  stranded its rationale on the branch it vacated — `deleted_evidence` and
  `unreadable_inventory` are produced only *after* the head is confirmed to be
  this worktree, yet still told the reader "discovery of the current worktree
  would answer about a different tree". Reproduced verbatim on a PR that deletes
  a project's only evidence file.
- **The same fix reaching one of two commands.** Marking adopted candidates in
  `init`'s printed refusal left `detect`'s summary saying `init --write` while
  its own JSON said `doctor`, because `_echo_agent_scope` recomputed the
  candidate line inline. Fixed by consolidating the formatter rather than
  copying the marker.

## Verification

- The issue's reproduction walked end to end: `fetch_base` → checkout → pinned
  rerun → `init --workspace python/agents/smart_closer --write --json` → manifest
  written → `human_review_required` on the `declared_purpose` a person owes.
- `samples/`: 21 routable candidates, 21 commands, all `doctor`, and the first
  one executed to exit 0.
- The `HEAD~1` walk, the pointer staleness, the entry-point identity split, the
  symlinked-manifest guard, and `--ci`/`--agent-instructions` flag handling each
  exercised directly.
- `ruff check .`, `scripts/generate_schemas.py --check`, `llms-full.txt`
  regenerated, full `pytest tests/`, all six CI jobs.

No new public surface: `fetch_base`, `advance_alternatives`, and `next_actions[]`
all already shipped. The `AgentControlAction` union is embedded in six durable
published schemas, so `fetch_base`'s documented meaning was widened — with its
consumers, including the adoption scorer — rather than a new action kind added.

[google/adk-samples#1745]: https://github.com/google/adk-samples/pull/1745

### DIFF
--- a/harness/adoption/scorer/rules.py
+++ b/harness/adoption/scorer/rules.py
@@ -1023,8 +1023,20 @@ def _action_satisfies_control(action: _TimelineItem, control: _ControlSnapshot) 
 
     # ``fetch_base`` is the sole structured input-recovery route that may omit
     # an exact command.  Everything else fails closed when no command exists.
+    #
+    # Two families of input satisfy it, because ``expects`` names an input and
+    # not always a ref: making history available (``git fetch``) and putting
+    # the evaluated commit in the worktree (``git checkout``), which is what
+    # ``verify --preview`` asks for when the head under review is not the
+    # commit checked out — project markers are read from the working tree, so
+    # a fetch would leave that request unmet (#397 review).
     if control.next_action.get("kind") == "fetch_base":
-        return bool(re.search(r"\bgit\s+(?:fetch|remote\s+update)\b", command))
+        return bool(
+            re.search(
+                r"\bgit\s+(?:fetch|remote\s+update|checkout|switch|restore)\b",
+                command,
+            )
+        )
     return False
 
 
--- a/harness/adoption/scorer/rules.py
+++ b/harness/adoption/scorer/rules.py
@@ -1030,10 +1030,15 @@ def _action_satisfies_control(action: _TimelineItem, control: _ControlSnapshot) 
     # ``verify --preview`` asks for when the head under review is not the
     # commit checked out — project markers are read from the working tree, so
     # a fetch would leave that request unmet (#397 review).
+    #
+    # ``checkout`` and ``switch`` and nothing else: those are the verbs that
+    # move ``HEAD``. ``git restore`` rewrites working-tree files and leaves
+    # ``HEAD`` where it was, so crediting it would pass a cell that never
+    # produced the commit the route asked for.
     if control.next_action.get("kind") == "fetch_base":
         return bool(
             re.search(
-                r"\bgit\s+(?:fetch|remote\s+update|checkout|switch|restore)\b",
+                r"\bgit\s+(?:fetch|remote\s+update|checkout|switch)\b",
                 command,
             )
         )
--- a/harness/adoption/scorer/rules.py
+++ b/harness/adoption/scorer/rules.py
@@ -1025,24 +1025,52 @@ def _action_satisfies_control(action: _TimelineItem, control: _ControlSnapshot) 
     # an exact command.  Everything else fails closed when no command exists.
     #
     # Two families of input satisfy it, because ``expects`` names an input and
-    # not always a ref: making history available (``git fetch``) and putting
-    # the evaluated commit in the worktree (``git checkout``), which is what
-    # ``verify --preview`` asks for when the head under review is not the
-    # commit checked out — project markers are read from the working tree, so
-    # a fetch would leave that request unmet (#397 review).
+    # not always a ref: making history available, and putting the evaluated
+    # commit in the worktree, which is what ``verify --preview`` asks for when
+    # the head under review is not the commit checked out (#397 review).
     #
-    # ``checkout`` and ``switch`` and nothing else: those are the verbs that
-    # move ``HEAD``. ``git restore`` rewrites working-tree files and leaves
-    # ``HEAD`` where it was, so crediting it would pass a cell that never
-    # produced the commit the route asked for.
+    # Which family is required is read from ``expects``, not accepted as a
+    # union.  Taking either satisfied a checkout request with ``git fetch`` and
+    # a fetch request with an unrelated checkout, so the criterion dropped an
+    # obligation the cell never met.
     if control.next_action.get("kind") == "fetch_base":
-        return bool(
-            re.search(
-                r"\bgit\s+(?:fetch|remote\s+update|checkout|switch)\b",
-                command,
-            )
-        )
+        requested = _CHECKOUT_REQUEST.search(control.next_action.get("expects") or "")
+        if requested is not None:
+            return _moves_head_to(command, requested.group("commit"))
+        return bool(re.search(r"\bgit\s+(?:fetch|remote\s+update)\b", command))
     return False
+
+
+# ``verify --preview`` spells a checkout request as "commit <id> checked out in
+# this worktree".  Matched as prose rather than imported, because this scorer
+# deliberately does not depend on the product schemas.
+_CHECKOUT_REQUEST = re.compile(
+    r"\bcommit\s+(?P<commit>[0-9a-f]{7,40})\b[^.]*?\bchecked out\b",
+    re.IGNORECASE,
+)
+
+
+def _moves_head_to(command: str, commit: str) -> bool:
+    """Whether ``command`` puts ``commit`` in the worktree.
+
+    Three near misses all read as checkouts and none of them satisfy the
+    request: ``git restore`` and ``git checkout -- <path>`` rewrite files while
+    leaving ``HEAD`` where it was, ``git switch main`` moves ``HEAD`` somewhere
+    else, and ``git checkout <other-sha>`` moves it to the wrong commit.  So the
+    verb has to be one that moves ``HEAD``, the path-restoring form is rejected,
+    and the target has to be the commit that was asked for — compared as a
+    prefix in either direction, since either side may be abbreviated.
+    """
+
+    if not re.search(r"\bgit\b.*\b(?:checkout|switch)\b", command, re.IGNORECASE):
+        return False
+    if re.search(r"\s--(?:\s|$)", command):
+        return False
+    commit = commit.casefold()
+    return any(
+        commit.startswith(token) or token.startswith(commit)
+        for token in re.findall(r"\b[0-9a-f]{7,40}\b", command.casefold())
+    )
 
 
 def _summary_satisfies_structured_request(summary: str, control: _ControlSnapshot) -> bool:
--- a/harness/adoption/scorer/rules.py
+++ b/harness/adoption/scorer/rules.py
@@ -1050,27 +1050,70 @@ _CHECKOUT_REQUEST = re.compile(
 )
 
 
+_ABBREVIATED_COMMIT = re.compile(r"[0-9a-f]{7,40}")
+# Git's own global options, and the ones among them that take a separate value.
+_GIT_GLOBAL_WITH_VALUE = frozenset(
+    {"-c", "-C", "--git-dir", "--work-tree", "--namespace", "--exec-path"}
+)
+_SHELL_SEPARATORS = frozenset({"&&", "||", ";", "|", "&"})
+
+
+def _git_invocations(command: str) -> list[tuple[str, list[str]]]:
+    """Each git subcommand in a possibly compound line, with its own arguments.
+
+    Parsed rather than pattern-matched because the pieces of a checkout are
+    otherwise free to come from three different commands: ``git log <sha> &&
+    npm run checkout-preview`` has a ``git``, the word ``checkout`` and the
+    requested commit in it, and checks nothing out.
+    """
+
+    tokens = command.split()
+    invocations: list[tuple[str, list[str]]] = []
+    index = 0
+    while index < len(tokens):
+        if tokens[index].rsplit("/", 1)[-1] != "git":
+            index += 1
+            continue
+        cursor = index + 1
+        while cursor < len(tokens) and tokens[cursor].startswith("-"):
+            cursor += 2 if tokens[cursor] in _GIT_GLOBAL_WITH_VALUE else 1
+        if cursor >= len(tokens):
+            break
+        name = tokens[cursor]
+        cursor += 1
+        args: list[str] = []
+        while cursor < len(tokens) and tokens[cursor] not in _SHELL_SEPARATORS:
+            args.append(tokens[cursor])
+            cursor += 1
+        invocations.append((name, args))
+        index = cursor
+    return invocations
+
+
 def _moves_head_to(command: str, commit: str) -> bool:
     """Whether ``command`` puts ``commit`` in the worktree.
 
-    Three near misses all read as checkouts and none of them satisfy the
-    request: ``git restore`` and ``git checkout -- <path>`` rewrite files while
-    leaving ``HEAD`` where it was, ``git switch main`` moves ``HEAD`` somewhere
-    else, and ``git checkout <other-sha>`` moves it to the wrong commit.  So the
-    verb has to be one that moves ``HEAD``, the path-restoring form is rejected,
-    and the target has to be the commit that was asked for — compared as a
-    prefix in either direction, since either side may be abbreviated.
+    Several near misses read as checkouts and none of them satisfy the request:
+    ``git restore`` and ``git checkout -- <path>`` rewrite files while leaving
+    ``HEAD`` where it was, ``git switch main`` moves ``HEAD`` somewhere else,
+    and ``git checkout <other-sha>`` moves it to the wrong commit.  So the
+    subcommand has to be one that moves ``HEAD``, the path-restoring form is
+    rejected, and the target has to be the commit that was asked for — read
+    from that invocation's own arguments, and compared as a prefix in either
+    direction, since either side may be abbreviated.
     """
 
-    if not re.search(r"\bgit\b.*\b(?:checkout|switch)\b", command, re.IGNORECASE):
-        return False
-    if re.search(r"\s--(?:\s|$)", command):
-        return False
     commit = commit.casefold()
-    return any(
-        commit.startswith(token) or token.startswith(commit)
-        for token in re.findall(r"\b[0-9a-f]{7,40}\b", command.casefold())
-    )
+    for name, args in _git_invocations(command.casefold()):
+        if name not in {"checkout", "switch"} or "--" in args:
+            continue
+        if any(
+            commit.startswith(argument) or argument.startswith(commit)
+            for argument in args
+            if _ABBREVIATED_COMMIT.fullmatch(argument)
+        ):
+            return True
+    return False
 
 
 def _summary_satisfies_structured_request(summary: str, control: _ControlSnapshot) -> bool:
--- a/src/agents_shipgate/cli/detect.py
+++ b/src/agents_shipgate/cli/detect.py
@@ -25,8 +25,8 @@ from agents_shipgate.cli.discovery import (
 )
 from agents_shipgate.cli.scope_routing import (
     MAX_LISTED_SCOPE_CANDIDATES,
+    candidate_caveats,
     describe_candidate,
-    is_adopted,
     scope_candidate_actions,
 )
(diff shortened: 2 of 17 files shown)


pengfei-threemoonslab (commented):
Reviewed the exact head `b4415a20f1a3f33c38b30187645770689859aad4` across preview control routing, setup identity, candidate recovery, downstream control consumers, tests, and docs. I found six issues that can leave the deterministic loop non-terminating or make its published identity/consumer behavior inconsistent; please address these before merge.

Submitting this as a comment review because the authenticated GitHub identity is also the PR author; the findings themselves are merge-blocking from this review.

Validation context: all three GitHub workflows are green; focused discovery/setup/control tests and Ruff passed; schema/`llms-full.txt` checks passed. The full Shipgate verifier reported `review_required` solely for the touched `AGENTS.md` trust root (3 review items, 0 blockers, no detected policy weakening).

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -3971,6 +3977,7 @@ def _preview_scope(
                 ),
             ),
             cause="head_mismatch",
+            head=head,
```
pengfei-threemoonslab, line 4088: [P1] Freeze commit-ish refs before asking for checkout

`--head` accepts revision expressions, not only stable branch names. With `--head HEAD~1`, the first preview evaluates parent P; checking out `HEAD~1` moves `HEAD` to P; the identical rerun now resolves `HEAD~1` to P's parent and emits this route again, walking backward instead of reaching scoped init. HEAD-relative `--base` can likewise change the evaluated range. Resolve the requested base/head to immutable commit IDs for the checkout and rerun, and add a `HEAD~1` regression.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -4019,22 +4026,62 @@ def _unresolved_scope_route(
...
 
     resolution = scope.resolution
-    if scope.cause in ("deleted_evidence", "unreadable_inventory", "head_mismatch"):
+    if scope.cause == "head_mismatch":
+        # ``head_mismatch`` is only produced for a non-None ``--head``: an
+        # absent one is the working tree by definition.
+        ref = scope.head or "the evaluated head"
+        return (
+            CodingAgentFetchBaseAction(
+                kind="fetch_base",
```
pengfei-threemoonslab, line 4175: [P1] Keep `fetch_base` consumers compatible with checkout semantics

This changes commandless `fetch_base` from “make a ref available” to “mutate the worktree checkout,” but the independent adoption scorer still accepts only `git fetch`/`git remote update` for this kind (`harness/adoption/scorer/rules.py:1024-1028`). It rejects the required `git checkout`/`git switch`, while accepting a fetch that leaves the new postcondition unmet. Introduce a distinct checkout/input action, or update every consumer plus the published semantics and conformance tests in this PR.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -4019,22 +4026,62 @@ def _unresolved_scope_route(
...
+                # A crisp noun phrase, because consumers interpolate it:
+                # the Claude Code stop hook renders `fetch_base` as
+                # "Make <expects> available locally".
+                expects=f"{ref} checked out in this worktree",
```
pengfei-threemoonslab: [P2] Make the requested checkout invalidate the old control pointer

Preview publishes this route with `head_commit_sha` and `head_tree_sha` unset. After checking out a stable `feature` ref, `agents-shipgate agent control` still exits 0 with the same `current_control_id` and the same unmet-looking checkout request. That contradicts the required refresh-after-checkout workflow and can make refresh-driven controllers repeat the action. Bind the source worktree's actual HEAD identity (distinct from the evaluated target ref), or otherwise ensure the expected checkout makes the preview pointer stale; test checkout followed by `agent control`.

src/agents_shipgate/cli/scope_routing.py
```diff
@@ -0,0 +1,127 @@
...
+    """
+
+    actions: list[NextAction] = []
+    # The workspace root is never offered as a command: it is the scope this
+    # run just refused, so running it again returns here. `.` stays in the
+    # reported candidate list because agent files that belong to no
+    # sub-project are real evidence of why the answer is unresolved, and
+    # `--allow-unresolved-scope` is the route that accepts them.
+    routable = [candidate for candidate in candidates if candidate.path != "."]
+    for candidate in routable[:MAX_LISTED_SCOPE_CANDIDATES]:
```
pengfei-threemoonslab: [P1] Do not apply the ten-item display cap to machine routes

The PR promises one exact command per candidate, but this slice silently drops candidate 11 onward. `detect --workspace samples --json` finds 22 candidates and emits only 10 commands; issue #397's own reproduction has 25. Selecting any omitted project still leaves no `command`/`executable`/`args`, recreating the dead end this change is meant to close. Keep the cap for human summaries only, or provide a target-specific/continuation mechanism that remains executable.

src/agents_shipgate/cli/scope_routing.py
```diff
@@ -0,0 +1,127 @@
...
+                    expects=f"An adoption kit reachable from {candidate.path}.",
+                )
+            )
+            continue
+        actions.append(
+            NextAction(
+                kind="command",
+                command=render_command(
+                    [
+                        "init",
```
pengfei-threemoonslab, line 320: [P1] Route already-adopted candidates through their manifest

`AgentProjectCandidate` can come from a nested `shipgate.yaml`, yet every candidate is sent to `init --write`. On this repository, all ten commands emitted by `detect --workspace samples --json` target directories that already have manifests; following one exits 2 as `config_already_exists` (and may update `.gitignore`) instead of advancing, while `expects` falsely promises a new manifest. Emit the existing-manifest `doctor`/verification handoff for adopted candidates.

src/agents_shipgate/cli/detect.py
```diff
@@ -104,6 +104,7 @@ def detect(
         advance=advance,
         advance_kind=advance_kind,
         advance_decision=advance_decision,
+        advance_alternatives=advance_alternatives,
```
pengfei-threemoonslab, line 122: [P2] Include the new alternatives in the setup identity

`advance_alternatives` are retargeted to the current invocation, but `input_id` hashes only the detection result and decision. I reproduced the same ambiguous detect under `AGENTS_SHIPGATE_CLI=agents-shipgate` and `/opt/custom/agents-shipgate`: candidate `command`/`executable` changed while `control.input_id` stayed identical. A cache keyed by the documented answer identity can therefore replay an unavailable command. Fold the selected advance and serialized alternatives into `routing_facts`, as the init/doctor paths do, and add the cross-entry-point identity test.

harness/adoption/scorer/rules.py
```diff
@@ -1023,8 +1023,25 @@ def _action_satisfies_control(action: _TimelineItem, control: _ControlSnapshot)
...
+    # the evaluated commit in the worktree (``git checkout``), which is what
+    # ``verify --preview`` asks for when the head under review is not the
+    # commit checked out — project markers are read from the working tree, so
+    # a fetch would leave that request unmet (#397 review).
+    #
+    # ``checkout`` and ``switch`` and nothing else: those are the verbs that
+    # move ``HEAD``. ``git restore`` rewrites working-tree files and leaves
+    # ``HEAD`` where it was, so crediting it would pass a cell that never
+    # produced the commit the route asked for.
     if control.next_action.get("kind") == "fetch_base":
```
pengfei-threemoonslab, line 1036: [P1] Match the recovery command to the requested input

This union still accepts an action from either family for every commandless `fetch_base`, without consulting `next_action.expects`. On this head, a checkout-shaped obligation (`commit 66f837355087 checked out in this worktree`) returns true for `git fetch origin main`, `git checkout -- AGENTS.md`, `git checkout deadbeef`, and `git switch main`; `respects_required_agent_action` then drops the obligation and its blocker criterion passes although the requested commit was never checked out. The new test even codifies the first false positive by pairing `git fetch` with checkout-shaped `expects`. Split matching by the requested input and reject path-only/wrong-target checkouts.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -3368,10 +3377,24 @@ def _current_control_workspace_identity(
     return CurrentControlWorkspaceIdentity(
         repository=_safe_repository_identity(git_root),
         head_ref=verifier.head_ref,
-        head_tree_sha=verifier.head_tree_sha,
+        head_commit_sha=_safe_worktree_sha(git_root, commit_sha),
```
pengfei-threemoonslab, line 3391: [P2] Bind the worktree content preview actually used

This fixes checkout invalidation by binding `HEAD`, but the fallback still leaves `snapshot_kind` and `worktree_overlay_sha256` null. I reproduced a preview whose route was selected from untracked `app/pyproject.toml` plus `app/agent.py`; deleting `app/agent.py` without moving `HEAD` left `agent control` exiting 0 with the same `current_control_id` and stale `initialize app` route. Preview reads project markers from the worktree, so bind an overlay/path-set identity (or equivalent) rather than only the committed HEAD/tree.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -4019,22 +4101,87 @@ def _unresolved_scope_route(
...
+            CodingAgentFetchBaseAction(
+                kind="fetch_base",
+                # A crisp noun phrase, because consumers interpolate it:
+                # the Claude Code stop hook renders `fetch_base` as
+                # "Provide <expects>".
+                expects=f"commit {target} checked out in this worktree",
+                # Short enough that an ordinary ref name leaves the whole
+                # route inside the envelope's 400-byte prose cap, and the
+                # instruction leads so a long one still survives truncation.
+                why=(
```
pengfei-threemoonslab, line 4189: [P2] Put the immutable rerun before truncatable ref prose

The compact control envelope caps `why` at 400 bytes, but `resolution.detail`—including the raw ref—comes before both the checkout instruction and pinned `--base`/`--head`. With a valid `feature/` + 220-character + `~1` ref, the bounded `why` contains neither `Check 123456789abc out` nor `--base abcdefabcdef --head 123456789abc`; `expects` names only the checkout commit. A controller that reruns the original HEAD-relative request can therefore recreate the backward-walking loop this fix targets. Lead with the immutable checkout/rerun, or publish the pins in untruncated structured data, and test the compact envelope.

src/agents_shipgate/cli/scope_routing.py
```diff
@@ -0,0 +1,230 @@
...
+    setup, so promising any here would be inventing it.
+    """
+
+    actions: list[NextAction] = []
+    # The workspace root is never offered as a command: it is the scope this
+    # run just refused, so running it again returns here. `.` stays in the
+    # reported candidate list because agent files that belong to no
+    # sub-project are real evidence of why the answer is unresolved, and
+    # `--allow-unresolved-scope` is the route that accepts them.
+    routable = [candidate for candidate in candidates if candidate.path != "."]
```
pengfei-threemoonslab, line 242: [P1] Keep a terminating route for the root candidate

`"."` remains a real entry in `agent_project_candidates`, and rank 1 tells the caller to choose from that full list, but this filter silently removes its only alternative. A root agent plus a nested project yields candidates `[".", "nested"]` and only the nested command; rerunning `init --workspace <root> --write` returns the same refusal. `--allow-unresolved-scope` is a separate whole-workspace semantic override (and can select the nested agent), not an exact route for the root candidate. Publish an explicit human-owned root resolution/restructure route, or stop presenting `"."` as a selectable project.

src/agents_shipgate/cli/scope_routing.py
```diff
@@ -0,0 +1,230 @@
...
+    for candidate in routable:
+        target = workspace / candidate.path
+        defines = ", ".join(candidate.agent_names)
+        manifest = is_adopted(target)
+        if manifest is not None and not init_refreshes_existing:
```
pengfei-threemoonslab, line 247: [P1] Preserve requested `--ci` setup for adopted candidates

This branch runs before `setup_flags` are applied. After `init --workspace <ambiguous-root> --write --ci --json` refuses the unresolved root (and writes no workflow), selecting an adopted candidate yields only `doctor --config … --json`; doctor can then hand off to verify while the requested CI setup was never installed. Directly, `scope_candidate_actions(..., setup_flags=["--ci"])` returns doctor args with no `--ci`, contradicting this function's flag-preservation contract. Route through an idempotent CI refresh or carry the missing CI obligation, and add adopted-plus-`--ci` coverage.

pengfei-threemoonslab (commented):
Re-reviewed the exact latest head `f1d03b2e371938f5ce075fe2355ff144aa34469d`. The new commits substantially address the six earlier threads, but five residual cases still break the deterministic recovery/control contract (three P1, two P2), so I do not consider this head merge-ready yet.

Submitting this as a comment review because the authenticated GitHub identity is also the PR author; GitHub does not permit a self `REQUEST_CHANGES` review.

Validation: all three GitHub workflows are green; the focused discovery/setup/scorer suite, Ruff, `git diff --check`, and schema-generation check passed. Full Agents Shipgate verification is current at this head and reports `review_publishable` / `review_required` solely because `AGENTS.md` is a touched trust root (3 review items, 0 blockers, no proven policy weakening).

PR #404, opened 2026-08-22
fix(evidence): account for every narrowing decision (#403)

### LINKED ISSUE #403: Epic: account for every narrowing decision — exclusion ledger, conservation invariant, and skip-requires-evidence
## The pattern

Across two first-time adoption walks — [google/adk-samples#1745](https://github.com/google/adk-samples/pull/1745) and [github/github-mcp-server#3076](https://github.com/github/github-mcp-server/pull/3076) — every failure had the same shape. Shipgate computed the correct signal, stored it, and did not connect it to the decision:

| Site | It knew | It did |
|---|---|---|
| trigger | the diff contains an MCP tool schema (`__toolsnaps__/delete_repository.snap`) | `should_run: false`, `skip_reason: "no_match"` |
| binding | `unbound_tools: 1` | `pass_eligible: true`, `gap_count: 0` |
| scope resolution (#394) | all 17 changed files share one directory prefix | routed setup to the repository root |
| project discovery (#395) | the walk stopped at the 1000-file cap | reported 53 projects as a complete list |
| reason text (#396) | `12/12 reachable, 0 unbound, 0 warnings` | "the tool surface could not be fully enumerated" |

These are not five unrelated bugs. They are one architectural habit: **a stage narrows the analysed surface, and the narrowing is neither recorded as evidence nor reachable by the release decision.** Each stage answers "is this in scope?" independently, and a negative answer removes the subject from every downstream stage — *including* from the evidence-gap accounting that exists to catch exactly this.

## Evidence A — a destructive tool addition, silently excluded

github-mcp-server#3076 adds `delete_repository` (`annotations: {destructiveHint: true, idempotentHint: false, readOnlyHint: false}`) to GitHub's official MCP server. Reproduced with the 117 checked-in `pkg/github/__toolsnaps__/*.snap` schemas merged into one `type: mcp` source, manifest reviewed at base declaring the 116-tool surface `complete: true`, head at 117.

One variable — whether the new tool appears in the declaration:

| | A: declaration lists 116 | B: declaration lists 117 |
|---|---|---|
| `binding_coverage` | `unbound_tools: 1`, `pass_eligible: **true**`, `gap_count: **0**` | `unbound_tools: 0` |
| `capability_review` | `added=0 modified=0 removed=0` | — |
| `delete_repository` mentions in `report.json` | **1** (only `tool_catalog[102].name`) | **41** |
| blockers naming it | **0** | **2** |
| verdict | `blocked` — for 14 pre-existing findings on the other tools | `blocked` — 16 |

Run B's blockers are exactly right:

```
SHIP-POLICY-APPROVAL-MISSING              delete_repository lacks a declared approval policy
SHIP-ACTION-DESTRUCTIVE-ROLLBACK-MISSING  delete_repository has destructive capability without required controls
```

The checks work. The tool is dropped from the analysed surface *before* they run, and the drop is counted (`unbound_tools: 1`) but not gated (`gap_count: 0`).

This is a fail-open in the reward-hacking shape the product exists to prevent, and it has a pincer: `shipgate.yaml` is a trust root, so updating the declaration returns *"This PR edits a release trust root — a coding agent cannot self-approve that change"*, while not updating it leaves the new destructive capability invisible. For an MCP **server**, "unbound" is not a meaningful safety state at all — a published tool is callable by any client.

## Evidence B — the gate never ran

Same PR, unmodified repository, full clone, worktree at the PR head:

```json
"diff_status": {"completeness": "complete"},
"trigger": {"should_run": false, "skip": true, "skip_reason": "no_match",
            "matched_rules": [], "diff_tokens": [],
            "rationale": "No rules matched; nothing in this PR signals a tool-surface change."}
```

`TRIGGER-MCP-EXPORT-CHANGED` matches on filename globs only — `**/*mcp*.json` and `.agents-shipgate/*.json`. The file is `pkg/github/__toolsnaps__/delete_repository.snap`: wrong extension, no `mcp` in the basename. Its *content* is unmistakably an MCP tool definition (`inputSchema` plus MCP annotation hints), but recognition is by naming convention. The same glob would also miss `mcp-server/tools.json`.

## Proposed work

**1. Exclusion ledger.** Every stage that removes a subject from the analysed surface emits one typed record — `{stage, subject, reason, source_ref}`. Trigger no-match, discovery cap truncation, unresolved scope, unbound tool, unparsed region, source-type completeness denial are all the same event class, currently handled five different ways (or not at all). One representation, carried on the report, reachable by `release_decision`.

- [ ] Define the record type and attach it to the report schema
- [ ] Emit from: trigger, discovery walk, scope resolution, binding, adapter parse, `_surface_is_complete`
- [ ] Surface the ledger in the human-facing reason text and in `next_action`

**2. Conservation invariant.** Enforce, in code and as a property test across every adapter and every fixture:

```
observed == analysed ∪ excluded
excluded ≠ ∅  ⟹  gap_count > 0
```

`binding_coverage` already computes both sides and reports `unbound_tools: 1` alongside `gap_count: 0`. The invariant makes that state unrepresentable rather than relying on each call site to remember.

- [ ] Property test over `samples/` and `benchmark/repos/` asserting conservation
- [ ] Negative controls — mutate a fixture to drop a tool and assert the ledger catches it
- [ ] Wire `excluded ≠ ∅` into the release decision so exclusions cannot silently pass

**3. `skip` requires positive evidence.** A diff containing files the trigger cannot classify is not evidence of irrelevance. `TRIGGER-DOCS-ONLY-NEGATIVE` is a legitimate skip; `no_match` is not. Route unclassified content to `unknown` and let the scan decide.

This is the #308 monotonicity rule — already accepted for diff *readability* — applied to diff *comprehension*. Content-shaped recognition (JSON carrying `inputSchema` plus MCP annotation hints is a tool surface, whatever it is named) follows naturally once the trigger has to justify a skip.

- [ ] `no_match` on unclassified changed files becomes `unknown`, not `skip`
- [ ] Record the unclassified files in the exclusion ledger
- [ ] Recognise MCP tool schemas by content, not filename

## Why this first

It is largely additive and it depends on nothing else being correct. Landing it does not require fixing extraction confidence (#393), scope resolution (#394), the discovery cap (#395) or the reason text (#396) — it makes each of those *visible* rather than silent. A gate that says "I saw 32 changed files and could not classify any of them" is honest and useful; one that says "nothing in this PR signals a tool-surface change" is not.

Instances and related: #393, #394, #395, #396, #397, #398.

---

Evidence gathered 2026-08-20/21 during two first-time-user adoption walks at CLI `0.16.0b7`, contract 24.

pengfei-threemoonslab:
Closes #403.

## The pattern

Every failure in the issue had the same shape: a stage computed the correct signal, stored it, and did not connect it to the decision. The narrowing was neither recorded as evidence nor reachable by `release_decision`.

## Evidence A — a destructive tool, silently excluded

`github/github-mcp-server#3076` adds `delete_repository` (`destructiveHint: true`) to a published MCP server. With the reviewed declaration still listing the base commit's 116-tool surface, the run reported `unbound_tools: 1` beside `gap_count: 0` and `pass_eligible: true`, and named the new tool exactly once in the whole report — as a `tool_catalog` row.

The checks are correct: declaring the tool produces `SHIP-POLICY-APPROVAL-MISSING` and `SHIP-ACTION-DESTRUCTIVE-ROLLBACK-MISSING`. The tool left the analysed surface before they ran, and it could not be declared without editing `shipgate.yaml` — a trust root a coding agent cannot self-approve. Both options were bad.

Reproduced as a test in [`tests/test_surface_exclusions.py`](tests/test_surface_exclusions.py); it fails on `main`.

## What landed

**1. The exclusion ledger.** `surface_exclusions` on the report (schema `0.34 → 0.35`), one typed `{stage, subject, reason, source_ref, detail, accounting}` record per subject a stage removed. `detect --json` and `trigger --json` emit the same record for the stages they own — a capped discovery walk, an unresolved manifest scope, a glob-matched source the adapter rejects, an unclassified change set — replacing four ad-hoc spellings of one event class.

It is **derived**, not emitted at each call site. An emitter has to be remembered, which is the habit that produced `unbound_tools: 1 / gap_count: 0` in the first place; a derivation reads the same facts the decision read.

`accounting` is what makes a record checkable rather than decorative:

| Value | Claim | Checked by |
| --- | --- | --- |
| `evidence_gap` | a gap row names this exact subject | the invariant below |
| `route_blocked` | the stage withheld its verdict; its `next_action` repairs it | the stage's own verdict |
| `not_claimed` | nothing claims the subject as capability, and this change did not introduce it | the base comparison |

**2. Conservation invariant**, enforced in `validate_semantic_consistency` at emission:

```
observed == analysed ∪ excluded
every excluded subject is in the ledger
every `evidence_gap` record is backed by a gap row with the same subject
a subject this change newly excluded is never `not_claimed`
```

Plus a negative control per clause: mutate a report to keep the record and drop the gate, and it is refused rather than written.

**3. `skip` requires positive evidence.** `TRIGGER-DOCS-ONLY-NEGATIVE` classifies every changed file and concludes — a legitimate skip. `no_match` classifies nothing. A non-empty change set no rule classified now returns `evaluation_status: "unclassified"` with `should_run: null` and a `next_action` routing forward to the scan; an empty change set keeps `no_match`, because there it is a fact about the PR. This is the #308 monotonicity rule applied to diff *comprehension* rather than *readability*.

Trigger catalog `0.3 → 0.4` also adds `TRIGGER-MCP-TOOL-SCHEMA-CONTENT`: an MCP input schema beside MCP annotation hints is a tool surface whatever the file is named. `TRIGGER-MCP-EXPORT-CHANGED` matches `**/*mcp*.json`, which misses `pkg/github/__toolsnaps__/delete_repository.snap` and `mcp-server/tools.json` alike.

## The one judgement call worth reviewing

The issue proposes wiring `excluded ≠ ∅` straight into the release decision. That would regress a deliberate boundary: `samples/large_multi_framework_agent` has **58** unbound OpenAPI operations by design, and #385 established that catalog membership is not evidence of capability — gating on those makes declaring a spec or an MCP server self-blocking.

So the discriminator here is the **base comparison**, not the source type. `binding_surface_diff` gains `added_unbound_tool_ids` (head exclusions minus base exclusions); a tool in that set raises a `missing_binding_evidence` gap naming it. A subject that was already excluded stays a reviewer-facing `not_claimed` row; a subject *this change* newly excluded is a gap. A plain `scan` has no base, so nothing fires on one.

Both halves are pinned by tests: the destructive-tool case gates, the pre-existing case does not, and the remedy the gap prescribes actually clears it (a prescribed fix whose side effect blocks it is a bug class this repo has hit twice — #385, #386).

## Verification

- Full suite green (`pytest tests/ -n auto`, exit 0); `ruff check .` clean.
- The Evidence A regression test fails with the fix disabled — and it fails at the *invariant*, which is where it should.
- All schema-bump touchpoints done: regenerated schemas (v0.34 frozen), 5 sample goldens, 8 codex-boundary goldens, privacy inventory, adoption-kit renderer source + render hash + `prior_render_sha256`, plugin skill copy, `llms-full.txt`, and the doc version stamps.
- Packet goldens and `run_id` are untouched — the ledger is top-level and outside the run identity hash.

## Surface discipline

Per `CONTRIBUTING.md` § Surface discipline: the headline metric is `blocked_recall` on real history. This is the first change that makes an unjudged capability reach the verdict at all, and it extends the one decision engine (`build_release_decision`) rather than adding a parallel one — `release_decision.decision` remains the only gate.


pengfei-threemoonslab:
## Review pass 1 — 3 findings, all addressed in 17753cd1

**1. (confirmed, high) The ledger could not join a tool with itself.**
`partial_binding_evidence` named its subject by the raw canonical tool id, and so did every binding graph-issue row; `_unreached_tool_gaps` and `_semantic_gap` rendered `name [provider]`. `build_surface_exclusions` looked up the second spelling, missed, and recorded a tool the decision *had* gapped as `not_claimed`:

```
binding_coverage.gap_count: 1
surface_exclusions.gated:   0
```

That is `unbound_tools: 1 / gap_count: 0` re-expressed one layer up — inside the mechanism built to make it unrepresentable — and the full suite stayed green on it. Reproduced against `build_release_decision` + `build_surface_exclusions` directly.

Fix: one `catalog_subject` spelling at every tool-scoped emitter. Regression test `test_a_possibly_reachable_tool_is_recorded_as_gated`; verified it fails when the raw-id spelling is restored.

**2. (confirmed) The invariant only checked one direction.** It rejected a record claiming a gap that does not exist, but nothing rejected a gap existing while the ledger said `not_claimed` — which is why #1 passed. Added two claims: no joinable gap may name a catalog tool by raw id (checked at the source, not by re-running the same join, which would repeat any shared mistake), and an excluded tool the decision gapped may never be recorded `not_claimed`.

**3. (plausible, latent) `gated` conflated `route_blocked` with `evidence_gap`.** The `gated > 0 ⟹ evidence_gaps` guard would have failed a scan the first time a report stage emitted a legitimate `route_blocked` row. Removed — claim 2 subsumes it exactly.

### Scoping note

The raw-id spelling rule is deliberately scoped to the gap kinds the ledger joins. Enforcing it globally trips on `inferred_policy_applicability`, which names tools by raw canonical id in `policy_evidence_gaps` — that reads badly in `Improve evidence:` and is worth its own issue, but it is an actionability question, not a conservation one, and fixing it here would churn an unrelated published surface.

Full suite green (exit 0), `ruff check .` clean, no golden drift.

pengfei-threemoonslab:
## Review passes 2 & 3 — coverage, bounds, adversarial probes

**Pass 2 (coverage + efficiency)**

- `build_detect_exclusions` had **no coverage at all**. The three paths it derives from — a capped walk, a contested scope, a rejected source candidate — were exercised only through fields it does not read, so a bug in the builder was invisible. Added a test per path plus the negative case (a settled workspace records nothing), each verified against real `detect` output rather than a constructed `DetectResult`.
- The trigger's ledger is now bounded at 25 entries, not the shared 200. When no rule matches, *every* changed file is unclassified, so those rows enumerate a list the same payload already carries in full under `changed_files` — while the result is embedded verbatim in `verifier.json` and in the Codex boundary payload written to stdout. A 500-file PR was adding ~200 copies of one identical sentence. `total` and `gated` stay exact, so nothing that reads the counts is affected.
- Added the epic's property sweep over every bundled manifest. The invariant is enforced at emission, so a scan returning at all already proves it — but that made the proof depend on which samples other tests happen to scan. The sweep also states the half emission cannot check for itself: a fixture with a non-empty excluded set never ships an empty ledger. `benchmark/repos/` is materialized from `samples/` (eight of its nine archetypes are copies, per its README), so this covers the benchmark corpus by construction rather than by a second sweep that would drift.

**Pass 3 (adversarial)**

Probed whether untrusted source content can escape the ledger or forge an accounting. An MCP export that self-declares `annotations.agent_bindings: {root: true}` has the reserved keys stripped (#268 holds), lands in `added_unbound_tool_ids`, is gated, and the verdict is `insufficient_evidence` — the ledger names it `newly_unbound_tool / evidence_gap`.

Subject forgery is closed structurally: the provider half of a subject comes from `tool_identity.bindings[].provider` or the configured `source_id` — both trust-root values, neither reachable from file content — and `newly_unbound_tool` must be `evidence_gap` regardless of the join, so a collision can only over-report gating, never under-report it.

**Merge + CI**

Merged `origin/main` (#402 landed after the branch point). Three prose conflicts — `AGENTS.md`, `CHANGELOG.md`, `llms-full.txt` — resolved keeping both entries, `llms-full.txt` regenerated from source. No source conflicts, and the full suite passes on the merge result.

All checks green: `test`, `verify`, `verify-self`, `clean-checkout-launcher`, `windows-launcher`.

Finally, the epic's "surface the ledger in the reason text and `next_action`" box, end to end on the reproduction:

```
DECISION: insufficient_evidence
REASON  : Insufficient evidence: the agent's tool bindings are unproven
          (delete_repository [s]). Fix at shipgate.yaml#agent_bindings.declarations.
          Context: 1 binding evidence gap(s) and 1 semantic evidence gap(s); …
```

pengfei-threemoonslab (commented):
Changes are needed before merge. I reviewed the exact head `50bc8bf823e6b53cf7e4d50f0fd46fee6ee87b22` with independent trigger/contract, release-decision, and schema/adversarial passes. The focused suite is green, but the following fail-open and wire-contract cases are reproducible and are not covered by the current tests.

### [P1] Do not let a negative detector discard an explicit capability match

At `src/agents_shipgate/triggers.py:622-628`, `stop_fired` wins before `has_force_run` or `has_run`. With the exact kind of path this PR adds content recognition for:

```python
detect = {
    "is_agent_project": False,
    "suggested_sources": [],
    "codex_plugin_candidates": [],
    "python_parse_truncated": False,
}
evaluate(
    paths=["pkg/github/__toolsnaps__/delete_repository.snap"],
    diff_text='+"inputSchema": {}\n+"destructiveHint": true\n',
    detect_result=detect,
)
```

the result contains `TRIGGER-MCP-TOOL-SCHEMA-CONTENT` with `action="run_shipgate"`, yet publishes `should_run=false`, `skip_reason="stop_conditions"`, `next_action.kind="stop"`, and an empty exclusion ledger. This preserves the exact #403 fail-open in the normal pre-adoption flow that supplies a complete negative detect result. A positive capability-content match disproves the detector's negative conclusion; the stop route should only be terminal when no run/force-run rule matched. Please add this exact regression.

### [P1] Fail closed when a requested base comparison is unavailable

`src/agents_shipgate/ci/release_decision.py:880-882` treats `binding_surface_diff.enabled == false` as equivalent to “no comparison requested.” It is also false when a valid supplied base report predates v0.31 and has no `binding_surface_facts`, and verify can reach the same state after a base scan failure.

I reproduced the production `run_scan(..., diff_from_path=...)` path with a valid v0.30 base and a head adding unbound destructive `delete_repository`. The scan exited 0 with:

```text
release_decision.decision = passed
binding_surface_diff.enabled = false
binding_coverage.unbound_tools = 1
binding_coverage.gap_count = 0
evidence_gaps = []
```

That contradicts `docs/engineering/ai-coding-workflow-verifier.md:77`, which says base-scan failure must never weaken the head release gate. Please distinguish “comparison not requested” from “requested comparison unavailable” and route the latter to insufficient evidence/human review. Cover both the legacy `--diff-from` report and verify base-scan-failure paths.

### [P1] Propagate the new tri-state trigger contract to every shipped consumer

The runtime adds `evaluation_status="unclassified"` and `should_run=null`, but the operational consumers still coerce the null to a skip:

- `src/agents_shipgate/cli/install_hooks.py:774-775` and `:924-925` both return when `not result.get("should_run")`. Feeding either hook the new unclassified result emits no guidance and exits 0.
- `skills/agents-shipgate/prompts/decide-shipgate-relevance.md:25,100-103,115-121,135-137` (and the adoption/plugin mirrors) still pins catalog 0.3 and instructs agents that no match means false / do not propose Shipgate.
- `docs/triggers.json:3` is 0.4, while `src/agents_shipgate/schemas/contract.py:106`, `.well-known/agents-shipgate.json:300`, `docs/agent-contract-current.md:240`, and the local contract still advertise 0.3. Directly, `load_triggers()` returns 0.4 and `build_contract_payload()` returns 0.3.
- `docs/triggers.json:428` still publishes `no rules matched (skip)`.

These are the supported entry points external agents actually follow, so they preserve the old fail-open even though the Python function changed. Please update the hooks to branch on `evaluation_status`/`next_action`, update and regenerate every prompt/kit/contract mirror, document the nullable fields and forward scan route, and add a cross-surface equality plus consumer integration test.

### [P1] A dry-run match must not hide unclassified sibling files

The unclassified conversion at `src/agents_shipgate/triggers.py:691` only runs when the whole result's `skip_reason` is `no_match`. Consequently:

```python
evaluate(
    paths=["requirements.txt", "pkg/github/__toolsnaps__/opaque.snap"],
    diff_text="+langchain==1.2\n+opaque=true\n",
)
```

returns `evaluation_status="evaluated"`, `should_run=false`, `skip_reason="dry_run_only"`, and an empty exclusion ledger. The dependency rule classified one signal; it did not classify the opaque sibling. Bundling an unknown capability file with a framework bump therefore restores a confident skip, and the current hook consumers suppress even the advisory action. Please track coverage per changed path (or conservatively withhold a dry-run-only skip when any sibling is uncovered), ledger those uncovered paths, and test pure dependency-only and mixed changes separately.

### [P2] Account for `invalid_binding_annotation` gaps

`BINDING_GAP_KINDS` at `src/agents_shipgate/core/surface_exclusions.py:64-72` omits `invalid_binding_annotation`, even though the release decision emits and routes that gap kind at `src/agents_shipgate/ci/release_decision.py:1138`. An unbound tool with a tool-scoped invalid binding annotation gets an `invalid_binding_annotation` evidence gap, but `_binding_exclusions` labels its ledger row `accounting="not_claimed"`; `validate_semantic_consistency()` accepts the contradiction. Add the missing kind and a regression proving the row is `evidence_gap`.

### [P2] Do not convert every source warning into a fabricated adapter omission

`src/agents_shipgate/core/surface_exclusions.py:202-218` maps every `source_warning` to “part of that input never entered the catalog.” `source_warning` is broader than partial parse/catalog loss. The committed `samples/simple_crewai_agent/expected/report.json` demonstrates this: `FileReadTool` is present in `tool_catalog` and `tool_inventory`, is structurally reachable, and has high-confidence identity/binding/effect evidence, yet the ledger says its adapter excluded part of the input and it “never entered the catalog.” The duplicate-base-action-ID warning produces the same false provenance. Please preserve a typed warning/excluded-subject fact and emit an adapter exclusion only for an actual omission.

### [P2] Tighten the v0.35 JSON Schema around the new safety fields

Although `surface_exclusions` is top-level-required, `$defs.SurfaceExclusionLedger` at `docs/report-schema.v0.35.json:5507-5535` requires none of `entries`, `total`, `gated`, or `truncated`; `BindingSurfaceDiff.added_unbound_tool_ids` is optional too. With `Draft202012Validator`, replacing a valid golden's ledger with `{}` produces zero validation errors, as does deleting `added_unbound_tool_ids`. Thus a nominally valid v0.35 report can erase the evidence/counts introduced by this PR. Please pin the nested required lists (and nonnegative/bounded/count constraints where expressible) in `scripts/generate_schemas.py`, then add negative schema tests.

### [P2] The cap still drops gated evidence

`src/agents_shipgate/schemas/exclusions.py:150-167` sorts gated rows first and then slices all rows with `rows[:limit]`. With `MAX_LEDGER_ENTRIES + 1` `evidence_gap` entries, the serialized ledger contains only 200 rows while reporting `gated=201`. That directly contradicts the adjacent guarantee that the cap cannot discard the row proving a subject was gated; semantic completeness is also skipped once `truncated` is true. Apply the cap only to `not_claimed` rows (or explicitly represent gated truncation) and add the 201-gated-row regression.

### Verification performed

- Focused tests: `tests/test_surface_exclusions.py`, `tests/test_trigger_command.py`, `tests/test_public_surface_contract.py`, and `tests/test_install_hooks.py` all pass on this head.
- `git diff --check` passes; all GitHub checks are currently green.
- Shipgate receipt reproduced successfully. Current control is `review_publishable`, release decision `review_required` because this PR changes trust roots; updating the PR is permitted, merge/completion remain human-owned.

The first four items are release-blocking because each can still turn an unjudged capability change into a skip or passed decision. The P2 items undermine the new conservation artifact's accuracy or its published validation contract.

pengfei-threemoonslab:
## All eight findings addressed in 120ccce5

Each one reproduced first, against the exact code paths named. Thank you — the four P1s were real fail-opens, and two of them (the `stop_conditions` precedence and the `dry_run` sibling) reinstated the very failure this PR exists to close, in routes I had not thought to probe.

### [P1] Stop conditions no longer discard an explicit capability match

Reproduced exactly: `TRIGGER-MCP-TOOL-SCHEMA-CONTENT` matched and the verdict was `should_run: false, skip_reason: "stop_conditions", next_action.kind: "stop"`.

The stop is now terminal only while nothing contradicts it. Your framing is the one I took into the code comment: the block's premise is *"this workspace is not an agent project"*, read by `detect` from the working **tree**; a matched run rule is evidence from the **diff**, which can carry what `detect` never saw — and the `.snap` file is precisely invisible to `suggested_sources`, so the block held while the rule that recognised it was thrown away.

Only `force_run`/`run_shipgate` override it; `dry_run` is advisory by construction and `skip_shipgate` agrees with the stop, so neither has anything to contradict it with. `stop_conditions_fired` keeps its raw meaning and a new `stop_conditions_terminal` says whether it decided. A plain non-agent file still stops. Regression added, plus the published `action_precedence` and the agent prompt's decision tree.

### [P1] A requested-but-unavailable base comparison fails closed

Reproduced with a v0.30 base: `unbound_tools: 1`, `gap_count: 0`, and the ledger asserting `not_claimed` — a claim that requires the comparison nobody ran.

`BindingSurfaceDiff.base_comparison_requested` separates the two states, and `VerificationContext.base_comparison_unavailable` carries the verify base-scan-failure path, which the head scan genuinely could not see (`diff_from_path` is just `None` there). That state now raises **one** gap naming the unusable base with a `provide_source` repair — one mechanism, one repair, one row, per the #361 rule — rather than fabricating per-tool provenance. Ledger rows become `unverified`, a fourth accounting value whose invariant is that the base gap must exist; `not_claimed` was the lie.

```
requested: True  enabled: False
unbound: 1  gap_count: 1
gaps: ['base comparison (report schema 0.30)', ...]
ledger: [('delete_repository [s]', 'unverified_unbound_tool', 'unverified')]
```

### [P1] The tri-state reaches the consumers that act on it

You were right that this was the load-bearing one — the Python function changing means nothing if the surfaces agents follow still coerce `null`.

- The hooks branch on `evaluation_status` and now word the withheld case honestly ("could not decide whether this diff is relevant"), because "trigger matched" misdescribes its own evidence. Two consumer tests: it speaks on withheld, stays silent on a real `false`.
- `decide-shipgate-relevance.md` teaches the tri-state, both withheld states, the new precedence, and *why* — regenerated through the kit renderer into all four mirrors, with the outgoing hash appended to `prior_render_sha256`.
- `trigger_catalog_schema_version` moved in step across `schemas/contract.py`, `.well-known`, `docs/agent-contract-current.md`, and the rendered local contract (outgoing hash appended there too, so a pinned repo can upgrade in place). **Nothing compared the two**, which is why it drifted — `test_every_surface_advertises_the_catalog_version_the_loader_returns` now asserts the loader's value against all five.
- `action_precedence` no longer publishes `no rules matched (skip)`.

### [P1] A `dry_run` match no longer hides an unclassified sibling

Coverage is per changed path now, not per change set. `TRIGGER-DOCS-ONLY-NEGATIVE` is unaffected by construction — its `every_file_matches` leg only fires when it classified every file, which is exactly the difference between a negative rule and an absent one. Tested separately: mixed withholds and ledgers only the uncovered sibling; pure dependency-only keeps `dry_run_only` with the advisory; docs-only keeps `skip_rule`.

One thing your report surfaced indirectly: a rule whose action this build does not recognise contributes no coverage either. It fired, but it decided nothing, and counting its files as classified would let a malformed catalog buy a skip it never earned.

### [P2] `invalid_binding_annotation`

Derived `BINDING_GAP_KINDS` from `AgentBindingIssue.kind` rather than fixing the copy — a hand-kept list is what drifted. The test asserts it stays derived.

### [P2] Fabricated adapter omissions — removed, not reworded

You are right, and `simple_crewai_agent` is the proof: `FileReadTool` is in `tool_catalog`, in `tool_inventory`, structurally reachable, high-confidence — and the ledger said part of that input never entered the catalog.

I looked for the typed fact to gate this on. There isn't one: `LoadedToolSource` carries `warnings: list[str]` and nothing else, so telling a real omission from a low-confidence note means reading prose. Every omission that *can* be proven (wildcard exports, unenumerated capability surfaces) already reaches the ledger through `surface_completeness`, whose signal is the tool's own `extraction.surface`, and the warnings remain `source_warning` evidence gaps the decision reads exactly as before. So the emitter is gone and the stage stays in `ExclusionStage` with the reasoning written down — restoring it should add an emitter, not a vocabulary. The decision loses nothing; the ledger loses a claim it could not support, which in a mechanism whose whole value is honest accounting is the more expensive of the two.

### [P2] Schema tightened

`SurfaceExclusionLedger`, `SurfaceExclusion`, and `BindingSurfaceDiff` all pin their required lists; `total`/`gated` are bounded below at zero. Negative tests cover erasing the ledger, dropping either count, dropping `added_unbound_tool_ids` or `base_comparison_requested`, and a negative count. `gated <= total` isn't expressible in JSON Schema — `validate_semantic_consistency` checks it.

### [P2] The cap

Fixed as you suggested: `evidence_gap` rows are never discarded, the cap applies to the rest. 201 gated rows → 201 entries, `truncated: false`. Regression added at exactly that boundary. The trigger's own 25-row bound still holds, since its rows are `route_blocked`.

---

Full suite green (exit 0), `ruff check .` clean, schemas regenerated, five sample goldens and eight codex-boundary goldens refreshed, `llms-full.txt` rebuilt.

pengfei-threemoonslab (commented):
Changes are still needed before merge. I re-reviewed the exact head `120ccce510681be5682e05b6974d30546ff2226e` with independent trigger/consumer, release-decision, and schema/adversarial passes. The remediation fixes the original producer paths, and the focused tests plus all GitHub checks are green, but the following edge cases still reproduce.

### [P1] Project the winning trigger verdict in the GitHub Action

`scripts/github_action_outputs.py:46-58` still reads the raw `stop_conditions_fired` bit before `should_run`, and maps every nonempty withheld result to `"none"`. On this head:

```text
runtime: should_run=true, stop_conditions_fired=true,
         stop_conditions_terminal=false
Action:  trigger_action="skip_shipgate"

runtime: evaluation_status="unclassified", should_run=null,
         next_action="agents-shipgate verify --preview --json"
Action:  trigger_action="none", should_run=""
```

The first case is the exact capability-content/negative-detect combination fixed in the runtime, but the Action reinstates the skip. The second collapses both `unclassified` and `not_evaluated` into an indistinguishable no-op; `action.yml` exposes no `evaluation_status` output from which a workflow can recover. Branch on `stop_conditions_terminal` / the winning verdict, export `evaluation_status` (or the exact typed next action), and add consumer tests for stop+run, unclassified, and partial input. The CLI text at `src/agents_shipgate/cli/trigger.py:224-225` and `src/agents_shipgate/triggers.py:1114-1115` should likewise stop saying a raw stop overrode a published `RUN` verdict.

### [P1] Do not advertise a repair command that removes the comparison

The new unavailable-base gap at `src/agents_shipgate/ci/release_decision.py:976-988` emits:

```text
agents-shipgate scan -c shipgate.yaml --format json
```

while `expects` says to regenerate the base and rerun the head with `--diff-from`. In the v0.30-base reproduction, the guarded run is `insufficient_evidence`, `base_comparison_requested=true`, with one binding gap and an `unverified` ledger row. Executing the advertised command on the same clean head drops the comparison and produces `passed`, zero binding gaps, and `not_claimed`. This command is also promoted into `fix_task.allowed_repairs`, so following the machine-readable repair clears the guard without repairing its evidence. Publish no command until the base workspace/path is known, or emit an exact base-pinned/two-step repair, and test that following it cannot clear the gap without a usable base.

### [P1] Keep first adoption out of the failed-base-comparison route

`src/agents_shipgate/cli/verify/orchestrator.py:971-979` marks every resolved base with no report as `base_comparison_unavailable`, including the expected `missing_manifest` state for first adoption. Reproduction: the base has an unchanged MCP catalog with reachable `list_issues` plus deliberately unwired `delete_repository`, but no `shipgate.yaml`; the head adds only `shipgate.yaml` and binds `list_issues`. Verify reports `manifest_introduced=true`, `base_status="missing_manifest"`, then `insufficient_evidence` with a request to regenerate a base report that cannot exist. This makes ordinary first adoption over a partial catalog permanently unfinishable unless unrelated catalog operations are falsely bound, contrary to the #385 boundary. Treat `manifest_introduced && base_status==missing_manifest` as the existing first-adoption human-review route, and only classify excluded tools from changed sources as newly unverified (or compare base inputs using the head manifest). Add paired unchanged-source/changed-source adoption tests.

### [P1] Enforce the converse of the unavailable-base ledger invariant

`src/agents_shipgate/core/semantic_consistency.py:240-246` checks only `accounting="unverified" => base gap`. It does not require excluded binding rows to remain unverified when the comparison was requested but unavailable. Starting from the generated one-tool unavailable-base report, changing its row to `reason="unbound_tool"`, `accounting="not_claimed"`, and `gated=0` is accepted by `validate_semantic_consistency()`, even though the base gap remains. That makes the new fail-closed state erasable without tripping the conservation check. Add the reverse invariant: an unavailable requested comparison cannot coexist with a `not_claimed` binding exclusion.

### [P2] Preserve a malformed `--diff-from` as a requested comparison

At `src/agents_shipgate/cli/scan/sanitization.py:147-163`, `base_comparison_requested` looks only at a successfully loaded `diff_reference`, not `diff_reference_error`. Supplying `{broken` as the base report while the head has an unbound destructive tool yields:

```text
base_comparison_requested=false
binding_coverage: unbound_tools=1, gap_count=0, pass_eligible=true
ledger: delete_repository / unbound_tool / not_claimed
```

The generic parse warning keeps the overall decision at insufficient evidence, but the new binding provenance says no comparison was requested and asserts the tool predates this change. Carry the supplied/requested bit independently of successful parsing, with regressions for malformed, missing, and legacy references.

### [P2] Do not join tools by the non-unique display label

`catalog_subject()` at `src/agents_shipgate/core/surface_exclusions.py:36-47` returns `name [provider]`, and that string is now the conservation join key. Two catalog IDs can legitimately share both fields. With unbound IDs `tool_v2:old` and `tool_v2:new`, both rendered as `dup [api]`, and only `new` in `added_unbound_tool_ids`, the decision emits one gap but the ledger marks both rows `evidence_gap` and reports `gated=2`; semantic validation accepts it. Join on the canonical tool ID (while retaining the label for display) and test the collision.

### [P2] Validate the ledger counts the schema tells consumers to gate on

`SurfaceExclusionLedger` at `src/agents_shipgate/schemas/exclusions.py:144-151` has no count validator, and `src/agents_shipgate/core/semantic_consistency.py:204-207` never inspects `gated`, despite `scripts/generate_schemas.py:615-617` saying it does. Both Pydantic and the generated schema accept `entries=[]`, `total=0`, `gated=999`, `truncated=false`; full semantic validation accepts it too. Enforce `0 <= gated <= total`, exact visible accounting when untruncated, and compatible lower bounds when truncated. While tightening the wire shape, `BindingSurfaceDiff.required` at `scripts/generate_schemas.py:626-636` also omits the emitted nullable `base_report_schema_version`; deleting it still validates, unlike the other stable diff fields.

### [P2] Preserve every kind counted as gated across truncation

The cap fix at `src/agents_shipgate/schemas/exclusions.py:173-188` preserves only `accounting="evidence_gap"`, while `gated` counts every value except `not_claimed`. With 201 `route_blocked` rows—or 201 `unverified` rows—the ledger keeps 200, reports `gated=201`, and sets `truncated=true`. This still contradicts `docs/report-reading-for-agents.md:183-184` ("the cap never drops" a gated row) and the class contract. Preserve every acted-on accounting value, or revise the published guarantee and give consumers an explicit gated-truncation representation; add both boundary cases.

### [P2] Record real adapter omissions instead of declaring the stage empty

The comment at `src/agents_shipgate/core/surface_exclusions.py:219-242` removes the false warning-derived emitter but leaves `adapter_parse` with no emitter. A concrete MCP input with one valid tool and one integer entry hits `src/agents_shipgate/inputs/mcp.py:85-88` (`Skipping non-object MCP tool entry`): the skipped entry never enters the catalog, the decision correctly carries a `source_warning` evidence gap, but `surface_exclusions` is `{entries: [], total: 0, gated: 0}`. The report therefore still does not account for every narrowed subject it claims to enumerate. Add a typed excluded-subject fact to adapter results and ledger only proven omissions; do not restore the blanket prose-warning mapping.

### [P2] Finish propagating trigger schema 0.4 and the new accounting value

`AGENTS.md:483` and `llms-full.txt:508` still advertise trigger schema `0.3` and explain only `not_evaluated`; `docs/agent-contract-current.md:605-613` likewise omits the complete-input `unclassified` route. This contradicts the runtime, `docs/triggers.json`, and `.well-known`, which are now 0.4. Separately, `docs/report-reading-for-agents.md:164-177` omits the new `unverified` accounting enum/table entry. These are authoritative agent-facing surfaces, so strict or prompt-driven consumers can still reject or mishandle the new wire values. Update/regenerate them and expand the version/enum parity test—the current “every surface” version test does not include `AGENTS.md` or `llms-full.txt`.

### Verification performed

- Focused trigger, Action-output, hook, public-contract, schema, exclusion-ledger, and verify suites pass; `git diff --check` passes; all GitHub checks are green.
- The adversarial reproductions above were run against this exact head, including both Action projections, malformed base input, duplicate display identities, count forgery, both capped accounting values, and an actual MCP adapter omission.
- Agents Shipgate verification succeeded and its receipt reproduced as valid. Current control is `review_publishable`, release decision `review_required` because this PR edits trust roots; updating the PR is permitted, while merge/completion remain human-owned.

The four P1s remain release-blocking. The P2s are contract/integrity gaps in the new conservation evidence and its supported consumers.

pengfei-threemoonslab:
## All ten findings addressed in 2aeb924f

Each reproduced first. Two of the P1s were mine to own twice over — the Action projection and the first-adoption route are both consequences of the review-1 fixes, in surfaces I did not follow the change into.

### [P1] The Action reinstated the skip the runtime refused

Reproduced exactly:

```
runtime: should_run=true, stop_conditions_fired=true, stop_conditions_terminal=false
Action:  trigger_action="skip_shipgate"      →  now "run_shipgate"

runtime: evaluation_status="unclassified", should_run=null
Action:  trigger_action="none"               →  now "withheld"
         + trigger_evaluation_status="unclassified"
```

`trigger_action` projects the winning verdict now, and `withheld` is a value that cannot be mistaken for a decision. `action.yml` exports `trigger_evaluation_status` so a workflow can tell "run the scan" from "repair the input" — they need different handling and `trigger_action` alone could not say which. The legacy shape still works: `stop_conditions_terminal` is absent from older payloads, and there a fired stop *was* terminal, which is the fallback. The CLI lines stop asserting an override that did not happen. Consumer tests for stop+run, both withheld states, and the legacy payload.

### [P1] A repair command that removed the comparison

This one is the worst of the ten, and you named why: the command reaches `fix_task.allowed_repairs`, so it is a machine-readable instruction to delete the evidence. Confirmed on the same clean head — following it took `binding gaps: 1` → `0` and `unverified` → `not_claimed`.

Both the new gap and the pre-existing base-regeneration one now publish **no** command. The distinction that decided it: the *warning prose* scopes the one-liner with "from its source workspace", but `next_action.command` carries no such qualification and is executed as written. The two steps live in `expects`, ending with "Rerunning without `--diff-from` clears this row without answering it", and `path: --diff-from` keeps the rows addressable and rankable.

### [P1] First adoption is not a failed comparison

You are right, and the distinction already existed one function over — `safe_recovery` excludes `missing_manifest` with the comment *"the base was read successfully and simply has no gate yet"*. My `bool(base) and base_report is None` swept it back in. Now driven by an explicit `_BASE_COMPARISON_FAILURES = {ref_missing, archive_failed, scan_failed}`, with a test pinning that `missing_manifest` and the never-asked states stay out.

### [P1] The converse invariant

Added: no binding row may be `not_claimed` while a requested comparison stands unperformed. Regression starts from the real generated report and rewrites only the accounting, so it isolates that claim instead of tripping an earlier one.

### [P2] The join key

Fixed at the root rather than by special-casing. `EvidenceGap.subject_id` carries the canonical tool id, and every ledger entry names the gap accounting for it through an explicit `accounted_by` pointer. That turned out to matter beyond the collision: the three accountings key off three different gap shapes — a tool's display label, a warning's text, the base comparison — and one join-by-subject could only have got all three right by luck. Collision test included (`tool_v2:old` / `tool_v2:new`, both `dup [api]`, one gapped).

### [P2] Forged counts

`0 <= gap_backed <= gated <= total`, exact when untruncated and lower-bounded when truncated, enforced in the Pydantic validator *and* the invariant *and* the schema minimums. `accounted_by` is required for `evidence_gap`/`unverified` and forbidden otherwise, so an accounting cannot claim a join it does not have.

### [P2] The cap guarantee

You were right that the guarantee was still false, and the fix is the one you offered second: `evidence_gap` rows carry per-row proof and are never dropped — that count is now separate, `gap_backed` — while `route_blocked` and `unverified` are counted in `gated` and may be capped, because their accounting is one whole-run fact a single row proves as well as five hundred. Preserving all three would have re-broken the trigger's own bound, where 500 `route_blocked` rows are 500 copies of one sentence. Both boundary cases tested; `docs/report-reading-for-agents.md` says exactly this now.

### [P2] Real adapter omissions

Done as you specified: a typed `SourceSurfaceOmission` on `LoadedToolSource`, recorded at the two MCP skip branches, carrying the warning it raised so the row joins to the `source_warning` gap by pointer. Your `[valid tool, 7]` input now yields:

```
adapter_parse  /tools/1  unreadable_entry  evidence_gap
  accounted_by: "Skipping non-object MCP tool entry"
```

Adapters not yet taught to record omissions contribute nothing, which keeps the blanket prose mapping from creeping back.

### [P2] Malformed `--diff-from`, and the docs

`base_comparison_requested` reads the error path too — whether the bytes parsed is not the same question as whether the caller asked. `base_report_schema_version` joins the required list. `AGENTS.md`, `llms-full.txt`, and `docs/agent-contract-current.md` teach catalog `0.4` and both withheld states, and the accounting enum is documented where agents read it.

On the parity test: it named only the machine payloads, which is exactly why the prose surfaces drifted past it. It now asserts the version *and* both withheld states across `AGENTS.md` and `llms-full.txt`, and a second test derives the accounting enum from `ExclusionAccounting` so a new value cannot ship undocumented.

---

Full suite green (exit 0), `ruff check .` clean. Schemas regenerated with the new required lists, five sample goldens, the packet goldens, and eight codex-boundary goldens refreshed.

PR #408, opened 2026-08-22
fix(evidence): a gap subject is a label, in every kind

pengfei-threemoonslab:
## Summary

- `#403` scoped the raw-id rule in `_validate_exclusion_ledger` to `LEDGER_JOINED_GAP_KINDS`, reasoning that a gap the exclusion ledger never joins may name its subject however it likes. **A guard scoped to a set of kinds passes vacuously for every kind outside it** — and the policy evidence gaps are exactly such a kind.
- So every row of `report.policy_evidence_gaps` kept a raw 64-hex canonical tool id in `subject`, in two shapes. `subject` is a display label: `evidence_gap_headline` prints it verbatim into the CLI's `Improve evidence:` line, `_decision_reason`, and the GitHub step summary. A reader got a digest where a tool name belongs.
- Both emitters now render through the shared `catalog_subject`, **and set `subject_id`** so identity is not dropped in the process. The rule widens to every gap kind.

### Three spellings, not two

The issue described one offending emitter. There were two, and the second used a third spelling:

| Emitter | `subject` before | `subject` after | `subject_id` |
|---|---|---|---|
| `cli/scan/decision.py` (per finding) | `tool_v2_2c9ee6ae…` | `send_email_preview [openai_sdk_static]` | was `null`, now `tool_v2_2c9ee6ae…` |
| `core/lenses/action_surface.py` (per action) | `support.search_kb [tool_v2_445a25…]` | `support.search_kb [support_mcp_tools]` | was `null`, now `tool_v2_445a25…` |

`samples/support_refund_agent` carried `support.search_kb` in **one gap list under both spellings at once** — `[support_mcp_tools]` from the binding stage, `[tool_v2_445a25…]` from the policy stage.

### Identity is kept, not dropped

Review 2 of #404 added `subject_id` as the field that carries identity and that the ledger now joins on, and documented `subject` as a display label. But `policy_evidence_gap()` had no parameter for it and left `subject_id: null` on every row it produced. It takes one now, and both call sites pass the tool id they already hold — so relabelling `subject` does not lose anything, it puts the id in the field built for it.

### Why widen the guard rather than add a third scoped set

Since review 2 moved the join onto `subject_id`, this rule is **no longer load-bearing for joinability at all**. What it does now is keep `subject` a *label*, which is the one thing the field is documented to be — and a kind added later arrives already conforming instead of needing this caught once per kind. `LEDGER_JOINED_GAP_KINDS` existed only to carry the carve-out (its docstring named `inferred_policy_applicability` explicitly as "a separate actionability question") and is removed with it. Nothing else read it.

### Reproduction note

The issue names `samples/conductor_agent` as the reproducer; its gap subjects were already clean. The real reproducers are `samples/support_refund_agent` and `samples/simple_openai_api_agent`.

## Type

## Verification

Additional local checks run:

- Full suite `pytest tests/ -q -n auto` — exit 0. `ruff check .` unpiped — clean.
- **Both halves proven load-bearing by reverting each independently.** Re-scoping the guard to `BINDING_GAP_KINDS` fails the new negative control (`test_a_raw_id_is_refused_for_a_gap_kind_the_ledger_never_joins`); reverting the two emitters fails the new end-to-end test (`test_a_policy_gap_labels_a_tool_the_way_every_other_gap_does`). Neither passes vacuously.
- Live scans before and after, confirming no `evidence_gaps[].subject` equals any `tool_catalog[].tool_id`, that every relabelled row now carries its id in `subject_id`, and that no rendered `evidence_gap_headline` contains a 32+ hex run.
- Goldens regenerated: `simple_openai_api_agent/expected/report.json`, `support_refund_agent/expected/report.json` and `packet.json`. No `report.md` golden moves — the markdown renderer does not carry these subjects.

## Release-readiness notes

- [x] New or changed check IDs are documented in `docs/checks.md` — n/a, no check IDs change

**`report_schema_version` stays `0.35`.** No field is added, removed, or retyped — `subject` is still a required string, and `subject_id` already shipped in `0.35`. Only which value lands in which field changes. Documented in STABILITY.md § *every gap subject is a label, never a raw id*, with a before/after table, following the house pattern for value-only changes. A consumer that joined `evidence_gaps[].subject` against `tool_catalog[].tool_id` should move to `subject_id` — that join was always ambiguous anyway, since two catalog ids can render the same `name [provider]`.

## Known follow-up (not in this PR)

`_semantic_gap` and the `low_confidence_tool` gap in `ci/release_decision.py` build their bracket as `provider or source_id or source_type`, where `catalog_subject` uses `provider` alone — so the two diverge for a provider-less tool. No bundled sample has one (verified: 0 provider-less `tool_catalog` rows across all goldens), and the conservation invariant would fail loudly rather than silently if one appeared.


pengfei-threemoonslab (commented):
Changes requested after reviewing only `82912b6f` against its first parent, as requested.

1. **[P1] The “every gap” invariant still accepts an existing policy-pack hash spelling.** In `src/agents_shipgate/core/semantic_consistency.py:254-256`, the new guard rejects only a subject that is *exactly* a catalog ID. `src/agents_shipgate/inputs/policy_packs.py:124` still emits `f"{subject.tool.name} [{subject.tool.id}]"`, so the wrapper defeats the guard. Using the existing `test_heuristic_policy_match_routes_to_evidence_gap_not_finding` fixture, the real scan produces catalog subject `create_refund [api]` but gap/headline `create_refund [tool_v2_6dcebe…]`; validation succeeds. This directly contradicts the new STABILITY claim and leaves the digest in `Improve evidence:`/GitHub output. Please route this emitter through the catalog subject path and add the policy-pack fixture as a negative control.

2. **[P1] Preserve canonical identity through `subject_id` when restacking onto current `main`.** Current `main` added `EvidenceGap.subject_id` specifically because two catalog IDs may legitimately share the same `name [provider]` label; `source_ref` can be shared too. The new migration note at `STABILITY.md:45-48` instead tells consumers to join on those non-unique display fields. On rebase, extend the policy-gap construction path to carry the canonical ID (`finding.tool_id`, `action.tool_id`, and `subject.tool.id`) and direct identity consumers to `evidence_gaps[].subject_id -> tool_catalog[].tool_id`. Keep `subject` as display only.

3. **[P2] Action-policy gaps can still use a second provider spelling.** `src/agents_shipgate/core/lenses/action_surface.py:294` renders from `ActionFact.provider`, which is whitespace-normalized by `_normalize_token`, while the catalog retains the original provider/source ID. A valid source ID such as `"my api"` yields `create_refund [my_api]` here but `create_refund [my api]` in catalog-backed gaps. The widened invariant accepts both. Resolve the display subject through `action.tool_id` and the original catalog row, and cover a whitespace-bearing provider in the regression test.

Integration note: the PR is currently `CONFLICTING` with `main`; #404 has landed with additional identity/accounting fixes that are absent from this stack, so it needs to be rebased/restacked before the result can be re-reviewed.

Validation performed: focused `test_surface_exclusions.py`, `test_policy_packs.py`, and `test_scan.py` pass; targeted Ruff, `compileall`, and `git diff --check` pass; live CI is green. Agents Shipgate reports `review_publishable` / `review_required` (10 trust-root review items, 0 blockers), which is consistent with requiring a human review of the protected surfaces.

pengfei-threemoonslab (commented):
Superseding update for current head `79f303289`: the PR is now rebased/mergeable, and the prior review's general `subject_id` migration concern is resolved. The current STABILITY guidance is correct. These findings still apply to the new head:

1. **[P1] A third production policy-gap emitter was missed.** `src/agents_shipgate/inputs/policy_packs.py:124` still supplies `subject=f"{subject.tool.name} [{subject.tool.id}]"` and omits the new optional `subject_id`. On this head, the existing heuristic policy-pack fixture emits `subject='create_refund [tool_v2_6dce…]'`, `subject_id=None`, while the catalog label is `create_refund [api]`; `_validate_exclusion_ledger` accepts it because the ID is wrapped. This falsifies the CHANGELOG/STABILITY claim that every `report.policy_evidence_gaps` row was migrated and still leaks the digest into `Improve evidence:`. Canonicalize this caller, pass `subject_id=subject.tool.id`, and assert both fields in `test_heuristic_policy_match_routes_to_evidence_gap_not_finding`. Making `subject_id` explicit/required at tool-scoped call sites would keep future producers from escaping silently.

2. **[P2] Action-policy labels still diverge for whitespace-bearing providers.** `src/agents_shipgate/core/lenses/action_surface.py:295` renders from normalized `ActionFact.provider`, while the catalog-backed path retains the original provider/source ID. A valid source ID `"my api"` produces `mixed_policy_evidence: create_refund [my_api]` here and `inferred_policy_applicability: create_refund [my api]` for the same `subject_id`; validation passes. Resolve the display label by `action.tool_id` against the original Tool/catalog row and add that counterexample.

3. **[P2] Unknown/stale plugin tool IDs can still become raw display labels.** `src/agents_shipgate/cli/scan/decision.py:53-57` deliberately falls back to a raw `finding.tool_id` when the ID has no current catalog row and no name. Plugin runtime validation checks the declared `check_id`, not tool membership, so a plugin finding carrying `tool_v2_<digest>` reaches `subject` verbatim; the invariant only recognizes IDs in the current catalog and accepts it. Validate plugin finding tool IDs against the scan catalog, or avoid the raw-ID display fallback and fail closed on a non-null `subject_id` that cannot resolve.

Validation on `79f303289`: focused surface-exclusion/policy-pack/scan suites pass; Ruff, `compileall`, and diff checks pass; both counterexample scans above reproduce. The deterministic PR verifier itself is `complete / passed / mergeable` with 0 blockers and 0 review items, which demonstrates these are coverage gaps rather than emitted Shipgate findings.

pengfei-threemoonslab:
Addressed all three findings in `7b872220`. Each was reproduced first, and each has a regression test that fails against the unfixed code.

**1. [P1] Third emitter — confirmed and fixed.** `inputs/policy_packs.py:124` was a real miss. Reproduced exactly as reported:

```
CATALOG LABEL : create_refund [api]
GAP subject   : create_refund [tool_v2_6dcebe422940254c0322fa8855af9c11b78edcd834c46d7da0c0aaca0b01287b]
GAP subject_id: None
```

Root cause on my side: the grep I enumerated emitters from was piped through `head -50`, which cut this file off. `test_heuristic_policy_match_routes_to_evidence_gap_not_finding` now asserts `subject == catalog[subject_id]` and that no digest reaches the headline; reverting the emitter makes the scan itself fail closed.

**2. [P2] Whitespace-bearing provider — confirmed and fixed.** Rather than special-case normalization, all three emitters now resolve the label from the tool catalog by `tool_id` through one shared index (`catalog_label_index` / `tool_label`). That also closes the second half of the same divergence I had not flagged: `_provider()` falls back to `source_id`/`source_type` where `catalog_subject` uses `provider` alone. New counterexample asserts `refund_status [my api]`, and fails with `refund_status [my_api]` against the old code.

**3. [P2] Unresolvable plugin ids — confirmed and fixed.** Taken further than `subject_id`-only validation, because a membership test structurally cannot see either of the two shapes that actually shipped. The invariant now matches a canonical id **by shape, anywhere in the label**, so the wrapped form from finding 1 and an id belonging to no catalog are both refused. `_gap_subject` falls through to the check id rather than printing an unresolvable id raw — identity is not lost, it stays in `subject_id`. Parametrized over bare / wrapped / not-in-catalog.

One thing worth flagging that you did not raise: the shared ledger fixture used `tool_v1:abc`, a synthetic id `tool_v1` appears nowhere in production code, samples, or docs. It was not exercising the real `tool_v2` + sha256 shape, and a shape-based guard would have passed it vacuously. Now a realistic id.

Docs corrected — the "every row" claim was false as written. CHANGELOG and STABILITY now say three emitters and describe the shape-based rule, and the STABILITY table gains the policy-pack row.

Verification: full suite `-n auto` exit 0, `ruff check .` clean unpiped, no golden drift (no bundled sample uses a policy pack or a whitespace-bearing provider).

Issue #410, opened 2026-08-23
RFC: evidence-first declarations — ask only what the scanner cannot prove, pin every answer to its evidence


pengfei-threemoonslab:
## Problem

The declaration workflow is the last wall between a first-time user and a gateable verdict, and it is currently shaped wrong in four measured ways (all from the 2026-08-20/23 adoption walks, CLI `0.16.0b7`, post-#403):

1. **It asks humans to re-state what the scanner already knows.** smart_closer: 12 tools × 4 fields ≈ 48 hand-typed values, while the same report already carries derived `risk_tags` for the interesting tools. The user's question "what should I write?" has no answer in the product; the vocabulary's *semantics* (`ambient` vs `unscoped`) appear in no glossary.
2. **It accepts contradictions silently** — #409. The declarant is the person the gate is blocking; the path of least resistance is `read`/`none`, and shipgate converts it into pass-eligibility with zero findings.
3. **It flattens three epistemically different statements into one form.** `effect` is a fact about code (derivable here). `authority`/`scopes` are facts about deployment (they live in Terraform/OAuth config, often unanswerable from the repo — both walk targets are *samples* whose honest authority is `none`, which trivially passes). `approval` is an organizational rule. Different owners, different sources of truth, different rot rates — one undifferentiated blank list.
4. **Nothing ever re-opens a declaration.** Matching is by name (`tool_id` optional, `schemas/manifest/action_surface.py:150`); a year-old `effect: write` keeps passing after the function's semantics change. Declarations rot, and a green gate at month twelve rests on descriptions of functions that no longer do what they did.

And for the coding-agent workflow specifically, the current model creates a pincer (hit live on github-mcp-server#3076): the agent must not edit `shipgate.yaml`, but its new tool is invisible until someone does.

## Principles

1. **Never ask a human what the scanner can prove.** Strong evidence establishes an effect with no declaration; questions exist only where evidence is ambiguous, absent, or being overridden.
2. **An agent may propose what evidence supports; only a human may assert against it.** A *content* rule, enforced by comparison against derived evidence — no authorship forensics (git can't provide them; agents commit as their user).
3. **Every answer is pinned to the evidence that justified it.** Evidence moves → the question re-opens. One comparator serves contradiction checking (#409) and drift detection.
4. **Attestation is the PR review of a protected file, not a separate ceremony.** The manifest stays the trust root; CODEOWNERS + branch protection supply the human-authority property in practice, and shipgate should say so.
[...]

Issue #357, opened 2026-08-09
ux(reporting): distinguish heuristic action projections from evidence-backed assertions


pengfei-threemoonslab:
## Problem

A tool such as `lookup_account` can carry a conservative action-surface heuristic like `effect: write` / `writes_data` while the semantic evidence section correctly reports its proven effect as `unknown`. Those are different epistemic levels, but current console, JSON, and packet vocabulary can make them look contradictory.

Split from #318 so the narrow framework-routing bug can ship independently.

> Product priority (2026-09-04): **P1 trust and first-value work** under #440. A reviewer cannot act reliably when one result appears to assert both `write` and `unknown`. Reuse the existing subject rollup and provenance attributes; this does not justify another report summary or a new verdict. #521 observes whether a reviewer can distinguish the projection from established evidence without maintainer translation.

## Product rule

Heuristic projections must look provisional without looking ignorable. They are conservative risk signals; they are not reviewed proof and must never satisfy semantic evidence requirements.

## Acceptance criteria

- [ ] Console, JSON, report/packet projections, and reviewer copy use one documented vocabulary for heuristic/provisional versus evidence-backed assertions.
- [ ] A reader can tell why an action is structurally reachable and conservatively risky yet still not pass-eligible.
- [ ] Policy severity cannot upgrade a heuristic predicate into typed evidence.
- [ ] Provisional wording does not imply the signal may be ignored.
- [ ] Applying a patch or inventory suggestion is never described as proving runtime behavior or clearing human-owned evidence by itself.
- [ ] Contract/schema changes, if any, pass the repository’s surface-discipline gate with a named metric.

## Related

- #318 owns framework-aware remediation routing.
- #329 audits adopter-facing internal vocabulary.

## Bounded implementation starting point — 2026-09-06

Follow the current projection from `semantic_assessment.conservative_effect` into `ActionFact.effect`, alongside the evidence-backed `semantic_assessment.effect`. First add one real scan fixture demonstrating the apparent write/unknown contradiction, then project the existing provenance consistently through console, report, PR and packet output. Assert that provisional wording remains visible and cannot satisfy effect evidence or make the action pass-eligible.

[...]

PR #411, opened 2026-08-24
fix(evidence): a declaration weaker than its evidence is never silent (#409)

### LINKED ISSUE #409: A declaration weaker than inferred evidence is accepted silently: `effect: read` on a tool tagged `external_write` becomes pass-eligible with zero findings
## Summary

A human declaration that **contradicts** the scanner's own derived evidence is accepted silently, closes the very gap that evidence raised, and makes the action pass-eligible. Declaring `effect: read` on a tool shipgate itself tagged `external_write` produces zero findings — the declaration wins, the evidence vanishes from the decision, and the report still carries the contradicting `risk_tags` on the same catalog row.

This is the trust anchor failing open. The whole declaration model rests on "a human said so" outranking "the scanner guessed" — which is right — but nothing tells the human, or their reviewer, that what they said is contradicted by what the scanner observed. The path of least resistance for anyone the gate is blocking (including an agent-assisted human rubber-stamping a drafted manifest) is to declare everything `read`/`none`, and shipgate will convert that into pass-eligibility without objection.

## Reproduction

CLI `0.16.0b7` at `4e712f54` (post-#403/#404), contract 24, report schema 0.35. Workspace: adk-samples#1745 `python/agents/smart_closer` with the standard scoped `init` manifest (12 tools, all reachable, `declared_purpose` filled).

Baseline scan: 24 semantic gaps, `0/12 actions pass-eligible`. `send_email`'s gaps include `inferred_effect_only` ("tool effect is supported only by heuristic evidence", `source_ref: risk_hint`) and `missing_authority_evidence`. Its catalog row carries `risk_tags: ["customer_communication", "external_write"]`.

Add a deliberately false declaration:

```yaml
action_surface:
  actions:
    - tool: send_email
      effect: read
      authority:
        mode: none
```

Re-run `scan`. Actual result:

- Decision text moves on to the *next* tool (`create_salesforce_quote`); `send_email` is done.
- `1/12 actions pass-eligible` (was 0/12); gaps 24 → 22.
- The pre-existing `inferred_effect_only` gap on `send_email` is **closed by a declaration that contradicts the heuristic that raised it**.
- Grep of the full `report.json` for any conflict/contradiction/mismatch mention involving `send_email`: **0 hits**.
- The same report still contains `risk_tags: ["customer_communication", "external_write"]` on the `send_email` catalog row. The contradiction is present in the artifact; nothing joins it.

## Mechanism

The contradiction check **exists** and is correct for the claims it can see:

- `core/semantic_assessment.py:462-495` — when a declaration is present, `contradictory = [claim for claim in [*structural, *inferred] if claim.policy_eligible and _EFFECT_RANK[claim.value] > _EFFECT_RANK[declared_effect]]`; a hit emits `conflicting_effect_evidence` ("declared effect is weaker than high-confidence source evidence").
- `core/domain.py:126` — `policy_eligible = basis in POLICY_ELIGIBLE_EVIDENCE_BASES and confidence == "high"`.

`send_email`'s `external_communication` claim comes from a name-heuristic risk hint (`semantic_assessment.py:386-412`), so it is not `policy_eligible` — and is therefore invisible to the `contradictory` filter. `_EFFECT_RANK` (`:25-35`) ranks `external_communication` at 6 vs `read` at 0; the rank comparison would have caught it if the claim had been admitted.

The exclusion is a deliberate #357 boundary: heuristic evidence must not *drive* policy (must never prove a read-only action or independently block). That is right. But the flag conflates two different powers:

1. **Driving a verdict** — heuristics rightly cannot.
2. **Challenging a human assertion** — heuristics should. A declaration sitting *below* an inferred claim is not the heuristic gating anything; it is a human statement contradicting an observation, which is precisely the situation a reviewer needs surfaced.

In #403's language: closing `inferred_effect_only` via a contradicting declaration is a narrowing decision with no record.

## Expected behaviour

A declaration ranked below an **inferred** (non-policy-eligible) claim should:

- be accepted as the operative effect (human authority stands — heuristics still don't drive policy), **but**
- emit a distinct review-level finding (e.g. `declaration_below_inferred_evidence`, severity review_required, never blocking) naming the declared value, the inferred value, and the hint source, and
- appear in the PR-comment / reviewer projection as an explicit override row (⚠ "declares `read`; heuristic evidence says `external_communication` (risk_hint: name)"),
- with an explicit `override: {evidence, reason}` in the declaration silencing nothing but downgrading the finding to an acknowledged override (still visible, never silent).

Acceptance criteria:

- [ ] The reproduction above yields a finding naming `send_email`, `read`, and `external_communication`; the action may remain pass-eligible only if declarations-with-acknowledged-override are pass-eligible by policy.
- [ ] `conflicting_effect_evidence` behaviour for policy-eligible claims is unchanged (regression: existing tests).
- [ ] A declaration ranked *above* all evidence (conservative escalation) stays silent — the rule is monotone, only de-escalation is flagged.
- [ ] Negative control: the finding does not fire when declaration and heuristic agree.

## Relationship to the RFC

This is Increment 1 of the evidence-first declaration workflow RFC (filed separately): the monotone declaration rule. It is independently shippable and closes the live hole; the RFC generalises the same comparator into drift detection and the proposal/confirmation flow.

---

Found 2026-08-23 while testing the #403 fixes, via a deliberate adversarial declaration on the smart_closer walk.

pengfei-threemoonslab:
## What

**Increment 1 of [#410](https://github.com/ThreeMoonsLab/agents-shipgate/issues/410) — the monotone declaration rule. Closes [#409](https://github.com/ThreeMoonsLab/agents-shipgate/issues/409).**

A human declaration that contradicts the scanner's own derived evidence was accepted silently, closed the very gap that evidence raised, and made the action pass-eligible. Declaring `effect: read` on a tool this scanner itself tagged `external_write` produced zero findings, while the same report kept the contradicting `risk_tags` on the same catalog row with nothing joining them.

Reproduced at unit level and end to end before the fix:

```
=== baseline (no declaration) ===
status inferred   conservative external_communication  pass False  issues ['inferred_effect_only']
=== adversarial declaration effect: read ===
status declared   conservative external_communication  pass True   issues []
```

## Why it happened

The contradiction check exists and is correct for the claims it can see. `_assess_effect` admits a claim into `contradictory` only when `claim.policy_eligible`, and `domain.py` grants that only to typed, high-confidence bases. Heuristic risk hints are deliberately excluded — a heuristic must never **drive** policy ([#357](https://github.com/ThreeMoonsLab/agents-shipgate/issues/357)).

That exclusion is right, but one flag governed two different powers:

1. **Driving a verdict** — heuristics rightly cannot. Unchanged here.
2. **Challenging a human assertion** — heuristics should. A declaration sitting *below* an inferred claim is not the heuristic gating anything; it is a human statement contradicting an observation, which is exactly what a reviewer needs surfaced.

## The rule

Effect declarations are now **monotone**.

| situation | before | after |
|---|---|---|
| declared **above** the evidence (conservative escalation) | silent | silent |
| declared **equal** to the evidence | silent | silent |
| declared **below policy-eligible** evidence | `conflicting_effect_evidence`, blocking | unchanged |
| declared **below inferred** evidence | **silent, pass-eligible** | `declaration_below_inferred_evidence` — review-level gap, not pass-eligible |
| …with `override: {evidence, reason}` | n/a | accepted, pass-eligible, reported as one semantic review concern |

The declaration remains the operative effect in every row — heuristics still do not drive the verdict, and the new row never blocks. What changes is that the de-escalation is on the record and owed an answer.

Two answers close it, and the reviewer owns the choice:

```yaml
action_surface:
  actions:
    - tool: send_email
      effect: read
      override:                                  # ← new, additive
        evidence: agent.py returns a canned status dict; no mail client is built
        reason: the name matches the comms heuristic but the body sends nothing
```

…or raise `effect` to the inferred value. Both `evidence` and `reason` are required and non-blank, and `override` requires a declared `effect` to acknowledge.

An acknowledged override is **accepted but never silent**: the action is pass-eligible again, and the run carries one `acknowledged_effect_override` semantic review concern, so it can never read `passed` — the same tier `unscoped`/`ambient` authority already occupies.

## What corroborating source evidence does — and does not — buy

The first draft **exempted** a declaration the source itself corroborates: `support.search_kb` declares `read` and carries `readOnlyHint: true`, so why make a reviewer defend a protocol annotation against a keyword? Self-review killed that, for two reasons:

- **The resolver already refuses to pass on that annotation alone.** With no declaration the same tool is `inferred_effect_only` and not pass-eligible, *precisely because* a hint outranks it. Letting a declaration that merely restates the annotation close that gap would buy what the annotation could not — #409's hole would move rather than close.
- **The corroboration came from the tool source's own content**, which is not conditioned on `tool_sources[].trust`. An MCP server can assert `readOnlyHint: true` about a destructive tool; under the exemption a `read` declaration then sailed through with zero rows. #410's non-goals rule that out explicitly.

So corroboration is **named in the row** instead — `declared effect 'read' is weaker than inferred 'financial_write' evidence (risk_hint:keyword); source evidence agrees with the declaration (mcp_annotation)` — which is what makes the override one line to write rather than an investigation to open.

Measured cost of dropping the exemption across the entire sample and fixture corpus: **two rows**, both true. Both now carry overrides in `samples/support_refund_agent` and `samples/ai_generated_refund_pr`, so the shipped samples are the worked example (`Reasons: acknowledged_effect_override=2, …` in the packet golden).

The manifest row's own `effect`, `risk_tags`, `scopes`, and `override` never count as agreeing evidence for itself — `DECLARATION_CLAIM_SOURCES`, now one spelling shared by the resolver and the action-surface lens, which previously kept its own copy of three of the four.

**An override never silences `conflicting_effect_evidence`.** Where policy-eligible evidence outranks the declaration the blocking conflict is unchanged, no acknowledgement attaches, and the row now *says* the override does not reach it — that field is the first thing a blocked reviewer reaches for, and silently discarding it left them re-running against a byte-identical message.

## Acceptance criteria (#409)

- [x] The reproduction yields a row naming `send_email`, `read`, and `external_communication` — `tests/test_declaration_monotonicity.py::test_declaration_below_inferred_evidence_is_reported`.
- [x] `conflicting_effect_evidence` for policy-eligible claims is unchanged — existing tests plus `test_override_cannot_silence_policy_eligible_contradiction`.
- [x] Conservative escalation stays silent — `test_conservative_escalation_is_silent`.
- [x] Negative control: nothing fires when declaration and heuristic agree — `test_declaration_matching_the_evidence_is_silent`.
- [x] The action is not pass-eligible without an acknowledgement; with one it is.

## Two deliberate deviations from the RFC text, flagged for review

1. **#410 §A says an unacknowledged de-escalation is a `config_error`.** It is a review-level evidence gap here instead. A hard error would let a name heuristic block a config outright, which #357 forbids and #409 explicitly rules out ("severity review_required, never blocking"). The gap yields `insufficient_evidence`, never `blocked`.
2. **The unacknowledged row is a gap (→ `insufficient_evidence`), not a review concern (→ `review_required`).** This matches its sibling `conflicting_effect_evidence` and every other semantic gap, and it is what makes the action non-pass-eligible, which #409's acceptance criteria require. The row is fully actionable — exact path, template, and both routes — so it is not a bare dead-end. Happy to move it to the review tier if you read #409 the other way.

### Known limitation

An `override` whose inferred evidence later stops firing stays accepted and unreported. Distinguishing a stale exception from one that never applied needs the `basis: confirmed:<derivation_id>` pin from increment 4 of #410, which is exactly what that increment is for. Raising a review concern here instead would strand a repo at `review_required` over a leftover YAML block precisely when its evidence improved. The behaviour is pinned by a test so it stays a decision rather than an accident.

## Human-facing result

```
Summary: Insufficient evidence: a declared effect is weaker than the evidence inferred for it
         (send_email [adk_agent]). Fix at shipgate.yaml#action_surface.actions[tool='send_email'].
Evidence coverage: static (1/1 catalog tools reachable; 1 semantic evidence gap(s); 0/1 actions pass-eligible)
Improve evidence: Raise action_surface.actions[].effect to 'external_communication', or acknowledge the
         difference with an override naming the evidence you checked and why it does not apply, then
         rerun verification. A ready-to-review block is written to suggested-declarations.yaml next to
         report.json.
```

The instruction names the value because every human surface renders the short per-kind headline (`evidence_gap_headline`), not the row's `why` — a first draft said "raise it to the inferred effect this row names" and named it nowhere the user could see. The value is derived through `strongest_effect_above_declaration`, which shares `claims_above_declared_effect` with the resolver, so the instruction and the reason cannot drift onto two derivations of the same comparison.

and with the override applied: `1 semantic review concern(s); 1/1 actions pass-eligible`.

The scaffold writes the block with its vocabulary attached, the same way #388 taught the other blanks to carry theirs:

```yaml
---
# closes: declaration_below_inferred_evidence
# merge into: shipgate.yaml#action_surface.actions[tool='send_email']
tool: send_email
tool_id: tool_v2_702b224a…
source_id: adk_agent
override:
  # what you checked to conclude the inferred effect does not apply — the
  # function body, the deployment, the upstream contract. Named so the next
  # reviewer can re-check it, not so this row goes quiet.
  evidence: <REVIEW_REQUIRED>
  # why that evidence does not establish the stronger effect for this action.
  # Raising `effect:` to the inferred value instead needs no override at all.
  reason: <REVIEW_REQUIRED>
```

`verify` routes it to a human: `fix_task.actor: human`, `next_action.actor: human`, `state: review_publishable`. The gap's template carries `suggested_patch_kind: manual`, `auto_apply: false`, `requires_human_review: true`, so `apply-patches` never writes an override — an agent may propose what evidence supports, only a human may assert against it (#410 principle 2).

## Compatibility

- **Additive manifest field** (`action_surface.actions[].override`); every existing manifest is unchanged and keeps its verdict.
- **Report schema `0.35` → `0.36`** for the new `evidence_gaps[].kind`. All bump touchpoints done: schema regen, `v0.35` frozen, sample goldens, doc stamps, `.well-known`, adoption-kit renderer source + outgoing hash recorded in `prior_render_sha256`, plugin skill copy, `llms-full.txt`, STABILITY.md, agent contract.
- **No new CLI surface**, no new next-action kind (reuses `resolve_semantic_conflict`) — per the surface-discipline gate.
- The `Action.effect` fact is unchanged in every case: it was already `conservative_effect`, the strongest claim across all evidence, so policy checks never saw the weakened value.

## Sample and fixture changes

- `samples/support_refund_agent` and the other four goldens: schema version only. The two rows the first draft raised there were both over-fires the corroboration guard now suppresses.
- `tests/test_google_adk.py`: the `_ADK_MANIFEST` fixture declared `read`/`write` against `external_communication` hints. `send_email` is now declared honestly and `get_manager_email` — whose name trips the comms heuristic although it only reads an address — carries an override, which also gives the feature an end-to-end exercise through a real scan.
- `tests/test_p0_policy_evidence_canaries.py::test_weaker_write_declaration_with_inferred_financial_effect_is_ie` now asserts the *stronger* property. `create_refund` declares `write` against a `financial_write` heuristic, so the discarded escalation is on the record rather than being carried only by the policy-evidence gap. Every headline assertion in that canary (exit 20, `insufficient_evidence`, no findings, the `mixed_policy_evidence` row) is unchanged.
- Packet label: "Known **authority** review concerns" → "Known review concerns", since the tier now holds two different kinds of concern. `_decision_reason` composes the sentence from `reason_counts` for the same reason (and its pre-existing "1 action use" grammar slip is fixed).

## Not in this PR

Increments 2–4 of #410 — evidence-first effects and the questionnaire scaffold, per-source authority and policy packs, `basis: confirmed:<derivation_id>` drift pinning, adoption rungs, `environment.target: template`, and the CODEOWNERS finding. This increment is the comparator that E (drift detection) reuses.

## Self-review rounds

Three rounds ran against this branch; each is its own commit with the reasoning, and every finding is either fixed or has its decision pinned by a test.

1. `7c5bf25d` — the corroboration exemption above, the silently-discarded override, the `review_concern_count` contract drift, an unnamed review-concern reason vanishing from `_decision_reason`, and the comparator's locals escaping the guard that gives them meaning.
2. `cba9b70f` — the read-versus-side-effect conflict branch owed the same ignored-override note as the ranked one; `docs/passed-verdict-contract.md` did not list an acknowledged override among the things that keep human review mandatory. Swept every sample and fixture manifest for rows the stricter rule newly raises: zero.
3. `93999566` — `expects` naming the effect it asks for, and one comparator behind both the instruction and the reason.
4. `17bf51d7` — the four findings from your blocking review, below.

## Blocking review — what changed

**[P1] Consume the acknowledgement in policy applicability.** You were right that the route did not deliver: policy applicability asks exactly what the override answers, so the reviewer followed the row's own instruction and traded `declaration_below_inferred_evidence` for `mixed_policy_evidence` — the same verdict under a different name. The override claim now carries `overridden_claim_ids`, and `_non_authoritative_effect_escalation_support`, the action-policy predicates (`_semantic_action_predicate`'s inputs), and capability-policy matching all read that one authored list rather than re-deriving the comparison at three sites. The `_ACKNOWLEDGED` fixture now reaches exactly `review_required` with `policy_gap_count == 0`, `policy_evidence_gaps == []`, and no semantic gaps — asserted in `test_acknowledged_override_is_accepted_and_reported`.

**[P1] Preserve each override in the reviewer projection.** `semantic_coverage.acknowledged_overrides[]` carries one row per exception: subject, `subject_id`, declared effect, inferred effect, the hint sources, any source evidence that agrees, and the human's evidence and reason plus the manifest path. The packet's §1 renders each row (visible in the refund golden), and `select_pr_items()` emits `SHIP-ACTION-EFFECT-OVERRIDE-ACKNOWLEDGED` per override. Packet schema `0.12` → `0.13` and verifier `0.9` → `0.10`, both with forward-reading paths for the frozen shapes — the field is absent there, and an empty list is the honest reading of a build that could not have recorded an exception.

**[P2] Reject visually blank override evidence.** Reproduced. `has_visible_content` moved to `schemas/text.py` so the schema layer can use it without importing `core` (`core.evidence_actions` re-exports it, so its callers are unchanged), and both fields now require visible content — whitespace, C0/C1 controls, bidi marks, and every Default_Ignorable code point, not only ASCII spaces.

**[P2] Keep the public manifest schema aligned with runtime validation.** The dependency is published as an `if`/`then` (`override` present and non-null ⇒ `effect` required and non-null), and the visible-content rule as a `pattern` generated from the same code-point table `has_visible_content` reads. BMP points are emitted as `\uXXXX` and astral points as literal characters — `\u{...}` is ECMA-262-with-`u`-flag only and Python's `re` rejects it outright, which would have made the published pattern unusable by the validators that consume this file; a negative control pins that. `tests/test_manifest_schema_parity.py` runs twelve payloads (valid override, missing effect, explicit-null effect, empty/whitespace/zero-width/bidi/control/variation-selector strings, padded-but-visible, no override) through both validators and requires them to agree.

Residual, stated rather than hidden: the runtime remains strictly stricter than the published pattern for surrogate, private-use, and unassigned code points, which no portable ECMA-262 character class can enumerate.

## Verification

`ruff check .` clean, `scripts/generate_schemas.py --check` clean, full suite green.


src/agents_shipgate/core/semantic_assessment.py
```diff
@@ -455,6 +465,93 @@ def _assess_effect(
...
+            acknowledged_override = declaration.override
+            if acknowledged_override is not None:
+                claims.append(
+                    _claim(
+                        "effect",
+                        declaration.effect,
+                        "high",
+                        "static_declaration",
+                        "reviewed_declaration",
+                        DECLARATION_OVERRIDE_SOURCE,
```
pengfei-threemoonslab, line 541: [P1] Consume the acknowledgement in policy applicability

This claim clears the new semantic issue, but downstream policy matching still sees the original higher heuristic as unresolved. Using this PR’s `_ACKNOWLEDGED` fixture, semantic coverage is 1/1 pass-eligible with one review concern, yet `policy_gap_count=1` (`mixed_policy_evidence`) and the final decision is `insufficient_evidence`. Applying the generated override therefore trades `declaration_below_inferred_evidence` for another gap instead of reaching the intended `review_required` route. `_non_authoritative_effect_escalation_support`, `_semantic_action_predicate`, and capability-policy matching need to recognize the acknowledged source/effect, and the E2E test should assert exact `review_required` with zero policy gaps.

src/agents_shipgate/ci/release_decision.py
```diff
@@ -1056,6 +1063,17 @@ def _semantic_coverage(
...
 
+        # An acknowledged effect override is accepted — the declaration stands
+        # and the action stays pass-eligible — but it is never silent (#409).
+        # A review concern is exactly that tier: it cannot reach ``passed`` and
+        # it cannot block, so the reviewer sees the exception and decides.
+        if any(
+            claim.source == DECLARATION_OVERRIDE_SOURCE
+            for claim in assessment.effect.claims
+        ):
+            review_concern_count += 1
```
pengfei-threemoonslab, line 1076: [P1] Preserve each override in the reviewer projection

This reduces acknowledged exceptions to a count and discards the data a reviewer must judge. On the acknowledged fixture, `select_pr_items()` contains no override item, while the packet/PR surfaces render only `acknowledged_effect_override=N`; they do not identify the action, declared and inferred effects, hint source, evidence, or reason. That misses #409’s explicit per-action override-row acceptance criterion. Please emit and project a structured concern for each acknowledged override, with PR-comment and packet coverage.

src/agents_shipgate/schemas/manifest/action_surface.py
```diff
@@ -143,6 +143,41 @@ class ActionAuthorityConfig(BaseModel):
...
+    #: What the reviewer actually checked — the function body, the deployment,
+    #: the upstream contract. Named so the next reviewer can re-check it.
+    evidence: str
+    #: Why the inferred evidence does not establish the stronger effect here.
+    reason: str
+
+    @field_validator("evidence", "reason")
+    @classmethod
+    def require_non_blank(cls, value: str) -> str:
+        text = value.strip()
```
pengfei-threemoonslab: [P2] Reject visually blank override evidence

`strip()` leaves default-ignorable characters such as U+200B and U+2060 intact. I reproduced an override using those two values: it validates, suppresses the mismatch, makes the action pass-eligible, and emits an override claim even though both fields render blank in review. Please use the repository’s visible-content semantics (or an equivalent schema-layer helper) and cover default-ignorables, bidi controls, and control-only strings—not only ASCII spaces.

docs/manifest-v0.1.json
```diff
@@ -171,6 +171,17 @@
           "default": null,
           "title": "Operation"
         },
+        "override": {
```
pengfei-threemoonslab, line 202: [P2] Keep the public manifest schema aligned with runtime validation

The published Draft 2020-12 schema accepts both an `override` with no `effect` and blank `evidence`/`reason` strings with zero validation errors, while `AgentsShipgateManifest.model_validate` rejects both. Because this schema is advertised for live editor validation, users can be told a manifest is valid and then have the CLI refuse it. Encode the dependency and string invariants in generated JSON Schema and add schema-level parity tests.

pengfei-threemoonslab (commented):
Blocking review: four findings (two P1, two P2) should be resolved before merge. GitHub does not allow this account to request changes on a PR it authored, so this is submitted as COMMENT; please treat the inline findings as change requests.

The monotone mismatch detection itself is directionally sound, but the acknowledged-exception route does not currently deliver its advertised decision semantics or reviewer evidence.

Validation performed:
- `tests/test_declaration_monotonicity.py` + `tests/test_semantic_assessment.py`: 59 passed
- Full `pytest -q` reached 100% with no test failures; its eight packaging setup errors were caused by the sandboxed isolated build being unable to fetch Hatchling, and `pytest tests/test_packaging.py -q` passed 8/8 when rerun with network access
- `ruff check .`, schema generation check, and `git diff --check` passed
- Agents Shipgate over `origin/main...HEAD`: `review_publishable` / `review_required` because protected trust roots changed; PR updates are permitted, merge and completion remain human-owned

PR #412, opened 2026-08-24
fix(evidence): a declaration below the evidence is recorded, not assumed (#409)

### LINKED ISSUE #409: A declaration weaker than inferred evidence is accepted silently: `effect: read` on a tool tagged `external_write` becomes pass-eligible with zero findings

pengfei-threemoonslab:
## Summary

- **A declaration weaker than the evidence Shipgate observed is no longer accepted in silence** (#409). Declaring `effect: read` on a tool Shipgate itself tagged `external_write` closed the `inferred_effect_only` gap the same heuristic had raised, turned the action pass-eligible, and named the disagreement nowhere. The path of least resistance for anyone the gate is blocking was to declare everything `read`/`none`.
- The contradiction check existed and was correct for the claims it could see: its filter admitted only `policy_eligible` claims, and #357 deliberately withholds that flag from heuristics so they cannot *drive* policy. One flag was carrying two different powers. **Driving a verdict and challenging a human assertion are not the same thing** — a declaration sitting below an inferred claim is not a heuristic gating anything, it is a statement contradicting an observation the report already carries.
- Heuristics still never drive policy. The declaration remains the operative effect and the new signal never blocks. New `action_surface.actions[].override` (`evidence` + `reason`) records the reviewed exception.

### One correction to the issue's framing

Reproducing against a pristine `HEAD` checkout showed the case was **not fully silent**: `mixed_policy_evidence` already fired and already produced `insufficient_evidence`. What was actually broken is that the action reported `pass_eligible: true`, the semantic dimension recorded nothing, and the generic policy row never named the declaration, the declared value, or the inferred one. That discovery is load-bearing — see "the remedy has to work" below.

### Behaviour

| | before | unanswered | reviewed override |
|---|---|---|---|
| `release_decision.decision` | `insufficient_evidence` | `insufficient_evidence` | `review_required` |
| `pass_eligible_actions` | **1/1** | 0/1 | 1/1 |
| names declared vs inferred | no | gap + finding | finding |

The rule is **monotone**: a declaration that agrees with or escalates above the evidence stays silent, so a conservative answer still costs nothing.

### The remedy has to work

The override also clears the matching `builtin-effect-control-applicability` policy gap. Without that, writing the override would restore pass eligibility while the verdict sat at `insufficient_evidence` anyway — a published next step that cannot change the answer. Unacknowledged escalations are untouched, so the P0 canary keeps its row.

### Override drift

`override.evidence` must name **exactly** the inferred effects observed above the declared effect — exact set match, not containment, in both directions. Containment would let one edit pre-acknowledge every effect in the vocabulary and go permanently silent. Exact match means new evidence re-opens the question *and* evidence that has since disappeared re-opens it too.

### One comparator

`core/effect_override.py` serves the resolver, the gap scaffold, the finding, and the PR-comment rows. There are two effect-rank tables in this repo and they disagree about `write` vs `privileged_data_access`; every surface now reads this comparison's answer instead of re-deriving it.

### Acceptance criteria (from the issue)

- [x] The reproduction yields a finding naming `send_email`, `read`, and `external_communication`; the action is pass-eligible only with an acknowledged override.
- [x] `conflicting_effect_evidence` behaviour for policy-eligible claims is unchanged (explicit regression test added).
- [x] A declaration ranked above all evidence stays silent.
- [x] Negative control: nothing fires when declaration and heuristic agree.

### Review loop — 3 passes, 7 findings, all fixed

Every finding was in the reviewer projection, and four lived in filters between the emitter and the reviewer's eyes:

- Stale overrides rendered ``evidence says `?` `` — three states reaching one renderer needed three sentences.
- The 3-row cap sorted by tool name (all rows share a check id and severity), so the one unanswered override on a late-alphabet tool was the row hidden behind `(+1 more)`.
- Unbounded `override_reason` on a surface whose budget truncates prose from the *end* evicted the trigger, base-diff, and artifact lines below it.
- An override on a `conflicting` row is correctly inert, but nothing said so — a consumed input was never accounted for.
- **`checks.ignore` took an acknowledged override off the PR comment entirely** (`decision: passed`, `gap_count: 0`, no block rendered). The override trades pass eligibility *for* visibility, so dropping suppressed rows removed the half of the bargain justifying the other half. Suppressed rows now render marked `(suppressed)`; the suppression still decides the verdict, not what is shown. The unacknowledged path was never exposed — its evidence gap is immune to suppression.
- The check catalog declared 5 of the 8 evidence keys the finding emits.
- `docs/checks.md` claimed the check "never blocks a release" without qualifying a user's own `fail_on: [medium]`.

Pass 3 found nothing.

### Surface discipline

Per [`CONTRIBUTING.md` § Surface discipline](../blob/main/CONTRIBUTING.md#surface-discipline): this moves **blocked-recall** — it closes a fail-open in the trust anchor, in exactly the reward-hacking shape the product exists to catch. It extends the one decision engine (`release_decision.decision`) and its existing projections; no second verdict, no new CLI command, no new adapter. The new surface is one gap kind, one check id, and one additive manifest field, each documented in `STABILITY.md`.

## Type

- [ ] Input adapter change
- [ ] Documentation only

## Verification

Additional local checks run:

- Full suite green (`pytest tests -n auto`, exit 0) after each of the three review passes.
- `ruff check .` clean; `scripts/generate_schemas.py --check` clean.
- End-to-end reproduction of the issue against a pristine `git archive HEAD src` checkout, before and after, plus the acknowledged-override and `checks.ignore` variants.
- Sample goldens regenerated (`support_refund_agent` report.md/report.json/packet.*); every sample swept for the new signal.

## Release-readiness notes

- [x] New or changed check IDs are documented in `docs/checks.md`

### Notes for the reviewer

- `report_schema_version` 0.35 → 0.36; `docs/report-schema.v0.35.json` is frozen and unchanged. Additive: one value in a frozen union.
- `samples/support_refund_agent` and `samples/ai_generated_refund_pr` were both already declaring below their own heuristics (`support.search_kb`, `send_email_preview`) and now carry reviewed overrides. Any repository in that state gains one gap, one finding, and one non-pass-eligible action.
- `tests/test_p0_policy_evidence_canaries.py` asserted `pass_eligible is True` on a discarded heuristic escalation — that assertion inverts here by design, and the canary's own P0 property (never a clean pass, exit 20) is unchanged.
- `tests/test_surface_exclusions.py` asserted its cross-stage label claim on `support.search_kb`, which no longer raises a policy gap; it is re-pointed at a sibling from the same source.

Closes #409. Increment 1 of #410.


docs/packet-schema.v0.12.json
```diff
@@ -557,6 +557,7 @@
             "missing_effect_evidence",
             "inferred_effect_only",
             "conflicting_effect_evidence",
+            "declaration_below_inferred_evidence",
```
pengfei-threemoonslab: [P1] Preserve published schema identifiers. This new emitted enum value also propagates into verifier v0.9, capability-lock v0.6, and capability-lock-diff v0.7, but all four artifacts keep their old version fields and overwrite the existing `docs/*-schema` files. I validated newly emitted artifacts against the corresponding `origin/main` schemas: the old readers reject them despite the identical version identifier. `STABILITY.md` requires prior published schemas to remain frozen. Please advance the packet, verifier, lock, and diff versions (as this PR already does for report 0.36), retain the prior files unchanged, and add a compatibility test against the frozen schemas.

src/agents_shipgate/core/effect_override.py
```diff
@@ -0,0 +1,251 @@
...
+            continue
+        # A protocol default is the absence of an observation, not one. An MCP
+        # tool with no annotations says nothing about its own effect, and the
+        # declaration is the answer to that silence, not a contradiction of it.
+        if claim.basis == "protocol_default":
+            continue
+        if claim.value not in EFFECT_EVIDENCE_VALUES:
+            continue
+        effect = cast(ActionEffect, claim.value)
+        if EFFECT_EVIDENCE_RANK[effect] <= declared_rank:
```
pengfei-threemoonslab: [P1] Do not let a total severity rank erase orthogonal obligations. This branch drops every inferred category ranked below the declaration, although the controls are not monotone across categories. In a full-scan repro, an MCP `send_email` action inferred as external communication was declared `financial_write` with the financial controls but no confirmation; Shipgate reported the action pass-eligible with zero semantic gaps/review concerns and emitted no external-confirmation finding, while the external-write risk tags remained present. Financial write does not subsume the confirmation obligation for communicating externally. Please preserve category-specific obligations (or define an explicit subsumption relation) and require an override for unequal orthogonal categories; add an end-to-end control test.

src/agents_shipgate/core/effect_override.py
```diff
@@ -0,0 +1,251 @@
...
+#: this comparator reads only those, so both call sites share one answer.
+EffectClaim = SemanticClaim | SemanticClaimEvidence
+
+#: Evidence strength order for one action's effect. Higher means "a stronger
+#: claim about what this action does in the world", so a declaration ranked
+#: below an observation is a de-escalation.
+EFFECT_EVIDENCE_RANK: dict[ActionEffect, int] = {
+    "read": 0,
+    "write": 1,
+    "privileged_data_access": 2,
```
pengfei-threemoonslab, line 53: [P1] Make policy-gap closure use this same comparison. `EFFECT_EVIDENCE_RANK` orders `write` below `privileged_data_access`, while `_non_authoritative_effect_escalation_support()` still uses `ACTION_EFFECT_RANK`, which orders them the other way. Repro: declared `privileged_data_access` plus inferred `write` is silent/pass-eligible here, but the policy path adds `mixed_policy_evidence` and the release decision becomes `insufficient_evidence`. That contradicts the documented promise that the override/assessment repairs the matching policy gap. Route both consumers through one assessment and cover the pairwise effect matrix.

docs/manifest-v0.1.md
```diff
@@ -705,6 +736,8 @@ Known `policies[].require` paths:
 | `evidence.runbook` | string |
 | `evidence.approval_ticket` | string |
 | `effect` | one of `read`, `write`, `destructive`, `external_communication`, `financial_write`, `production_operation`, `privileged_data_access`, `code_execution`, `identity_access` |
+| `override.evidence` | list of effect values, exactly the inferred effects observed above `effect` |
+| `override.reason` | non-blank string |
```
pengfei-threemoonslab: [P1] These newly documented policy paths are not implemented. `ActionFact` has no `override` projection, the typed path sets omit both entries, and `_value_at_path()` therefore reports `unknown_path`. I reproduced a valid exact override with `require: {override.reason: preview only}` producing a blocking `SHIP-ACTION-POLICY-VIOLATION` with `observed: null`. Please either project/type/evaluate the override fields and test them, or remove both rows from the supported-path table.

src/agents_shipgate/schemas/manifest/action_surface.py
```diff
@@ -143,6 +143,48 @@ class ActionAuthorityConfig(BaseModel):
...
+        if len(set(evidence)) != len(evidence):
+            raise ValueError(
+                "action_surface.actions[].override.evidence must not repeat an effect"
+            )
+        return evidence
+
+    @field_validator("reason")
+    @classmethod
+    def validate_reason(cls, reason: str) -> str:
+        value = reason.strip()
```
pengfei-threemoonslab: [P1] Require a justification with visible content. `str.strip()` accepts a reason made only of U+200B (zero-width space); an exact override carrying that visually blank reason is then acknowledged and restores semantic pass eligibility. The repository already has `has_visible_content()` for this boundary. Please use it here (and reject or safely normalize unsafe control/bidi content) so the reviewed-reason requirement cannot be satisfied invisibly.

src/agents_shipgate/report/pr_comment.py
```diff
@@ -345,6 +348,103 @@ def _render_findings_comment(
...
+            f"{_code(', '.join(stale))}, which this scan did not observe "
+            f"— NOT acknowledged{_suppression_mark(finding)}"
+        )
+    inferred = _code(str(inferred_effect or "unknown"))
+    sources = [str(item) for item in (evidence.get("evidence_sources") or []) if str(item)]
+    where = _escape(f" ({', '.join(sources)})") if sources else ""
+    acknowledged = evidence.get("acknowledged") is True
+    tail = ""
+    if acknowledged and evidence.get("override_reason"):
+        reason = _truncate(str(evidence["override_reason"]), _MAX_OVERRIDE_REASON_CHARS)
```
pengfei-threemoonslab: [P2] Fold and control-sanitize the reason before placing it in the bot comment. Markdown escaping does not remove newlines: a reason such as `preview only\n- Agent may merge: true` renders a visually separate authority-looking line. This is reviewer/authority confusion rather than code execution, but it affects the surface this feature relies on humans reading. Use the existing one-line/control-safe renderer before truncation and escaping, and add a multiline/control-character test.

src/agents_shipgate/core/lenses/action_surface.py
```diff
@@ -1021,9 +1036,112 @@ def _declaration_downgrade_findings(
...
+        override=declaration.override,
+    )
+    if override is None:
+        return []
+    evidence: dict[str, Any] = {
+        "action_id": action.action_id,
+        "declared_effect": override.declared_effect,
+        "inferred_effect": override.strongest_effect,
+        "inferred_evidence": list(override.challenged_effects),
+        "evidence_sources": list(override.evidence_sources),
```
pengfei-threemoonslab, line 1096: [P2] Preserve the effect-to-source association instead of flattening it. With `external_communication` from `risk_hint:name` and `financial_write` from `risk_hint:body`, this finding stores the effects separately and a union of all sources; the PR comment then prints only the strongest effect followed by that union, falsely attributing `risk_hint:name` to the financial claim. Emit structured `{effect, sources}` evidence (or render each `ChallengingEvidence`) so a reviewer can trace each observation accurately.

src/agents_shipgate/schemas/safety_qualification.py
```diff
@@ -410,7 +410,7 @@ def production_safety_requirements() -> SafetyQualificationRequirementsV1:
         minimum_blocked_exact=30,
         minimum_review_exact=19,
         minimum_insufficient_evidence_exact=19,
-        required_report_schema_version="0.35",
+        required_report_schema_version="0.36",
```
pengfei-threemoonslab, line 413: [P2] Update the qualification runbook with this contract bump. `benchmark/safety-qualification/README.md:28` still requires report schema 0.35, while this runner now requires 0.36. Following the checked-in procedure therefore produces receipts the current runner rejects.

pengfei-threemoonslab (commented):
I would not merge this revision yet. I found eight actionable issues: five affect gate correctness or wire compatibility, two make the reviewer-facing evidence unsafe or inaccurate, and one leaves the qualification runbook inconsistent.

The highest-risk defects are the loss of orthogonal effect obligations (a financial declaration can hide external-communication confirmation), mutation of four published schema versions in place, and disagreement between the new override comparator and the existing policy-gap comparator. I also reproduced the documented override policy paths false-blocking valid declarations and a visually blank override reason restoring pass eligibility.

Validation performed: 191 focused schema/override/config/packet/capability-lock/safety tests passed; `scripts/generate_schemas.py --check`, Ruff, and `git diff --check` passed; GitHub CI is green. Those checks do not exercise the adversarial and compatibility cases in the inline comments. The source verifier reports `review_publishable` / `human_review_required` for the touched trust roots.

This review is posted from the PR author's GitHub account, so GitHub will not accept a self-authored `REQUEST_CHANGES` review. Please treat the P1 comments as changes requested before merge.

docs/packet-schema.v0.12.json
pengfei-threemoonslab: Fixed in 94743b56. You were right that `--check` cannot catch this: it proves committed == generated, never that a *content* change moved the version.

All four documents are restored byte-for-byte from `origin/main` and the versions advance — packet `0.12 → 0.13`, verifier `0.9 → 0.10`, capability lock `0.6 → 0.7`, capability-lock diff `0.7 → 0.8`.

Two things the fix needed beyond the bump, both found while making prior artifacts still readable:

- The verifier's legacy-reader set had never gained `0.9`, so a stored v0.9 artifact would have failed `Literal["0.10"]` outright. Added, with the same rationale the `0.8` entry carries.
- A v0.6 `capabilities.lock.json` was rejected outright after the bump — the normalizer only handles `0.1`–`0.4`. That would have broken every adopter with a committed lock, a far larger break than the bump repairs. Added `_STRUCTURALLY_CURRENT_LOCK_SCHEMAS` for prior versions whose payload shape is identical, and mapped `0.6` explicitly in `_CAPABILITY_STANDARD_BY_LOCK_SCHEMA` rather than leaving it to the fallback. (v0.5 was already rejected on `main`; left as-is rather than widening scope.)

Compatibility tests added in `tests/test_declaration_override.py`: one parametrised over all five (frozen, current) pairs asserting the new enum reaches only the current document *and* that each frozen document still matches `origin/main` byte-for-byte, plus one that validates real emitted `report.json` and `packet.json` against their own current schemas.

src/agents_shipgate/core/effect_override.py
pengfei-threemoonslab: Fixed in 94743b56. This was the sharpest one — the monotone rule was built on a total order, and obligations are not totally ordered.

`declaration_covers` now requires both:

- the declaration ranks at or above the observation, and
- it obliges at least that observation's built-in controls.

`financial_write` {approval, audit, idempotency} does not contain
`external_communication` {audit, confirmation}, so your repro is now an evidence gap and the action is not pass-eligible. The finding gets its own title and remedy for that state — "raise the effect" would be wrong advice when it is already higher.

Rather than invent a subsumption lattice, the relation is derived from the obligations themselves: `BUILTIN_EFFECT_OBLIGATIONS` in `core/action_semantics.py`. I did not refactor the four control branches to read it (that is gate-critical code and the risk did not seem worth it in this PR); instead `test_the_builtin_obligation_table_matches_the_controls_that_fire` walks every entry through a real scan and asserts the emitted `missing` set, so the table cannot drift from the branches it mirrors.

Your comment also led to a second defect: coverage was reading `effect` alone, but `risk_tags: [financial_action]` produces a policy-eligible `financial_write` claim and applies the financial-write controls. Coverage now reads every policy-eligible claim — the same set `_control_effects` unions. The shipped `ai_generated_refund_pr` fixture declares exactly that pair and caught it.

src/agents_shipgate/core/effect_override.py
pengfei-threemoonslab: Fixed in 94743b56. `_non_authoritative_effect_escalation_support()` now calls `declaration_covers` over the policy-eligible effect set instead of comparing `ACTION_EFFECT_RANK` itself, so both consumers make one comparison.

On which table wins: neither. `declaration_covers` requires coverage under **both** orders. Picking `EFFECT_EVIDENCE_RANK` would have loosened an existing gate path (`write` inferred under a `privileged_data_access` authoritative claim stops being indeterminate); picking the other would have contradicted the #409 rule. Requiring both makes them agree without weakening either, and the new predicate is a superset of the old one — nothing that raised `mixed_policy_evidence` before stops raising it.

Pairwise matrix covered: `test_both_rank_tables_must_agree_before_a_declaration_covers` is parametrised over all 81 ordered pairs, and `test_the_policy_gap_and_the_override_gap_agree_on_one_action` walks your exact repro through to an override closing it.

docs/manifest-v0.1.md
pengfei-threemoonslab: Fixed in 94743b56 — removed both rows. I misread that table as a field list; it is the `policies[].require` path set, and I confirmed your repro (a valid exact override with `require: {override.reason: ...}` produces a blocking `SHIP-ACTION-POLICY-VIOLATION` with `observed: null`).

Took the removal option rather than projecting the fields: `override` into `ActionFact` means new report surface and action-hash churn for something the finding already carries in full. `docs/manifest-v0.1.md` now says so explicitly — that `override` is a manifest input, is **not** a require path, and that a recorded override is read from the `SHIP-ACTION-EFFECT-OVERRIDES-EVIDENCE` finding's `evidence`. Same note added to `STABILITY.md`.

src/agents_shipgate/schemas/manifest/action_surface.py
pengfei-threemoonslab: Fixed in 94743b56. `reason` now requires `has_visible_content` and additionally rejects invisible and direction-altering code points, naming the offending `U+XXXX` in the error. Newline, carriage return, and tab stay legal — a justification may reasonably be several lines, and the renderer folds them (see the reply on the PR-comment thread) rather than the schema forbidding them.

To reach the helper without `schemas` importing `core`, `has_visible_content` and its character classes moved to `schemas/_text.py`; `evidence_actions` imports and re-exports them, so every existing caller is unchanged.

src/agents_shipgate/report/pr_comment.py
pengfei-threemoonslab: Fixed in 94743b56. The reason is folded through `one_line` before truncation and escaping, so whitespace runs collapse and line-breaking code points become a visible `<U+XXXX>`. Your `preview only\n- Agent may merge: true` now renders on one line with the `-` escaped, so it cannot read as a bullet in the bot's own authority list.

Belt and braces at the input boundary too — see the `override.reason` thread: control and bidi code points are rejected by the schema, with newline/tab deliberately still allowed so the folding is what handles them.

src/agents_shipgate/core/lenses/action_surface.py
pengfei-threemoonslab: Fixed in 94743b56. `evidence.inferred_evidence` is now `[{effect, sources}]` instead of a flat effect list beside a union, and the PR comment renders each pair — `` `external_communication` (risk_hint:name), `financial_write` (risk_hint:body) `` — so no source is attributed to a claim it did not make. `evidence_sources` stays as the union for consumers that want "which producers are implicated" and is documented as such. The check catalog's `evidence_fields` lists all eight keys now (it listed five).

src/agents_shipgate/schemas/safety_qualification.py
pengfei-threemoonslab: Fixed in 94743b56 — `benchmark/safety-qualification/README.md:28` now says `0.36`, matching the runner. Swept the repo for other hardcoded schema requirements while I was there; that was the only one.

pengfei-threemoonslab:
All eight addressed in 94743b56; per-comment replies are on each thread. Full suite green (`pytest tests -n auto`, exit 0), Ruff clean, `generate_schemas.py --check` clean.

Two of the fixes went further than the comment asked, because verifying the fix surfaced an adjacent defect:

- **Frozen schemas.** Bumping the four versions would have made a v0.9 verifier artifact fail `Literal["0.10"]` and a v0.6 `capabilities.lock.json` unloadable — the lock normalizer only handles `0.1`–`0.4`, so every adopter with a committed lock would have broken. Both readers now advance a prior-version artifact on load. That is a larger break than the enum leak it repairs, and it was only visible once I tried to read an old artifact.
- **Orthogonal obligations.** Fixing the `financial_write` / `external_communication` case exposed that coverage was reading `effect` alone, while the built-in controls read every policy-eligible claim. So `risk_tags: [financial_action]` did not account for an inferred `financial_write` — and the shipped `ai_generated_refund_pr` fixture declares exactly that pair. Coverage now reads the same set `_control_effects` unions.

One place I deliberately did not go as far as the comment allowed: on P1-2 you offered "preserve category-specific obligations (or define an explicit subsumption relation)". I derived the relation from `BUILTIN_EFFECT_OBLIGATIONS` rather than refactoring the four control branches to read it — that is gate-critical code, and a behavioural pin (`test_the_builtin_obligation_table_matches_the_controls_that_fire`, which walks each entry through a real scan and asserts the emitted `missing` set) buys the same anti-drift guarantee without the regression risk. Happy to do the refactor as a follow-up if you would rather have the single source of truth in code.

Also worth flagging for your re-review, since it is a scope judgement rather than a fix: on P1-3 I made `declaration_covers` require coverage under **both** rank tables rather than choosing one. Choosing `EFFECT_EVIDENCE_RANK` would have loosened an existing gate path; choosing `ACTION_EFFECT_RANK` would have contradicted the #409 rule. The conjunction agrees with both and is a strict superset of the old policy-path predicate, so nothing that gated before stops gating — but it does mean a declared `privileged_data_access` over an inferred `write` is now a question where it used to be silent on one path.

pengfei-threemoonslab:
Superseded by #411, which landed a parallel implementation of #409 and closed the issue. Closing this rather than resolving the 33-file conflict — the overlap is two independent implementations of the same feature (both add `declaration_below_inferred_evidence`, both add `action_surface.actions[].override`, both bump report to 0.36), and hand-merging those would produce a worse result than either.

Three P1 defects this PR's review found are **live on `main`** after #411, so I am opening a focused follow-up carrying only those, rebuilt on #411's implementation:

1. **Four published schema documents were mutated in place.** #411 added the new enum value to `docs/packet-schema.v0.12.json`, `docs/verifier-schema.v0.9.json`, `docs/capability-lock-schema.v0.6.json`, and `docs/capability-lock-diff-schema.v0.7.json` while keeping their version identifiers. It bumped packet to 0.13 and verifier to 0.10, but the two capability-lock documents have **no successor version at all** — a consumer pinned to any of the four now rejects artifacts that document is supposed to describe. `generate_schemas.py --check` cannot catch this: it proves committed == generated, never that a content change moved the version.

2. **The total effect rank erases orthogonal obligations.** `claims_above_declared_effect()` compares `_EFFECT_RANK` alone, so declaring `financial_write` over an inferred `external_communication` reads as escalation and stays silent. Reproduced against `main`: `pass_eligible_actions: 1 of 1`, `gap_count: 0`, and no `SHIP-ACTION-EXTERNAL-COMMUNICATION-AUDIT-MISSING` — `financial_write` requires approval/audit/idempotency but not confirmation, which is exactly what communicating outward requires.

3. **The two comparators still disagree.** The declaration rule uses `_EFFECT_RANK`; `_non_authoritative_effect_escalation_support()` still uses `ACTION_EFFECT_RANK`, and the two order `write` and `privileged_data_access` oppositely — so an override cannot always close the policy gap raised on the same action.

Plus one P2: `AcknowledgedOverride` carries a single `inferred_effect` beside a flat `inferred_sources` list, so a tool with two inferred effects from two different hints attributes every source to the strongest effect.

The other four review findings are already handled on `main` — the blank-reason check (`schemas/text.py`), the newline-in-reason case (rejected at the schema boundary), the qualification runbook stamp, and the `override.*` require-paths, which #411 never documented.

PR #413, opened 2026-08-24
fix(evidence): a declaration cannot discharge a category it does not cover (#409 follow-ups)

pengfei-threemoonslab:
## Summary

- Three follow-ups to #411, each a defect that shipped with the monotone declaration rule. Two affect gate correctness; one is a live wire-compatibility break on `main`.
- Found while reviewing #412, which implemented #409 in parallel. That PR is closed as superseded; this carries only the residual fixes, rebuilt on #411's implementation.

### 1. A declaration cannot discharge a category it does not cover

`claims_above_declared_effect()` compared `_EFFECT_RANK` alone. Effects are risk-ordered; their **obligations are not**. `financial_write` outranks `external_communication` and obliges approval, audit, and idempotency — but not confirmation, which is exactly what communicating outward requires.

Reproduced against `main`:

```
declared: financial_write     inferred: external_communication (risk_hint:keyword)
pass_eligible_actions: 1 of 1     gap_count: 0
findings: no SHIP-ACTION-EXTERNAL-COMMUNICATION-AUDIT-MISSING
```

…while the tool's `external_write` risk tag sat untouched in the same report.

A declaration now accounts for an observation only when it ranks at or above it **and** obliges at least that observation's built-in controls. The obligations move into `BUILTIN_EFFECT_OBLIGATIONS` (`core/action_semantics.py`).

I did **not** refactor the four inline control branches to read the table — that is gate-critical code and the risk did not seem worth it here. Instead `test_the_builtin_obligation_table_matches_the_controls_that_fire` walks every entry through a real scan and asserts the emitted `missing` set, so the table cannot drift from the branches it mirrors. Say the word if you would rather have the single source of truth in code.

Coverage also reads **every policy-eligible claim** rather than the `effect` field alone. `risk_tags: [financial_action]` produces a policy-eligible `financial_write` claim and applies the financial-write controls, so a heuristic reading the same effect is already accounted for. That is also the clean remedy for the cross-category case, and it is why the two "escalation is silent" fixtures now carry a `risk_tags` entry — declaring `destructive` alone genuinely does not carry the audit obligation an `external_communication` observation would.

### 2. The two comparators disagreed

`_non_authoritative_effect_escalation_support()` still compared `ACTION_EFFECT_RANK`, which orders `write` and `privileged_data_access` opposite to `_EFFECT_RANK`. So a declaration could read as covered by the declaration rule and raise `mixed_policy_evidence` on the same action — a verdict no override could reach, contradicting what the row's own instruction promises.

Both now call `declaration_covers`, which requires **both** orders to agree. Picking one would have loosened an existing gate path (`EFFECT_EVIDENCE_RANK`) or contradicted the #409 rule (`ACTION_EFFECT_RANK`); the conjunction is a strict superset of the predicate it replaces, so nothing that gated before stops gating. The pairwise matrix is covered over all 81 ordered pairs, plus a test asserting the two surfaces reach the same answer on one action.

### 3. Four published schema documents were mutated in place

#411 wrote `declaration_below_inferred_evidence` into `packet-schema.v0.12.json`, `verifier-schema.v0.9.json`, `capability-lock-schema.v0.6.json`, and `capability-lock-diff-schema.v0.7.json` while each kept its version identifier. It bumped packet to 0.13 and verifier to 0.10, but **the two capability-lock documents have no successor version at all** — so a consumer pinned to any of the four currently rejects artifacts that document is supposed to describe.

`generate_schemas.py --check` cannot catch this: it proves committed == generated, never that a *content* change moved the version.

All four are restored byte-for-byte. The capability lock advances `0.6` → `0.7` and its diff `0.7` → `0.8`.

**One thing the bump needed beyond the version edit:** a lock written under `0.6` was rejected outright, because `_normalize_capability_lock_payload` handles only `0.1`–`0.4`. That would have made every committed `capabilities.lock.json` unloadable — a far larger break than the leak it repairs. Prior-version locks are now advanced on read, the way a `0.12` packet and a `0.9` verifier artifact already are. (`0.5` was already rejected before this change; left as-is rather than widening scope.)

### Also

- The row names **every** uncovered observation rather than the strongest alone, so a reviewer is not asked to acknowledge evidence the row never showed them.
- The published remedy is true of its state. `strongest_effect_above_declaration` becomes `effect_remedy_instruction`, which names the missing controls instead of telling a reviewer to raise an effect that already outranks the observation.

## Type

- [ ] Input adapter change
- [ ] Documentation only

## Verification

Additional local checks run:

- Full suite green (`pytest tests -n auto`, exit 0); `ruff check .` clean; `generate_schemas.py --check` clean.
- Every defect reproduced against a `git archive origin/main` tree first, and re-run after the fix.
- All 21 samples swept: the two shipped acknowledged overrides still hold and no new gap appears.

## Release-readiness notes

- [x] New or changed check IDs are documented in `docs/checks.md` — no new check IDs

### Notes for the reviewer

- `report_schema_version` (0.36), `packet_schema_version` (0.13), `verifier_schema_version` (0.10), and `contract_version` are **unchanged**. Only the two capability-lock schemas move.
- **Behaviour change worth a second look:** a repository declaring a higher-risk effect *across* categories — `financial_write` or `destructive` on a tool inferred to communicate externally — now raises `declaration_below_inferred_evidence` where it was silent. That is the defect being fixed, but it does narrow #411's "conservative escalation costs nothing" promise to "escalation that still carries the observation's obligations costs nothing". The remedy is one `risk_tags` entry, and the row names which controls are missing.
- Two existing tests encoded the old promise with a cross-category pair (`destructive` over an inferred `external_communication`); both now carry `risk_tags: [external_write]`, and the cross-category case is covered as its own documented behaviour.

Follows up #411. Related: #409, #410.


pengfei-threemoonslab (commented):
Review result: changes needed before merge.

I reviewed d73d64bc against 0c5f40fc with independent semantic, schema-compatibility, and adversarial-test passes. CI and the focused suites are green, but I found the following correctness and contract gaps.

1. [P1] The emitted semantic repair is not guaranteed to close the row (src/agents_shipgate/core/semantic_assessment.py:835-845; src/agents_shipgate/ci/release_decision.py:1308-1339).

   Repro: declare read with heuristic financial_write and external_communication observations. effect_remedy_instruction says to raise the effect to financial_write; applying exactly that edit and rerunning still emits declaration_below_inferred_evidence for external_communication, because financial_write lacks confirmation. A single-claim rank-disagreement case is also impossible: privileged_data_access plus inferred write says “Declare the write controls”, while builtin_obligations("write") is empty.

   The structured action compounds this: accepted_values contains only effect values and the template offers only override, while the non-override orthogonal-category repair is to retain the original effect and add the reviewed category/risk tag plus its controls. Please derive a repair from the full uncovered set, only advertise a candidate that covers every observation, and update the remaining “weaker/higher effect” copy in evidence_actions.py:257 and release_decision.py:1317.

2. [P2] A multi-effect override hides secondary observations from the reviewer (src/agents_shipgate/core/semantic_assessment.py:545-557; src/agents_shipgate/ci/release_decision.py:1132-1138).

   With read plus financial and external heuristic claims, the override claim records two overridden_claim_ids but only one overridden_effect/source pair. After acknowledgment, semantic gaps are empty and the reviewer-facing acknowledged_overrides row, PR projection, and packet show only financial_write/body; external_communication/name disappears even though it was also suppressed. Please project every acknowledged effect/source pair, or emit one reviewer row per effect.

3. [P2] The frozen-schema byte guard silently skips in normal CI (tests/test_effect_coverage.py:426-435).

   It shells out to git show 0c5f40fc~1 and calls pytest.skip when the object is unavailable. The main test job uses actions/checkout with its default fetch depth of 1 (.github/workflows/ci.yml:18-20), so those historical objects are absent in ordinary PR/push CI and all five freeze cases skip exactly where this invariant must be enforced. Prefer checked-in golden hashes/fixtures, or fetch sufficient history and fail rather than skip.

4. [P2] The obligation-table drift test protects only one direction (tests/test_effect_coverage.py:319-385).

   obligations.issubset(missing) still passes when the policy branch adds a required control that BUILTIN_EFFECT_OBLIGATIONS omits; deleting an effect from the table skips that effect entirely. That is the unsafe drift direction because declaration_covers can then discharge the observation without the new control. Iterate every ActionEffect and compare the exact per-effect missing set.

5. [P2] The prior lock schema is tied to a mutable current-standard constant (src/agents_shipgate/core/capability_lock.py:269-280).

   The new entry maps "0.6" to CAPABILITY_STANDARD_VERSION. On the next capability-standard bump, legacy v0.6 locks will therefore be relabeled as the new current standard and may be compared instead of requiring re-export—the opposite of the adjacent comment. Pin the historical entry to literal "0.5"; only the current lock-schema key should follow the current constant.

6. [P2] The new current examples and canonical docs disagree with runtime versions.

   docs/examples/capability-lock-diff.v0.8.example.json:6,14 embeds lock schema 0.6, but the loader normalizes v0.6 to v0.7 and _lock_ref emits v0.7; the CLI cannot emit this tuple. docs/examples/capability-lock.v0.7.example.json:4 also attributes v0.7 output to 0.16.0b6, which emitted v0.6. In addition, STABILITY.md:2348-2357, README.md:714 and 859-860, and docs/architecture.md:329-335 still advertise older schemas as current. Please update these public contract surfaces together and add a runtime-semantic example assertion, not only JSON Schema validation.

Validation performed: ruff, compileall, generate_schemas.py --check, targeted semantic/schema/capability suites, all GitHub checks, and the repository-wide pytest run. The local run's only setup errors were the sandboxed isolated wheel builder being unable to download its backend; tests/test_packaging.py passed 8/8 when rerun with allowed network access. Agents Shipgate reports review_publishable / review_required with the expected trust-root human-review route.

I recommend addressing the six items above before merge.

pengfei-threemoonslab:
All six addressed in 803b0659. Full suite green (`pytest tests -n auto`, exit 0), Ruff clean, `generate_schemas.py --check` clean.

**1 [P1] — the repair now closes the row.** Both repros confirmed. `effect_remedy_instruction` becomes `effect_repair`, deriving the route from the **full** uncovered set:

- A raise is advertised only when one *observed* effect covers every uncovered observation **and** the value already declared. Restricting candidates to the observations matters — rank alone nominates an unrelated effect that merely sits higher (`identity_access` "covers" `privileged_data_access` + `write`), which is a different claim, not a repair. Requiring it to cover the declared value stops "raise the effect" quietly asking for a *lower* assessment.
- Otherwise the row publishes the `risk_tags` route. That is not a formality: a declared tag is policy-eligible evidence, so it both accounts for the observation and makes that category's built-in controls apply — the outcome the uncovered obligation was asking for, and the answer to your `builtin_obligations("write") == set()` case.

`accepted_values` and the scaffold template follow the route (risk-tag values, filled `risk_tags` list beside the override block), and the scaffold hint no longer names `effect:` as the only alternative. `_GAP_PHRASE` and the row's `why` drop "weaker" for "does not account for".

`test_the_published_repair_closes_the_row_it_is_printed_on` applies the published repair to every gapped declared/observed combination — 243 with one- and two-observation sets — and asserts the row is gone. It was 0/243 closing before; it is 243/243 now.

**2 [P2].** Resolver records `overridden_observations`; the projection emits one `AcknowledgedEffectOverride` row per suppressed observation, each with its own sources. No schema change — the field set is unchanged and the list simply gets more rows. The singular pair stays readable so an older report projects one row rather than none. This is the one behaviour change worth noting downstream: a consumer counting rows per action may now see more than one, keyed by `subject_id`.

**3 [P2].** Confirmed — `fetch-depth: 1` meant all five freeze cases skipped. Replaced with checked-in sha256 hashes (verified equal to the `0c5f40fc~1` bytes), plus `test_every_frozen_schema_has_a_pinned_hash` so a frozen document cannot sit outside the guard.

**4 [P2].** Now iterates every `ActionEffect` (not just table keys), scopes to the four built-in control checks so unrelated `missing` fields do not pollute the comparison, and asserts exact set equality. Verified by perturbing the table — dropping `production_operation`'s obligation fails with `says [] but the built-in controls reported ['approval.required']`.

**5 [P2].** Pinned to literal `"0.5"`. My comment claimed the opposite of what the code did on a future bump; good catch.

**6 [P2].** Both examples corrected (lock refs `0.6` → `0.7`; `cli_version` → `0.16.0b7`) and `test_the_published_examples_are_artifacts_this_runtime_could_emit` asserts the runtime semantics, not just the schema. `STABILITY.md`, `README.md`, and `docs/architecture.md` now advertise the current pair — worth noting they were stale *before* this change too (v0.4/v0.5 and v0.5/v0.6, so two bumps behind), which is why the drift was not obvious from this diff alone.

One flow-on the goldens record: adding `overridden_observations` to the override claim changes its `claim_id`, and therefore the refund sample's `run_id`. That is the claim legitimately carrying more evidence, not churn — `overridden_claim_ids` is computed from the challenged claims, so there is no circularity.

Issue #329, opened 2026-08-07
Audit adopter-facing output for internal identity-model vocabulary


pengfei-threemoonslab:
> Depends on #328’s failure-attribution taxonomy so copy changes preserve the distinction between product defects, unsupported valid inputs, user configuration, and environment failures.

## Summary

Adopter-facing output exposes Agents Shipgate's internal identity model. Proceeding through the adoption flow should never require understanding it.

Invariant 5 of #327.

## Evidence

The clearest instance is the failure in #321:

```text
Duplicate tool observation identity: source_type='google_adk_function', source_id='adk_python_agents_smart_closer_agent', native_locator='python/agents/smart_closer/agent.py#map_salesforce_account_to_sap_bp'
```

Three internal concepts — `source_type`, `source_id`, `native_locator` — are printed at someone who has just run the tool for the first time on their own repository. None of them appears in the manifest they wrote. Two of them are derived. The one recoverable fact in that message, that a function is bound to several agents, is the one thing it does not say.

#321 will fix the underlying collision, but the message shape is a separate and more general problem: nothing tracks whether adopter-facing strings speak the adopter's vocabulary.

## Scope

An audit of strings an adopter is expected to read and act on — console output, agent-mode `next_action` and error envelopes, `agent-handoff.json`, and PR comment text. Report internals (`report.json` evidence blocks, catalog identity fields) are legitimately internal and out of scope: they are consumed by tooling, not read for a next step.

The distinction to enforce: internal identifiers may appear as *evidence* in machine-read artifacts; they must not appear in anything whose purpose is to tell a person what to do next.

## Requested

- Enumerate adopter-facing surfaces and audit them for internal-model vocabulary.
- Where an internal identifier is load-bearing for diagnosis, keep it in the structured payload and write the human-facing sentence in terms of files, functions, agents, and manifest keys — things the adopter can see.
- Add a guard test over the adopter-facing string set, in the spirit of the existing declaration-template guard test, so new messages cannot reintroduce the vocabulary.

## Acceptance criteria

- [ ] No adopter-facing message requires knowledge of `source_type`, `source_id`, `native_locator`, observation IDs, or fingerprints to act on.
- [ ] Diagnostic identifiers remain available in the structured payload for machine consumers and bug reports.
- [ ] Every adopter-facing failure names a file, symbol, or manifest key the user can locate.
- [ ] A guard test enumerates the adopter-facing strings and fails on reintroduced internal vocabulary.

[...]

PR #422, opened 2026-08-26
fix(reporting): group human-facing findings by subject; recommend only what is missing (#364)

### LINKED ISSUE #364: Roll up human-facing findings by subject; derive recommendation from evidence.missing
## Summary

Human-facing output lists findings flat, ordered by severity, so one fact about four tools is presented as 17 separate rows across 5 check families. Separately, `recommendation` text is templated and restates requirements the user has already satisfied, contradicting the same finding's `evidence.missing`.

## Evidence (reproduced 2026-08-10, CLI 0.16.0b7)

Final scan on `google/adk-samples` PR #1917 — 17 findings, 4 subjects:

| check_id | sev | count | subject |
|---|---|---|---|
| `SHIP-ACTION-FINANCIAL-WRITE-CONTROL-MISSING` | critical | 4 | the 4 spraay tools |
| `SHIP-AUTH-MISSING-SCOPE` | high | 4 | the same 4 |
| `SHIP-SIDEFX-IDEMPOTENCY-MISSING` | high | 4 | the same 4 |
| `SHIP-MANIFEST-HIGH-RISK-OWNER-MISSING` | high | 4 | the same 4 |
| `SHIP-ADK-GUARDRAIL-EVIDENCE-MISSING` | high | 1 | aggregate over the same 4 |

`Top findings:` shows 3 rows — three instances of the *same* check on three different tools — so the human summary spends its entire budget on one check family and never mentions idempotency, scope, owner, or guardrails.

Recommendation drift, same finding:

```json
"evidence": {"missing": ["safeguards.audit_log", "safeguards.idempotency"]},
"recommendation": "Declare approval.required, safeguards.audit_log, and safeguards.idempotency
                   for this financial write action."
```

`approval.required: true` was already declared in `action_surface.actions`; that is precisely why it is absent from `evidence.missing`. The recommendation instructs the user to re-declare it.

## Impact

Direct user report from the walkthrough: *"痛点不聚焦…看到了 finding 也不知道改什么"* ("the pain isn't focused… I see the findings and still don't know what to change"). The underlying report is correct and complete; the rendering scatters a single decision ("these four tools move money with no enforced controls") across 17 rows sorted by an axis (severity) that is orthogonal to how the reader acts. Following the drifting recommendation wastes a round and teaches distrust of the text.

## Proposal

1. Group the human summary (`stdout`, `report.md`, `pr-comment.md`) by **subject** — tool, then aggregate findings — with severity as an attribute rather than the sort key.
2. Render `recommendation` from `evidence.missing` rather than a static per-check template; drop satisfied requirements.
3. `Top findings:` selects distinct subjects (or distinct check families), never N instances of one check on sibling tools.

## Before / After

**Before**
```
Top findings:
- SHIP-ACTION-FINANCIAL-WRITE-CONTROL-MISSING: spraay_batch_eth - ...
- SHIP-ACTION-FINANCIAL-WRITE-CONTROL-MISSING: spraay_batch_eth_variable - ...
- SHIP-ACTION-FINANCIAL-WRITE-CONTROL-MISSING: spraay_batch_token - ...
```

**After**
```
4 tools move funds with no enforced controls (financial_write, ambient authority):
  spraay_batch_eth, spraay_batch_eth_variable, spraay_batch_token, spraay_batch_token_variable
    BLOCKS RELEASE  missing: audit_log, idempotency
    review          no auth scopes · no idempotency evidence · no owner declared
    review          no ADK callback/plugin guardrail evidence for this group
```

## Acceptance criteria

- [ ] Repro's stdout presents 4 tool groups + 1 aggregate instead of 17 flat rows; blocking status stays visible per group.
- [ ] No `recommendation` names an item absent from that finding's `evidence.missing`.
- [ ] `report.json` `findings[]` is byte-identical — per-finding structure and count unchanged.
- [ ] `report.md` / `pr-comment.md` goldens updated; SARIF unchanged.

## Scope / non-goals

Presentation only. No change to finding identity, fingerprints, counts, severities, `blocks_release`, or the machine contract. `report.json` remains the flat per-finding record that automation consumes.

## Headline metric

**Time-to-first-correct-remediation** on a blocked verdict, and removal of one guaranteed wasted iteration (the already-satisfied `approval.required`). No schema bump; renderer plus recommendation-construction change.

pengfei-threemoonslab:
Closes #364.

## The two defects

**One fact, seventeen rows.** A scan of four money-moving tools produced 17
findings across 5 check families. `Top findings:` showed three — all three
the *same* check on three sibling tools — so idempotency, scopes, owners and
guardrails were never mentioned at all. The report was correct and complete;
the rendering sorted it along an axis (severity) orthogonal to how anyone acts
on it. A reader opens one tool, fixes what is wrong with it, and moves on.

**A recommendation that contradicts its own evidence.** On the finding that
reader would open first:

```json
"evidence": {"missing": ["safeguards.audit_log", "safeguards.idempotency"]},
"recommendation": "Declare approval.required, safeguards.audit_log, and
                   safeguards.idempotency for this financial write action."
```

`approval.required` was already declared — which is precisely why it is absent
from `evidence.missing`. Following that sentence costs a round and returns you
to the same finding.

## What changed

**Subject is the group key.** `scan` stdout, `report.md` and the PR comment
read one projection, `core.findings.subject_rollup`, which groups active
findings by the tool they are about (or by the agent, for findings that name
no tool). Severity and blocking status become attributes of a row; the
location is hoisted to the heading when every row in the group shares one, and
kept on the rows when they differ (which is what distinguishes
`samples/conductor_agent`'s two identically-titled findings). Every truncation
states what it hid.

One selection rule replaces three. Each surface previously had its own rule
*and* its own limit, so one scan reported a different "top" three ways; the
shared rule is "critical or high, or named by the release decision", which is
a superset of all three.

**The sentence is derived from `missing`.** Deriving it at the emit site is
not enough on its own — the caller could still pass a different list to
`evidence`. `_builtin_control_finding` takes **one** `missing` argument and
builds the evidence, the sentence, and the `missing_builtin_controls`
predicate row from it, so the three agree by construction rather than by
review. `SHIP-ACTION-POLICY-VIOLATION` also stops naming both high-impact
effects on an action that has one.

Rendering order is canonical rather than the caller's, because two branches
collect the same set in different orders and one set of missing controls has
to have one sentence. Where nothing is declared the sentence is byte-identical
to the old literal — which is why **no shipped sample's `recommendation`
moved**.

## Before / after

Before, on the reported repository:

```
Top findings:
- SHIP-ACTION-FINANCIAL-WRITE-CONTROL-MISSING: spraay_batch_eth - ...
- SHIP-ACTION-FINANCIAL-WRITE-CONTROL-MISSING: spraay_batch_eth_variable - ...
- SHIP-ACTION-FINANCIAL-WRITE-CONTROL-MISSING: spraay_batch_token - ...
```

After:

```
Top findings (14 findings across 5 subjects):
- spraay_batch_eth [spraay] (at agent.py) — BLOCKS RELEASE (1 critical, 2 high)
    critical SHIP-ACTION-FINANCIAL-WRITE-CONTROL-MISSING (blocks release) — missing: safeguards.audit_log, safeguards.idempotency
    high SHIP-MANIFEST-HIGH-RISK-OWNER-MISSING — spraay_batch_eth is high-risk but has no owner
    high SHIP-SIDEFX-IDEMPOTENCY-MISSING — spraay_batch_eth lacks idempotency evidence
  … three more tool subjects …
- spraay-agent (agent-wide) (at shipgate.yaml) — review (1 high, 1 medium)
    high SHIP-ADK-GUARDRAIL-EVIDENCE-MISSING — High-risk Google ADK tools lack static guardrail evidence
    medium SHIP-ADK-EVAL-COVERAGE-MISSING — Google ADK eval coverage is not declared
```

## Acceptance criteria

- [x] The repro's stdout presents 4 tool groups + 1 aggregate instead of 17
      flat rows; blocking status stays visible per group.
- [x] No `recommendation` names an item absent from that finding's
      `evidence.missing` — swept over every shipped sample, with a
      non-vacuity assertion so an empty sweep fails.
- [x] `report.json` `findings[]` unchanged: same count, same order, same
      fields, same fingerprints. Every sample golden `report.json` is
      byte-identical (only `manifest_dir` moves on regen, and that is reverted).
- [x] `report.md` / PR-comment goldens updated; SARIF untouched.

## What the review rounds found

Five defects in this change, all fixed here with regression tests.

**Round 1 — two claims that are not the same claim.**

1. **`finding.blocks_release` is not "the decision blocks on this."** A
   baseline separates them: `ci.release_decision` files a policy finding whose
   debt a baseline has accepted as a *review item* while the flag stays true.
   Reading the flag printed BLOCKS RELEASE two lines under a verdict that had
   accepted it. The decision is now the only authority; the flag is consulted
   only when there is no decision to contradict.
2. **A shared title, then a shared fingerprint, borrowed another finding's
   verdict.** The join to a decision item is now three tiers in descending
   precision — id, then fingerprint for an item with no id, then check id and
   title for an item with neither — each holding only the items that could not
   supply the tier above. `samples/conductor_agent` ships two findings with
   one check id and one title; `assign_finding_ids` exists because two
   findings can share a fingerprint.
3. **`Declare  for this financial write action.`** An empty `missing` is
   unreachable (every branch is inside `if missing:`), so it is a wiring
   mistake rather than a state — and the useful answer to one is the pre-#364
   sentence, not a hole. Raising would abort a scan over a rendering detail.

**Round 2 — the same second-spelling failure, from the other side.**

4. **One tool, two headings.** `checks.baseline_integrity` emits findings with
   a tool *name* and no tool id, so a tool with one of those and one ordinary
   finding got both `create_refund` and `create_refund [stripe]` — exactly what
   `catalog_label_index` exists to prevent. A name-only finding now joins the
   tool its name resolves to, where the name resolves to exactly one. An
   ambiguous name is left unresolved rather than guessed: filing a finding
   under a tool it is not about is worse than a heading the reader reconciles.
   (And ambiguity is two *tools* answering to a name, not two catalog rows.)
5. **`report.md` said the missing list twice.** It annotated every row with
   its recommendation, including rows whose detail already *is* the `missing`
   list that recommendation is derived from. The reason the compact surfaces
   skip the sentence is that it repeats the row in different words, and that
   reason does not stop applying inside a report. The full sentence is still
   in `## Recommended Next Actions`.

The three-surface test now renders all three and pins what has to agree: each
surface shows the first N subjects of one projection, never a different N.

## Review round three — the six findings on `2801becd`

All six confirmed against the code and fixed; each has a regression test.

1. **[P1] The default PR comment had none of this.** Correct, and the most
   important one. `render_pr_comment` defaults to `capability-review`; the
   block landed only in `_render_findings_comment`, which the CLI reaches only
   under `--pr-comment-style findings` — the style its own help text calls
   "legacy v1 … available for one minor release cycle". So the comment a
   reviewer actually gets said what *moved* and nothing about what is wrong
   per tool. Both styles now render the same projection at the same budget,
   nested under "Human summary" and omitted when nothing is selected. A test
   calls `render_pr_comment` with no `style`; another pins that the 6000-char
   truncation still preserves the agent instruction block, since a machine
   reads that.

2. **[P1] A blocking row could be truncated away.** Correct. Blocking rows now
   sort ahead of severity within a group — the same order the groups are in,
   one level down. A heading has to be able to show its own evidence. Covered
   at equal severity with the blocker sorting last by check id.

3. **[P1] Not every `evidence.missing` entry is an absent thing.** Correct on
   all three counts. The fix is the structural discriminator rather than a
   check-id allowlist: **only a list of plain strings is read as a missing
   list.** `_missing_requirements` writes `{"path", "expected"}` rows, so
   policy findings now keep their own title *and* their adopter-authored
   recommendation. That also keeps the framework metadata checks working —
   they write plain strings, and `missing: description, parameters` is
   faithful for those.

4. **[P2] A name is not a binding.** Correct, and it makes my previous round's
   fix the wrong idea. `BaselineFinding.tool_id` exists, but propagating it
   into the emitted `Finding` moves that check's **fingerprint** (it is
   `check_id` + `tool_id` + evidence), which is a baseline-compatibility
   change and not a rendering one — so I took your second option. A name-only
   finding keeps its own subject, and the absent `[provider]` qualifier is the
   signal that the two are not known to be the same tool.

5. **[P2] `location` skipped, and identity trimmed.** Correct, and this PR's
   own repro was hitting it: every ADK finding carries
   `ref="agent.py"` + `location="agent.py:5"` with `path` unset, so four
   findings on four functions rendered one suffix — and then *shared* it, so
   it was hoisted to the heading as if they were one place. Chain is now
   path+pointer / path+line / path / location / ref, with emptiness decided by
   `has_visible_content` and the stored value passed to `display_literal`
   untrimmed. Visible in the golden as
   `agents/refund_agent.py` → `agents/refund_agent.py:5`.

6. **[P2] `info` missing from the histogram.** Correct. Fixed as a rule rather
   than an entry: the display order is derived from `SEVERITY_ORDER`, so it
   cannot fall behind the table again. Regression test drives an `info`
   finding through the decision.

One more found while fixing (3): `report.md` decides "this sentence only
repeats the row" by looking the row's items up in the sentence — asking for
the raw *path* missed every control the sentence renames, so
`confirmation.required` printed twice on every external-communication
finding. It now asks through `control_phrase`, the same table that renders
them.

## Notes for review

- **A group heading is the most adopter-facing string this projection emits**,
  so #329's rule applies to it. Two id-leak paths are closed and tested:
  `catalog_subject` falls back to the tool id for a nameless catalog row
  (right where it joins two surfaces by value, wrong in a heading), and an
  agent name is adopter-controlled so it can *look* like a derived id.
- `core.surface_exclusions.agent_label_index` indexes the **binding graph**,
  keyed by `agent_v1:` digests, while `Finding.agent_id` is the
  manifest-derived `agent:<project>/<agent>`. The two id spaces never
  intersect — reaching for that index here reads correct and resolves nothing.
- A path in a row goes through `display_literal`, so a zero-width character
  stays visible and recoverable rather than silently naming a different file.
  Findings are already redacted in scan phase 7, before `build_report`, so the
  console and the comment print sanitized values.
- `report.md` keeps each row's recommendation as a nested line; the console
  and the PR comment do not, because since this change that sentence is
  derived from the same `missing` list the row already prints.
- `core/findings/subject_rollup.py` is in the `tests/test_adopter_vocabulary.py`
  sweep. Markdown report layout is explicitly non-contract in
  [`STABILITY.md`](STABILITY.md); no stability claim moves.

## Verification

- `ruff check .` clean.
- `pytest -n auto -m "not perf" --ignore=tests/test_adapter_static_only.py`
  green, plus the perf and static-only runs separately, with
  `GITHUB_STEP_SUMMARY` set the way CI has it.
- `./shipgate self-check` ready; `./shipgate verify --config shipgate-self.yaml`
  → `mergeable` / `passed`.


pengfei-threemoonslab (commented):
Changes requested after reviewing exact head 2801becd8793d56ae13d8db1bf7188e1b8bfcd9a against issue #364 and tracing all three human-facing renderers.

1. [P1] Wire the subject rollup into the default PR-comment path ([pr_comment.py:358](https://github.com/ThreeMoonsLab/agents-shipgate/blob/2801becd8793d56ae13d8db1bf7188e1b8bfcd9a/src/agents_shipgate/report/pr_comment.py#L358)). This new block is inside _render_findings_comment, which is reached only when style is findings; render_pr_comment and the CLI still default to capability-review. A direct default invocation contains neither Top findings nor the grouped subjects, while the same call with style=findings does. Consequently the normal verifier PR comment does not receive the behavior this PR and issue #364 promise. Please wire the projection into the default renderer (or intentionally change the default) and add a test that exercises render_pr_comment without overriding style.

2. [P1] Keep an actual blocking row visible before applying row_limit ([subject_rollup.py:226](https://github.com/ThreeMoonsLab/agents-shipgate/blob/2801becd8793d56ae13d8db1bf7188e1b8bfcd9a/src/agents_shipgate/core/findings/subject_rollup.py#L226)). Rows are sorted only by severity, check id, and title, then top_findings_block takes the first three/five. With five high findings on one subject and only the alphabetically last one named by release_decision.blockers, the output says BLOCKS RELEASE but shows three non-blocking rows and hides the only blocker under “and 2 more findings.” Sort blocking rows first (or reserve a visible slot for one) before truncation, and cover the equal-severity case.

3. [P1] Do not interpret every evidence.missing entry as an absent control ([subject_rollup.py:152](https://github.com/ThreeMoonsLab/agents-shipgate/blob/2801becd8793d56ae13d8db1bf7188e1b8bfcd9a/src/agents_shipgate/core/findings/subject_rollup.py#L152), [markdown.py:318](https://github.com/ThreeMoonsLab/agents-shipgate/blob/2801becd8793d56ae13d8db1bf7188e1b8bfcd9a/src/agents_shipgate/report/markdown.py#L318)). Action-policy evaluation puts both unknown paths and present-but-wrong values in evidence.missing and adds evidence.observed for the latter. The new projection can therefore say “missing: safeguards.dry_run” when dry_run exists but differs from the expected value; it also suppresses the adopter-authored policy recommendation. Distinct custom policies requiring the same path become identical rows. Restrict the shorthand to the built-in generated missing-control checks, or render policy id plus expected/observed and retain custom titles/recommendations.

4. [P2] Do not bind historical baseline findings to a current tool by name ([subject_rollup.py:418](https://github.com/ThreeMoonsLab/agents-shipgate/blob/2801becd8793d56ae13d8db1bf7188e1b8bfcd9a/src/agents_shipgate/core/findings/subject_rollup.py#L418)). Baseline-integrity findings intentionally arrive with tool_id=None and a name copied from the historical BaselineFinding. If the old tool was removed and a new provider now exposes the same unique name, this lookup labels an expired/stale entry as belonging to the new provider. Preserve and propagate BaselineFinding.tool_id through the integrity issue/Finding, or leave name-only baseline findings unbound; current-catalog uniqueness cannot establish historical identity.

5. [P2] Preserve legacy source line data and exact source identity ([subject_rollup.py:468](https://github.com/ThreeMoonsLab/agents-shipgate/blob/2801becd8793d56ae13d8db1bf7188e1b8bfcd9a/src/agents_shipgate/core/findings/subject_rollup.py#L468)). Existing adapters and shipped reports commonly use SourceReference(ref="agent.py", location="agent.py:5", path=None). This fallback ignores location and renders only agent.py, so findings on different lines become indistinguishable and may be hoisted as one shared location. The strip calls also collapse valid identity-bearing paths/pointers that differ by whitespace before display_literal can preserve them. Prefer structured path/start_line, then location, then ref, and pass the exact stored identity to display_literal.

6. [P2] Include info in the severity histogram ([subject_rollup.py:429](https://github.com/ThreeMoonsLab/agents-shipgate/blob/2801becd8793d56ae13d8db1bf7188e1b8bfcd9a/src/agents_shipgate/core/findings/subject_rollup.py#L429)). Severity supports info, and roll_up_findings deliberately includes a lower-severity finding when the release decision names it. An info finding named as a blocker currently renders BLOCKS RELEASE () because info is absent from the display order. Add info and a decision-selected regression test.

Validation performed: 35 focused subject-rollup tests passed; the related report, CLI, action-surface, release-decision, and privacy tests passed; Ruff and git diff --check passed; all GitHub checks are green. A full local run reached 100% with no assertion failures; its eight test_packaging.py setup errors were solely the isolated wheel builder being unable to download hatchling in the restricted environment. Agents Shipgate verification at this exact head returned control=complete, merge_verdict=mergeable, and release_decision=passed for its static scope.

PR #423, opened 2026-08-26
feat(evidence): pin an answer to its evidence, and name the rung you stand on (#410 increment 4)

pengfei-threemoonslab:
Sections **E** and **G** of [#410](https://github.com/ThreeMoonsLab/agents-shipgate/issues/410) — the RFC's final increment. E is what makes the system safe to run for years; G is what makes every state of an adoption a place with a name and a next step instead of a verdict that reads like a failure.

## §E — a confirmed declaration is pinned to the evidence behind it

Declarations are matched by name (`tool_id` optional) and nothing ever re-opened one, so a green gate at month twelve could rest on a description of a function that no longer does what it did.

`action_surface.actions[].basis` records **which evidence an effect answer was given against**, as `confirmed:<digest>`. Every scan re-derives it and compares. Equal is complete silence; different re-opens the question as a `declaration_drift` evidence gap that names what the action reads as *now*, carries the new pin, and is closed by re-reading and re-confirming.

Three choices decide whether this helps or becomes a treadmill, and each has a test:

- **What is digested is what the reviewer was shown** — the effects the scan *observed*, which is exactly the readings the questionnaire prints above the answer.
- **Producers are not digested.** A second heuristic reading an effect somebody already answered is not new information; digesting it would let a shipgate release re-open every pinned declaration on every adopter at once.
- **The pin does not move when the answer arrives.** Parametrised over every declaration shape plus the MCP protocol-default path, because a false drift row on somebody's own answer would teach every adopter to ignore the row.

Additive and unpinned-by-default. New answers arrive pinned because `suggested-declarations.yaml` stamps it; to pin an existing declaration, write any short placeholder (`basis: confirmed:0`) and rescan — the drift row names the value.

`declaration_drift` is a different statement from `declaration_below_inferred_evidence`. That one asks whether the declaration is *weaker than* today's evidence; this one asks whether today's evidence is what was answered at all. Each is closed by a different edit.

## §G — the adoption ladder, and the trust root it ends at

**`environment.target: template`.** A sample has no deployment, so it has no credentials, and asking each of its actions which credential it runs with asks a question the repository cannot answer in principle. Declaring `template` answers the authority dimension once, and **only where nothing else answers it**: an action row's `authority`, a `tool_sources[].authority` block, an action's own `scopes:` list, and anything the *source* publishes all win over it. The reviewed record supplies mode, auth type, credential mode and permission list **together**, so applying it over a source that published any of the four would erase what that source proved. And **never silently**: every action it answers for is one semantic review concern, so a template repository reaches `review_required` and never `passed`.

**`SHIP-TRUST-MANIFEST-UNPROTECTED`.** Attestation is the PR review of a protected file, and CODEOWNERS is the half a checkout can prove. Fires only where the manifest is load-bearing — the manifest's own `ci.mode: strict`, from the declared block rather than `--ci-mode` — and never moves a verdict. Fails closed on both halves it cannot establish: outside a git checkout there is no pull request for a rule to gate, and a token GitHub would not accept as an owner assigns nobody.

**`doctor` names the rung.** 0 Audit · 1 Gate the delta · 2 Answer on touch · 3 Strict, with what each is worth and **only the conditions actually unmet** to reach the next one, published as `adoption` on `doctor --json`. A rung describes what a repository has **declared** — it does not predict a verdict, and it does not claim enforcement: `ci.mode: strict` is the manifest's statement, while the workflow passes its own `ci_mode` (the generated one ships `advisory`) and branch protection is a repository setting no file can read. Rung 3 names both limits. The rungs are not cumulative, so a structurally-resolved repository that is never asked a question is not stranded at rung 1.

## The loop, walked end to end

| step | verdict | questions |
|---|---|---|
| 1 · baseline, nothing declared | `insufficient_evidence` | 1 of 3 answered (the template answered authority) |
| 2 · paste the two scaffold blocks verbatim | `blocked`, risk named | **3 of 3 answered, 0 open** |
| 3 · a tool's description gains "permanently delete" | `blocked` | 2 of 3 — `declaration_drift` re-opens |
| 4 · take the value the drift row publishes | `review_required` | back to answered, silent |

Pinned as a test that walks all four hops through real scans.

## Review

**Round 1 (self)** found a fail-open the perturbation sweep surfaced by returning *zero* failures for a guard I believed I had written: `template` was erasing a permission a source published, dropping a `high` finding. **Round 2** added the four-hop loop test and closed the last two unguarded mechanisms. **Round 3** found three rungs describing the RFC's intent rather than the product.

**Round 4 — the review on this PR.** Nine comments, three of them blockers; each fix has a guard that fails without it.

- **[P1] v0.7 capability locks stopped loading.** The bump froze `0.7` without adding it to the compatibility set or the historical standard map — `0.7` had been covered only by the *current constant* key, so v0.8 taking that key orphaned every committed v0.7 lock and would have relabelled it as current-standard. The old guard pinned `0.6` and kept passing throughout; the replacement walks the published `docs/capability-lock-schema.v*.json` set and requires every version describing the current standard to load.
- **[P1] `template` erased a source-published `credential_mode`.** `_source_authority` never reads that field, so a tool publishing `service_account` and nothing else graded `unknown` and had its credential mode cleared — taking the capability out of reach of `credential_modes` policy selectors. The guard is keyed off `ReviewedAuthority`'s own fields: the record replaces four, so a source publishing any of the four disqualifies it and a fifth cannot be forgotten.
- **[P1] Rung 3 claimed an enforcement it cannot see.** Fixed as described above.
- **[P2] ×6** — a control-only action row no longer advances the rung; rung 1 no longer names a verdict; rung 0's next step is the rendered `init --workspace … --write` (routed through the glob-aware helper `doctor` already uses); adoption reads the manifest the user wrote rather than the copy with unresolved sources removed; CODEOWNERS fails closed outside a checkout and on non-owner tokens; and the contract bump reached the prose in `STABILITY.md`, `docs/capability-standard.md`, `README.md`, and every shipped skill copy.

Three **prose-drift parity rules** were extended so that last class stops recurring: version literals are now recognised in their `v0.16` form (the `\bv` boundary is precisely why a stale "Packet v0.15" beside a v0.16 link was invisible), a line naming a superseded version must name the one it actually superseded, and the capability-lock schemas are covered at all.

## One permission list, across a third site

`environment.target` is a third manifest site a reviewed authority can come from, so it is threaded into `resolve_action_scopes` rather than re-derived without it by the action lens. `CapabilityFactV1` requires the action fact's `required_scopes` and the authority dimension's `scopes` to be one list and only checks on the base-comparison path, so this was verified with a **real `verify --base` run** on a template manifest, plus a sweep across (declaration × source × target).

## Verification

- Full suite green (`-n auto -m "not perf"`, the static-only split, the perf split), `ruff check .` clean, `generate_schemas.py --check` clean, `verify --base` exit 0.
- Report schema 0.38 → 0.39, packet 0.15 → 0.16, verifier 0.12 → 0.13, capability lock 0.7 → 0.8, capability-lock diff 0.8 → 0.9. All additive, all prior versions frozen and byte-pinned, all readable forward.
- Merged with `main` (#421, #422) — a conflicting PR stops CI silently, which is why no checks ran on two commits until the merge landed.

## Found here, filed separately

[#424](https://github.com/ThreeMoonsLab/agents-shipgate/issues/424) — **the `declare_risk_tags` repair cannot close the row it is printed on.** Applying it exactly as instructed replaces the row with a blocking `conflicting_effect_evidence`: the `contradictory` filter reads the reviewer's own `risk_tags` as source evidence contradicting their own `effect`, one branch over from where the same class was already fixed. Measured: **0 of 92** first-answer proposals affected, **7** repair shapes affected. Predates this branch, and the fix turns a blocking verdict into a passing one, so it is filed with a reproduction rather than folded in.

## Not in this PR

**Rung 0 as `scan` with no manifest.** `scan` requires a manifest today, and synthesizing one is a real feature with its own decisions — what the audit report claims, whether it writes, and what happens when `-c` is simply mistyped. Rung 0 names what actually runs without a manifest today (`shipgate detect`, `shipgate audit --host`).

Advances [#410](https://github.com/ThreeMoonsLab/agents-shipgate/issues/410).


src/agents_shipgate/schemas/capabilities.py
```diff
@@ -14,8 +14,8 @@
 from agents_shipgate.schemas.semantic import ToolSemanticEvidence
 from agents_shipgate.schemas.surfaces import ActionEffect
 
-CAPABILITY_LOCK_SCHEMA_VERSION = "0.7"
-CAPABILITY_LOCK_DIFF_SCHEMA_VERSION = "0.8"
+CAPABILITY_LOCK_SCHEMA_VERSION = "0.8"
```
pengfei-threemoonslab, line 17: [P1] Preserve immediately prior v0.7 locks. This bump makes the model accept only 0.8, but the compatibility normalizer still recognizes only 0.6 and the historical standard map omits 0.7. Loading the shipped `docs/examples/capability-lock.v0.7.example.json` now raises `InputParseError` (expected 0.8), so existing committed locks break in capability diff/verifier. Add 0.7 to `_STRUCTURALLY_CURRENT_LOCK_SCHEMAS`, map 0.7 to capability standard 0.5, and make the prior-schema test exercise 0.7 rather than 0.6.

src/agents_shipgate/core/semantic_assessment.py
```diff
@@ -1377,9 +1508,54 @@ def reviewed_authority(
...
+    ``_source_authority`` reports ``"unknown"`` status exactly when the source
+    published nothing usable — including the ambiguous and invalid cases, which
+    are *something published*, badly, and equally not this claim's to overwrite.
+    """
+
+    if environment_target != TEMPLATE_ENVIRONMENT_TARGET:
+        return False
+    if _reviewed_scopes(declaration, None):
+        return False
+    return _source_authority(tool)[1] == "unknown"
```
pengfei-threemoonslab, line 1709: [P1] Do not let template authority erase a source-published credential mode. `AuthInfo` legally permits `credential_mode="service_account"` without mode/type/scopes; that is retained locally with a missing-authority gap, but this predicate labels the source `unknown`, installs `mode:none`, clears `credential_mode`, and lets `credential_modes` policy selectors stop matching. Treat any non-empty `credential_mode` as source evidence (at least partial) that disqualifies the template fallback, and add a local-vs-template monotonicity test for the policy finding set.

src/agents_shipgate/core/adoption_ladder.py
```diff
@@ -0,0 +1,193 @@
...
+    protection: ManifestProtection,
+) -> AdoptionRung:
+    """The highest rung whose conditions this repository already meets.
+
+    Conditions are cumulative and each is a fact about the manifest or the
+    checkout: a rung is never awarded for something a scan would have to prove,
+    because the rung has to be answerable before the first scan runs.
+    """
+
+    strict = manifest.ci.mode == "strict"
```
pengfei-threemoonslab: [P1] This does not establish an enforced gate. A repository with no Shipgate workflow—or with the generated workflow still passing `ci_mode: advisory`—gets `strict=True` from the manifest alone, even though the workflow/CLI override is the mode that controls the exit code. With one declaration and CODEOWNERS, doctor then promises “CI fails on a blocking verdict” when no such enforcement exists. Include the installed/effective CI mode in the rung conditions, or describe this rung only as declared strict intent.

src/agents_shipgate/core/adoption_ladder.py
```diff
@@ -0,0 +1,193 @@
...
+    """Whether anyone has answered a declaration question in this manifest.
+
+    ``environment.target: template`` counts, and counts as one answer for the
+    whole repository: it is a reviewed statement that nothing here holds a
+    credential, which is exactly the authority question every action would
+    otherwise be asked (#410 §G).
+    """
+
+    return bool(
+        manifest.action_surface.actions
```
pengfei-threemoonslab: [P2] Count only rows that actually answer a declaration question. A valid approval-only or safeguards-only action row makes `bool(actions)` true here and advances doctor to rung 2 even though neither effect nor authority has been answered and the questionnaire remains open. Check effect/risk tags/authority specifically (or reuse the counterfactual declaration-question logic) instead of treating every action row as semantic evidence.

src/agents_shipgate/core/adoption_ladder.py
```diff
@@ -0,0 +1,193 @@
...
+            blocking=blocking,
+        )
+    return AdoptionRung(
+        number=1,
+        name="Gate the delta",
+        # What running the gate here actually produces, which is not yet a
+        # gateable verdict: with nothing declared the answer is
+        # ``insufficient_evidence``, and promising otherwise would be the same
+        # over-claim the audit rung was rewritten to avoid.
+        you_get=(
```
pengfei-threemoonslab, line 165: [P2] Rung 1 cannot promise this verdict or file from manifest shape alone. A fully structural tool surface can have no reviewed declarations, zero declaration questions/evidence gaps, and a `review_required` or `passed` scan; in that case `suggested-declarations.yaml` has nothing to list and the verdict is not `insufficient_evidence`. Keep doctor’s copy agnostic about the eventual scan result, or derive the rung from resolved semantic coverage.

src/agents_shipgate/core/adoption_ladder.py
```diff
@@ -0,0 +1,193 @@
...
+#: one call site that can observe it, so the ladder is one list in one place.
+AUDIT_RUNG = AdoptionRung(
+    number=0,
+    name="Audit",
+    you_get=(
+        "a read-only look with no manifest and nothing written: `shipgate "
+        "detect` for the agent frameworks and tool sources present, `shipgate "
+        "audit --host` for the host grants."
+    ),
+    exit_criterion=(
```
pengfei-threemoonslab: [P2] This next step does not advance the rung: bare `agents-shipgate init` is a dry run, so it does not “write a manifest”, and the fixed string also drops doctor’s explicit workspace. Emit a context-aware `agents-shipgate init --workspace <workspace> --write` action (or reuse doctor’s structured recovery route).

src/agents_shipgate/cli/scan/inspect.py
```diff
@@ -145,6 +147,7 @@ def inspect_sources(
     # Some adapters expose the same warnings through both LoadedToolSource
     # and the artifact bag; keep doctor warning output stable and unique.
     warnings = list(dict.fromkeys(warnings))
+    rung = adoption_rung(manifest, manifest_protection(config_path))
```
pengfei-threemoonslab: [P2] Compute adoption from the manifest the user actually wrote. When a required source is unresolved, lines 94–107 remove it from this object before this call; a missing source that carries `authority: {mode: none}` is therefore reported as rung 1 even though the original manifest contains a reviewed authority answer and `adoption_rung(original)` returns rung 2. Keep an original manifest for adoption and a separate filtered copy for source loading.

pengfei-threemoonslab (commented):
I reviewed head `01b0461e` across declaration drift, authority resolution, adoption/doctor behavior, CODEOWNERS protection, serialized contracts, and generated schemas.

Validation is green: the focused regression suite passed; Ruff passed; generated schemas are current; 6,320 non-packaging tests passed with 7 skipped; all 8 isolated packaging tests passed when rerun with network access; and the GitHub Actions checks on this head are successful.

I found two release-blocking regressions plus several adoption/contract correctness issues in the inline comments. The blockers are (1) existing v0.7 capability locks become unreadable after the 0.8 bump, and (2) `environment.target: template` can erase a source-published `credential_mode` and bypass credential-mode policy selectors. I would not merge this head until the P1 items are fixed and the adoption claims are tightened.

Agents Shipgate’s own diff verifier reports `review_publishable / review_required` (0 blockers, 9 protected trust-root review items), as expected for the `AGENTS.md` / `SKILL.md` changes. This is submitted as a COMMENT review because the authenticated GitHub identity is also the PR author; the P1 labels are still intended as blocking.

src/agents_shipgate/core/manifest_protection.py
```diff
@@ -0,0 +1,197 @@
...
+def manifest_protection(config_path: Path) -> ManifestProtection:
+    """Read this checkout's CODEOWNERS and say whether it covers the manifest.
+
+    Never raises: an unreadable or absent file is reported as no coverage,
+    which is what it is. A repository that is not a git checkout has no
+    CODEOWNERS semantics at all and is reported the same way.
+    """
+
+    manifest = config_path.resolve()
+    root = git_root_for(manifest.parent) or manifest.parent
```
pengfei-threemoonslab: [P2] Fail closed on whether CODEOWNERS is effective. If `git_root_for()` returns `None`, this falls back to `manifest.parent`, so a non-git directory with a stray CODEOWNERS is reported protected. In a git checkout, any trailing token—including `not-an-owner`—also makes `bool(owners)` true even though GitHub skips invalid ownership syntax. Both cases suppress the guidance finding and can award rung 3. Return uncovered outside a git checkout and validate GitHub-supported owner forms (and ideally self-coverage/size) before setting `covered`.

STABILITY.md
```diff
@@ -2499,15 +2499,15 @@ those identities should not be recorded.
 envelope to `.agents-shipgate/capabilities.lock.json` and, by default, a
 byte-identical generated mirror at
 `agents-shipgate-reports/capabilities.lock.json`. The current lock schema is
-[`docs/capability-lock-schema.v0.7.json`](docs/capability-lock-schema.v0.7.json)
+[`docs/capability-lock-schema.v0.8.json`](docs/capability-lock-schema.v0.8.json)
```
pengfei-threemoonslab, line 2502: [P2] Complete the contract bump, not just the link. This section still says exports emit lock 0.7, normalize to 0.7, and diffs emit 0.8, while this PR’s runtime emits 0.8/0.9. The same stale lock/diff values remain in `docs/capability-standard.md` and all shipped skill copies; those skills also label the packet v0.15 while linking v0.16 (and call v0.37 the report predecessor). Update every synchronized surface and extend the prose-drift test so agents are not directed to strict prior schemas.

src/agents_shipgate/core/semantic_assessment.py
```diff
@@ -1222,6 +1283,70 @@ def effect_readings(effect: EffectSemanticAssessment) -> list[EffectReading]:
...
+      row back on the very next scan, which is the one thing a pin may not do.
+
+    Producers are deliberately **not** digested, only the readings themselves.
+    A second source corroborating a reading the reviewer already answered is
+    not new information about the action, and a shipgate release that adds a
+    heuristic would otherwise re-open every pinned declaration on every
+    adopter at once.
+    """
+
+    observed = sorted({reading.effect for reading in readings if reading.observed})
```
pengfei-threemoonslab: [P1] Reopen when authoritative evidence is replaced. This digest hashes only the distinct effect strings, not the evidence class or strength. Repro: pin `effect: read` while an MCP tool has `readOnlyHint=True` plus a medium `read_only` keyword hint, then remove `readOnlyHint`. Both pins remain `confirmed:8a2d0b87e8c7`; the stale declaration has no issues, stays pass-eligible, and semantic coverage reports zero gaps. The authoritative evidence the reviewer saw has disappeared while the safety-sensitive `read` classification survives. Include a stable provenance/strength class in the digest (while keeping purely additive corroboration quiet) and add a structural-to-heuristic replacement test.

pengfei-threemoonslab (commented):
Second pass on `ff42e0f7`. The nine findings from the first review are materially addressed, but this pass found two remaining P1 blockers and five P2 correctness/contract gaps. I am submitting a COMMENT because the authenticated account is also the PR author; please treat the P1 comments as blocking.

Validation completed: the full local suite (with packaging run separately), focused drift/adoption/public-contract suites, Ruff, schema-generation check, `git diff --check`, all three PR workflows, and `verify --base origin/main --head HEAD` are green. Shipgate reports `review_publishable` / `review_required`, so a human review remains required before merge.

src/agents_shipgate/core/adoption_ladder.py
```diff
@@ -0,0 +1,237 @@
...
+    return AdoptionRung(
+        number=0,
+        name="Audit",
+        you_get=(
+            "a read-only look with no manifest and nothing written: `shipgate "
+            "detect` for the agent frameworks and tool sources present, "
+            "`shipgate audit --host` for the host grants."
+        ),
+        exit_criterion=(
+            f"run `{init_command}` to write a manifest, and the next pull "
```
pengfei-threemoonslab: [P1] Do not promise PR gating from manifest creation alone. The doctor route supplied here is `init --workspace … --write`; `--ci` is a separate flag, so a repository with no workflow gets no verdict on its next PR. `adoption_rung()` then infers rung 1 from the manifest alone and says the gate is running on every PR, although it never inspects a workflow or trigger. Either install and validate CI as part of this transition/rung, or describe local manifest/scan capability instead. Add a no-workflow doctor regression.

src/agents_shipgate/core/manifest_protection.py
```diff
@@ -0,0 +1,226 @@
...
+    directory_only = pattern.endswith("/")
+    core = pattern.strip("/")
+    if not core:
+        return False
+    # A pattern with no separator matches at any depth, exactly as gitignore
+    # does: ``shipgate.yaml`` covers ``apps/api/shipgate.yaml``.
+    anchored = "/" in pattern.rstrip("/")
+    body = _translate(core)
+    prefix = "" if anchored else r"(?:.*/)?"
+    tail = "/.*" if directory_only else "(?:/.*)?"
```
pengfei-threemoonslab: [P2] Match GitHub's one-level wildcard semantics. The unconditional descendant tail makes `/docs/* @docs` match `docs/sub/shipgate.yaml`; [GitHub explicitly documents](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) that `docs/*` does not match further-nested files. This false positive sets `covered=True`, suppresses `SHIP-TRUST-MANIFEST-UNPROTECTED`, and can award rung 3 without a code-owner review. Recurse only for patterns whose semantics actually include descendants, and add this exact nested case.

src/agents_shipgate/core/manifest_protection.py
```diff
@@ -0,0 +1,226 @@
...
+
+import re
+from dataclasses import dataclass
+from pathlib import Path
+
+from agents_shipgate.core.agent_controls import git_root_for
+
+#: Where GitHub looks for CODEOWNERS, in its own precedence order: a file in
+#: ``.github`` wins over one at the root, which wins over one in ``docs``.
+CODEOWNERS_LOCATIONS: tuple[str, ...] = (
```
pengfei-threemoonslab, line 42: [P2] Protect the new protection input itself. These three files now decide whether manifest protection is credited, but `trust_root_class_for()` returns `None` for every one because none is in the canonical trust-root table. With the recommended manifest-only rule, CODEOWNERS also does not self-own, so a later PR can weaken or remove the rule without `SHIP-VERIFY-TRUST-ROOT-TOUCHED` or host-boundary routing. Add all three locations to the shared trust-root/protected-file catalog and require the selected file to cover itself before awarding protection.

src/agents_shipgate/core/manifest_protection.py
```diff
@@ -0,0 +1,226 @@
...
+        relative = manifest.relative_to(root).as_posix()
+    except ValueError:  # pragma: no cover - resolve() keeps these under root
+        relative = manifest.name
+
+    for location in CODEOWNERS_LOCATIONS:
+        candidate = root / location
+        if not candidate.is_file():
+            continue
+        try:
+            text = candidate.read_text(encoding="utf-8", errors="replace")
```
pengfei-threemoonslab: [P2] Fail closed when GitHub will not load the file. GitHub ignores CODEOWNERS files that are 3 MB or larger, but this path parses them normally; a 3 MiB+ file ending with `/shipgate.yaml @security` returns `covered=True` locally even though GitHub requests no owner review. Check the first-precedence candidate's size before reading it, return uncovered rather than falling through, and add boundary tests around the 3 MB limit.

src/agents_shipgate/core/adoption_ladder.py
```diff
@@ -0,0 +1,237 @@
...
+
+    ``environment.target: template`` counts, and counts once for the whole
+    repository: it is a reviewed statement that nothing here holds a
+    credential, which is exactly the authority question every action would
+    otherwise be asked (#410 §G).
+    """
+
+    return bool(
+        any(
+            action.effect is not None or action.risk_tags or action.authority is not None
```
pengfei-threemoonslab: [P2] Base rung advancement on an actually answered semantic dimension. Any nonempty `risk_tags` list currently advances to rung 2, but valid tags such as `network_access` and `customer_data` have no `_TAG_EFFECTS` mapping and produce no reviewed effect claim. A manifest containing only `risk_tags: [network_access]` reaches “Answer on touch” while the corresponding action still reports both `missing_effect_evidence` and `missing_authority_evidence`. Count only tags that resolve an effect answer, or derive this from questionnaire coverage, and test the non-effect tags.

tests/test_public_surface_contract.py
```diff
@@ -3669,15 +3687,40 @@ def test_action_yml_status_output_marked_legacy():
     "docs/report-reading-for-agents.md",
     "docs/passed-verdict-contract.md",
     "docs/INDEX.md",
+    # Carries the lock and lock-diff numbers in prose, which the report/packet
+    # bump used to leave behind: "New exports use 0.7" survived a 0.7 → 0.8
```
pengfei-threemoonslab, line 3691: [P2] Extend the prose guard to the stale claims this bump still leaves behind. The runtime emits report `0.39`, packet `0.16`, and lock `0.8`, but `docs/architecture.md:335` still says new exports use v0.7, while the authoritative `docs/passed-verdict-contract.md` still says report v0.38 (line 3) and current packet v0.15 (line 42). The focused guard passes because it recognizes same-line links and quoted `<kind>_schema_version` fields, not these natural-language/wrapped claims. Update the three statements and cover forms such as `current packet schema vX`, `runtime … report schema vX`, and `new exports use vX`.

PR #399, opened 2026-08-20
fix(discovery): route preview to the changed project and stop reporting a capped walk as complete (#394, #395)

pengfei-threemoonslab:
Closes #394. Closes #395.

Two adoption-path bugs found in the same first-time-user walk of [google/adk-samples#1745](https://github.com/google/adk-samples/pull/1745). Both are cases of a surface asserting more than the walk behind it established.

## #394 — preview routed a monorepo PR to a root `init` that always refuses

`verify --preview` emitted `init --workspace <repo root> --write`, and running that exact command printed *"holds 53 self-contained projects that define agents, and one manifest describes one agent surface."* The resolver had everything it needed: all 17 changed paths shared the prefix `python/agents/smart_closer/`. It returned nothing because the preview call site omitted `evidence_dirs`, and that parameter is what unlocks `WEAK_PROJECT_MARKERS` — the entire boundary a `requirements.txt` beside `agent.py` has. `detect` built and passed it and reported `smart_closer` correctly, so the two commands an adopter runs in sequence answered the same question differently.

`init` at the root is already treated as a known-bad recommendation two branches above the one that emitted it.

**Fix.** A new `signals.weak_marker_evidence_dirs(root, changed_files)` builds the evidence at the preview call site. Rather than re-running whole-workspace detection (~9s on a large monorepo, against a routing path that must stay cheap), it asks the question only where the answer can change anything:

- collect the directories at or above the changed paths;
- keep only those with **no strong marker** (already a project root without evidence) and **one weak marker** (the only case worth reading files for) — in practice a handful, often none;
- answer from the Python files each directory *directly* holds, via `_score_python_signals` — the one place in the module that decides whether a file is an agent file.

Non-recursive is exact, not an approximation: `detect`'s own `evidence_dirs` are the immediate parents of its evidence files, so both ask the same question of the same files. Measured at **60ms on a 1271-file change set** (the resolver it feeds already costs 300ms), and 318ms in a pathological 2000-flat-modules-with-a-requirements-file directory.

On the reported PR the first command an adopter sees goes from an eight-step recovery to `init --workspace python/agents/smart_closer --write`, verified end to end — the emitted command now exits 0 and writes the manifest.

## #395 — a truncated candidate list was reported as complete

```python
if len(candidates) > 1:
    return "ambiguous"
if parse_truncated and project_roots > 1:
    return "unknown"
```

[...]

PR #426, opened 2026-08-26
feat(policy): the rules layer is a pack, chosen once at init (#410 §F)

pengfei-threemoonslab:
Closes the last open section of #410: **§F, policy packs as the rules layer.** Deliberately deferred from increment 3 because it reshapes what shipped check ids fire on; that reshaping is the substance of this PR and is bounded below.

Rebased onto `main` after #423; report schema renumbered to **`0.40`** with `0.39` frozen and hash-pinned.

## The problem

Shipgate has always had a rule saying *a financial write needs approval, an audit log, and idempotency*. It was written **four times** in the engine:

- `BUILTIN_EFFECT_OBLIGATIONS` in `core/action_semantics.py`,
- inline literals in three branches of `_current_action_policy_findings`,
- the effect set inside `subject_requires_approval_review`,
- the effect set inside `subject_requires_confirmation_review`.

The last three were hand-maintained projections of the first. And the rule was not selectable, not named, and stated nowhere the adopter reads — so a scan of four money-moving tools said *"lacks a declared approval policy"* seven times and never once said **what** requires one. Every organisation got Shipgate's opinion, and no organisation could see it.

## The change

`policies.control_pack` names the effect → required-controls rule set for the repository:

```yaml
policies:
  control_pack: default   # default | financial-strict | read-only-agent
```

| Pack | The posture it encodes |
|---|---|
| `default` | Today's rules, exactly. Money, destruction, production operations, code execution, and outbound communication carry controls. |
| `financial-strict` | **Recoverability and record.** Every state change is logged and retry-safe, production operations are reversible, financial writes are confirmed, privileged reads leave a trail. |
| `read-only-agent` | **Permission.** This agent reads; any state change, outbound message, or privileged read is an exception carrying approval and an audit trail. |

The full effect × pack matrix is in [`docs/manifest-v0.1.md`](docs/manifest-v0.1.md#control-packs), rebuilt from the pack objects and compared row by row by a test — a pack is chosen by reading the docs, so a drifted row is a wrong answer given confidently.

`shipgate init --control-pack <id>` writes the selection with a glossary of the alternatives above it (rendered from the pack table, so it cannot go stale), and every recovery route it emits repeats the selection. `init --json` reports `requested`, `selected` (what the manifest **on disk** carries), and `available`. `doctor` reads the answer back. Omitting the key means `default`.

## The properties that make this safe

**A pack can only add obligations.** Every built-in pack requires at least what `default` requires. Asserted at import, with a test that feeds the guard a weakened pack; and pinned by a **real scan over every (pack × effect) pair** plus a whole-finding-set comparison, on synthetic workspaces *and* on all five shipped samples.

**A rule nobody can satisfy is a trap, not a gate.** All 27 (pack × effect) pairs declare exactly what the finding tells them to declare and require every control family to go quiet. `confirmation.required` is the one control not writable on the action row, which is exactly where an unsatisfiable instruction would hide (#399).

**A pack decides which findings fire, never what a declaration means.** The obligation lattice `declaration_covers` reads stays the built-in table, so a pack requiring identical controls for two effects still cannot let a declaration of one discharge the other (#413).

**A pack move is a gate weakening, and the comparator sees it.** Every other field in the effective-policy snapshot answers *does the same finding still block?*; a control pack answers *does the same action still produce the finding?*. `effective_policy.control_pack` publishes the pack in force and `SHIP-VERIFY-POLICY-WEAKENED` gains `kind: control_pack_weakened` — one finding per pack move (the `fail_on_loosened` shape), carrying `removed_controls[] = {effect, controls}`. A base snapshot with no pack is compared *as* `default`, which puts the add-only invariant in the comparison rather than beside it; a base naming a pack this build cannot resolve routes to `SHIP-VERIFY-POLICY-BASE-ABSENT` with `kind: control_pack_unrecognized`, because "cannot compare" is not "no weakening".

**Identity follows the rules.** `run_id` hashes the pack's id, version, and canonical obligations, so two manifests enforcing different policy are different runs even where neither produces a control finding — and a release that changes what `default` requires moves it too.

## What the adopter sees

`scan` stdout and `report.md` name the rule that wanted each missing control, **once per rule** instead of once per tool:

```
## Control Pack

`default` — Shipgate default controls v1. Shipgate's built-in requirements: …

- external communication requires confirmation policy and safeguards.audit_log — 2 actions short
- financial write requires approval.required, safeguards.audit_log, and safeguards.idempotency — 1 action short
```

Both surfaces render **one** projection, so they cannot report different counts from one report — including for a suppressed mandatory control, which the release decision keeps blocking and the section therefore keeps explaining. The tool-level `SHIP-POLICY-*` findings carry the pack but are not rule rows: a rule row is about an *action*, so the discriminator is an `action_id`.

## Surface discipline

- **No new CLI command**; `scan`, `init`, `doctor`, `report.md`, and the verify comparator extended.
- **No new check ids.** Effects with no control check of their own (`write`, `privileged_data_access`, `identity_access`) report a pack obligation through `SHIP-ACTION-POLICY-VIOLATION` at `high`, naming the rule in `evidence.policy_id` as `control-pack:<effects>` — no pack name, because `policy_id` is a fingerprint input and two packs asking the same thing state one rule. Effects obliging the same controls are **one rule and one finding**.
- Those obligations are **mandatory current-surface controls**: a `checks.ignore` entry records the exception without waiving the blocker. That is decided from `evidence.control_pack`, which only the engine writes, and the `control-pack:` prefix is reserved against user-authored `action_surface.policies[].id` at manifest load — the way `SHIP-` already is for check ids.
- `SHIP-ACTION-POLICY-VIOLATION`'s built-in title says "without required controls" rather than "without approval" — under a stricter pack that was telling an action which *has* approval that it does not (#364, one field over).

## Compatibility

Omitting `policies.control_pack` is today's behaviour: the shipped samples' findings, severities, `blocks_release`, and release decisions are unchanged.

- Report schema `0.39 → 0.40` (additive; `0.39` frozen, hash-pinned, read forward) for `effective_policy.control_pack`. Full bump checklist applied.
- `evidence.control_pack` and `evidence.control_effects` are excluded from the finding fingerprint, so baselines recorded before the fields keep matching.
- Switching to a *stricter* pack re-fingerprints a control finding whose required set grew, re-opening a baseline entry that accepted the narrower gap. A move between packs that require the same controls re-opens nothing.
- An older CLI *rejects* a manifest carrying the field (strict loader, routable exit 2) rather than ignoring a release rule. `init --json`'s `control_pack.available` is the capability probe.

## Verification

- Full suite green (`pytest -n auto -m "not perf"` plus the two runs CI splits out), `ruff` clean, `generate_schemas.py --check` clean, `self-check` ready.
- `tests/test_control_packs.py` (209 cases).
- **Twenty-one perturbations across two review rounds; every fix has a guard that fails when the fix is reverted.** Four perturbations returned zero failures on the first pass — the tool-level checks' pack-awareness was unproven, the console line was asserted only through its shared projection, the rule-row discriminator was an unreachable check-id branch, and the missing-base rule could not be told from skipping it. All four are now guarded.
- Measured: `support_refund_agent` goes `default` 5 blockers → `financial-strict` 11 → `read-only-agent` 12 with nothing dropped at any step; `verify --base` runs clean under every pack, and a pack downgrade lands `review_required` with `policy_weakened: true`.

## Relationship to #423

#423 added `environment.target: template`. The RFC's §F list mentions a fourth `template` choice; implemented as a *control pack* it would be a repository-wide claim that **subtracts** obligations — exactly the fail-open increment 4's own perturbation sweep found. That claim belongs to `environment.target`, where #423 pairs it with the verdict cap that makes it honest, so it is not duplicated here.


pengfei-threemoonslab:
Review complete — three rounds, all findings addressed on this branch. CI green on `0fb5c403`.

**Round 1 — perturbation sweep (14 perturbations, 3 returned zero failures).** Each zero was a real gap:
- the tool-level `SHIP-POLICY-*` checks' pack-awareness was unproven (reverting them to their literal effect sets broke nothing);
- the console line was asserted only through the projection it shares with `report.md`, in a test whose name claimed both surfaces;
- the rule-row discriminator was an unreachable check-id branch — a guard scoped to one id shape, vacuous for the next check that carries a pack. It is now an `action_id`, which is the structural reason it was excluding those rows.

Also added: all 27 (pack × effect) pairs declare exactly what the finding tells them to declare and require every control family to go quiet, because a rule nobody can satisfy is a trap rather than a gate — and `confirmation.required`, the one control not writable on the action row, is exactly where an unsatisfiable instruction would hide.

**Round 2 — a fail-open this branch opened.** `policies.control_pack` can *weaken* the gate, and `EffectivePolicy` — the snapshot `SHIP-VERIFY-POLICY-WEAKENED` compares base against head — had no pack field. Every other field there answers *does the same finding still block?*; a control pack answers *does the same action still produce the finding?*. Fixed, which forced the report-schema bump this PR had deferred (`0.38 → 0.39`, full checklist, v0.38 frozen and hash-pinned) — and carried the auditability the deferral was also postponing.

Same round: `init --json`'s `control_pack.selected` reported `--control-pack` even on `skipped_existing`, describing a file the run did not write (#399 one field over). It now follows `_manifest_placeholders`' doctrine — the manifest on disk is the authority — with `requested` beside it.

**Round 3 — copy and a fourth zero.** The weakening emitter first produced one finding *per effect*: one changed line, eight `high` findings repeating one sentence, which is the shape this issue exists to remove. Now one edit is one finding. And "resolve a `None` base pack to `default` rather than skip it" perturbed to zero failures — no pair of today's packs can distinguish the two, so the test now registers a weaker-than-default pack and compares through it.

**Not fixed here, stated instead:** the report-schema collision with #423. Both branches land on `0.39`; whichever merges second renumbers to `0.40`. There is no other overlap.

pengfei-threemoonslab:
## Expert review — changes requested

I reviewed head `0fb5c403` in three independent passes: policy invariants, CLI/schema compatibility, and identity/edge-case testing. I reproduced both P1 issues locally.

### Blocking correctness and contract issues

1. **[P1] Preserve the selected `--control-pack` in every `init` recovery route.**

   `src/agents_shipgate/cli/_register_init.py` registers the option, but `_requested_setup_flags()` and the dry-run/scoped recovery route builder do not carry it forward. A dry run with `--control-pack financial-strict --json` emits a machine-readable `next_action` without the pack; executing that exact action writes `control_pack: default`. Different requested packs also receive the same route `input_id`. The invalid-pack recovery command separately drops `--workspace`, `--write`, and `--json`.

   This breaks the agent-mode promise that emitted recovery argv faithfully completes the requested setup. Please thread the selected pack through all rerun/candidate argv and route identity, with a dry-run → execute regression test plus unresolved-scope/truncation coverage.

2. **[P1] Include the resolved control pack in `run_id`.**

   `src/agents_shipgate/cli/scan/final_report.py` computes the run identity before `effective_policy` is assigned in `effective_policy.py`. Two otherwise identical manifests that select `default` and `financial-strict` can therefore produce the same `run_id` (`agents_shipgate_9c59190903267771` in my reproducer), even though the reports enforce different policy. This also contradicts `STABILITY.md`, which states that the pack participates in run identity.

   Hash the resolved pack ID/version or its canonical obligations before finalizing `run_id`, and add a clean-surface pack-switch test where neither pack emits a finding.

### Additional correctness issues

3. **[P2] Fail closed when the base report names an unknown pack.** `_weakened_pack_rules()` currently returns an empty delta when either pack is unknown. A forward-compatible base report such as `future-strict → default` consequently loses the specific `control_pack_weakened` / `policy_weakened` evidence. Route an unknown base pack through the existing base-unavailable/fail-safe path instead of treating it as no weakening.

4. **[P2] Do not invent high-impact effects and controls in the summary.** The summary path in `control_packs.py` returns all `HIGH_IMPACT_EFFECTS` for any built-in high-impact finding. A `financial-strict` finding caused only by `code_execution` is reported as both `code_execution` and `production_operation`, and gains rollback even though the actual rule only requires approval plus audit. Stamp/read the matched effects and controls rather than expanding to the entire category.

5. **[P2] Keep suppressed mandatory pack blockers visible in `control_rule_summaries`.** The summary builder skips every suppressed finding, while release-decision logic correctly keeps recognized mandatory control findings blocking. The resulting report can be blocked without a Control Pack section or console line explaining the active blocker. Share the mandatory-control predicate with release-decision logic or include these suppressed-but-nonwaivable controls explicitly.

6. **[P2] Do not infer engine provenance from a policy-ID string alone.** `is_control_pack_policy_id()` accepts the exact engine-minted grammar, but user-authored `action_surface.policies[].id` is not prevented from colliding with it. A user policy such as `control-pack:read-only-agent:write` can become unexpectedly nonwaivable—and suppression can turn it into a blocker. Require canonical `evidence.control_pack` provenance, or reserve and reject this prefix for user IDs.

7. **[P2] Stabilize pack-only fingerprints across equivalent obligations.** `control_pack` is intentionally excluded from the fingerprint, but `policy_id` still embeds the pack name. The same missing control for the same action therefore churns fingerprints when switching between packs with equivalent rules. Base the fingerprint on canonical obligation identity, not the display/source pack ID.

### Branch state

The branch now conflicts with `main`. PR #423 has already taken report schema `0.39`, so this PR needs the anticipated rebase/conflict resolution and schema renumber to `0.40` before re-review.

### Verification performed

- `tests/test_control_packs.py`: all 164 tests passed.
- Schema generation check and focused Ruff checks passed.
- The full non-performance suite completed with code tests passing. Eight packaging cases initially could not fetch the isolated-build dependency in the sandbox; all 8 passed when rerun with network access.
- The full PR-diff Shipgate receipt reproduced as valid. Its control state was `review_publishable`, with 9 trust-root human-review items and 0 static blockers; merge/completion remained human-controlled.

I am not approving this revision because the two P1 bugs violate the documented agent-control and report-identity contracts. Please ping for re-review after the fixes and rebase.

pengfei-threemoonslab:
## Review addressed — all seven, plus the rebase

Rebased onto `main` (merge commit `23c06ab8`) and renumbered to report schema **`0.40`**, with `0.39` frozen and hash-pinned. `doctor` and `inspect_sources` now carry both this PR's control pack and #423's adoption rung; every derived artifact was regenerated from the merged tree rather than merged textually.

### P1 — both reproduced and fixed

**1. `--control-pack` in every recovery route.** `_requested_setup_flags` now carries it, so the dry-run advance, the unresolved-scope candidates, and the scoped-refusal rerun all repeat it — and route identity follows, because the identity already hashes the selected action. Verified by running the emitted argv rather than a reconstruction of it: a dry run at `financial-strict` now hands back a command that writes `financial-strict`, and the three packs produce three distinct `input_id`s.

It is repeated **only when it is not the default**, which is what makes the recovery faithful rather than merely verbose: the manifest written is byte-identical either way, while repeating it unconditionally would rewrite every existing route string — and every route identity — to say what the previous string already meant. This rests on "omitting the key means `default`", which `STABILITY.md` pins.

Both early-validation routes are fixed, not just mine: the invalid `--agent-instructions` selector beside it dropped the same three flags. One `_recovery_command` builds both, because two routes spelling one recovery two ways is how one of them stays wrong.

**2. `run_id` covers the pack.** `_run_id` hashes the pack's id, version, **and canonical obligations**, resolved through the same single reader every other consumer uses. Obligations rather than the name alone, so a release that changes what `default` requires moves the identity too. Your reproducer is now a test: a workspace clean under all three packs yields three distinct `run_id`s.

### P2 — all five

3. **Unknown base pack fails closed.** Routed to `SHIP-VERIFY-POLICY-BASE-ABSENT` with `kind: control_pack_unrecognized` — the reason code whose whole job is saying the comparison could not be made, under the same touched-policy-surface guard the no-base route uses.
4. **No invented effects or controls.** Findings stamp `evidence.control_effects`; nothing is reconstructed from an id. A `code_execution`-only action under `financial-strict` reports `code_execution` alone and does not gain rollback.
5. **Suppressed mandatory blockers stay explained.** `is_mandatory_current_control` moved to `core.control_packs` and both the release decision and the summary read it, so a report cannot say BLOCKED and name nothing that blocks it.
6. **Provenance is the stamp, not the string.** Non-waivability is decided from `evidence.control_pack`, which only the engine writes — and `action_surface.policies[].id` now rejects the reserved `control-pack:` prefix at manifest load, the way `SHIP-` already is for check ids. Both layers: the validator is the guarantee, the predicate is what a report loaded from disk is judged on.
7. **Equivalent rules keep one fingerprint.** The pack-only id is `control-pack:<effects>` with no pack name. Where two packs require *different* controls the `missing` rows still differ, so the fingerprint moves for the reason that actually changed — pinned in both directions.

### Verification

Every fix is guarded by a test that fails when the fix is reverted — seven perturbations, seven distinct failures. `test_control_packs.py` is at 209 cases; the full non-perf suite, the two split CI runs, `ruff`, `generate_schemas.py --check`, and `self-check` are all green, and the end-to-end `verify --base` pack-downgrade run still reports `review_required` with `policy_weakened: true`.

One note on the fingerprint case you raised: `external_communication` looked like a second instance of it but goes through a *dedicated* check whose id never carried the pack name, so that parametrisation was vacuous for the concern. It now names the route it exercises and asserts the finding came from it.

Ready for re-review.

PR #427, opened 2026-08-26
feat(control): an agent may draft what the evidence supports (#410 §D)

pengfei-threemoonslab:
Implements **§D of #410** — the coding-agent loop. Three surfaces, one rule.

The questionnaire (#410 increment 2) tells a person which blanks are owed. It has no way to say that most of them are the scan restating what it just read, and no way for the agent already holding the branch to write those down. On `github-mcp-server#3076` that was a pincer: the agent must not edit `shipgate.yaml`, but its new tool is invisible until someone does.

## What lands

**1. `next_action.kind: "confirm_declarations"`** — `verify` publishes it when the verdict is `insufficient_evidence` and at least one open declaration question is one the scan can answer from its own evidence. It carries the exact `apply-patches` command that writes those answers, every open question tagged `authorable_by`, and the two counts (`agent_authorable` / `human_authorable`) that say whether the printed list is the whole list.

```json
{
  "kind": "confirm_declarations",
  "actor": "coding_agent",
  "command": "agents-shipgate apply-patches --from …/report.json --kinds declare_action --confidence high --apply",
  "expects": "1 declaration(s) written into the manifest and committed to this branch; 1 question(s) then remain for a human.",
  "agent_authorable": 1,
  "human_authorable": 1,
  "questions": [
    {"subject": "send_email [adk_agent]", "dimension": "effect",
     "answer_path": "shipgate.yaml#action_surface.actions[tool='send_email']",
     "authorable_by": "coding_agent"},
    {"subject": "adk_agent [tool_source]", "dimension": "authority",
     "answer_path": "shipgate.yaml#tool_sources[id='adk_agent'].authority",
     "authorable_by": "human"}
  ]
}
```

**2. `authorable_by` tags** on `evidence_gaps[].next_action` and on `declaration_questions.open_questions[]`.

**3. A `declare_action` patch kind** for `apply-patches`, published on the gap row it answers.

## Authorship is decided by content, never by who is running

A row is `authorable_by: "coding_agent"` only where the scan filled **every** blank in its `declaration_template` — an effect, drawn from the closed `ActionEffect` vocabulary, never weaker than any reading it observed (all four properties are already proven by `propose_effect_declaration` from increment 2). Every authority block, override and inventory keeps its `<REVIEW_REQUIRED>` blank and its `"human"` tag, because filling one would be an assertion drawn from an absence.

The rule is enforced **on the model**, not in the one builder that sets it today, so a gap kind added later cannot inherit a licence its author never considered. A question carries the *conjunction* of the tags on the rows it folds: a block whose effect the scan can propose but whose authority it cannot is still a human's edit.

## What the agent gains is a pen, not a decision

- `declare_action` is **outside the default `--kinds`**, so no existing `apply-patches` or `bootstrap` invocation starts writing manifests.
- It writes **only into fields the manifest leaves silent**. A row that already answers one differently, or two rows naming one tool, is refused outright — **exit 5, nothing written**, because an agent that re-ran verify after a silent no-op would loop against an unchanged file forever.
- `requires_human_review` stays `true` on every evidence-gap row, and the route's `permissions` are publish-only (`edit`/`commit`/`push`/`update_pr`; never `merge`/`report_complete`). Writing a declaration touches the trust root, so a person still merges it.
- A weakening written by hand is not blocked at the file — it is answered at the next scan by `declaration_below_inferred_evidence` (#409), which is where the content rule has always lived. `test_a_forged_weakening_still_meets_the_gate` pins exactly that boundary.

## Where the route fires, and why each condition is there

`insufficient_evidence` (the verdict a declaration moves, and the one that says nothing is *wrong*), no blockers, no proven policy weakening, a **working-tree** run (a ref-bound rerun would re-scan the commit the edit is not in yet — the same `_repair_subject_available` precondition the mechanical repair route has always carried), and a report that actually carries the patches the command would apply.

That last one is why declaration patches are emitted on **every** scan rather than under `--suggest-patches`: a published route may not name a step the report it points at does not contain. `--suggest-patches` gates the large per-finding surface; a declaration patch is the machine-applicable form of a `declaration_template` the row already publishes unconditionally.

`trust_root_touched` is deliberately **not** a condition. A PR that already edits the manifest is the normal case here — an agent declaring the tool it just added is exactly the #3076 pincer this dissolves — and the touch is what routes the merge to a human either way.

## Also in this PR, because the route needs them

- **The envelope budget is re-derived, 4 KiB → 6 KiB.** A `confirm_declarations` route carries a question list, which is a third variable component the published 4096 was never derived against; measured, a row renders at ~0.4 KiB. The two variables were traded rather than one let run: the list is capped at **6** rows — already more than the measured worst case produces, since per-*source* authority folding turns github-mcp-server's 117 tools into one question — which keeps the increase to a half-step instead of the near-double a ten-row list would have cost. A real envelope from the walk below measures 4258 bytes. `test_a_declaration_route_fits_the_published_budget` renders the full-cap shape and pins it.
- **The printed prefix is human-owned rows first**, each half keeping the report's ranking. Not a re-ranking of the questionnaire — it decides only which rows survive the cap, and the two halves are not equally droppable: a drafted row is answered by the command whether or not it is printed, while a human-owned row is the thing the agent has to hand a person.
- **A ref-bound run no longer leaks its archive path.** `_project_archived_report_paths` mapped `findings[].patches` only; a declaration patch is the second surface carrying a target, and an unmapped one publishes a temp directory name into receipt-bound evidence. Both surfaces are now enumerated in one function, so a third that starts publishing a patch is a change to it rather than a silent omission. Perturbation-checked: reverting the fix makes the new test print `…/agents-shipgate-verify-head-*/head/shipgate.yaml`.
- **A stale report exits 5 too, not just a conflicting one.** `skipped_drift` writes nothing, and an agent that re-ran verify on a silent exit 0 would come back with the identical route.

## Two fixes that fell out of building it

- **`VerifierArtifact` forbade the route it was supposed to allow.** It demanded that an `agent_action_required` repair route's command equal the fix task's *rerun* command — while `_verifier_control` routes the first `allowed_repairs[].command` and `test_control_next_action_follows_agent_safe_fix_task` has always asserted the two differ. No verify run had reached the mechanical route and built an artifact from it, so nothing caught the contradiction. The invariant now says what it always meant: a route may not name a command the fix task does not authorize.
- **`apply-patches` reformatted the manifest.** ruamel re-emits every sequence at one global indent, so writing one declaration also re-indented every unrelated list and buried the change in whitespace — which defeats the review surface this whole design rests on. Pinned to the style `init` writes (`  - id:`), so a Shipgate-generated manifest round-trips byte-identical apart from the patch.

## Contract / schema

Report `0.38 → 0.39`, packet `0.15 → 0.16`, verifier `0.12 → 0.13`, runtime contract `24 → 25`. `minimum_control_contract_version` stays `21` for the third time and for the same reason: the shared `AgentControl` union is byte-identical — the control holds the step as the `repair` command it truthfully is, and the envelope (stdout-only, never stored) publishes the richer form. Exactly the split `SetupEditAction` uses.

All fields are additive and default to the human reading, so a v0.38 report and a v0.12 verifier read forward unchanged.

## Verification

Full suite green (`-n auto -m "not perf"`, plus the two split runs), `ruff`, `compileall`, `generate_schemas.py --check`, and `shipgate self-check`. 44 new tests across `tests/test_declaration_authoring.py` and `tests/test_declaration_confirmation_route.py`, including the adversarial ones: a forged weakening, a stale report pointed at an answered manifest, two rows sharing a tool name, a tool name carrying newlines and list markers into the trust root, and a sweep asserting no published patch anywhere carries a value a human still owes.

Walked end to end on a real workspace: `verify` → `confirm_declarations` (1 agent, 1 human) → `apply-patches` (a six-line diff, nothing else touched) → `verify` → `blocked`, the gateable verdict declaring the outward communication makes reachable, with the remaining authority question named.

**Perturbation sweep.** Each of the twenty invariants this PR adds was deliberately broken and the tests re-run; every one is caught. Seven returned zero failures on the first pass and each was chased rather than waved through, which is where five of the tests above came from:

| what it found | fix |
|---|---|
| No fixture reached a **non-`insufficient_evidence`** verdict while still owing an agent-authorable question — the case the verdict gate exists for | two fixtures added (`blocked` via an unclosed control gap, `review_required` via an elevated severity), isolating the blocker clause from the verdict clause |
| The **patch-presence** and **agent-authorable-question** guards are unreachable end to end (they are each other's redundancy) | reached directly, by stripping one and then the other off a real report and calling `build_fix_task` |
| The **envelope's command join** was never exercised on a mismatch | a test that perturbs the fix task's command and asserts the envelope falls back to the plain `repair` action |
| `test_a_declaration_route_cannot_be_published_by_setup` **passed for the wrong reason** — an unrelated setup-provenance validator raised first, so a bare `pytest.raises(ValueError)` was green whether or not the route was guarded at all | `input_id` supplied and the message matched; verified by perturbing *both* layers |
| The new `VerifierFixTask` validator had no test | one added |

Three rows still show no single-perturbation failure, and each was confirmed guarded at a second layer by perturbing both together: the builder's gap-kind gate (the row model enforces it), the projection's verify-only check (the envelope variant enforces it), and `decision.decision` beside `merge_verdict`.

## Not in this PR

- §D's **PR-comment declaration diff** (✓ evidence-consistent / ⚠ override rows). Overrides already render there from #409; the ✓ half needs a base-vs-head manifest comparison and a third "declared but unverified" bucket — collapsing that into ✓ is the one misreading the surface exists to prevent. Filed as #428 rather than half-done here.
- The `basis: confirmed:<derivation_id>` pin is §E, in #423. The patch value is a plain mapping, so stamping it there is additive.

## Ordering note

Branched from `main`. #423 (increment 4) is open and currently conflicting against `main`; it bumps the same three schema versions. Whichever lands second re-bumps to `0.40` / `0.17` / `0.14` — mechanical, and #423 needs a rebase regardless.

Closes nothing on its own; advances #410.


pengfei-threemoonslab (commented):
I reviewed `2552f6e` against current `origin/main` (`0d4044b`) and reproduced several blocking issues. The overall control design is thoughtful, but I do not think this is safe to merge yet.

1. **[P1] Receipt-bound artifacts still contain deleted, random archive paths.** `verify` calls `_project_archived_report_paths()` only on the in-memory head report after `run_scan` has already serialized `packet.json`. The later packet rewrite at `src/agents_shipgate/cli/verify/orchestrator.py:3263-3271` updates identity fields only. In a ref-bound repro, `report.json` correctly named the checkout's `shipgate.yaml`, while `packet.json` named `/.../agents-shipgate-verify-head-<random>/head/shipgate.yaml`. Two identical runs produced different packet SHA-256 values and different receipt `artifact_set_id`s solely because the temp directory changed. The base side has the same class of bug: declaration patches are attached unconditionally, but `_prepare_base_report()` clears only `manifest_dir`/generated outputs, so cached base reports and `verification-base-report.json` retain a deleted `/.../agents-shipgate-verify-<random>/base/shipgate.yaml`. Please project/rebuild the packet before receipt hashing, strip non-applicable patches from base evidence (and invalidate the cache epoch if needed), and test packet/base artifacts plus repeated-run determinism.

2. **[P1] Valid same-name tools make the new route unexecutable.** `_declare_action()` at `src/agents_shipgate/cli/apply_patches.py:692-717` matches existing rows by display `tool` alone. A fresh scan of two supported providers both exporting `send_email` emits two valid patches distinguished by `tool_id`/`source_id`. Applying the batch appends the first row, then treats it as the second patch's match and raises `DeclarationConflict`; exit 5 leaves the manifest unchanged. Same-name tools are already supported elsewhere. Match compatible qualified selectors (or append a distinct qualified row) and reserve refusal for genuinely ambiguous/overlapping selectors; add an end-to-end two-provider test.

3. **[P1] An unchanged CRLF manifest can never apply a generated declaration patch.** Generation hashes raw bytes in `checks/patches.py`, but `_apply_one_file()` hashes `path.read_text().encode()` at `src/agents_shipgate/cli/apply_patches.py:500-504`, normalizing CRLF to LF. The untouched target therefore always reports `skipped_drift`; rescanning recreates the same failure. Hash the original bytes during application and preserve the input newline style. A CRLF scan/apply regression test should cover this loop.

4. **[P1] The model does not bind the executable patch to the evidence-derived template.** `EvidenceGapAction._authorship_matches_the_template()` validates the `authorable_by` tag and template sentinel, but never cross-validates `patch.selector`/`patch.declaration`/confidence against that template or `observed_readings`. I confirmed that a `coding_agent` `declare_action_effect` row can validate while its patch writes an authority block, an invalid effect, or an effect weaker than the readings. `fix_task` then authorizes the route merely because some patch exists. This contradicts the stated invariant that the model protects rehydrated payloads and that the agent writes only what evidence supports. Require an exact template split, closed/conservative effect validation, and high confidence before a patch can authorize this route.

5. **[P1] The shipped agent instructions still tell agents to refuse this feature.** `AGENTS.md:194-201` still forbids auto-filling every effect declaration. All shipped `fix-top-finding.md` copies (root, skill, adoption kit, and Claude Code plugin) say every declaration row is manual, "never write the content yourself," and that no evidence-gap row reaches a coding agent. An agent following those authoritative/bundled instructions will ignore `confirm_declarations`. Add the precise exception already present in the updated `SKILL.md`: execute only the current `confirm_declarations` control action for rows explicitly marked `authorable_by=coding_agent`; authority, binding, inventory, override, and human-owned rows remain prohibited. Please add renderer/copy parity coverage.

6. **[P1] Rebase integration with #423 needs fresh schema identities and a human-only drift rule.** Current `main` already published report `0.39`, packet `0.16`, and verifier `0.13` with different wire additions (`basis`, `policy_eligible`, and `declaration_drift`), so combining this PR under those identifiers would mutate frozen schemas. Preserve main's artifacts and advance these three families (normally to `0.40` / `0.17` / `0.14`) before regeneration; contract v25 itself is still available. Also, main's `declaration_drift` uses `kind=declare_action_effect` with a complete `{effect,basis}` template. The generic predicate at `src/agents_shipgate/ci/release_decision.py:1896-1900` would therefore mark it coding-agent-authorable after rebase, contrary to main's explicit human-review rule; applying it would then fail because `_declare_action()` refuses to replace the stale `basis`, creating an exit-5 loop. Keep drift human-owned and test both drift and normal basis-carrying drafts.

Validation performed: the targeted declaration/control tests, lint, generated-schema check, static-only adapter suite, packaging suite, and self-check passed. The broader non-performance suite had no implementation failures; its packaging fixture initially could not download an isolated build dependency in the sandbox and passed when rerun with network access. Shipgate itself reports `control_state=review_publishable`, `decision=review_required`, and `permissions.update_pr=true`; it does not authorize merge or completion. The branch also currently conflicts with `main`.

pengfei-threemoonslab:
Thank you — this was a good review. All six are addressed on `47899fdb`, rebased onto current `main` (`d228c9ee`, so #423 **and** #426). Two of your findings turned out to be wider than reported and one narrower; details below.

## 6 — rebase and the drift rule

Rebased. `main` had moved twice, so the identifiers are `0.41` / `0.17` / `0.14`, not `0.40` / `0.16` / `0.13`. Main's `report-schema.v0.40.json`, `packet-schema.v0.16.json` and `verifier-schema.v0.13.json` are byte-unchanged and now pinned by `_PUBLISHED_SCHEMA_SHA256`; contract v25 was still free.

The drift rule is in, and you were right about the mechanism: main's `declaration_drift` is spelled `declare_action_effect` with a complete `{effect, basis}` template, so every input the content rule reads says "drafted". The distinction only exists on the **gap** kind, so that is where the gate went — `HUMAN_ONLY_GAP_KINDS`, checked in the builder *and* on `EvidenceGap` itself, because "an agent re-stamping the pin closes the request the row exists to make" is a rule about the question, not about the repair. Covered end to end by a scan of a manifest with a stale `basis:` (`test_a_real_drift_row_is_published_human_owned`) plus the model-level rejection.

## 1 — receipt-bound artifacts: wider than reported, and fixed at the root

Reproduced, and there was a **third** site you did not list: `report.sarif` also embeds the gap row, and it was the one actually moving `artifact_set_id` run to run.

That made stripping the wrong fix — three consumers today, unbounded tomorrow, each having to remember. The root cause is that the patch carried an **absolute** path at all. `DeclareActionPatch.target_path` is now relative to `manifest_dir`:

- no temp archive path in the packet, the SARIF, or a cached base report — the field reads `shipgate.yaml` everywhere;
- deterministic across runs and machines;
- containment becomes structural rather than a check — `apply-patches` resolves it *under* `manifest_dir`, so it can no longer be handed a path that escapes;
- `_project_archived_report_paths` needs no second surface, which is what its own docstring warned about.

Base evidence additionally drops the patch outright (a base report describes a commit nobody is editing), and `BASE_CACHE_KEY_EPOCH` → `4` so no entry written mid-flight is served back. `test_a_ref_bound_run_never_publishes_an_archive_path` asserts no `agents-shipgate-verify-` substring survives into any of the three artifacts.

**One correction.** Repeated-run determinism: my first measurement reproduced your result, then I found my own methodology was the variance — I had run into two different `--out` directories, which legitimately changes `generated_reports`. Re-measured into a fixed directory: unmodified `main` is stable, my branch was **not**, and the sole cause was the SARIF row. It is stable now. So the packet and SARIF halves were mine; had `artifact_set_id` been unstable on `main` too I would have said so rather than claim a fix.

## 2 — same-name tools

Confirmed, and it was my rule that caused it: matching on the display name alone meant the first patch appended a row that the second then read as its own mismatched match. Rows are now matched on the qualifiers they *share* — a row spelling a different `tool_id`/`source_id` is a different capability and is skipped, not refused; a row spelling `tool:` alone is the same action less precisely named and is written into (without back-filling the qualifiers, which are this scan's identity rather than a correction to the reviewer's spelling). Refusal is reserved for genuinely ambiguous cases: two *equally compatible* rows.

Two tests: the applier on a two-patch batch, and generator→applier (`attach_declaration_patches` over two same-named rows, then applied). I did try the full two-provider scan fixture you asked for and stopped — getting two same-named tools past the binding graph turned into fixture engineering rather than a test of this defect, and the generator→applier test covers the same path. Say the word if you want the scan-level version anyway.

## 3 — CRLF

Confirmed. `_apply_one_file` now hashes `path.read_bytes()` — what the generator hashes — and re-emits the newline style it found, so the round trip does not rewrite every line ending. `test_a_crlf_manifest_can_answer_its_own_question` pins both halves. This was pre-existing for the pointer kinds too (silent `skipped_drift`); the declaration route only made it fatal.

## 4 — binding the patch to the template

Fixed on the model. `patch` must now be *exactly* `declaration_template`, split into selector and declared halves, at high confidence, with the effect from the closed vocabulary and no identifying key smuggled into the written half. Seven parametrised cases, each of which validated before.

One boundary I want to state rather than imply, because I could not close it here: "not weaker than `observed_readings`" needs the effect-covering relation, which lives in `core.semantic_assessment` and may not be imported by a schema module. What the model checks itself is the one direction where accepting the weaker answer *loses* safety — a declared `read` against a non-`read` observation (#357). The full covering property is enforced where the template is produced. If you would rather have it on the model, the honest options are lifting the rank tables into `schemas/` or moving the check to a post-construction validator in `ci/`; I did not want to pick that on my own.

## 5 — shipped instructions

Fixed in `AGENTS.md` and all four copies of `fix-top-finding.md` (via the `adoption-kits/` renderer source, with render hashes and `prior_render_sha256` updated). The exception is stated narrowly — execute only the current `confirm_declarations` action, only for rows tagged `authorable_by: "coding_agent"` — and every prohibition around it is kept, with `declaration_drift` named explicitly.

Parity coverage added, and it immediately found something: comparing the kit source byte-for-byte was **my** error, since it is a template with `{{ … }}` placeholders. The test now pins the three *rendered* copies identical to each other and asserts all four state the exception plus its prohibitions.

## Verification

Full suite, both CI split runs, `ruff`, `generate_schemas.py --check`, `self-check` — all green. Shipgate on itself: `review_publishable` / `review_required`, no merge authority.

The perturbation sweep is now 27 invariants, all caught. It earned its keep again: after writing the fixes for points 4 and 6 it returned **seven** zero-failure rows — every guard I had just added, none of them yet tested. Those seven tests exist now. The last one to fall was the builder's drift gate, which needed a real scan-produced drift row to distinguish "unreachable" from "unguarded".

PR #435, opened 2026-08-27
feat(binding): a published tool surface is one declaration, not one row per tool (#432)

### LINKED ISSUE #432: An MCP server needs 116 per-tool selectors to bind its own published surface; the two shorter spellings both dead-end
## Summary

Binding an MCP server's own published surface requires naming **every tool individually** in `agent_bindings.declarations[].tools`. For `github/github-mcp-server` that is 116 selector rows to state a fact that is structurally true of the source: a server publishes its tools, and every published tool is callable by any client connected to it.

The two shorter spellings a reader would reach for both fail, and neither failure explains what to do instead. Until one of them is written, `reachable_tools: 0`, nothing is evaluated, and the verdict is `insufficient_evidence` regardless of what the PR did.

This is the same shape as [#410 increment 3](https://github.com/ThreeMoonsLab/agents-shipgate/issues/410) — *"authority follows credentials, not functions"* — applied to the binding dimension. Asking per-tool for something that is one fact about the source is what breeds copy-paste, and 116 hand-written rows is where an adopter stops.

## Reproduced, three spellings

`github/github-mcp-server` at [#3020](https://github.com/github/github-mcp-server/pull/3020), 116 tools merged from the repository's checked-in `__toolsnaps__/*.snap` into one `type: mcp` source. CLI `0.16.0b7` at `9233c6b8`.

**1. No `agent_bindings` at all** — the manifest states source-level `authority` (increment 3) and a `control_pack`, nothing else:

```json
"binding_coverage": {"total_catalog_tools": 116, "reachable_tools": 0, "unbound_tools": 116,
                     "pass_eligible": false, "gap_count": 2,
                     "reason_counts": {"ambiguous_root_agent": 1, "missing_binding_evidence": 1}}
```

**2. `agent_bindings.root` naming the source:**

```yaml
agent_bindings:
  root:
    object: github_mcp
    source_id: github_mcp
```

```
ambiguous_root_agent | agent binding graph | "No root agent matched the configured selector."
reachable_tools: 0
```

`root.object` is documented as *"Exact selector for the statically reviewed root agent"* — a code object the scanner observed. A JSON tool export produces no code objects, so this field cannot be satisfied for a pure MCP source by construction.

**3. A wildcard selector:**

```yaml
tools:
  - {tool: "*", source_id: github_mcp}
```

```
unresolved_bound_tool: 1
reachable_tools: 0
```

`"*"` is matched as a literal tool name. `BoundToolSelectorConfig` requires a `tool` string and offers no wildcard or source-only form.

The only route that works is the one `samples/ai_generated_refund_pr` demonstrates — `agent: root, complete: true` with an explicit `tools:` list. That fixture has two tools. This server has 116.

## Why "unbound" is not a meaningful state here

For an agent, binding is real information: a catalog may hold 63 OpenAPI operations of which the agent wires 5, and #385 drew the boundary there deliberately. For a **tool server**, there is no such gap. The repository under review *is* the tool surface; anything in its published `tools/list` is callable by any client that connects. There is no root agent, and there is nothing to select between.

That reasoning is already accepted elsewhere in the codebase — `schemas/exclusions.py` records it as the motivating case for the exclusion ledger:

> `github/github-mcp-server#3076` adds `delete_repository` (`destructiveHint: true`) to a 117-tool MCP server. The scan computed `unbound_tools: 1` and reported `gap_count: 0` […]

The ledger fixed the *accounting*. The declaration required to make those tools reachable in the first place is still per-tool.

## What is working, for the record

This run also validated the #403 discriminator on a repository it had never seen. All 116 tools were excluded (nothing bound), and the ledger gated exactly **one**:

```
surface_exclusions: {"total": 116, "gated": 1, "gap_backed": 1}
"1 of 2 evidence gap(s) are new in this diff"
```

`find_duplicate`, the tool this PR adds — with the 115 pre-existing correctly spared, `blockers: 0`, and `control.state: review_publishable` rather than a stop. Proportionate on a benign change. The problem is only that nothing downstream could run.

## Options

**A. A source-level binding statement.** `tool_sources[].binding: complete` (or equivalent) — one reviewed human declaration that this source's published surface is the surface under review. Mirrors `tool_sources[].authority` from increment 3 exactly: same place, same shape, same argument.

**B. A source selector in `BoundToolSelectorConfig`.** Allow `{source_id: github_mcp}` with `tool` omitted, meaning every tool this source contributes. Smaller schema change; keeps the declaration inside `agent_bindings` where a reader already looks.

**C. Make `root.object` satisfiable for artifact sources**, or fail with a message that says it cannot be — currently it reports `ambiguous_root_agent` as though a better selector existed.

Recommend **A**, for the symmetry with authority. Either way, C's error text is worth fixing independently: two separate walks reached for `root.object` first, and the message sent both of them looking for a selector that does not exist.

Whichever is chosen, it must stay a **human declaration** — inferring "the source binds everything" from source content is the #268 attack, and the point here is to make the statement one line rather than to remove it.

## Acceptance

- [ ] A 116-tool MCP source reaches `reachable_tools: 116` from one reviewed declaration.
- [ ] An agent that binds a *subset* of a catalog still reports the unbound remainder (#385 regression).
- [ ] `root.object` against an artifact-only source either resolves or explains that it cannot, rather than reporting an ambiguous root.
- [ ] Untrusted source content cannot produce the binding statement (#268).

---

Found 2026-08-24 walking `github-mcp-server#3020`; the same wall was hit on `#3076` in an earlier walk, where it was worked around by generating the 116-row list programmatically.

pengfei-threemoonslab:
Closes #432.

Binding an MCP server's own surface required naming every tool individually under `agent_bindings.declarations[].tools` — 116 selector rows for `github/github-mcp-server` to state a fact that is structurally true of the source, and the point at which an adopter stops. The two shorter spellings a reader reaches for both dead-ended: `agent_bindings.root` naming the source reported `ambiguous_root_agent` as though a better selector existed, and a `"*"` selector was matched as a literal tool name. Until one of them was written `reachable_tools` was `0`, nothing downstream ran, and the verdict was `insufficient_evidence` whatever the change did.

## The declaration

```yaml
tool_sources:
  - id: github_mcp
    type: mcp
    path: mcp/tools.json
    binding:
      complete: true
      reason: >-
        This repository is the server; every tool in its published tools/list
        is callable by any client that connects to it.
```

Option **A** from the issue, for the symmetry it argues for. It sits where `tool_sources[].authority` sits and makes the same shape of argument: binding, like authority, is a fact about the source rather than about each function it exposes. `complete` is spelled exactly as `agent_bindings.declarations[].complete` and means the same thing — the block's presence is the closed-world claim, so `true` is the only value; `reason` records how the surface was reviewed.

For an **agent** that argument does not hold — a catalog may hold 63 OpenAPI operations of which the agent wires 5, and #385 drew that boundary deliberately — so the block is opt-in per source and changes nothing where it is not written.

## Additive and widening

The block can only move tools *into* the analysed surface, where every check then judges them; it never moves one out. That is the safety argument for allowing it on any source type, and it is asserted as a comparison rather than as a count.

It also settles the #3076 case from the other end. On a real `verify --base` run over a declared 20-tool MCP server whose head adds `repo.delete_repository` (`destructiveHint: true`), the new tool is now reachable, judged, and **blocks**:

```
decision: blocked
blockers: SHIP-POLICY-APPROVAL-MISSING, SHIP-ACTION-DESTRUCTIVE-ROLLBACK-MISSING
binding_coverage: {total: 21, reachable: 21, unbound: 0, pass_eligible: true, gap_count: 0}
```

Before, it left the surface before any check could see it and was reported as an evidence gap by the #403 ledger.

## The dead end is a route

When nothing observed an agent object, the root gap no longer prescribes a selector that cannot exist:

```
why:  No agent object was observed anywhere in this scan, so no
      agent_bindings.root selector can match one. State the reviewed surface in
      shipgate.yaml instead: tool_sources[].binding for a source whose published
      tools are the surface under review, or one agent_bindings.declarations
      entry naming agent: root.
path: shipgate.yaml#tool_sources[].binding
```

and `suggested-declarations.yaml` carries one block per configured source with the ids read off the surface and `<REVIEW_REQUIRED>` for both halves of the judgement. A catalog whose tools come from a per-scan adapter has no row to declare on, so it keeps the closed-world `agent_bindings.declarations` route instead of being handed a remedy the schema rejects — the `path` an agent routes on and the block a human pastes are read from one helper so they cannot disagree. Once a source is declared, `root: {object: <id>, source_id: <id>}` resolves to it.

## It stays a human declaration

Inferring "this source binds everything" from the source's own content is the #268 attack. The block is refused by the agent-authored `tool_sources` proposal allowlist, is a human-owned placeholder so `doctor` will not publish an executable edit for it, and an MCP export that spells `binding` in its own annotations reaches exactly `0` bound tools. A reviewed declaration that binds no tool fails closed with `missing_binding_evidence` naming the source, rather than proving a binding graph over an empty analysed surface.

## Acceptance

- [x] A 116-tool MCP source reaches `reachable_tools: 116` from one reviewed declaration.
- [x] An agent that binds a subset of a catalog still reports the unbound remainder (#385 regression).
- [x] `root.object` against an artifact-only source either resolves or explains that it cannot, rather than reporting an ambiguous root.
- [x] Untrusted source content cannot produce the binding statement (#268).

## Notes for review

- **No published schema changes.** The manifest field is additive; nothing in `report.py` moved, so there is no `report_schema_version` bump. A CLI that predates the field rejects the key with a routable `ConfigError` (exit 2) rather than ignoring a reviewed claim.
- **A surface node is not an agent.** It carries a disjoint id namespace and is excluded from name resolution, so a `tool_sources[].id` that happens to share a name with an observed agent cannot rewire that name — the #385 "a prescribed fix whose own side effect blocks it" class, guarded by a test.
- **One node per declared source, not one shared node.** A repository may publish two servers, and electing one the root of the other would be a claim nobody made; `_reachable_agents` seeds every declared surface alongside the root.

Test plan: `tests/test_source_binding.py` (26 cases, organised by the four acceptance properties), full suite green (`pytest -n auto -m "not perf" --ignore=tests/test_adapter_static_only.py`, plus the perf and static-only splits), `ruff check .` clean, and the real `verify --base` run quoted above.


pengfei-threemoonslab:
## Review rounds

Three self-review passes since the first push. What they found, and the method that found it.

### Round 1 — a `path` and its scaffolded block contradicting each other

A reviewed source binding that reaches no tool is raised against its **surface node** — which is the graph's *root* whenever it is the only declared surface. That put it inside `_binding_declaration_template`'s root-scoped branch, so a gap whose own `path` named `shipgate.yaml#tool_sources[].binding` shipped an `agent_bindings.declarations` block listing an *unrelated* source's tools under an `agent:` naming a tool source, which resolves to nothing:

```
KIND missing_binding_evidence | PATH shipgate.yaml#tool_sources[].binding
  TEMPLATE {"agent_bindings": {"declarations": [{"agent": "empty_source", …}]}}
```

Asserting the `path` alone passed. Asserting the template alone passed. They are now checked as a **pair**, by a rule over every shape that raises a binding gap rather than at the site that was wrong.

Also in this round: `tools: [{tool: "*", …}]` — the other spelling the issue reported — now names the statement it was reaching for, at the binding site only (a pattern in an `action_surface` row is a different misunderstanding with a different repair). And `root.object` searched every node, so adding a `binding` block to a source whose id happened to match an observed agent turned a selector that used to resolve into `ambiguous_root_agent`; an observed agent now wins the name outright.

### Round 2 — one sentence wrong three ways

The row a `tool_sources[].binding` declaration raises read:

> the agent's tool bindings are unproven (empty_source [empty_source])

- **It spoke in the voice of an action** about a subject that has no agent. `_GAP_PHRASE` is keyed by kind alone; increment 3 already built the mechanism (`subject_kind: tool_source` selecting a source-scoped phrase), and the row now carries that kind.
- **The subject repeated its own qualifier.** `agent_subject` qualifies a name with the source it was read from, and a declared surface is named by its own source id. A qualifier that *is* the name separates nothing.
- **It published a source id in a field two joins read as a tool id.** `surface_exclusions` and `semantic_consistency` join a binding gap's `subject_id` against canonical tool ids and neither checked the namespace — the exact hazard the schema's own comment names.

The namespace guard is proven by the collision **constructed rather than argued about**: name the declared-but-empty source after the canonical id of the tool that is genuinely excluded. Without the guard:

```
accounting: evidence_gap | accounted_by: tool_v2_6f6f9cbc… | gated: 1 gap_backed: 1
```

The ledger reporting gating it does not have — the #404 class. With it, `not_claimed` / `gated: 0`.

### Round 3

A punctuation seam in the pattern hint, and pinning `status: declared` beside `pass_eligible: false` as a decision rather than an accident.

### Perturbation sweep

Seventeen mechanisms disabled one at a time; every one fails at least one test.

| | mechanism | failing |
|---|---|---|
| 1 | surface id namespace | 1 |
| 2 | surfaces excluded from name resolution | 1 |
| 3 | surfaces seed reachability | 5 |
| 4 | empty declaration fails closed | 4 |
| 5 | root heuristics ignore surfaces | 1 |
| 5b | observed agent wins the selector | 1 |
| 6 | dead-end message | 1 |
| 7 | pattern hint | 1 |
| 8 | gap routes to the new block | 4 |
| 9 | declarable-source discriminator | 1 |
| 10 | no template for a source-binding issue | 2 |
| 11 | human-owned placeholder | 1 |
| 12 | `rooted` pass_eligible | 1 |
| 13 | subject qualifier drop | 2 |
| 14 | `subject_kind: tool_source` | 1 |
| 15 | tool-id namespace guard | 1 |
| 16 | source-scoped phrase | 1 |

**Two guards were vacuous on the first attempt, for opposite reasons.** The name-collision test could not reach the code it was written for — its fixture collided on a name whose source contributed no agents at all, so the dedupe it was guarding never ran; it now uses a Google ADK source named `front_desk` holding an agent named `front_desk`, whose observations agree on *both* keys the graph dedupes on. And the namespace test re-implemented the filter it was checking instead of observing its product. Neither showed up in a green suite; only the sweep did.

### Safety properties probed directly

- **A declaration cannot launder an unenumerable surface.** An ADK source with a dynamic `MCPToolset` stays `partial` / `pass_eligible: false` with the block written — "everything this source publishes is under review" is not an answer to "nobody can read what it publishes".
- **What it closes, it replaces with obligations.** `wire_money` on a sub-agent the root cannot reach goes from an excluded subject nothing judges to a judged action owing effect and authority, and the verdict stays `insufficient_evidence`.
- **A real `verify --base` run** over a declared 20-tool MCP server whose head adds `repo.delete_repository` blocks on `SHIP-POLICY-APPROVAL-MISSING` + `SHIP-ACTION-DESTRUCTIVE-ROLLBACK-MISSING`.

### Noticed in passing, not fixed here

`report/markdown.py` prints `Root agent: agent\_v1:7205d836…` — a derived id where a name belongs, in every shipped sample. It is pre-existing and unchanged by this diff, and the vocabulary sweep misses it only because Markdown escaping breaks the pattern it matches on. Worth its own change: the fix is one line (`agent_label_index`) plus a golden regen and a sweep that un-escapes first.

pengfei-threemoonslab:
## Round 4, and the finish line measured

**A documentation example that does not run.** The manifest reference example wrote `path: __toolsnaps__` — the directory `github/github-mcp-server` keeps its tool snapshots in, which is where the 116 tools in the issue came from. An `mcp` source names one protocol artifact, so a reader copying that block got `static input is not a regular file` before any of the prose about it could be true. Fixed here and in the PR body above.

The guard for it was **vacuous on the first attempt**, and instructively so: lifting the block out of the section and materialising what it names creates a *file* called `__toolsnaps__`, which loads perfectly. No scan of an example can catch the shape of its own path. So it is split — run the block for "does it parse and do what the prose says", and *state* the path rule from where the codebase already states it (`manifest_proposals._FILE_SOURCE_TYPES`: those types name one artifact, so the documented path carries a suffix) — and verified by putting the original defect back and watching it fail.

That is the third vacuous guard the perturbation method caught in this PR, and none of the three showed up in a green suite.

---

**The route is a finish line, not just a step.** A 116-tool read-only MCP server with the two one-line blocks — `binding` and increment 3's `authority` — and nothing else:

```
Evidence coverage: static (116/116 catalog tools reachable; 116/116 actions pass-eligible)
Decision: review_required
Next action: Walk the 1 review item(s) starting with SHIP-INVENTORY-TOOL-SURFACE-TOO-LARGE;
             release is allowed but the human reviewer should weigh in.
```

Zero evidence gaps. The one review item is an honest, separate judgement about a 116-tool surface — which the gate could not previously make, because it could not see the surface at all. Before this change the same repository reported `reachable_tools: 0` and `insufficient_evidence` regardless of what the PR did.

src/agents_shipgate/core/agent_bindings.py
```diff
@@ -377,6 +409,43 @@ def resolve_agent_binding_graph(
                 )
             )
 
+    for source, pointer, observation in surfaces:
+        bound = [tool for tool in tools if source.id in tool.configured_source_ids]
```
pengfei-threemoonslab: **[P1] Preserve the configured source identity for Codex-plugin inventories**

A configured `codex_plugin` row cannot match the MCP inventories it contributes at this join. In a package source `id: browserish` whose `mcp_tool_inventories` publishes `open_page`, the adapter mints the loaded source ID `codex_plugin:browserish/browser:inventory`, so the tool never carries `browserish` in `configured_source_ids`. I reproduced that the scan is structural/pass-eligible without the new block, but adding only `binding` makes it fail with `missing_binding_evidence` claiming the source contributed nothing. That violates the advertised additive guarantee for a supported built-in source. Please propagate the configured parent row ID through these inventory loads and add this adapter regression, or constrain `binding` to source types that can establish this join.

src/agents_shipgate/core/agent_bindings.py
```diff
@@ -377,6 +409,43 @@ def resolve_agent_binding_graph(
...
+                        f"contributed nothing to the tool catalog."
+                    ),
+                    agent_id=observation.agent_id,
+                    source=TOOL_SOURCE_BINDING_DECLARATION,
+                    source_pointer=pointer,
+                )
+            )
+            continue
+
+        for tool in bound:
```
pengfei-threemoonslab, line 467: **[P1] Fail closed when source and agent declarations conflict**

If source `srv` publishes `a,b`, has this `binding`, is selected by `agent_bindings.root`, and a complete `agent: root` declaration lists only `a`, the resolver reports `declared` / `pass_eligible=true` with no issues and silently unions both tools. These source edges are appended after the existing declaration-vs-structural conflict comparison, so two reviewed closed-world statements about the same node can disagree without `conflicting_binding_evidence`. Please compare these declaration sets before granting pass eligibility, or reject an agent declaration that targets a source-surface node.

src/agents_shipgate/ci/release_decision.py
```diff
@@ -552,6 +553,31 @@ def _inventory_declaration_template(
...
+    ``tool_identity`` binding merged across two configured sources is only
+    fully covered when both of them declare.
+    """
+
+    if not tool_catalog:
+        return None
+    ids: set[str] = set()
+    for tool in tool_catalog:
+        configured = set(tool.configured_source_ids)
+        if len(configured) != 1:
```
pengfei-threemoonslab: **[P2] Scaffold every source contributing to a merged tool identity**

A canonical tool can legitimately carry `configured_source_ids=['east','west']` after a reviewed `tool_identity` merge. This cardinality check treats that valid case as undeclarable and routes an artifact-only repository to `agent_bindings.declarations` with no template, even though adding `binding` to both source rows closes the gap and produces 1/1 reachable/pass-eligible coverage. That also contradicts this helper's docstring that both contributors must declare. Return `None` only when a catalog tool has no configured source; otherwise union all contributing IDs.

src/agents_shipgate/schemas/manifest/tool_sources.py
```diff
@@ -92,6 +92,53 @@ def validate_co_requirements(self) -> SourceAuthorityConfig:
...
+    the source's own content is the #268 attack, and the point of this block is
+    to make the statement one line, not to remove it.
+    """
+
+    model_config = STRICT_MODEL_CONFIG
+
+    #: Spelled exactly as ``agent_bindings.declarations[].complete``, and for
+    #: the same reason: the closed-world assertion is the thing being reviewed,
+    #: so there is no ``false`` to write. Presence of the block is the claim.
+    complete: Literal[True] = True
```
pengfei-threemoonslab, line 130: **[P2] Keep the new declaration in runtime/schema parity**

The two validators currently disagree in both directions: the runtime coerces `complete: 1` to `True` while `docs/manifest-v0.1.json` rejects it as non-boolean, and the published schema accepts `reason: ''` / whitespace while the runtime rejects those values. The runtime also accepts zero-width, bidi-control, or control-only reasons, which render blank despite this field being the review record. Please make `true` strict, reuse the repository's visible-content invariant for `reason` in both validators, and add full-manifest JSON-Schema/runtime parity cases rather than only asserting that the property exists.

src/agents_shipgate/ci/release_decision.py
```diff
@@ -748,13 +807,55 @@ def _agent_subject(issue: AgentBindingIssue) -> str:
...
+                    "shipgate.yaml#agent_bindings.declarations with agent: root, "
+                    "then rerun verification."
+                )
+        elif issue.source == TOOL_SOURCE_BINDING_DECLARATION:
+            # A reviewed source binding that binds nothing. The block is where
+            # the reader starts either way — the declaration is wrong about
+            # what the source publishes, or the source is not reading it.
+            action_kind = "declare_agent_bindings"
+            path = "shipgate.yaml#tool_sources[].binding"
+            accepted_values = ["complete:true", "reason"]
```
pengfei-threemoonslab: **[P2] Publish remediation values that can actually close this gap**

This branch correctly says the reviewer must fix the source path/artifact or remove the binding, but then advertises `complete:true` and `reason`—the fields already present on the declaration that raised the issue. The durable `fix_task` renders those verbatim as accepted values, directing a reader to reaffirm a declaration that still binds no tool. Publish no declaration values here, or use remediation choices such as `correct_source_path` / `remove_binding`; targeting the exact source row would make the route safer as well.

pengfei-threemoonslab (commented):
I reviewed this end-to-end across adapter provenance, binding-graph construction, remediation routing, manifest/runtime-schema parity, public reporting semantics, and regression coverage. The direction is valuable, but I reproduced five correctness/contract defects in the new source-wide declaration path; each is attached inline.

There is also a public representation issue around the new rootless entry-point state. A multi-source declared graph can now be `pass_eligible=true` with `root_agent_id=null`, while `report.md` renders that as `Root agent: unresolved` and `docs/passed-verdict-contract.md` still describes inventory/actions as root-reachable and reviewed declarations as living under `agent_bindings`. Please distinguish a deliberate source-entry-point graph from an unresolved graph (and give source pseudo-nodes an explicit public kind), or preserve an unambiguous root identity invariant.

The release notes should also avoid saying “No published schema changes” while `docs/manifest-v0.1.json` adds `tool_sources[].binding`; if the intended statement is that there is no schema-version bump, say that directly.

Validation performed:

- Targeted binding, source-authority, release-decision, declaration-scaffold, surface-exclusion, semantic-property, and manifest-schema-parity suites passed.
- `ruff check .` passed.
- The full non-performance suite reached the end with only eight packaging setup errors caused by the sandbox blocking the `hatchling` download; all eight packaging tests passed when rerun with network access.
- All GitHub checks are green.

Agents Shipgate returned `review_publishable` / `review_required`. This PR changes `AGENTS.md`, so that trust-root change still requires human review and cannot be self-approved by a coding agent.

Because the authenticated GitHub account owns this PR, GitHub does not permit a formal `REQUEST_CHANGES` review from it; this `COMMENT` review is the strongest non-approval review state available. These findings should be resolved before merge.

pengfei-threemoonslab:
Thanks — all five inline findings reproduced before fixing, and both contract points addressed. Report schema `0.41 → 0.42`; v0.41 frozen and hash-pinned.

## [P1] Codex-plugin inventories — and the class behind it

Reproduced exactly as described: `plain(reach=1, elig=True)` → `declared(reach=1, elig=False)` with the declaration claiming the source contributed nothing while its tool sat reachable in the catalog.

Rather than fix the one instance, I probed **all nine built-in `tool_sources` types** for the additive guarantee:

```
mcp                OK    openapi        OK    openai_agents_sdk  OK
google_adk         OK    langchain      OK    crewai             OK
conductor          OK    codex_plugin   *** BINDS NOTHING ***
codex_config       SETUP-ERROR (pre-existing, see below)
```

So the class is exactly one type wide, and the cause is the one you named: the dispatcher attributes pass-2 results by the `(adapter source type, minted source id)` pair, and a Codex plugin MCP inventory is `codex_plugin:<plugin>/<server>:inventory` with source type `codex_plugin_mcp_inventory` — neither half. Fixed by recording the association in the frame that knows (the adapter's own `tool_sources` loop), with `_absorb` no longer overwriting an id an adapter set. That also silently repairs **`tool_sources[].authority`**, which had the identical exposure since #410 and failed quietly rather than loudly.

A parametrised sweep now asserts the guarantee per source type, so a new adapter minting derived ids fails here rather than in an adopter's repository.

**`codex_config` is excluded, and filed separately.** A row of that type plus any `.mcp.json` fails *every* scan with `InputParseError` — `mcp_manifest` returns tools carrying `mcp_json:<server>` inside a `LoadedToolSource` named `codex_config_mcp:<path>`, and `tool_identity._observations` rejects the mismatch. Reproduced on a clean `origin/main` worktree; the existing `codex_config` fixtures all use workspaces where the loader mints no tools, so nothing catches it.

## [P1] Conflicting closed-world declarations

Reproduced: `status: declared`, `pass_eligible: true`, zero issues, both tools unioned. Now:

```
status: conflicting | pass_eligible: False
conflicting_binding_evidence — Closed-world declaration for 'root' does not match
the published surface that entry reviews at tool_sources[].binding.
```

Placed beside the structural conflict rule it mirrors. Negative control in the test: two statements that *agree* are still not a conflict, so the rule is about the disagreement rather than about declaring both.

## [P2] Merged tool identities

You're right, and the docstring was the tell — it said both contributors must declare while the code returned "undeclarable". It now returns `None` only where a catalog tool has no configured source at all, and the test walks the loop: an undeclared merge is offered the route with **both** ids in the scaffolded block, and taking it reaches `1/1` reachable with zero gaps. The docstring now also says why this is deliberately *not* the rule `authority` obeys — authority replaces published evidence, so applying one source's credential to a tool another also contributes would be a claim about the wrong deployment.

## [P2] Runtime / schema parity

Eleven shapes, **zero mismatches** in both directions:

```
{"complete": 1, ...}              runtime=False schema=False
{"complete": "true", ...}         runtime=False schema=False
{"complete": true, "reason": ""}  runtime=False schema=False
{"reason": "​⁠"}        runtime=False schema=False
```

`strict=True` cannot be applied to a `Literal` schema (pydantic raises at class construction), so the boolean rule is a before-validator; the published schema already carried `type: boolean`. `reason` now carries `VISIBLE_CONTENT_PATTERN`, reusing the invariant `action_surface.actions[].override` already uses, so the pattern reaches `docs/manifest-v0.1.json` and the validator keeps owning the message.

`agent_bindings.declarations[].complete` has the same coercion and I left it alone deliberately — tightening a field that has shipped is a compatibility decision of its own. Say the word if you'd rather it moved in this PR.

## [P2] Remediation values

Now `correct_source_path` / `remove_binding`, and the path targets the exact row:

```
PATH   shipgate.yaml#tool_sources[id='empty_source'].binding
VALUES ['correct_source_path', 'remove_binding']
```

spelled the way `action_surface.actions[tool='…']` already is.

## The rootless entry-point state

Taken as the contract defect it is, and published rather than inferred:

- **`binding_surface_facts.entry_point_agent_ids`** — every node the reachability walk started from. Empty *exactly* when nothing rooted the graph; equal to `[root_agent_id]` for every graph a prior release could produce, so a consumer adopting it needs no special case for the old shape. Both halves are pinned by tests.
- **`agents[].kind`** (`agent` | `tool_source`, default `agent`) — the explicit public kind you asked for. The contract doc needed it anyway: it was already hand-waving about the skill-only Codex plugin package root as a node in `agents[]` that "is not a runtime agent".
- **`report.md`** stops calling the deliberate state unresolved:

```
Status: declared
Root agent: none (graph rooted by declared tool sources)
Entry points: server_a, server_b
Pass eligible: true
```

- **`docs/passed-verdict-contract.md`** now says reachable *from the graph's entry points*, names both declaration sites, and spells out which `root_agent_id: null` is which.

## Release notes

Corrected — it now states the report bump `0.41 → 0.42` and says the manifest field is additive, rather than claiming no published schema changed.

---

Full CI split green locally (`pytest -n auto -m "not perf" --ignore=tests/test_adapter_static_only.py`, plus perf and static-only), `ruff check .` clean, `generate_schemas.py --check` clean. The schema bump touched the usual chain — regenerated schemas, four sample `report.json` + five `report.md` + the refund packet trio, doc stamps, the adoption-kit renderer source with its outgoing hash recorded, `llms-full.txt`, the safety-qualification pin, and the newly frozen v0.41 byte hash. One checklist item was not on the known list and is worth adding: `tests/test_effect_coverage.py` hardcodes the current schema filename in two places.

pengfei-threemoonslab:
One follow-on from the contract point, pushed separately.

`docs/passed-verdict-contract.md` was not the only place stating the criterion as *root*-reachable. The same sentence appears in four more surfaces a reader is more likely to open first —

```
README.md          `passed` still requires a complete root-reachable binding graph …
AGENTS.md          A `passed` result requires a complete root-reachable static binding graph …
skills/…/SKILL.md  A `passed` decision requires a complete root-reachable static binding graph …
docs/examples.md   … root-reachable binding graph and normalized semantic assessments
```

— and my own manifest-reference section had added a fifth ("Every tool the source contributes becomes root-reachable"), which is precisely the thing a published tool surface is not.

A repository that publishes tool surfaces has no root agent by construction, so leaving those would have told exactly the adopter this change is for that they cannot get there. All five now read "from its entry points", which is the same requirement stated in terms `entry_point_agent_ids` names, and is unchanged for an agent application whose only entry point is its root. The Claude Code skill went through the renderer-source path with its outgoing hash recorded, and the plugin mirror was re-copied.

Full CI split green locally again after the change.

pengfei-threemoonslab:
## Merged `main` — and why CI had gone quiet

[#434](https://github.com/ThreeMoonsLab/agents-shipgate/pull/434) landed while this was in review, which put the PR in `DIRTY` and stopped GitHub producing a merge ref — so the two commits before this one never got a CI run at all. Resolved; the PR is `CLEAN` again and checks are running.

Only `CHANGELOG.md` conflicted textually (both entries kept). But the two changes meet somewhere more interesting than the changelog: **both touch `core/surface_exclusions.py`, at the same pointer.** #434's new headline clause selects rows by `accounted_by`; this branch filters what may enter `binding_gap_subject_by_id`, which is what sets it. An auto-merge is not evidence those agree, so I checked rather than assumed:

- **The guard still holds.** The adversarial fixture — a `tool_sources[].id` spelled as the canonical id of the genuinely excluded tool — still leaves the row `not_claimed` / `accounted_by: null` / `gated: 0` on the merged tree. So #434's clause can never name a *source* for a *tool* exclusion.
- **Nothing #434 needs was filtered away.** A legitimate tool-scoped row keeps its pointer:

```json
{"stage": "binding", "subject": "wired.new_delete [agent_tools]",
 "reason": "newly_unbound_tool", "accounting": "evidence_gap",
 "accounted_by": "wired.new_delete [agent_tools]"}
```

- **A/B over one fixture**, same commits, same manifest but for the block `main` cannot parse:

```
MAIN    clause=False  … the agent's tool bindings are unproven (wired.new_delete [agent_tools]) …
                          1 binding evidence gap(s) and 1 semantic evidence gap(s)
BRANCH  clause=False  … the agent's tool bindings are unproven (wired.new_delete [agent_tools]) …
                          1 binding evidence gap(s) and 2 semantic evidence gap(s)
```

Same clause behaviour, same named subject. The one difference is the extra semantic gap on the branch, which is the declared source's tool being *judged* rather than excluded — the widening this change is for, showing up as an obligation.

Perturbation sweep re-run on the merged tree: all **25** mechanisms still fail at least one test. Full CI split green locally, `ruff` clean, `generate_schemas.py --check` clean.

PR #438, opened 2026-08-27
fix(reporting): the Root agent line names the agent, not its digest (#329)

pengfei-threemoonslab:
## Summary

- **`Root agent:` in `report.md` named a derived digest.** `_append_binding_surface` printed `Root agent: agent_v1:7205d836e4b3fee257d90695` at the head of the Agent Binding Surface section, and all five shipped samples carried it. That is the #329 invariant — proceeding must not require understanding the internal identity model — and a digest appears in no file the adopter has, while being the label for the agent the *whole section* is about. It now resolves through `core.surface_exclusions.agent_label_index`, the same index `ci/release_decision.py` uses for binding gaps:

  ```diff
  - Root agent: agent\_v1:7205d836e4b3fee257d90695
  + Root agent: durable\_order\_agent \[conductor\_workflows\]
  ```

- **`Entry points:` (#432) resolves through the same index too.** Its fallback was the raw `agent_v1:` id, which is the same class. Both lines now go through one `_agent_display`, so the section cannot spell the same agent two ways — which it was doing on `main`, one line apart.

- **`unresolved`, and `Root agent:` itself, are only ever said about an agent.** Two states reach this line with a truthy `root_agent_id` and no agent object behind it. Both found in review (see below).

- **The guard that should have caught this was near-vacuous.** `test_sample_markdown_speaks_the_adopters_vocabulary` does sweep `samples/*/expected/report.md`, but the renderer escapes every value it prints, so the line arrived as `agent\_v1:7205d836…` and `AGENT_ID_PATTERN` never fired. This is wider than the one line: **every** shape in `INTERNAL_ID_SHAPES` and every term in `MANIFEST_SPELLED_TERMS` contains an underscore, so the sweep was blind to roughly half the report — everything passing through `_safe_markdown_text`. It now un-escapes each line first, against a `_unescape_markdown_text` defined as the escaper's inverse beside it.

## Type

- [ ] Input adapter change
- [ ] Documentation only

## Verification

Additional local checks run:

- `pytest -n auto -m "not perf" --ignore=tests/test_adapter_static_only.py`, `tests/test_latency_budget.py -m perf`, `tests/test_adapter_static_only.py` — all pass
- `ruff check`, `compileall -q src tests`, `scripts/generate_schemas.py --check` — pass
- **The un-escaped sweep run against all five samples *before* fixing the renderer**: `Root agent:` was the only violation, so the class is contained to that one line.
- Confirmed the sweep fails on the stale goldens before regenerating them — the guard demonstrably changed state.
- **30,000-case fuzz** over `_unescape_markdown_text(_safe_markdown_text(s)) == s`: zero mismatches.
- Swept the artifacts the test does *not* cover (`packet.md`, `packet.html`) with the un-escaping matcher: only finding fingerprints, which are the deliberately exempt kind. No hidden instances of this class.

Tests added:

1. the root line names the agent, asserted on the un-escaped text (what a reader sees rendered);
2. the unnameable-root shapes, parametrized, never fall back to the id;
3. the `Entry points:` line has the same fallback;
4. a **negative control** asserting `internal_vocabulary()` returns `()` on the raw escaped string while the sweep rejects it — without this the guard can silently go vacuous again;
5. an escape/un-escape round trip over every `MARKDOWN_ESCAPE_CHARS` entry including a literal backslash, which is what makes the escaper's ordering load-bearing;
6. the compat-graph state, asserted both on the assessment and through the schema default the way production reaches it;
7. the single-declared-surface state, asserted on the producer's shape and then on the rendered line.

## Review rounds

**Round 2 — the two review comments on this PR.** Both fixed.

- **[P2] A single declared surface is the same state as several.** `_select_root` returns the sole surface node when a repository declares exactly one reviewed `tool_sources[].binding` and nothing observed an agent (`core/agent_bindings.py:1025-1026`), so — unlike the two-surface form — that graph has a *truthy* `root_agent_id` pointing at a node whose `kind` is `tool_source`. Keying the line on "is a root id set" printed `Root agent: github_mcp` beside `Pass eligible: true`, announcing an agent object the repository does not have, while the two-source form of the same repository printed the correct sentence. Reproduced at this head before fixing. Whether a repository has an agent cannot depend on how many sources it declared, so the answer is now read from the root node's `kind`, and the sentence is one constant rather than a literal in each of the two branches that reach it — the drift *was* the defect. Both forms now render `none (graph rooted by declared tool sources)`, with `Entry points:` still naming the surface that roots each. Regression test added beside its two-source twin, asserting the producer's shape first so it cannot pass by accident.

  The renderer was chosen over the producer deliberately: changing which node `_select_root` returns would move `root_agent_id` in `report.json`, a published field `ci/release_decision` reads — a contract change rather than a rendering fix.

- **[P3] Contradictory release note.** The entry listed `legacy_direct` under what `unresolved` covers while the paragraph below it, and the code, render `none (tools bound directly, no agent graph)`. Removed from the list, the new single-source state recorded alongside it, and the privatised helper's name corrected there too.

**Round 1 — a self-review pass** found one real regression this PR had introduced, and a docstring asserting that regression could not happen.

- **`legacy_direct` rendered as `unresolved`.** It is the `root_agent_id` of the compatibility assessment a report carries when no agent graph was resolved at all — `core/findings/report_builder.py` builds it whenever its `binding_surface_facts` argument is `None`, and it is the schema default for a `report.json` written before the field existed. It is truthy and no node carries it, so resolving it like an agent id printed `Root agent: unresolved` directly between `Status: structural` and `Pass eligible: true`, describing a failure the report was not reporting. Confirmed by rendering a bare `ReadinessReport`. It now reads `Root agent: none (tools bound directly, no agent graph)` — the same shape #432 gave the tool-source state one branch below.
- **A docstring claim with nothing holding it up.** `_agent_display` asserted its fallback "is a fail-safe rather than a state the scanner can reach". True of a graph the walk built; false of the sentinel, which lands on it immediately. That claim is how the branch got written. Rewritten to state what callers owe the function.
- **A test pinning the wrong answer.** The `legacy_sentinel` parametrized case sat under "every way the graph can fail to name a root", which `legacy_direct` is not. Split into its own pair of tests.
- **New public surface for a test helper.** `unescape_markdown_text` was public while the escaper it inverts is private. Renamed `_unescape_markdown_text`.

## Release-readiness notes

- [x] Report/schema changes are additive or documented in `STABILITY.md` — **no schema moves.** Markdown-only: report, packet and verifier schema versions and the runtime contract are all unchanged, and no `report.json` field is added, renamed or removed.

## Reviewer notes

- **Goldens.** The five sample `report.md` files are regenerated; **no `report.json` differs from `main` at all**, which confirms the change is confined to the Markdown renderer.
- **Flagged, deliberately not fixed here:** for an *agent*-rooted graph, `Entry points:` is now a verbatim repeat of `Root agent:`, since `entry_point_agent_ids` is documented as `[root_agent_id]` there and both lines resolve it the same way. The redundancy predates this branch — it was only masked because the two lines used to spell the same agent differently. (It does not arise in either surface-rooted form, where the two lines now say different things.) Suppressing the line is one condition and no test asserts it for a rooted graph, but it is #432's rendering decision, so it is left for its author to settle.
- **Pre-existing, out of scope:** `samples/conductor_agent/expected/current-control.json`'s `artifacts.report` digest is computed before the `<REPO>` substitution, so it describes a file that is never committed and can never match the shipped `report.json` (50876 claimed vs 50809 actual). `main`'s value was equally wrong. The `report_markdown` half is correct and matches byte for byte, both before and after this change.


src/agents_shipgate/report/markdown.py
```diff
@@ -1092,10 +1133,22 @@ def _root_agent_line(graph: AgentBindingGraphAssessment) -> str:
...
+    above: ``Status: structural`` and ``Pass eligible: true`` are printed
+    around this line, and ``unresolved`` between them describes a failure that
+    did not happen.
     """
 
+    if graph.root_agent_id == _LEGACY_DIRECT_ROOT_AGENT_ID:
+        return "none (tools bound directly, no agent graph)"
     if graph.root_agent_id:
-        return graph.root_agent_id
+        return _agent_display(graph.root_agent_id, agent_label_index(graph.agents))
```
pengfei-threemoonslab, line 1177: [P2] Preserve the no-agent state for a single bound tool source

This branch treats every truthy `root_agent_id` as an agent, but `_select_root()` returns the sole `kind="tool_source"` node when a repository has exactly one reviewed `tool_sources[].binding` surface and no observed agent. The result is `Root agent: github_mcp` beside `Pass eligible: true`, even though the published contract says this graph has no agent object; the two-source form correctly renders `Root agent: none (graph rooted by declared tool sources)`. Please inspect the matched node's `kind` before using `_agent_display()` (or correct the producer's published root semantics) and add the missing single-source Markdown regression test.

CHANGELOG.md
```diff
@@ -2,6 +2,49 @@
...
+- **The report's `Root agent:` line names the agent instead of hashing it.**
+  (#329) Every shipped sample printed `Root agent: agent_v1:7205d836…` at the
+  head of the Agent Binding Surface section — a derived digest that appears in
+  no file the adopter has, telling them which agent the whole section is about
+  in the one vocabulary they cannot look anything up in. It now reads
+  `Root agent: durable_order_agent [conductor_workflows]`, resolved through the
+  same agent label index a binding gap uses, so the section header and a
+  finding about that agent cannot spell it two different ways. `unresolved`
+  covers every way the graph can fail to name a root — no root, the
+  `legacy_direct` sentinel, or a root id no node carries; chaining back to the
```
pengfei-threemoonslab: [P3] Keep the compatibility sentinel out of the unresolved list

This says `legacy_direct` renders as `unresolved`, but the implementation at `_root_agent_line()` and this same entry's lines 30–38 deliberately render it as `none (tools bound directly, no agent graph)`. Please remove the sentinel from this list so the release notes state one behavior.

pengfei-threemoonslab (commented):
Review found one user-visible correctness issue that should be fixed before merge, plus a contradictory release-note statement called out inline. GitHub does not permit the connected author account to submit a formal change-request event on its own PR, so this is posted as a comment review.

I reproduced the single-source case at this head: a valid graph whose sole entry point is `kind="tool_source"` renders `Root agent: github_mcp`, `Entry points: github_mcp`, and `Pass eligible: true`, even though there is no agent object. The two-source form already preserves that distinction.

Validation performed: all three GitHub workflow runs are green; local Ruff, compileall, schema generation, the 126-test adopter-vocabulary suite, performance/static-only suites, and the full non-performance suite passed (the isolated packaging tests were rerun successfully with build-network access). Agents Shipgate verification reports `control.state=complete`, `merge_verdict=mergeable`, and `release_decision=passed`; this renderer-semantic case is outside that static gate's findings.

I also verified the PR-described `current-control.json`/normalized `report.json` digest mismatch, but it exists on `main` and is explicitly acknowledged as out of scope here, so I did not count it as a new blocker.

Issue #431, opened 2026-08-27
MCP servers whose tool surface exists only as code have no route: `detect` reports the official MongoDB and Grafana servers as not agent projects


pengfei-threemoonslab:
## Summary

Shipgate reports that the **official MongoDB MCP server** is not an agent project:

```
$ agents-shipgate detect --workspace mongodb-mcp-server --json
is_agent_project : false
frameworks       : []
suggested_sources: []
next_action      : "Stop: Workspace has no framework imports, no tool artifacts,
                    and no prompt directory."
```

That repository publishes **49 MCP tools**, including `drop-database`, `drop-collection`,
`delete-many`, `update-many`, `atlas-create-cluster` and `atlas-create-db-user`. It is
precisely the kind of surface this product exists to gate, and there is no route to it today.

The important finding is what the repository actually contains, because it changes what the
ask should be. **This is not a Zod problem.**

## The gap, measured

Reproduced on `9233c6b8`, against
[mongodb-js/mongodb-mcp-server#1417](https://github.com/mongodb-js/mongodb-mcp-server/pull/1417)
("request elicitation for aggregations that include write stages") — a PR that *adds* a human
confirmation control to write-stage aggregations.

`verify --preview` on that diff:

```
diff completeness: complete   files: 17
should_run       : None   evaluation_status: unclassified   matched_rules: []   diff_tokens: []
```

The abstention itself is correct — #403's "skip requires positive evidence" is doing its job,
and this reports `unclassified` rather than claiming nothing changed. But no rule classified a
diff that adds a confirmation control to MCP write tools, and `detect` cannot see the project
at all, so there is nothing downstream to abstain *into*.

The only route the suggested-source probe found was filename-shaped, and it correctly rejected
what it found:

```json
"excluded_sources": [{"type": "mcp", "path": "packages/mongodb-atlas-mcp-remote/package.json",
                      "reason": "MCP tools file must contain a tools array"}]
```

## What the repository already declares statically

Every tool is a class with two literal static fields:

```ts
export class AggregateTool extends MongoDBToolBase {
    static toolName = "aggregate";
    public description = "Run an aggregation against a MongoDB collection";
    public argsShape = { ...ConnectionIdArgs, ...CollOperationArgs, ...AggregateArgs };
    static operationType: OperationType = "read";
}
```

Counted across the tool tree:

| | |
|---|---|
| unique `static toolName = "…"` literals | **49** |
| `static operationType` literals | 54 |
[...]

Issue #395, opened 2026-08-20
Project discovery reports a truncated candidate list as complete, omitting the project under review


pengfei-threemoonslab:
## Summary

On a repository large enough to hit the Python-file cap, `detect`/`init` report a **truncated** list of agent project candidates as if it were complete, with no cap warning and no `--max-python-files` hint. The project actually under review was missing from the list the user was told to choose from.

## Evidence

`src/agents_shipgate/cli/discovery/signals.py:2006`:

```python
if len(candidates) > 1:
    return "ambiguous"
if parse_truncated and project_roots > 1:
    return "unknown"
return "single"
```

The `ambiguous` check short-circuits. `"unknown"` — the state whose entire purpose is to say the parse was cut short — is reachable only when **one or fewer** candidates were found. Its own docstring makes the argument that applies equally to the candidate list:

> Python parsing stops at `max_python_files`, so on a large repository the evidence behind a `"single"` verdict may simply be the part of the tree that got read first — and filesystem ordering is not a safety property.

The candidate list is also just "the part that got read first". The cap-reporting branch at `signals.py:363` and the `--max-python-files` remedy it prints are therefore unreachable on exactly the repositories most likely to be truncated.

## Reproduction

[google/adk-samples#1745](https://github.com/google/adk-samples/pull/1745), CLI `0.16.0b7`.

```
agents-shipgate init --workspace <repo> --write --json
```

```
Refusing to write shipgate.yaml: <repo> holds 53 self-contained projects
that define agents, and one manifest describes one agent surface.
Candidate project directories:
  - . (food_tour_agent)
  ...
```

- `agent_project_candidates` contains exactly 53 entries and ends at `python/agents/podcast_transcript_agent`.
- `ls -d python/agents/*/ | wc -l` → **74**. Twenty-two directories sort after `podcast_transcript_agent`, including **`python/agents/smart_closer`**, the project this PR adds.
- `agent_scope: "ambiguous"` (not `"unknown"`); no cap warning anywhere in the output.
- An earlier `detect` on the same repository reported `workspace_signals.python_file_count: 1000` — exactly the cap set at `src/agents_shipgate/cli/first_look.py:104`.

So the user was handed an authoritative-looking list of 53 projects, told "pick the one this change belongs to", and the correct answer was not in it.

## Why this matters

[...]

PR #443, opened 2026-08-27
fix(init): a field the tool chose without evidence is marked as one (#441)

### LINKED ISSUE #441: `detect` declines a workspace and `init` writes a manifest for it anyway — with a source `type` it guessed and did not flag
## Summary

`detect` declines a workspace outright. `init` — the command the control loop routes to from `verify --preview` — writes a manifest for it anyway, and the manifest it writes declares a **source type it did not infer and did not flag for review**.

A user following the documented loop on a Python MCP server ends up committing an artifact that `detect` had just said was pointless, describing an OpenAPI spec that does not exist in the repository.

## Reproduction

[awslabs/mcp#4489](https://github.com/awslabs/mcp/pull/4489) — a FastMCP Python MCP server under `src/billing-cost-management-mcp-server`. CLI `0.16.0b7` at `fc96bad9`.

**Step 1 — preview routes to `init`.** Scope resolution works correctly here, picking the right sub-project out of a monorepo with dozens of servers:

```
control.next_action.command:
  shipgate init --workspace …/src/billing-cost-management-mcp-server --write --json
why: "Every changed path that carries a capability surface belongs to
      src/billing-cost-management-mcp-server…"
```

**Step 2 — `detect` on that exact path declines it:**

```
is_agent_project : false
frameworks       : []
suggested_sources: []
next_action      : "Stop: Python project with no agent framework, prompts, or tool
                    surface — not a Shipgate target."
```

**Step 3 — `init` on that exact path writes a manifest:**

```
created: true    manifest_status: written    agent_name: None
placeholders: ['agent.name', 'agent.declared_purpose[0]',
               'tool_sources[0].id', 'tool_sources[0].path']
```

```yaml
tool_sources:
  - id: CHANGE_ME
    type: openapi
    path: CHANGE_ME.yaml
```

## Three distinct defects

**1. Two commands answer one question differently.** `detect` says "not a Shipgate target"; `init` proceeds. Because `verify --preview` routes to `init` and not to `detect`, a user following `control.allowed_next_commands` never sees the refusal. The product leads them to create the artifact its own discovery command says has no purpose.

**2. `type` is a silent guess where `id` and `path` are flagged.** Both `CHANGE_ME` fields appear in `placeholders[]`; `type: openapi` appears in neither, so nothing tells the reader it needs review. Filling in the two flagged blanks yields a schema-valid manifest declaring an OpenAPI source for a FastMCP Python server — the exact outcome [quickstart](../blob/main/docs/quickstart.md) warns about: *"the manifest validates but downstream consumers see meaningless defaults."*

The fallback is deliberate and its reasoning is sound — `cli/discovery/template.py:94-102`:

```python
# Schema requires ≥ 1 of tool_sources / openai_api / anthropic /
# google_adk / langchain / crewai. Empty workspace → emit a
# CHANGE_ME stub identical to the legacy --minimal template's
# fallback so users get an actionable starting point.
```

Satisfying the schema is the right instinct. Picking one of six possible types and not marking it as a guess is the defect: the stub is "actionable" only if the reader knows to question a field the tool did not question.

**3. `has_tools_dir: false` with a `tools/` directory present.** The server's tools live at `awslabs/billing_cost_management_mcp_server/tools/`; the conventional-dir check looks only at the workspace root, so the one structural signal this repository does offer is missed. Contributing cause to defect 1's verdict, minor on its own.

## Why this class matters more than one repository

Four vendor MCP servers walked, three languages:

| repo | language | registration | `detect` sees it? |
|---|---|---|---|
| mongodb-mcp-server | TypeScript | `static toolName = "aggregate"` | no |
| mcp-grafana | Go | `mcpgrafana.MustTool("update_incident", …)` | no |
| **awslabs/mcp** | **Python** | `@server.tool(name='budgets', …)` | **no** |
| github-mcp-server | Go | generates `.snap` JSON | yes |

Three of four land in exactly this path — `detect` declines, `init` writes a stub anyway. That is the default experience for an MCP server today (#431), which makes this the shape most first-time adopters of that class will hit.

## Expected

- `init` on a workspace `detect` declines either refuses with the same reason, or writes and states plainly that discovery found no tool surface and the source block is a scaffold rather than an inference.
- Any field the tool chose without evidence is marked as needing review — either `type: CHANGE_ME` with the accepted values listed (as the declaration questionnaire already does for `effect`), or a comment saying which types are available and that this one was not inferred.
- `detect` and `init` reach the same conclusion about whether a workspace is a target, or the disagreement is stated where the reader is.

## Acceptance

- [ ] The reproduction above produces either a refusal or a manifest whose `type` is visibly unresolved.
- [ ] A workspace `detect` accepts is unaffected (regression: the ADK and MCP samples).
- [ ] `placeholders[]` includes every field the template chose without evidence.
- [ ] A nested `tools/` directory contributes its conventional-dir signal.

---

Found 2026-08-24 walking `awslabs/mcp#4489`. Related: #431 (why three of four MCP servers reach this path), #328 (product defect vs. user configuration error in adopter-facing output), and cause 2 of #440 (two surfaces answering one question with no precedence).

pengfei-threemoonslab:
Closes #441.

## The reproduction

A FastMCP Python MCP server (`awslabs/mcp`'s `src/billing-cost-management-mcp-server`). Before / after, same workspace:

| | before | after |
|---|---|---|
| `detect` → `has_tools_dir` | `false` (tools live at `awslabs/billing_cost_management_mcp_server/tools/`) | `true` |
| `detect` → `conventional_dirs` | `[]` | `["awslabs/billing_cost_management_mcp_server/tools"]` |
| `detect` → verdict | `Stop: Python project with no agent framework, prompts, or tool surface` | `Stop: Workspace has awslabs/billing_cost_management_mcp_server/tools/ but no framework imports and no parseable tool artifact… A conventional directory alone is not one.` |
| `init --write` → manifest | `id: CHANGE_ME` / `type: openapi` / `path: CHANGE_ME.yaml` | `id`, `type`, `path` all `CHANGE_ME`, under a comment listing the accepted types |
| `init --write` → `placeholders[]` | `agent.name`, `agent.declared_purpose[0]`, `tool_sources[0].id`, `tool_sources[0].path` | …plus `tool_sources[0].type` |
| `init --write` → what it says | `Wrote …/shipgate.yaml` | `Wrote …/shipgate.yaml` + *"Discovery found no tool surface here, so the tool_sources block is a scaffold, not an inference: every value in it is a placeholder."* |

## The three defects

**1. Two commands answered one question differently.** Resolved by stating the disagreement where the reader is, which is `init` — the control loop routes there from `verify --preview` and never runs `detect`. `init --json` gains `tool_surface_origin`: `"detected"` when every source was read out of the workspace, `"scaffold"` when none was. The renderer decides it (`RenderedManifest`), because the four block builders that make the fallback decision are the only things that know; no caller re-derives it. The same fact is prose in `manifest_message`, on stdout, and in `control.reason`. It is `null` when the render reached neither disk nor the payload — the authority rule `placeholders` already follows, so `skipped_existing` does not describe somebody else's manifest from a render nobody kept.

`init` was **not** made to refuse. Three of the four vendor MCP servers walked in the issue land on this path (#431), so refusing would close adoption for the shape most first-time adopters hit. What was wrong was writing a scaffold and reporting it exactly as an inference.

**2. `type` was a silent guess.** `type: CHANGE_ME` now, so it reaches `placeholders[]` through the existing collector rather than a special case. `path` loses its `.yaml` too — the extension belonged to the guessed type; an MCP export is JSON. The accepted values render from `BUILTIN_TOOL_SOURCE_TYPES` via one function, so the comment cannot describe a set the loader rejects. Both renderers (auto and `--minimal`) share one scaffold block — `--minimal` is the recovery `init` publishes when the auto render fails validation, so leaving the guess there would have handed the recovery path the defect.

**3. Nested `tools/`.** Conventional-dir discovery reads the whole inventory, deduplicated by directory *name* — a monorepo with thirty `tools/` directories contributes the one weak signal a single root `tools/` does. `workspace_signals.conventional_dirs` now carries each one's workspace-relative **path**: once the check reads the whole tree a bare name is no longer a location, and `["tools"]` pointed at a directory the reproduction does not have. The root `is_dir()` check stays, because the inventory is a list of *files* and an empty `prompts/` has no entry in it.

The widening cannot flip `is_agent_project`: detection requires `score ≥ 2.0 AND has_strong`, and every strong signal is worth exactly 2.0, so `has_strong` already implies the threshold.

## Two defects found while fixing those

- **`AdapterRegistry.require('CHANGE_ME')` sent the reader after a package that does not exist.** `type: CHANGE_ME` is the first failure a scan of the scaffold hits — ahead of the missing `path`, which has always been routed as a placeholder — and both the message and `SHIP-DIAG-UNKNOWN-ADAPTER-SOURCE-TYPE` offered "enable third-party adapter discovery" and "pip install" for a value Shipgate itself wrote. Both now route the placeholder to an edit. Both prose copies of the built-in type list had also dropped `codex_config` and `conductor`; all three copies now render from `builtin_tool_source_types_text()`.
- **`init --minimal` produced a manifest the schema rejects.** The `elif not api_artifacts:` guard selecting the fallback tested the artifact *dict*, which has fixed keys and is therefore always truthy — so the fallback was unreachable and every source-less workspace got an empty `openai_api:` block with no `tool_sources`. Both renderers now ask "is this an OpenAI API project?" through one shared tuple of anchor keys, which also stops `--minimal` declaring an OpenAI surface for a bare `prompts/` (a signal an Anthropic-only project carries too).

## What the self-review then found

Three commits after the first, from reviewing the diff and benchmarking it:

- **`conventional_dirs` named a path that resolves nowhere.** Fixed by carrying located paths (above). A directory at the root is still spelled as its bare name.
- **`SHIP-DIAG-PURE-PROMPT-EXPERIMENT` fired on non-Python repos.** It read the widened `has_prompts_dir`, but *"only prompts/ is present"* is a claim about the shape of a workspace: a thirty-file TypeScript MCP server with `src/prompts/` was reported as a prompt experiment — the `mongodb-mcp-server` shape the issue's own table names. That control asks about a **root** `prompts/` again.
- **The tree walk cost 4.4 s on a 120k-file inventory.** `path.relative_to(workspace).parts` per entry, on a whole-workspace pass `detect` already runs for exactly the monorepos #363 and #395 are about. Walking distinct parent directories on strings brings the same answer in **42 ms** — against the 100 s the existing glob pass spends on the same input.

## Acceptance

- [x] The reproduction produces a manifest whose `type` is visibly unresolved.
- [x] A workspace `detect` accepts is unaffected — every sample `detect` accepts still renders `tool_surface_origin: "detected"`, ADK and MCP included, pinned by a regression test. (`clean_read_only_agent` renders `scaffold`, and correctly: `detect` declines it too, because its `tools.json` matches no MCP glob. The two commands now agree.)
- [x] `placeholders[]` includes every field the template chose without evidence.
- [x] A nested `tools/` directory contributes its conventional-dir signal.

## Verification

Full suite green (`pytest -n auto`, perf included), `ruff check .` clean, `self-check` ready.

**Perturbation sweep** — 20 mutants, one per new guard, each reverted after. 17 caught; the 3 survivors are guards whose removal is behaviour-preserving by construction (the parent-directory dedup and the two fast-path `continue`s), which is the correct outcome and is said as such in the code and the tests.

The sweep paid for itself three times. The first pass had one survivor, and it was the important one: `_scaffold_next_action` was applied at `init`'s call site while `_init_advance` owns route selection, so the test exercising it asserted nothing — it compared the advance to the scan action it had just passed in. Both halves of that answer now live in `_init_advance`. It then caught two more tests of mine that pinned an implementation detail rather than a behaviour; both were rewritten to assert the answer.

That route is currently unpublished — the human-owned `declared_purpose` placeholder outranks it on every freshly written manifest — which the docstring says and a separate test pins as the behaviour a live payload actually shows.

## Surface

One new `init --json` field (`tool_surface_origin`), the machine-readable half of the issue's "state plainly" requirement. One semantic change to an existing `detect --json` field (`workspace_signals.conventional_dirs`: directory names → workspace-relative paths), documented in `docs/diagnostics.md` and the CHANGELOG; it has no production consumer. No verdict, count, gap, finding, or permission moves; `report_schema_version`, `contract_version`, the verifier artifact version, and every published schema document are unchanged.


pengfei-threemoonslab (commented):
Self-review result: **changes requested**.

GitHub does not allow the PR author to submit the formal `REQUEST_CHANGES` state, so this is posted as a comment review. I found two blocking contract regressions and five normal-priority correctness/documentation issues.

1. **[P1] Preserve the documented public renderer return contract** — `src/agents_shipgate/cli/discovery/artifacts.py:264-268`, `src/agents_shipgate/cli/discovery/template.py:48-53`, `src/agents_shipgate/cli/discovery/__init__.py:3-6`

   `render_manifest_template()` changed from returning `str` to returning `RenderedManifest`, even though the discovery package explicitly promises this import as a stable public API. All in-tree callers were changed to `.text`, which hides the downstream break. On this head:

   ```python
   yaml.safe_load(render_manifest_template(Path(tempfile.mkdtemp())))
   ```

   raises `AttributeError: 'RenderedManifest' object has no attribute 'read'`; ordinary consumers using `.splitlines()`, `.rstrip()`, or writing the returned string fail similarly. Keep the public string-returning wrappers and add separate provenance-aware internal/result APIs for `init` (or otherwise provide true string compatibility).

2. **[P1] Bind the new provenance answer into `control.input_id`** — `src/agents_shipgate/cli/_register_init.py:1456-1490`

   A dry-run answer can change while retaining the same cache identity. Reproduction: run `init --minimal --json` in an empty directory, add a valid OpenAPI file, and rerun. `tool_surface_origin` changes `scaffold -> detected` and `control.reason` loses the scaffold disclosure, but `control.input_id` is byte-identical. The same occurs in auto mode for an Anthropic-import workspace when a valid Anthropic tools artifact is added. `routing_facts` omits `tool_surface_origin` even though `_init_reason()` consumes it. Include the origin (or a digest of the complete provenance-bearing routing result) and add a two-state regression test.

3. **[P2] Do not claim `--minimal` performed full framework discovery** — `src/agents_shipgate/cli/discovery/artifacts.py:279-317`, `src/agents_shipgate/cli/discovery/manifest_scaffold.py:72-80`

   The minimal renderer probes MCP/OpenAPI/Conductor and simple OpenAI artifacts; it does not run Python framework detection. Yet it now emits “Discovery found no tool surface here” and “No framework import ... was read,” and the docs say `detect` on the same workspace will be negative. For `samples/simple_langchain_agent`, canonical `detect_workspace()` returns `['langchain']` and auto rendering is `detected`, while minimal rendering returns `scaffold` with that negative prose. Distinguish “minimal mode did not inspect/infer this surface” from “full discovery inspected and found none,” either with a separate cause/origin or mode-specific summary.

4. **[P2] Keep scaffold source types open to supported third-party adapters** — `src/agents_shipgate/cli/discovery/manifest_scaffold.py:95-107`, `src/agents_shipgate/cli/_register_init.py:590-603`

   The generated comment says `type` must be “one of” `BUILTIN_TOOL_SOURCE_TYPES`, and `_scaffold_next_action` requires a value listed in that comment. But `ToolSourceConfig.type` is intentionally open so an enabled, installed third-party adapter can register its own source type. Custom adapters are especially likely to reach this scaffold because built-in discovery cannot recognize their artifacts. Say “one of these built-ins, or a source type registered by an installed third-party adapter,” and make the control postcondition match.

5. **[P2] Make the `CHANGE_ME` diagnostic title match its selected route** — `src/agents_shipgate/cli/diagnostics.py:144-175,259-265`

   The new placeholder branch correctly selects an `edit`, but the diagnostic title still says “install/enable the adapter, or fix a typo” — exactly the two remedies the branch says do not apply. Use a placeholder-specific title and test it, not only `next_actions`.

6. **[P2] Port the nested conventional-directory semantics to the zero-install detector** — `src/agents_shipgate/cli/discovery/signals.py:412-435`, `tools/shipgate-detect.py:1766,2010-2012`, `tests/test_zero_install_detector.py:271-276`

   For a workspace containing `pkg/tools/helper.py`, canonical detection now returns `has_tools_dir: true` and `conventional_dirs: ['pkg/tools']`; `tools/shipgate-detect.py` still returns `false` and `[]`. The parity test compares only signal keys, so it stays green while the same named fields have different meanings. Port the location logic and add a value-level parity fixture for the three changed fields.

7. **[P3] Document the output states the implementation actually emits** — `CHANGELOG.md:23-28`, `docs/agent-recipes.md:220`, `docs/quickstart.md:241-246`

   Two claims need narrowing: (a) a normal fresh `init --write --json` scaffold does not carry the scaffold fact in `control.reason` because the human-owned `declared_purpose` route replaces `_init_reason` (it is present in `manifest_message`); and (b) `tool_surface_origin` is also `null` on `skipped_existing`, which the implementation and regression test intentionally pin, but the response docs advertise only `"detected" | "scaffold"`. Document those route-dependent/null cases or change the payload to satisfy the current prose.

Verification performed on `dc14480659eee2c5e6b9fad5b3963cbc80144578` against `fc96bad917f4f5ae614f7b1bfd25820144406ea4`: three independent expert passes; focused detect/init tests green; `ruff check .` green; all GitHub checks green; Agents Shipgate control `complete`, release decision `passed`, merge verdict `mergeable`. Those checks do not cover the public API compatibility and cache-identity reproductions above.

pengfei-threemoonslab:
All seven findings addressed in `bd15895e`. Verification for each, on the reproduction each one named:

**1 [P1] — public renderer return contract.** `RenderedManifest` is now a `str` subclass, so the published call works again and the provenance still travels with the text:

```python
>>> r = render_manifest_template(Path(tempfile.mkdtemp()))
>>> yaml.safe_load(r)['version'], r.splitlines()[0][:24], r.tool_surface_origin
('0.1', '# yaml-language-server:', 'scaffold')
```

Choosing this over string-returning wrappers plus parallel `*_result` APIs deliberately: two names for one render is the second-implementation shape this repo keeps paying for, and the subclass keeps `yaml.safe_load`, `.splitlines()`, `.rstrip()`, `write_text`, equality and hashing all working. Pinned by `test_both_renderers_still_return_a_usable_string`, which makes the consumer call rather than the `.text` one.

**2 [P1] — cache identity.** Reproduced exactly as described, then fixed by adding `tool_surface_origin` and `scaffold_summary` to `routing_facts`. Before: both runs `sha256:b8db5b73…`. After: `sha256:…` differs, `scaffold` → `detected`. Two-state regression test added (`test_input_id_covers_the_provenance_it_publishes`).

**3 [P2] — `--minimal` claiming discovery it never ran.** Correct and the sharpest of the seven: on `samples/simple_langchain_agent`, `detect` returns `['langchain']` while that renderer scaffolds. Each renderer now carries its own summary, chosen where the mode is known rather than looked up by a caller:

```
# `--minimal` does not run framework detection, so the tool_sources block is
# a scaffold, not an inference: every value in it is a placeholder. Only
# MCP/OpenAPI/Conductor exports and simple OpenAI API artifacts were probed,
# and none matched. Re-run `init` without `--minimal` to detect frameworks
# from source.
```

`RenderedManifest.__new__` rejects a summary that contradicts the origin, so the two cannot drift.

**4 [P2] — the open source type.** Fixed in all four places that had narrowed it: the scaffold comment, `_scaffold_next_action.expects`, the placeholder diagnostic, and `AdapterRegistry.require`. All now read *"one of the built-ins (…), or a source type registered by an installed third-party adapter"*. Your point that custom-adapter repos are **more** likely to reach the scaffold is what makes this load-bearing rather than pedantic.

**5 [P2] — diagnostic title.** Placeholder-specific title, with the generic one unchanged for a real unknown type:

```
CHANGE_ME   -> tool_sources[].type is still the 'CHANGE_ME' placeholder in shipgate.yaml (name the source this repository publishes)
openapii    -> No adapter handles tool_sources[].type 'openapii' in shipgate.yaml (install/enable the adapter, or fix a typo)
```

**6 [P2] — zero-install detector.** Ported `_conventional_dir_locations` to `tools/shipgate-detect.py`; script and CLI now agree on the reproduction (`has_tools_dir: true`, `conventional_dirs: ['awslabs/billing_cost_management_mcp_server/tools']`). The parity test now compares those three fields **by value**, and a dedicated nested-directory fixture was added — the samples all keep their conventional dirs at the root, so the per-sample check could never have seen this.

**7 [P3] — docs.** Narrowed in all three surfaces: `control.reason` carries the scaffold fact only on the routes where init's own reason is the envelope's (`manifest_message` is the field that always does), and `tool_surface_origin` is documented as `"detected" | "scaffold" | null`.

---

**Verification.** Full suite green (`pytest -n auto`), `ruff check .` green, `self-check` ready. A perturbation sweep over the seven fixes reverts each guard in turn — all seven caught, plus the new provenance invariant.

Two incidental fixes the work surfaced: `textwrap` was splitting "third-party" across comment lines, and the added imports shifted the line-pinned `subprocess` allowlist in `test_adapter_static_only.py` again.

One judgement call worth flagging: I did **not** weaken `test_trust_claims_carry_meta_cli_qualifier`, which fired on my phrase *"the minimal template does not run framework detection"* in `docs/agent-recipes.md`. That guard is a false positive here, but it is doing its job broadly, so I reworded the doc to "skips framework detection entirely" instead. The same phrase stays in `MINIMAL_SCAFFOLD_SUMMARY`, which is Python source and outside that guard's surfaces — say the word if you would rather the summary avoid it too.

src/agents_shipgate/cli/discovery/manifest_scaffold.py
```diff
@@ -0,0 +1,195 @@
...
+    """
+
+    tool_surface_origin: ToolSurfaceOrigin
+    #: The one sentence this render's mode is entitled to say about a scaffold,
+    #: or ``None`` when the tool surface was detected. Carried rather than
+    #: looked up, because the renderer is the only thing that knows both that
+    #: the block is a scaffold and what it actually inspected.
+    scaffold_summary: str | None
+
+    def __new__(
```
pengfei-threemoonslab, line 84: [P2] Preserve the public renderer's copy/pickle behavior

`RenderedManifest` is a `str` subclass, but inherited string reduction reconstructs it with only the text; this `__new__` also requires keyword-only provenance. Consequently, results from both renderers fail under `copy.copy`, `copy.deepcopy`, and `pickle.loads(pickle.dumps(...))` with `TypeError: ... missing ... tool_surface_origin`. All three worked when this public API returned plain `str`, so the compatibility fix remains incomplete for caches, multiprocessing, and generic copying code. Please define a reconstruction hook (for example `__getnewargs_ex__`/`__reduce_ex__`) that preserves both provenance fields and add regressions for both renderers.

src/agents_shipgate/cli/discovery/manifest_scaffold.py
```diff
@@ -0,0 +1,195 @@
...
+#: ``*_DETAIL``, which only the manifest comment carries — the artifact has no
+#: byte budget.
+DISCOVERY_SCAFFOLD_SUMMARY = (
+    "Discovery found no tool surface here, so the tool_sources block is a "
+    "scaffold, not an inference: every value in it is a placeholder."
+)
+
+#: The evidence behind :data:`DISCOVERY_SCAFFOLD_SUMMARY`. Auto mode has
+#: actually looked for each of these.
+DISCOVERY_SCAFFOLD_DETAIL = (
```
pengfei-threemoonslab, line 155: [P2] Do not claim that no framework import was read

Auto mode can read and classify a framework import yet still need the scaffold. For example, a workspace containing only `agent.py` with `import anthropic` produces an `anthropic` framework detection, but no renderable Anthropic artifact block, so this scaffold is emitted with “No framework import ... was read.” The scaffold decision is valid; the evidence sentence is false. Phrase this in terms of no renderable/tool-surface source being found, and cover an import-only recognized framework with a regression.

src/agents_shipgate/cli/discovery/manifest_scaffold.py
```diff
@@ -0,0 +1,195 @@
...
+#: on `samples/simple_langchain_agent` the canonical `detect` reports
+#: ``langchain`` while this renderer scaffolds (#441 review).
+MINIMAL_SCAFFOLD_SUMMARY = (
+    "`--minimal` does not run framework detection, so the tool_sources block "
+    "is a scaffold, not an inference: every value in it is a placeholder."
+)
+
+#: The evidence behind :data:`MINIMAL_SCAFFOLD_SUMMARY`, and the way out of it.
+MINIMAL_SCAFFOLD_DETAIL = (
+    "Only MCP/OpenAPI/Conductor exports and simple OpenAI API artifacts were "
```
pengfei-threemoonslab, line 176: [P2] Distinguish matched non-anchoring artifacts

A prompt-only workspace populates `discover_openai_api_artifacts()["prompt_files"]`, but because prompt files are intentionally not anchor keys the minimal renderer correctly scaffolds and then says no simple OpenAI API artifacts matched. Response-format-only and trace-only workspaces have the same contradiction. Say that no anchoring/unambiguous tool-surface artifact matched (rather than that none matched), and add negative-control coverage for these non-anchor matches.

docs/quickstart.md
```diff
@@ -238,6 +238,21 @@ Walk `placeholders[]`, edit each one in `shipgate.yaml`, then re-run
 `scan`. Skipping this step leaves an invalid adoption artifact — the
 manifest validates but downstream consumers see meaningless defaults.
 
+**When discovery reads no tool surface at all.** `init --json` reports
+`tool_surface_origin`: `"detected"` when every source in the manifest was read
```
pengfei-threemoonslab, line 242: [P3] Document the `null` origin in the quickstart

This describes `init --json` as reporting only `"detected"` or `"scaffold"`, but the implementation returns `null` when this run's render is not authoritative (`skipped_existing` and `refused_unresolved_scope`). The agent recipe and changelog now document that state; the quickstart should either include it and its route/authority semantics or explicitly scope this paragraph to runs whose render reaches the payload/disk.

CHANGELOG.md
```diff
@@ -2,6 +2,84 @@
...
+  declared `type: openapi` for a repository containing no OpenAPI spec. `id`
+  and `path` were flagged in `placeholders[]`; `type` was in neither
+  `placeholders[]` nor a comment, so filling in the two flagged blanks yielded
+  a schema-valid manifest describing a source that does not exist.
+
+  Three changes, one per defect:
+
+  - The fallback `tool_sources` block is now `id`/`type`/`path` all
+    `CHANGE_ME`, all three in `placeholders[]`, under a comment that lists
+    every accepted `type` (rendered from `BUILTIN_TOOL_SOURCE_TYPES`, so it
```
pengfei-threemoonslab: [P3] Do not describe the built-in tuple as every accepted type

`ToolSourceConfig.type` remains open to source types registered by installed third-party adapters, so `BUILTIN_TOOL_SOURCE_TYPES` cannot enumerate “every accepted type.” This contradicts the newly corrected generated comment and recovery guidance. Say the comment lists every built-in type and also notes that registered third-party types are valid.

pengfei-threemoonslab (commented):
Re-review of `bd15895e447839831458073611865045bf21945c`: the prior findings are substantially improved, but this head still has three behavioral issues and two documentation contradictions. I reproduced each behavioral case and left inline comments.

Merge readiness is also blocked independently of the code findings: GitHub currently reports this PR as `CONFLICTING`; a local merge-tree against current `origin/main` reproduces the conflict in `CHANGELOG.md`. The latest head currently has no reported status checks. Please rebase/resolve, address the findings below, and rerun CI.

Validation performed:

- `agents-shipgate verify` against current `origin/main`: `control_state=complete`, release decision `passed`.
- Focused pytest: `test_init_scaffold_disclosure.py`, `test_zero_install_detector.py`, `test_init_auto.py`, and `test_adapter_static_only.py` passed.
- Ruff on the changed implementation and focused tests passed.
- Independent reproductions confirmed the copy/deepcopy/pickle failure, the import-only Anthropic false evidence claim, and the prompt-only minimal-mode false evidence claim.

This is a comment review because GitHub does not allow the authenticated PR-author account to submit `REQUEST_CHANGES` on its own PR.

pengfei-threemoonslab:
Round 2 addressed in `560c2a0a`, on top of a merge of current `origin/main` (`cfac2649`). PR is `MERGEABLE` again and CI is running on `560c2a0a`.

**Merge conflict.** Both sides added at the top of `## Unreleased`; neither replaced the other, so the resolution keeps both with this PR's entry first. Nothing else conflicted — `cli/diagnostics.py` auto-merged against #442/#445/#438 and the full suite is green on the merged tree. One caveat on history: that merge commit also carries three of the round-2 fixes, because I resolved and fixed in the same working tree before committing. Given this repo squash-merges, the boundary vanishes on merge, but flagging it rather than rewriting a pushed branch.

**[P2] copy/pickle.** Confirmed exactly as reported, for both renderers:

```
copy      TypeError: RenderedManifest.__new__() missing 1 required keyword-only argument: 'tool_surface_origin'
deepcopy  TypeError: ...
pickle    TypeError: ...
```

`__getnewargs_ex__` now carries both provenance fields. The regression covers both renderers **and both provenances** — `__new__` validates the pair, so a round trip that dropped `scaffold_summary` would still have constructed successfully on a `detected` render and gone unnoticed.

**[P2] "No framework import ... was read".** Confirmed: a workspace whose only file is `agent.py` with `import anthropic` scores `anthropic` at 2.0, and with no `tools/anthropic-tools.json` there is nothing to point at — scaffold, with a sentence denying the import it had just classified. Now:

> No tool export, API spec, or framework source file was found for the manifest to point at — a framework import on its own is not one.

The summary moved with it (`no tool surface` → `no tool source to declare`), since the same objection applied one clause up.

**[P2] "and none matched".** Confirmed: `prompts/system.md` populates `prompt_files`, so the probe matched something. Now *"none that anchors a tool surface matched — a prompts/ directory on its own does not"*, which also answers the question the old sentence provoked. Response-format-only and trace-only workspaces take the same path.

**[P3] docs.** Quickstart documents `null` and scopes its paragraph to runs whose render reaches the payload; the CHANGELOG says the comment lists every **built-in** type and notes registered third-party types are equally valid.

---

**Verification.** Full suite green on the merged tree (`pytest -n auto`), `ruff check .` green, `self-check` ready. Perturbation sweep over the four round-2 guards — all four caught, including reverting `__getnewargs_ex__` and each of the two prose fixes.

One thing worth flagging from writing the copy/pickle test: my first version put the `detected` fixture *inside* `tmp_path`, which made the supposedly-bare workspace detect the spec too — both renders came back `scaffold` and the test would have passed for the wrong reason had I asserted less. It now asserts the set of provenances it exercised is `{scaffold, detected}` before checking anything else.

PR #444, opened 2026-08-27
fix(control): one large file must not withdraw the declaration route (#429)

### LINKED ISSUE #429: Trust-root routing pre-empts the §D declaration route: never offered on first adoption, and forbids the commit it just instructed
## Summary

Trust-root routing pre-empts the §D declaration route in both directions, so the agent-drafting loop is unusable on the run that needs it most and dead-ends immediately after the run that uses it.

1. **Before** — on first adoption the manifest is new, so there is no base policy to compare against, the run escalates to human review, and the `declare_action` route is never published. That is the run with **all thirteen** questions open, three of them agent-answerable.
2. **After** — the agent applies the patch the route authorized, and the next `verify` returns `human_review_required` with **every permission false** and **no allowed next command** — including the `commit` the route explicitly told it to perform.

The information needed to separate these from a real weakening is already computed: `policy_weakened: False`. Shipgate authored the edit and still stops as if it might have weakened the gate.

## Design intent

`9233c6b8` (§D) states the target state explicitly:

> `requires_human_review` stays true and the route's permissions are **publish-only**: writing a declaration touches the trust root, so a person still merges it.

Publish-only is `review_publishable` — commit/push/update_pr permitted, merge denied. The implementation lands on `human_review_required`, which permits nothing.

## Reproduction

`google/adk-samples#1745`, `python/agents/smart_closer`, CLI `0.16.0b7` at `9233c6b8`. Scoped `init`, `declared_purpose` filled, nothing else declared.

**Manifestation 1 — route never offered.** Manifest present in the working tree, not yet committed:

```
agents-shipgate verify --workspace <smart_closer> --config shipgate.yaml --no-base --ci-mode advisory --format json
```

```
merge_verdict           : insufficient_evidence     ← §D precondition met
trust_root_touched      : True
policy_weakened         : True
policy_weakening_proven : False
control.state           : review_publishable
control.next_action.kind: review        ← not the declaration route
```

`report.json` nonetheless carries all thirteen questions with `authorable_by` resolved (10 `human`, 3 `coding_agent`). The route's inputs exist; the route is not published.

**Manifestation 2 — the loop dead-ends after being followed.** Commit the manifest, re-run with a base, and the route now appears correctly:

```
next_action.kind   : repair
next_action.command: apply-patches --from …/report.json --kinds declare_action --confidence high --apply
next_action.why    : "Apply the 3 declaration(s) this scan derived from its own evidence,
                      commit them to this branch, and re-run verification."
```

Run exactly that command — it writes three `external_communication` rows with `basis:` pins, correctly declining the nine it cannot answer. Then re-run `verify` as instructed:

```
control.state              : human_review_required
control.permissions        : {"edit": false, "commit": false, "push": false,
                              "update_pr": false, "merge": false, "report_complete": false}
control.allowed_next_commands: []
control.next_action.kind   : stop
trust_root_touched         : True
policy_weakened            : False          ← not a weakening
```

The agent was instructed to commit, performed the authorized edit, and is now forbidden to commit it, with no route forward for either actor.

## Why this matters beyond ergonomics

§D exists to dissolve the pincer recorded on `github/github-mcp-server#3076`: an agent's capability change stays invisible unless the manifest is updated, and the manifest is a trust root the agent must not touch. The resolution was to decide authorship by **content** — an evidence-consistent write is fine, a weakening is not. That rule is implemented correctly at the patch level (`declare_action` refuses conflicting answers, writes only into silent fields). It is then overridden at the routing level, where any trust-root touch escalates identically regardless of whether the content weakened anything.

The practical effect is that the loop cannot complete a cycle: the only path that produces the state §D wants is one where the agent never edits the manifest, which is the thing §D was built to let it do.

## Expected

- A trust-root edit whose `policy_weakened` is `False` and whose content is an evidence-consistent declaration resolves to `review_publishable` — commit/push/update_pr permitted, `merge` and `report_complete` denied, human merges.
- The `declare_action` route is reachable on first adoption. If an unreviewed manifest must not be extended by an agent, the route should say so by name rather than be silently absent while its inputs are published.

## Acceptance

- [ ] Applying the route's own command and re-running `verify` yields `review_publishable` with `commit: true`.
- [ ] A manifest edit that *does* weaken the gate still yields `human_review_required` (regression).
- [ ] On a first-adoption run with agent-answerable questions, either the route is published or `control` names why it is withheld.

---

Found 2026-08-24 on the sixth adoption walk of `adk-samples#1745`, following the control loop's own instructions. Everything else in the RFC verified working on the same walk — questionnaire ordering (#419), per-source authority, `basis` drift pinning, adoption rungs, control packs, and `declare_action` content rules. Walk numbers: https://github.com/ThreeMoonsLab/agents-shipgate/issues/410#issuecomment-5401529032

pengfei-threemoonslab:
Closes #429.

> **Round-4 review response** (blocking review on `8a6d71a1`). All seven findings are fixed, each reproduced first, and the branch is rebased onto `3e8f1cb7`. The continuation now works on every supported shape — first adoption, an already-committed manifest, a scoped monorepo manifest, and the applier's field-fill patch — and first-party output validates against its own published schemas.
>
> **Responses**
>
> **[P1] Merge-base comparison for the introduction proof** — fixed. The proof asked the base *tip* while the diff and removal guard were taken from the merge base, so a diverged history where `main` deletes a gate the merge base still carries read as an adoption. `_introduction_comparison_ref` now returns the effective comparison for both modes, and both the presence check and the whole-tree probe use it. `test_a_diverged_base_tip_does_not_make_an_edit_an_adoption`.
>
> **[P1] Resolve loaded packs relative to `report.manifest_dir`** — fixed. `_repository_relative` takes the anchor a value was recorded against; report-owned paths get `report.manifest_dir`, CLI paths keep their own. `test_a_scoped_pack_path_is_seen_in_the_repository_spelling`.
>
> **[P1] Scoped receipt paths in the verifier's coordinate system** — fixed, and the root cause was worse than the report: `_repository_relative` resolved a *relative* value bare first, and `Path("shipgate.yaml").resolve()` answers against the **process directory**. Run from the repository root it relativized cleanly, so the check passed on a path nobody had written and failed on the real one. A relative value is now resolved against its anchor, then the repository root, never bare. `test_the_continuation_survives_a_scoped_manifest` and `test_a_relative_recorded_path_is_never_resolved_against_the_process`.
>
> **[P1] Do not anchor the before state to mutable `HEAD`** — fixed for the absent case, which is the one that dead-ended: a first adoption has no earlier manifest for any ref to carry, so the reader no longer looks for one and the *introduction proof* carries the claim there instead. The recorded before-digest is deliberately not consulted on that branch — it names bytes nothing committed, and a value only the writer knows is not an anchor. The comparison ref is also the effective one now, so a worktree run with a base reads the merge base, where the pre-adoption manifest still lives. `test_a_first_adoption_can_still_publish_its_proposal`, `test_an_absent_before_state_needs_the_introduction_proof`.
>
> **[P1] Accept the field-fill patch shape** — fixed. A parsed manifest spells "silent" as a present key with a `None` value, so a raw dict comparison read the authorized fill as a changed answer. The rule is now: every earlier row survives in a head row that agrees with it everywhere it *spoke*, and a field it already answered may not change. `test_the_continuation_accepts_a_filled_row`, `test_a_changed_answer_is_not_a_filled_blank`.
>
> **[P1] Published schemas admit the exception** — fixed in all three, written as part of the `if` condition rather than an exception in the consequent, and phrased so a payload that *omits* the field still matches and is still held to the original requirement. `test_a_published_continuation_validates_against_its_own_schemas` runs a real continuation and validates all three artifacts; `test_a_blocked_run_without_a_continuation_still_fails_the_schema` is the dual.
>
> **[P2] Current-contract schema targets** — fixed: `0.15`, and the v0.15/v5/v8 labels now link to their own files. `llms-full.txt` regenerated.
>
> One thing I removed rather than kept: the applier briefly shelled out to `git` to record whether the manifest was tracked. `tests/test_adapter_static_only.py` forbids that in a scanner source, and it was unnecessary — the reader can tell from the ref it already reads.

## The three walls

§D's loop had three separate walls between an agent and a declaration a human could review. Each round of review found the next one.

**1. The route was unreachable on a first adoption.** It was refused on `capability_review.policy_weakened` — the *fail-closed routing flag*, raised whenever the direction of a policy change could not be established. Establishing it means proving no file in the tree parses as a manifest under any name, and that proof reads the tree: one blob past the probe's per-candidate bound ends it, and `google/adk-samples` carries **35**. So the run with all thirteen questions open was the one run that could never be offered the route that answers them.

**2. The route named a step the protocol forbids.** `apply-patches` edits `shipgate.yaml`, so `agent control` refuses with `workspace_changed` the instant it succeeds. "Apply, commit, re-run" — and the "commit and publish first" I briefly replaced it with — both spend permissions computed against a manifest that no longer exists.

**3. The proposal could never reach review.** The rerun is a fresh decision, and a declaration that makes a risk judgeable turns it `blocked`, which authorizes nothing.

## What this changes

1. **A separate, cheaper, sound introduction fact.** The route asks: does this diff introduce the gate it is judged by? The configured manifest is in the evaluated diff, absent at the comparison ref, the diff deletes and renames away **nothing at all** (suffix-agnostic, over the comparison the run actually evaluated), and every policy input this run **resolved** — the fixed globs, `checks.policy_packs[].path` at any legal path, and `--policy-pack` / `--baseline` — is that manifest. A proven weakening refuses regardless. The whole-tree probe, `policy_weakened`, the verdict and the adoption wording are untouched: only who may draft the blanks moves.

2. **Rerun, then publish** — in the instruction, the typed `expects`, and the bundled prompt. When the refresh refuses, the recovery it advertises is the producing run's own exact local rerun, recovered from the artifacts the refusal had already hash-validated.

3. **The declaration continuation** (`declaration-continuation.json`, `shipgate.declaration_continuation/v1`). `apply-patches` leaves a receipt pinning the manifest by **byte** digest on both sides of the write. A later working-tree run honours it only when both digests still match *and* the two manifests differ by nothing but added `action_surface.actions` rows. On that proof the blocked run is `review_publishable` — `edit`/`commit`/`push`/`update_pr`, `merge` and `report_complete` still denied. Without a receipt, `blocked` authorizes nothing exactly as before.

4. **A withheld route names itself**, with blockers checked before the verdict so the documented cause is one a run emits.

Contract 25 → 26; verifier 0.14 → 0.15, handoff v7 → v8, verify-run v4 → v5 (each carries the `declaration_continuation` boolean the shared publication invariant reads).

## Responses to the round-3 review

**[P1] Include dynamically configured policy inputs in this proof — fixed.** Reproduced your fixture exactly (oversize blob forcing the probe closed, critical `org-rules.yml`, `shipgate.yaml` referencing it while emptying it): `agent_action_required` / `confirm_declarations` with commit/push. The packs now come from the run's own record of what it loaded — `report.loaded_policy_packs[].path` — rather than a glob set or a second resolution, plus the CLI `--baseline` / `--policy-pack` / `--diff-from` inputs. A run with no report cannot enumerate them and fails closed. That fixture is now human-routed; `test_an_introduction_that_also_empties_a_referenced_pack_is_refused`.

**[P1] Check renames over the actual evaluated diff, suffix-agnostically — fixed.** Both holes reproduced. `removes_a_yaml_file` is no longer what the route rests on: a new `removes_any_tracked_path` refuses on *any* deletion or rename-away, over the comparison the run evaluated (worktree overlay included). No suffix list, no parsing, no size bound — the cost is that an adoption PR which also deletes a file loses its drafting route, which is the fail-closed direction you asked for. `test_a_staged_rename_onto_the_configured_path_is_refused` (your `--base main` staged case) and `test_a_json_named_gate_rename_is_refused`.

**[P1] The fresh control still cannot publish the proposal — fixed, via the continuation.** You are right that `configured_gate_introduced` does not help here: your fixture's manifest is already committed, so it is not an introduction. Hence the receipt. `test_the_drafted_proposal_reaches_review` asserts the full permission vector rather than `merge == false` alone — the gap you named in `test_the_loop_ends_with_the_agent_out_of_moves_and_a_named_residue`.

  The receipt is **provenance, not a signature**, and the bound is deliberate: anyone who can write the manifest can write a receipt whose digests match it. What they cannot do is make the delta *parse* as declarations. `test_a_forged_receipt_cannot_publish_a_loosened_gate` mints exactly that forgery over a lowered severity override and asserts publication is refused. So a forged receipt buys putting a manifest change in front of a person — which is what it is for — never a loosened gate.

**[P2] Carry the exact rerun through the stale-pointer boundary — fixed.** A currency refusal is not an integrity one: the pointer and every artifact it binds were hash-validated in the same pass before the workspace comparison ran. `CurrentControlUnavailable` now carries that validated capture, and `agent control` recovers `fix_task.verification_command` from it — the producing run's own `--config … --workspace … --out … --no-base`. It falls back to a command rebuilt from the request when the set could not be validated. Verified end to end: the advertised recovery is no longer `--base origin/main --head HEAD` with the default reports directory.

## Verification

- Full suite green, plus `tests/test_adapter_static_only.py` and the perf split separately, on the final committed tree. `ruff check .` clean.
- **Perturbation sweep, 8 guards across the three rounds, every one confirmed non-vacuous** by breaking it and watching a named test fail. This round: the `configured_gate_introduced` exemption, the absent-at-ref check, the removal check, the policy-input isolation, the blocker order, the receipt's after-digest, the semantic delta comparison, and the publishable carve-out.
- Walked end to end on an adk-samples-shaped fixture: route offered → `apply-patches` → `agent control` correctly refuses (4) and names the right rerun → re-run → `blocked`, **publish-only**, control current again.


src/agents_shipgate/cli/verify/git.py
```diff
@@ -1021,14 +1041,33 @@ def carries_manifest_like_yaml(
...
+            # runs once the *configured* manifest path is already proven absent
+            # on the ref, so a skipped blob is by construction not at that path
+            # — and no scan loads a manifest from any other path. A blob this
+            # size is also one `read_file_at_ref` refuses at this same
+            # constant, so nothing here could be read as a manifest at a ref
+            # anyway. The rename case an oversize gate could otherwise hide has
+            # its own fail-closed guard in ``removes_a_yaml_file``, and a file
+            # under a canonical manifest name returned ``True`` above before
+            # its size was ever consulted.
+            continue
```
pengfei-threemoonslab: [P1] Keep oversized arbitrary-name manifests fail-closed

This `continue` treats every blob over 512 KiB as absent, but `load_manifest` accepts YAML content under any filename. `protected_paths` covers only the currently configured path, while `removes_a_yaml_file` catches only `.yaml`/`.yml`. I reproduced a base that retains a valid padded `old-gate.json` and a worktree that adds `shipgate.yaml`: the probe returns `False`, `_manifest_introduced` becomes `True`, `policy_weakened` becomes `False`, and control grants `confirm_declarations` with commit/push authority. The same prior gate below the size limit stays human-routed. An unread candidate cannot prove that no prior manifest exists; please return `None` for this ambiguous case or add a sound prior-manifest identity mechanism, and cover the oversized + custom-name cross-product.

src/agents_shipgate/cli/verify/fix_task.py
```diff
@@ -430,10 +490,16 @@ def _declaration_instructions(
...
     out = [
+        # Kept under ``MAX_ENVELOPE_PROSE_BYTES``: this sentence *is*
+        # ``control.next_action.why``, and the envelope truncates prose past
+        # that bound — which would cut the ordering clause this route depends
+        # on right off the end.
         f"Apply the {len(drafts)} declaration(s) this scan derived from its own "
-        "evidence, commit them to this branch, and re-run verification. Each "
-        "one restates an effect the scan already read; the merging human "
-        "reviews them as part of the manifest change.",
+        "evidence, commit and publish them on the authority this run grants, "
```
pengfei-threemoonslab: [P1] The publish-before-rerun sequence relies on stale authority

The exact `apply-patches` command edits `shipgate.yaml`. Contract v20 requires `agent control` to be refreshed after every Shipgate command/worktree mutation and before commit, push, or PR update. I reproduced: initial `confirm_declarations` has commit/push true; applying the emitted command succeeds; the mandatory refresh then exits 4 with `workspace_changed` and says the prior decision never saw `shipgate.yaml`. Rerunning verify restores currency but reaches the documented blocked/human route. Thus the old run cannot legally authorize the subsequent commit/publication. This needs a protocol-level continuation/atomic publication mechanism or a fresh post-apply control generation, not only reordered prose. If this sequencing survives, the typed `expects` should also model publication rather than only a local commit.

prompts/fix-top-finding.md
```diff
@@ -14,7 +14,7 @@ You are working in a repo with `shipgate.yaml` already in place. Run a scan and
...
...s `next_action.authorable_by: "coding_agent"` with `suggested_patch_kind: "declare_action"`, and `verify` routes it as `control.next_action.kind: "confirm_declarations"` — an exact `apply-patches` command plus the list of questions, each tagged. Run that command, commit and publish the result, and **only then** re-run verification: the route grants `edit`/`commit`/`push`/`update_pr` but never `merge`, and those grants belong to the run that issued them — the re-run judges the manifest you just wrote, and a declaration that makes a risk judgeable is exactly what can move the verdict to one only a person may clear. A human reviews the manifest change at the PR, because writing the manifest touches the trust root. Everything in the bullets above still holds for every other row: a question tagged `"human"`, any authority or `agent_bindings` block, an `override`, and a `declaration_drift` row asking someone to re-confirm an answer are still yours to surface, never to write. Do not fill a blank the scan left, do not weaken a declaration the manifest carries, and do not reconstruct the edit by hand when the route is absent — its absence is the answer, and `control.reason` names the cause.
```
pengfei-threemoonslab: [P2] Do not promise that `control.reason` always names the withholding cause

This prompt makes the cause unconditional, but the implementation and new tests explicitly allow that sentence to be dropped by the shared 400-byte headline budget; `docs/agent-contract-current.md` even says not to branch on its presence. Qualify this as best-effort in all four rendered copies, or publish a dedicated typed field if the cause must be guaranteed.

pengfei-threemoonslab (commented):
Blocking review on PR #444. I found two independent P1 contract/safety regressions and one P2 agent-instruction inconsistency. The current CI and focused tests are green, but neither blocker is covered by the new tests: one needs the oversized/custom-name cross-product, and the other needs an end-to-end current-control refresh after applying the declaration patch. Please address these before merge.

Validation: reviewed exact head `68a5675e6fb4a90e27f7b4c481310004391108b8`; `ruff check .` passed; focused declaration/adoption/envelope/verify tests passed; static adapter/render tests passed; the full non-perf suite passed apart from a sandboxed isolated-build dependency fetch, and `tests/test_packaging.py` passed when rerun with network access. The deterministic verifier returned `review_required` / `review_publishable` for the protected prompt surfaces, as expected.

One additional non-blocking consistency issue: `declaration_route()` checks the verdict mismatch before `decision.blockers`, so the documented example wording `a blocker is open` is not reachable for a normal blocked report; either reorder those checks or use an emitted example.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -1635,6 +1655,109 @@ def _rerun_options(
...
+        )
+        for path in changed_files
+    ):
+        return False
+    ref = _introduction_comparison_ref(base_status=base_status, base=base, head=head)
+    if ref is None:
+        return False
+    if path_present_at_ref(git_root, ref, config_relative) is not False:
+        return False
+    policy_paths = set(touched_policy_surfaces(changed_files))
```
pengfei-threemoonslab: [P1] Include dynamically configured policy inputs in this proof

`touched_policy_surfaces(changed_files)` is only the fixed `_POLICY_SURFACES` glob set; it does not include a manifest-declared policy pack at an arbitrary legal path, nor CLI `--policy-pack` / `--baseline` paths. I reproduced a base containing a >512 KiB unrelated blob (forcing the old manifest probe to fail closed) and a critical blocking `org-rules.yml`, then added `shipgate.yaml` referencing that pack while changing the pack to `rules: []`. The result has `capability_review.policy_weakened=true` and a reason saying a coding agent cannot self-approve, yet this exemption yields `agent_action_required` / `confirm_declarations`, `human_review.required=false`, and commit/push/update-PR permissions. Pass every resolved manifest/CLI policy input into this predicate and fail closed if any of them changed; otherwise the claimed “only the configured manifest changed” fact is false.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -1635,6 +1655,109 @@ def _rerun_options(
...
+    if any(
+        not is_configured_manifest(
+            config_relative,
+            str(path).replace("\\", "/"),
+            workspace=git_root,
+        )
+        for path in policy_paths
+    ):
+        return False
+    removed = removes_a_yaml_file(
```
pengfei-threemoonslab: [P1] Check renames over the actual evaluated diff, suffix-agnostically

This guard misses moved existing gates in two ways. First, `removes_a_yaml_file()` recognizes only `.yaml` / `.yml`, while valid configured manifests can be JSON or otherwise custom-named: a staged `R089 old-gate.json -> new-gate.json` that also lowers a severity returns `configured_gate_introduced=true` and receives `confirm_declarations` despite `policy_weakened=true`. Second, when a worktree run also supplies `--base`, line 1738 compares `base...HEAD`, omitting the index/worktree overlay already represented in `changed_files`; a staged `R086 old-gate.yml -> shipgate.yaml` with `--base main` reaches the same route. Inspect deleted/renamed sources over the effective diff being evaluated, and parse arbitrary-name candidates or fail closed when they cannot be ruled out.

src/agents_shipgate/cli/verify/fix_task.py
```diff
@@ -430,8 +520,14 @@ def _declaration_instructions(
     ]
     human = [item for item in confirmation.questions if item.authorable_by == "human"]
     out = [
+        # Kept under ``MAX_ENVELOPE_PROSE_BYTES``: this sentence *is*
+        # ``control.next_action.why``, and the envelope truncates prose past
+        # that bound — which would cut the ordering clause this route depends
+        # on right off the end.
         f"Apply the {len(drafts)} declaration(s) this scan derived from its own "
-        "evidence, commit them to this branch, and re-run verification. Each "
+        "evidence, then re-run verification before committing or pushing: the "
```
pengfei-threemoonslab, line 528: [P1] The fresh control still cannot publish the proposal

Rerunning before publication fixes the stale-authority wording, but it does not close the loop. Using this PR's own `send_email` fixture, the declaration command succeeds and the correct fresh worktree verification makes the newly declared external-communication risk judgeable as `blocked`. `_derive_verifier_control()` excludes blocked decisions from `subject_evaluated`, so the fresh control is `human_review_required` with `edit`, `commit`, `push`, and `update_pr` all false. The agent therefore cannot put the generated manifest change into the PR for the “merging human” to review. `test_the_loop_ends_with_the_agent_out_of_moves_and_a_named_residue` asserts only `merge == false`, so it misses this dead end. This needs a protocol-level/atomic continuation or a narrowly publishable post-repair state that keeps merge denied while authorizing the proposal to reach review.

src/agents_shipgate/core/agent_control_envelope.py
```diff
@@ -267,16 +267,16 @@ def _declaration_route(
...
+            f"{len(drafts)} declaration(s) written into the manifest, and this "
+            # It promises no commit, and that is the correction, not an
+            # omission. The command edits ``shipgate.yaml``, so ``agent
+            # control`` refuses with ``workspace_changed`` the moment it
+            # succeeds, and contract v20 requires that refresh before any
+            # commit, push, or PR update. The permissions printed beside this
+            # route were computed against a manifest that no longer exists;
+            # promising a publication they cannot authorize is the dead end
+            # #429 reported, one turn later than it looked (#429 review).
+            "control superseded — re-run verification before committing"
```
pengfei-threemoonslab, line 279: [P2] Carry the exact rerun through the stale-pointer boundary

The typed action says to rerun, but carries only the apply command, and the compact envelope omits the underlying `allowed_next_commands`. Once apply invalidates current control, the mandatory `agent control` refresh exits `workspace_changed` and advertises a hard-coded PR verify (`--base origin/main --head HEAD`, default reports directory). That recovery is wrong for this working-tree-only route: it fails where `origin/main` is absent, otherwise scans committed `HEAD` and misses the uncommitted manifest edit, drops `--no-base`/policy/baseline options, and cannot refresh a custom `--out` directory. Preserve the exact local verification command in a provenance-bound continuation (or make apply+verify atomic) so the published route has a runnable recovery after its own command supersedes it.

pengfei-threemoonslab (commented):
Blocking re-review of exact head `3bb563847a1fd197f8bf63eee4da0d30c96bf54f`. The prompt-budget correction and blocker-before-verdict reordering resolve the prior P2 and non-blocking notes, but three P1 safety/protocol failures remain, plus one P2 continuation defect. Two independent holes let `_configured_gate_introduced` override `policy_weakened=true`; the apply/rerun sequence still leaves the generated manifest change unpublishable; and the required exact local rerun is lost when the apply command invalidates current control.

Validation: `ruff check .` passed; focused declaration/adoption/envelope/verify tests passed; static-adapter and rendered-instruction tests passed; the broad non-performance suite passed apart from eight isolated wheel-build setup errors caused by sandboxed dependency fetching, and all eight packaging tests passed when rerun with network access. I also ran four end-to-end repros for the findings below. Hosted checks are green. Deterministic PR verification returned `review_required` / `review_publishable`; its receipt reproduced successfully, with `update_pr=true` and `merge=false`. GitHub currently reports this branch as conflicting with the advanced `main`, so it also needs an update before merge.

Because the authenticated reviewer is also the PR author, GitHub only permits a `COMMENT` review here; please treat these as blocking findings rather than an approval.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -1684,14 +1779,135 @@ def _manifest_introduced(
         for path in changed_files
     ):
         return False
+    ref = _introduction_comparison_ref(base_status=base_status, base=base, head=head)
+    if ref is None:
+        return False
+    return path_present_at_ref(git_root, ref, config_relative) is False
```
pengfei-threemoonslab, line 1795: [P1] Use the same merge-base comparison for the introduction proof

This checks configured-path absence at the base tip, while the worktree diff and removal guard are evaluated from the merge base. In a diverged history where the merge base contains a strict `shipgate.yaml`, `main` deletes it, and the feature worktree weakens that long-lived gate, verification reports `base_status=missing_manifest`, clears `policy_weakened`, and offers `confirm_declarations`. The gate was present at the actual diff base, so it is not an adoption. Please use the effective comparison ref/merge base consistently for path presence and the whole-tree adoption probe.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -1684,14 +1779,135 @@ def _manifest_introduced(
...
+
+    def _is_the_manifest(path: str) -> bool:
+        return is_configured_manifest(
+            config_relative, path.replace("\\", "/"), workspace=git_root
+        )
+
+    changed = {str(path).replace("\\", "/") for path in changed_files}
+    inputs = set(touched_policy_surfaces(sorted(changed)))
+    for pack in report.loaded_policy_packs:
+        inputs.add(_repository_relative(pack.path, git_root))
```
pengfei-threemoonslab: [P1] Resolve loaded packs relative to `report.manifest_dir`

`loaded_policy_packs[].path` is recorded relative to the manifest directory, but `_repository_relative()` treats a relative value as repository-root/CWD-relative. With `services/mailer/shipgate.yaml` loading `org-rules.yml`, the report says `org-rules.yml` while `changed_files` says `services/mailer/org-rules.yml`; emptying that loaded pack therefore still makes this helper return true and publishes `confirm_declarations` even though `policy_weakened=true`. Join report-owned pack paths to `report.manifest_dir` before relativizing them to `git_root`.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -1923,6 +2139,160 @@ def _stable_subject(subject: str) -> str:
...
+    try:
+        receipt = DeclarationContinuationV1.model_validate_json(
+            receipt_path.read_bytes()
+        )
+    except (OSError, ValueError):
+        return False
+    # The receipt names its own target, and it must be the gate this run is
+    # judged by. A receipt for some other file is not evidence about this one.
+    if not is_configured_manifest(
+        config_relative, receipt.manifest_path, workspace=git_root
```
pengfei-threemoonslab: [P1] Put scoped receipt paths in the verifier's coordinate system

The applier stores `manifest_path` relative to `report.manifest_dir`, whereas this comparison treats it as git-root-relative. For a scoped manifest at `services/closer/shipgate.yaml`, the receipt contains `shipgate.yaml`; the configured path here is `services/closer/shipgate.yaml`, so the exact apply/rerun route returns `declaration_continuation=false` and removes every publication permission. Store a git-root-relative identity or resolve the receipt path through its manifest directory before comparing.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -1923,6 +2139,160 @@ def _stable_subject(subject: str) -> str:
...
+    ):
+        return False
+    try:
+        current = config_path.read_bytes()
+    except OSError:
+        return False
+    if hashlib.sha256(current).hexdigest() != receipt.manifest_sha256_after:
+        return False
+    previous = read_bytes_at_ref(git_root, comparison_ref, config_relative)
```
pengfei-threemoonslab: [P1] Do not anchor the continuation's before state to mutable `HEAD`

This makes the handoff valid only during the immediate pre-commit rerun for an already-committed manifest. On first adoption there are no prior manifest bytes at `HEAD`, so the advertised apply/rerun path is immediately `blocked/human_review_required`. On an existing manifest the immediate rerun is publishable, but after the required commit the pointer is stale; its exact recovery rerun now sees the after bytes at `HEAD`, fails the before digest, and revokes `push`/`update_pr`. The continuation therefore cannot complete the mandatory commit-before-push workflow. Bind/validate the actual pre-apply state independently of the live head, including the absent-file case.

src/agents_shipgate/cli/verify/orchestrator.py
```diff
@@ -1923,6 +2139,160 @@ def _stable_subject(subject: str) -> str:
...
+        return False
+    head_surface = head.pop("action_surface", None) or {}
+    base_surface = base.pop("action_surface", None) or {}
+    if head != base:
+        return False
+    head_actions = head_surface.pop("actions", None) or []
+    base_actions = base_surface.pop("actions", None) or []
+    if head_surface != base_surface:
+        return False
+    if len(head_actions) <= len(base_actions):
```
pengfei-threemoonslab: [P1] Accept the supported existing-row field-fill patch shape

`_declare_action()` explicitly supports filling missing `effect`/`basis` fields on an existing action row. That exact authorized patch leaves the list length unchanged and changes the base row, so this predicate returns false; the rerun becomes `blocked/human_review_required` with no publication permissions. Validate that only previously absent derived declaration fields were filled without changing existing values, in addition to accepting appended rows.

src/agents_shipgate/schemas/verifier.py
```diff
@@ -732,6 +732,15 @@ class VerifierArtifact(BaseModel):
...
     diff_status: VerifierDiffStatus
+    #: Set when this run's trust-root delta is a declaration continuation:
+    #: ``apply-patches`` left a receipt, its two byte digests pin the manifest
+    #: on both sides of the write, and the delta parses as additions to
+    #: ``action_surface.actions`` and nothing else. It is the one fact that
+    #: lets a *blocked* decision authorize publication — putting the proposal
+    #: in front of a person — while ``merge`` and ``report_complete`` stay
+    #: denied. False everywhere else, so nothing about publication changes for
+    #: any other run (#429).
+    declaration_continuation: bool = False
```
pengfei-threemoonslab, line 750: [P1] Make the published schemas admit the new continuation exception

The Pydantic validators now allow a blocked decision to publish when this flag is true, but the JSON-Schema `allOf` conditions still unconditionally require a non-`blocked` decision whenever `update_pr=true` (the same stale condition exists in agent handoff and verify-run). An end-to-end valid continuation emits `verifier.json`, `agent-handoff.json`, and `verify-run.json` that each fail their newly published v0.15/v8/v5 schema. Update all three schema conditions and regenerate the frozen JSON files so first-party output validates against its advertised contract.

docs/agent-contract-current.md
```diff
@@ -270,17 +327,17 @@ Downstream repos generated with
...
 - Latest release: `v0.15.0`
 - In-tree runtime: `0.16.0b7` — see [pyproject.toml](../pyproject.toml)
-- Runtime contract: `25` (minimum control contract: `21`)
+- Runtime contract: `26` (minimum control contract: `21`)
 - Current report schema: `0.42` — [`docs/report-schema.v0.42.json`](report-schema.v0.42.json)
 - Current packet schema: `0.17` — [`docs/packet-schema.v0.17.json`](packet-schema.v0.17.json)
 - Current shared agent result schema: `agent_result_v3` — [`docs/agent-result-schema.v3.json`](agent-result-schema.v3.json)
-- Current verifier schema: `0.14` — [`docs/verifier-schema.v0.14.json`](verifier-schema.v0.14.json) (v0.13 and earlier stay frozen; `0.14` adds `authorable_by` to declaration questions and `fix_task.declaration_confirmation`)
-- Current verify-run schema: `shipgate.verify_run/v4` — [`docs/verify-run-schema.v4.json`](verify-run-schema.v4.json)
+- Current verifier schema: `0.14` — [`docs/verifier-schema.v0.15.json`](verifier-schema.v0.14.json) (v0.13 and earlier stay frozen; `0.14` adds `authorable_by` to declaration questions and `fix_task.declaration_confirmation`)
```
pengfei-threemoonslab: [P2] Point current-contract readers at the bumped schemas

This still calls the verifier schema `0.14`, the v0.15 label links to `verifier-schema.v0.14.json`, and the next line's verify-run v5 label links to v4; line 693 repeats the stale verifier target. `llms-full.txt` mirrors the same generated text. Update the versions/targets before regenerating the aggregate so consumers do not validate new artifacts against frozen prior schemas.

pengfei-threemoonslab (commented):
Blocking third-pass review of exact head `8a6d71a1b51789f617c2bcc76f1439c7bfffc982`. The four findings from the prior review are resolved on this head, but I found six new P1 protocol/control regressions and one P2 contract-documentation error. In particular, the new continuation works only for a root-level, already-committed manifest when the patch appends a row and verification is rerun before committing; several supported paths still dead-end, and the first-party continuation outputs do not validate against their newly published schemas. Please address the inline findings before merge.

Validation: hosted checks are green; `git diff --check`, `ruff check .`, the focused declaration/control/schema suites, and six independent end-to-end repros passed. The broad non-performance suite passed apart from the same eight isolated wheel-build setup errors caused by sandboxed dependency fetching, and all eight packaging tests passed when rerun with network access. Deterministic PR verification returned `review_required` / `review_publishable`; its receipt reproduced successfully with `update_pr=true` and `merge=false`. The branch is still reported `DIRTY` against the advanced `main`. Because the authenticated reviewer is also the PR author, GitHub permits only a `COMMENT` review here; please treat these findings as blocking rather than as an approval.

Issue #424, opened 2026-08-26
The declare_risk_tags repair cannot close the row it is printed on


pengfei-threemoonslab:
## What happens

`declaration_below_inferred_evidence` publishes two routes. The second one — *"Add `action_surface.actions[].risk_tags: [X]` so the X controls apply to this action"* — is the route the row itself names, and applying it exactly as instructed replaces the row with a **blocking `conflicting_effect_evidence`** instead of closing it.

Reproduced on `aa6bd618`; **re-confirmed byte-identical at `71254cb4`** (2026-08-29 — same repair kind, same instruction, same `conflicting_effect_evidence` after applying it). No branch needed:

```python
from agents_shipgate.core.domain import Tool, ToolRiskHint
from agents_shipgate.core.semantic_assessment import assess_tool_semantics, effect_repair
from agents_shipgate.schemas.manifest import ActionDeclarationConfig

tool = Tool.model_validate({
    "id": "t", "name": "send_email", "source_type": "mcp", "source_id": "s",
    "extraction_confidence": "high", "extraction": {"surface": "enumerated"},
    "risk_hints": [
        ToolRiskHint(tag="external_write", source="keyword", confidence="medium", basis="inferred_keyword"),
        ToolRiskHint(tag="destructive",   source="keyword", confidence="medium", basis="inferred_keyword"),
    ],
})

weak = ActionDeclarationConfig(tool="send_email", effect="external_communication")
print(effect_repair(assess_tool_semantics(tool, weak).effect))
# EffectRepair(kind='declare_risk_tags',
#              instruction='Add action_surface.actions[].risk_tags: [destructive] so the
#                           destructive controls apply to this action', ...)

repaired = ActionDeclarationConfig(tool="send_email", effect="external_communication",
                                   risk_tags=["destructive"])
print([i.kind for i in assess_tool_semantics(tool, repaired).effect.issues])
# ['conflicting_effect_evidence']   <- was declaration_below_inferred_evidence
```

The action goes from **not pass-eligible with a review-tier row** to **not pass-eligible with a blocking conflict**, and the message blames the reviewer's own manifest: *"declared effect is weaker than high-confidence source evidence"*, attributed to `action_surface_declaration`.

## Who hits this, and why it is first in the queue

[...]

PR #455, opened 2026-08-29
feat(release): approve and enforce an explicit pre-1.0 evidence bar (#341)

### LINKED ISSUE #341: release policy: decide the approved evidence bar for pre-1.0 tags
## Decision required

The v0.16 release workflow now enforces the 100-case safety-qualification policy that was introduced after v0.15. No prior release has exercised this gate end to end, and the currently labeled miner corpus is not large enough to produce the required qualification artifact.

This is a **product/trust-policy decision**, not yet an implementation issue. An agent must not choose a weaker evidence bar simply because the current corpus cannot satisfy the approved one.

## Current contract

The repository currently has one release policy:

- `production_safety_requirements()` defines exactly 100 cases across 28 profile × decision strata;
- the release verifier requires `qualification_tier == "beta"` and `production_qualified == true`;
- `QualificationTier` distinguishes `beta` from `test`; there is no latent, already-approved “production tier” to select at 1.0;
- distribution documentation describes the 100-case artifact as the protected release input.

The 32 rows under `benchmark/miner/results/*.labels.csv` are useful product measurements, but they are not interchangeable with qualification cases. Qualification additionally binds adjudicated labels and a terminal verifier receipt for each case.

## Product choices

A human product/security owner must approve one of these routes:

1. **Retain the 100-case production-beta policy.** v0.16 remains blocked while a separate P0 corpus-delivery issue produces the required independently labeled, adjudicated, receipt-bound artifact.
2. **Approve an explicit pre-1.0 policy.** Define a versioned policy with a smaller evidence-coverage requirement for 0.x tags while preserving the non-negotiable strictness invariants (including zero unsafe auto-passes). The implementation follows only after the thresholds, rationale, and promotion path are approved.

Do not infer the choice from the tag and do not rename the current 100-case policy after the fact.

## Acceptance criteria

- [ ] A named human product/security owner records the selected route and rationale.
- [ ] The applicable case counts, origin counts, agreement threshold, holdout rule, and zero-unsafe-pass invariant are explicit.
- [ ] The policy states which versions/tags it governs and how promotion to the 1.0 bar works.
- [ ] Documentation, schema vocabulary, qualification generator, release verifier, and runbook agree.
- [ ] A separate implementation/corpus-delivery issue is created after the decision; this decision issue does not silently implement a lower bar.
- [ ] The non-publishing rehearsal in #355 proves the chosen policy fails closed.

## Related

- #312 re-runs the smaller labeled product corpus after behavior changes; it is useful measurement, not a substitute for release qualification.
- #342, #343, #344, #355, and #356 are the other v0.16 release-readiness workstreams.

pengfei-threemoonslab:
Implements the decision recorded for #341. **Deliberately does not close it** —
the required failed-bar Release Rehearsal is still unrun, and that is the last
box on the issue. Close #341 once the rehearsal is linked.

## The decision

**Route 2 — an explicit pre-1.0 policy.** Recorded by **Pengfei Hu
(`pengfei-threemoonslab`), product/security, 2026-08-29**, in
[`docs/release-evidence-policy-decision.md`](docs/release-evidence-policy-decision.md).

The repository had exactly one release policy — 100 adjudicated, receipt-bound
cases, written as the claim `1.0` should make — enforced for every tag. No
corpus met it, so nothing published and evaluators kept installing `v0.15.0`:
an older, *less*-verified build than the one being withheld.

## What the new policy is

A second **named** policy, `pre_1_0`, governs `0.x` tags. It reduces how much
evidence a tag must carry. It reduces nothing about how that evidence is judged.

| | `beta` (production) | `pre_1_0` |
|---|---|---|
| Governs | `1.0`+ — and any tag, if offered | `0.x` only |
| Adjudicated cases | 100 | **56** |
| Strata | 28, weighted | **28, two each** |
| Qualifying origins | ≥ 40 (40%) | **≥ 23** (40%) |
| κ floor / holdout / unsafe auto-passes | 0.80 / 20% / 0 | **unchanged** |
| Per-case receipt, `static_only` | required / true | **unchanged** |
| Safe passes | ≥ 27/30 (90%) | **≥ 13/14** (92.9%) |
| Blocked / review / IE exact | 30/30, 19/20, 19/20 | **14/14, 14/14, 14/14** |

Two per stratum is the smallest allocation keeping all 28 profile × decision
cells non-empty *and* leaving each cell one tuning and one holdout case. A
zero-count cell would delete every observation of a profile/outcome pair — a
reduction in **strictness**, not coverage. Exact-match floors are the
production *rates* rounded up, so three of four land on 100%.

## How it is enforced

**The version decides the policy; the artifact never does.** Epoch 0 with major
0 admits `pre_1_0` or the stronger `beta`; everything else — including any
version that is not a complete PEP 440 version — admits `beta` only. Both gates
derive this independently, and an artifact naming a tier its version does not
admit is rejected **and then measured against the production policy**, so a bad
tier can never shrink what is checked. The same rule binds the producer, on both
`--policy-tier` and the `requirements=` keyword.

`production_qualified` keeps meaning "met the 100-case bar" —
`SafetyQualificationResultV1` refuses to *construct* an artifact that says
otherwise, so the producer cannot emit the inconsistency at all.

`shipgate.safety_qualification` advances **v4 → v5**: the new tier value and the
new cross-field invariant are grammar a v4 reader rejects. v4 stays readable;
the corpus and receipt-index envelopes deliberately do not move, because their
grammar did not change. Migration note in `STABILITY.md`.

## Review round 2 — three findings, all reproduced first

1. **The sealer restated a case count, not a policy.** It never checked the
   strata or the exact-match floors, so a 56-case `pre_1_0` artifact with 12/14
   safe passes passed the *sealing* gate while the exhaustive gate rejected it,
   and 56 identical rows in no stratum at all passed too — making the
   dependency-compromise boundary that file exists to hold decorative. It now
   restates a full `QualificationPolicy` per tier (strata, per-outcome floors,
   per-stratum holdout, origin and κ floors), re-derives them from the raw
   cases, and every field is bound to the real constructors by test. The two
   pre-existing sealer tests used 100 identical rows; they use conforming
   corpora now, which is what makes them meaningful.
2. **The version rule was anchored only at the start.** `0garbage`,
   `0.16.0garbage`, `0..1`, `0-`, `0x1` parsed as major 0 and bought the
   *cheaper* policy — the inversion of the documented fallback, in a helper both
   gates share. Now a complete PEP 440 parse; leading zeros still normalize.
3. **New grammar under the frozen v4 envelope** — fixed by the v5 bump above.

## Fail-closed proof

- `test_a_pre_1_0_artifact_publishes_a_0_x_tag_and_nothing_later` — the same
  bytes that pass `v0.16.0b7` fail `v1.0.0`.
- `test_the_sealer_reads_the_governing_policy_from_the_version_not_the_artifact`
  and `test_the_sealer_enforces_the_strata_and_floors_not_just_a_case_count`.
- `test_a_conforming_56_case_corpus_qualifies_a_0_x_wheel` — end to end: the bar
  is satisfiable, exits `0`, reports `production_qualified: false`.
- `test_only_a_complete_pep440_version_can_reach_the_cheaper_policy` — 11
  accepted, 20 rejected, including every malformed case from the review.
- `test_the_stdlib_policy_table_matches_the_named_policies`,
  `test_the_qualification_envelope_advanced_for_the_new_grammar`,
  `test_the_production_flag_cannot_disagree_with_the_tier`,
  `test_the_version_rule_binds_the_requirements_keyword_too`,
  `test_the_pre_1_0_policy_is_never_laxer_than_production_per_rate`.

Seventeen perturbations of the new guards have been run across two rounds; all
seventeen fail the suite.

## Follow-ups

- #456 — corpus delivery for the 56-case artifact and the 100-case 1.0 bar.
- [ ] Dispatch **Release Rehearsal** against an artifact that misses the bar,
      then close #341.

## Verification

`ruff check .` clean. Full suite green: `pytest -n auto -m "not perf"
--ignore=tests/test_adapter_static_only.py`, plus `tests/test_adapter_static_only.py`.

### DIFF
--- a/.github/workflows/release-verify.yml
+++ b/.github/workflows/release-verify.yml
@@ -488,18 +488,25 @@ jobs:
         # and confusion matrix) imports the project's pydantic schemas, so it
         # runs in the `policy` job where the project is installed. What is
         # restated here, with the standard library only, are the claims that
-        # actually delegate publication authority: beta/qualified/
-        # production_qualified, static-only, no failures, 100 cases and
-        # receipts, zero unsafe auto-passes, and the wheel name/version/digest
-        # binding. A signed-but-weakened artifact therefore cannot pass on the
-        # strength of its signature alone.
+        # actually delegate publication authority: a qualification tier the
+        # tag's version admits, static-only, no failures, exactly as many cases
+        # and receipts as that tier requires, zero unsafe auto-passes, and the
+        # wheel name/version/digest binding. A signed-but-weakened artifact
+        # therefore cannot pass on the strength of its signature alone. Which
+        # tier a version admits is in docs/release-evidence-policy-decision.md.
+        id: binding
         env:
           RELEASE_TAG: ${{ steps.candidate.outputs.release_tag }}
         run: |
+          set -euo pipefail
           python scripts/verify_qualification_binding.py \
             --qualification qualified-dist/safety-qualification.json \
             --wheel "${QUALIFIED_WHEEL}" \
             --tag "${RELEASE_TAG}"
+          # Read back only for the run summary, and only after the gate above
+          # accepted the tier. Nothing downstream trusts this value.
+          tier="$(python -c 'import json,sys; sys.stdout.write(str(json.load(open("qualified-dist/safety-qualification.json")).get("qualification_tier")))')"
+          echo "qualification_tier=${tier}" >> "${GITHUB_OUTPUT}"
 
       - name: Bind the qualified wheel to the tagged source tree
         # Without this the pipeline tests the checkout and publishes a wheel
@@ -613,6 +620,7 @@ jobs:
           WHEEL_FILENAME: ${{ steps.qualified.outputs.wheel_filename }}
           WHEEL_SHA256: ${{ steps.handoff.outputs.wheel_sha256 }}
           MANIFEST_SHA256: ${{ steps.handoff.outputs.manifest_sha256 }}
+          QUALIFICATION_TIER: ${{ steps.binding.outputs.qualification_tier }}
         run: |
           {
             echo "## Release candidate ${RELEASE_TAG} (${MODE})"
@@ -623,7 +631,7 @@ jobs:
             echo "| CHANGELOG section for the tag | pass |"
             echo "| Dependency locks match declarations | pass |"
             echo "| Qualification signature identity | pass |"
-            echo "| Production qualification policy | pass |"
+            echo "| Qualification policy (${QUALIFICATION_TIER}) for this version | pass |"
             echo "| Wheel bound to tagged source | $(python -c 'import json,sys; sys.stdout.write(json.load(open("provenance.json"))["provenance_mode"])') |"
             echo "| Correctness suite (-n auto, not perf) | pass |"
             echo "| Dependency audit | pass |"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -137,7 +137,19 @@
   binds its restated numbers to the real constructors so the two copies cannot
   drift. Along the way the exhaustive verifier stopped hard-coding `100` in six
   places and now derives every count, metric denominator and confusion-matrix
-  profile from the governing policy.
+  profile from the governing policy. Review turned up a **seventh**:
+  `benchmark/safety-qualification/README.md`, the corpus owner's runbook, which
+  decides what actually gets built — a change that left it behind would have
+  aimed the whole corpus effort at the wrong shape, which no gate can detect.
+  It now documents both policies, and its pre-existing claim that receipts must
+  carry report schema `0.40` (the gate pins `0.42`) is corrected.
+
+  **`production_qualified` is now unrepresentable when it disagrees with the
+  tier.** Every gate rejected such an artifact, but only after it had been
+  signed and handed to the release. `SafetyQualificationResultV1` refuses to
+  construct one, so the producer cannot emit it — and the version rule binds
+  the `requirements=` keyword too, not just `--policy-tier`, which previously
+  let a caller score a `1.0` wheel against the pre-1.0 policy and exit `0`.
 
 
 - **The declaration continuation: a drafted proposal can now reach the person
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -168,6 +168,32 @@
   (its vocabulary is a strict subset) and the corpus and receipt-index
   envelopes deliberately unmoved, since their grammar did not change.
 
+  **A second review round found the bump had a hole and the sealer had four
+  more.** Upgrading a legacy envelope *unconditionally* rewrote v4 to v5 before
+  anything looked at the tier, so a conforming pre-1.0 artifact could keep
+  claiming a v4 an old reader cannot parse — the exact combination the bump
+  existed to eliminate. A legacy envelope is now read only when the payload
+  uses the vocabulary that envelope can express, and both gates enforce the
+  pairing. In the sealer: floors count *matches*, so a case with a null
+  `actual_decision` merely failed to count toward its floor rather than being
+  rejected, and 56 rows sharing one id looked like 56 cases — case identity and
+  terminal decisions are now checked before any floor. `>= 0.80` admitted
+  `inf`, since the JSON literal `1e309` loads as a float that satisfies every
+  lower bound; κ and the origin count are now bounded on both sides and
+  required to be finite and integral respectively. And the report schema
+  version has no representation in `cases` at all, so the approved `0.42` could
+  be restated as `0.1` and still seal — the sealer now compares the artifact's
+  whole declared `requirements` block, field for field, against its
+  restatement of the policy.
+
+  **The 1-tuning/1-holdout split was documentation, not policy.** What is
+  enforced is a per-cell holdout *floor*; a corpus that marks more cases
+  holdout is accepted, deliberately — holdout evidence was never tuned on, so
+  more of it is stronger, and a floor on tuning cases would be a ceiling on
+  holdout. The decision document, the schema docstring and the corpus runbook
+  said "leaves each cell one tuning and one holdout case" as though that were
+  checked; they now say what is actually enforced and why.
+
 - **The declaration continuation: a drafted proposal can now reach the person
   it was drafted for.** (#429 review) `apply-patches --kinds declare_action`
   writes into `shipgate.yaml`, which is the trust root — so the control that
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,46 @@
 
   [#461]: https://github.com/ThreeMoonsLab/agents-shipgate/issues/461
 
+- **A `0.x` tag now has an evidence bar it can actually meet, and it is not a
+  weaker judgement.** (#341) The only release policy was 100 adjudicated,
+  receipt-bound cases — the claim `1.0` should make — enforced for every tag.
+  No corpus met it, so nothing published, and evaluators kept installing
+  `v0.15.0`: an older, less-verified build than the one being withheld.
+
+  A second **named** policy, `pre_1_0`, governs `0.x` tags: 56 cases, two in
+  each of the same 28 profile × decision strata, with the origin floor at the
+  same 40% share. Everything that decides whether evidence is *believed* is
+  byte-identical to the production policy — zero unsafe auto-passes per profile
+  and overall, a unique terminal verifier receipt per case, the 20% holdout
+  fraction, the κ floor, `static_only`, and full re-derivation by the verifier.
+  Every exact-match floor is the production rate rounded up, which at 56 cases
+  means three of the four land on 100%: a smaller corpus buys less tolerance
+  for error, not more. The route, the numbers, the rejected alternatives and
+  the promotion path are recorded by a named owner in
+  [`docs/release-evidence-policy-decision.md`](docs/release-evidence-policy-decision.md).
+
+  **The version decides the policy; the artifact never does.** Epoch 0 with
+  major 0 admits `pre_1_0` or the stronger `beta`; everything else — including
+  a version that will not parse — admits `beta` only. Both release gates derive
+  this independently, and an artifact naming a tier its version does not admit
+  is rejected *and* then measured against the production counts, so a bad tier
+  cannot shrink what is checked. `production_qualified` keeps meaning "met the
+  100-case bar": a `pre_1_0` artifact reports it `false`, and claiming
+  otherwise is itself a rejection.
+
+  **The sealing gate was the sixth definition site, not the fifth.** The
+  standing brief listed five places the bar is defined; `verify_qualification_binding.py`
+  — the standard-library gate that seals a release without importing the
+  project — hard-coded `REQUIRED_CASE_COUNT = 100` and `tier == "beta"`
+  independently. A change that moved the other five would have failed there,
+  at the last step before publication, with a bare case-count error. It is
+  per-tier now, and `test_the_stdlib_case_counts_match_the_named_policies`
+  binds its restated numbers to the real constructors so the two copies cannot
+  drift. Along the way the exhaustive verifier stopped hard-coding `100` in six
+  places and now derives every count, metric denominator and confusion-matrix
+  profile from the governing policy.
+
+
 - **The declaration continuation: a drafted proposal can now reach the person
   it was drafted for.** (#429 review) `apply-patches --kinds declare_action`
   writes into `shipgate.yaml`, which is the trust root — so the control that
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -151,6 +151,22 @@
   the `requirements=` keyword too, not just `--policy-tier`, which previously
   let a caller score a `1.0` wheel against the pre-1.0 policy and exit `0`.
 
+  **Review found three more, all in the parts that decide what publishes.**
+  The sealing gate restated only a *case count*, so it could not tell 56
+  correctly stratified cases from 56 identical ones, and accepted a corpus two
+  safe passes below its floor that the exhaustive gate rejected — the
+  dependency-compromise boundary it exists to hold was decorative. It now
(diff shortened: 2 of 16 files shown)


pengfei-threemoonslab (commented):
I found three blocking release-contract issues:

1. [P1] The stdlib sealer does not enforce the selected tier outcome and stratum thresholds (scripts/verify_qualification_binding.py:107-204).

The new mapping selects only a total case count. The sealer then checks count, receipts, runtime failures, and unsafe auto-passes, but never checks the required profile/outcome strata or the safe, blocked, review, and insufficient-evidence exact-match floors. I reproduced a schema-valid 56-case pre_1_0 artifact with 12/14 safe passes: verify_qualification_binding accepted it, while the exhaustive verifier rejected it for the safe-pass threshold. This defeats the stated dependency-compromise boundary of the stdlib sealer. The new test at tests/test_release_pipeline.py:1248-1282 also has the sealer accept 56 all-blocked cases, which do not satisfy the actual policy. Please restate the per-tier strata and floors in the stdlib policy table, derive them from raw cases, and bind those copied constants to the schema constructors in tests.

2. [P2] Malformed numeric-prefix versions receive the cheaper policy (scripts/_release_support.py:45-60).

_VERSION_PREFIX is anchored only at the start and release_version_is_pre_1_0 uses match(), so trailing invalid syntax is ignored. Values such as 0garbage, 0.16.0garbage, and 0..1 all return both beta and pre_1_0. I reproduced both release gates accepting a conforming 56-case artifact for wheel metadata Version: 0garbage and tag v0garbage. That contradicts the approved fail-closed rule that every unparsable version gets beta only, and both gates share this helper, so they fail together. Please validate the complete supported PEP 440 form before reading epoch/major and add malformed-leading-zero cases.

3. [P2] The new artifact grammar is emitted under the frozen v4 schema ID (src/agents_shipgate/schemas/safety_qualification.py:14,37,626).

The base v4 reader admits only beta/test, while this branch emits pre_1_0 and adds a new cross-field invariant without advancing shipgate.safety_qualification/v4. STABILITY.md:956-964 explicitly says the v4 safety envelope is frozen. A genuine pre_1_0 artifact therefore claims a schema ID whose existing reader rejects. Please advance the qualification envelope version and define the v4 migration/reader behavior.

Process note: the PR says Closes #341, but the issue and decision document still leave the required failed-bar Release Rehearsal unchecked. Please run and link that rehearsal before merge, or avoid closing #341 in this PR.

Validation on 0b7cf7b: 200 focused release/qualification tests passed; changed-file Ruff checks passed; git diff --check passed. Shipgate verify completed with release decision review_required and merge verdict human_review_required because release-verify.yml is a touched release trust root.

pengfei-threemoonslab:
All three findings confirmed and fixed in f2f8af5, plus the process note. I
reproduced each one before changing anything.

## 1 [P1] The sealer restated a case count, not a policy

Reproduced exactly as described. Against `verify_qualification_binding` at
`0b7cf7b`:

```
conforming             -> ACCEPTED
12/14 safe passes      -> ACCEPTED     <-- exhaustive gate rejects this
56 all-blocked         -> ACCEPTED     <-- no strata at all
```

You're right that this makes the boundary decorative, and right that my own new
test enshrined the weak shape. Two things I'd add: the gap predates this PR
(the sealer only ever checked `len(cases) == 100`), and both *pre-existing*
sealer tests used 100 identical rows too — so the fixture pattern was hiding it
in three places, not one.

`QUALIFICATION_CASE_COUNTS` is now a `QualificationPolicy` per tier carrying
strata, per-outcome exact-match floors, per-stratum holdout, and the origin and
κ floors. The sealer re-derives all of them from the raw `cases` array. The two
floors that genuinely cannot be derived there — origins and κ, since no case row
carries an origin or a label — are read from the summary and *named as such* in
the code rather than quietly skipped.

Same three artifacts now:

```
conforming             -> ACCEPTED
12/14 safe passes      -> rejected: passed exact-match floor failed: 12 of 14, need 13
56 all-blocked         -> rejected: case profile/outcome strata do not match the pre_1_0 policy; ...
```

`test_the_stdlib_policy_table_matches_the_named_policies` binds **every** field
to the real constructors — strata, case count, each floor, origins, κ, holdout
fraction, max unsafe passes, and `minimum_holdout()` against the verifier's own
`ceil`. All three sealer tests now build conforming corpora.

## 2 [P2] Malformed numeric-prefix versions

Confirmed — `0garbage`, `0.16.0garbage`, `0..1`, and also `0-` and `0x1`. This
is worse than a fail-open: it *inverts* the rule my own decision document
states, and as you note both gates share the helper.

Now a complete PEP 440 parse, anchored at both ends. Leading zeros still
normalize per the spec (`00.1` is pre-1.0; `01.0.0` is 1.0.0 and is not), and
the permissive leading `v` is deliberately not accepted — refusing it only ever
selects the stricter policy.
`test_only_a_complete_pep440_version_can_reach_the_cheaper_policy` covers 11
accepted and 20 rejected forms, including every case you named plus the
leading-zero pairs.

## 3 [P2] New grammar under the frozen v4 envelope

Agreed, and this one I should have caught: my own memory of this repo is that
frozen unions widen silently. `shipgate.safety_qualification` now advances
**v4 → v5**.

Reader behaviour is defined in `STABILITY.md`: v5 accepts and reads v4, because
every v4 payload *is* a v5 payload with identical meaning — v4's tier vocabulary
is a strict subset, and the producer that emitted v4 always satisfied the new
`production_qualified`/tier invariant. Upgrading grants nothing, since the
payload still has to satisfy every v5 rule and `schema_version` is
attacker-controlled text either way. The reverse direction is the reason the
version had to move. v1/v2 still read, v3 still does not — both unchanged.

The corpus and receipt-index envelopes deliberately **stay at v4**: their
grammar did not change, and these versions track grammar rather than release
batches. That deviates from the v3 → v4 note, which moved all three together, so
I've said so explicitly in the migration note.

## Process note

Taken. The PR no longer says `Closes #341` — it says it implements the recorded
decision and deliberately leaves the issue open, with the rehearsal as the last
box. I can't dispatch workflows, so that one needs your hands; the deterministic
half is covered by
`test_a_pre_1_0_artifact_publishes_a_0_x_tag_and_nothing_later` and
`test_the_sealer_enforces_the_strata_and_floors_not_just_a_case_count`.

## Verification

`ruff check .` clean. Full suite green (`pytest -n auto -m "not perf"
--ignore=tests/test_adapter_static_only.py`, plus `test_adapter_static_only.py`).
Seven perturbations of the new guards — sealer skipping strata / floors /
holdout / origins, the regex losing its end anchor, the stdlib table drifting
one floor, the envelope reverting to v4 — each fail the suite. Seventeen across
both review rounds.

`release-verify.yml` is still touched, so shipgate's own verdict on this branch
stays `review_required` by design.

pengfei-threemoonslab (commented):
Re-review of current head `f2f8af527`.

The malformed-version finding is fixed: the parser is now fully anchored and fail-closed (including differential checks against `packaging.version.Version`). The producer also emits the new v5 envelope, and the stdlib sealer now re-derives the advertised strata and exact-match floors. I still found the following blocking issues in received-artifact validation:

1. **[P1] A genuine `pre_1_0` artifact can still claim the frozen v4 schema.** In `src/agents_shipgate/schemas/safety_qualification.py:647-672`, the field validator rewrites v4 to v5 before it can see `qualification_tier`. Relabeling a conforming 56-case artifact from v5 to `shipgate.safety_qualification/v4` is accepted by `SafetyQualificationResultV1` and the exhaustive release verifier; the stdlib sealer also accepts it because it never checks `schema_version`. This preserves the incompatible v4/`pre_1_0` combination the bump was intended to eliminate, so old v4 readers can still receive an artifact they cannot parse. Upgrade legacy envelopes only when their payload uses the legacy vocabulary, and enforce the same schema/tier pairing in the raw sealer.

2. **[P1] The stdlib sealer still accepts non-terminal or duplicate case evidence.** Around `scripts/verify_qualification_binding.py:194-219`, floors are counted without first requiring a unique, nonblank case id and a terminal `actual_decision`. Changing one expected-passed case to `actual_decision: null` leaves 13/14 safe exact matches, so the sealer accepts the artifact while the exhaustive gate rejects “a case without an actual verifier decision.” Likewise, setting all 56 case ids to the same value while retaining distinct receipt digests passes the sealer and fails the exhaustive gate. Validate case identity uniqueness and require every actual decision to be one of the four terminal decisions before applying the floors.

3. **[P1] Non-finite κ satisfies the newly added sealer floor.** `scripts/verify_qualification_binding.py:80-83,251-263` treats any non-bool float as a number and checks only `>= 0.80`. The valid JSON number `1e309` loads as `inf`, passes the sealer, and is rejected by the exhaustive verifier's `<= 1.0` invariant. Require a finite κ in the valid range (and an actual integer, bounded by the population, for `qualified_origin_cases`).

4. **[P1] The dependency-free gate does not bind the report-schema requirement.** `QualificationPolicy` in `scripts/_release_support.py:48-68` omits `required_report_schema_version`, and the sealer ignores the artifact's `requirements`. Changing only that claim from the approved `0.42` to `0.1` passes the sealer while the exhaustive gate rejects the policy mismatch. Since report schema `0.42` is explicitly part of both named policies, restate/bind it alongside the other copied fields and reject a raw requirements block that disagrees.

5. **[P2] The documented one-tuning/one-holdout layout is not enforced.** The source-of-truth decision says each two-case cell leaves one tuning and one holdout case, but `scripts/run_safety_qualification.py:704-726` and both verifiers enforce only the minimum holdout count. A 56-case corpus with every case marked `holdout` receives exit 0 and is accepted by both gates, with every stratum reporting `tuning_count=0, holdout_count=2`. Either enforce at least one tuning case per cell or revise the decision/runbook statements that promise the 1/1 split.

Integration note: the branch now conflicts with current `main` in `CHANGELOG.md` (`git merge-tree --write-tree origin/main HEAD` reproduces the conflict), so it also needs an update before GitHub can merge it.

Verification performed on this exact head:
- focused qualification/release-pipeline pytest suite: passed
- Ruff on changed Python surfaces: passed
- `git diff --check`: passed
- Agents Shipgate verifier: `review_publishable` / `human_review_required` because the PR changes a release trust root

GitHub will not let the authenticated author account submit a formal “request changes” review on its own PR, so this is posted as a comment review; the P1 items above are blocking in substance.

pengfei-threemoonslab:
All five confirmed and fixed in `1b868c7`, and the branch is rebased onto
`main` (`73aa994`) with the `CHANGELOG.md` conflict resolved — both Unreleased
entries kept, #424's first. `git merge-tree --write-tree origin/main HEAD` is
now clean.

## 1 [P1] A `pre_1_0` artifact could still claim v4

Correct, and my previous reply was wrong on exactly this point: I said
"upgrading grants nothing", and it granted precisely the thing the bump existed
to prevent. As a bare `schema_version` field validator it rewrote v4 → v5
before anything could see the tier:

```
pydantic  v4 + pre_1_0  -> ACCEPTED (read back as v5)
exhaustive gate         -> ACCEPTED, tier pre_1_0
sealer                  -> ACCEPTED (never read the envelope at all)
```

Now a model-level before-validator: a legacy envelope is upgraded **only when
the payload uses the vocabulary that envelope can express** (`beta`/`test`),
and otherwise raises. The sealer enforces the same pairing on raw JSON via a
shared `qualification_envelope_admits_tier`. All three reject it now; a v4+beta
artifact still reads as v5.

## 2 [P1] Non-terminal / duplicate case evidence

Reproduced both. The root cause is that the floors count *matches*, so a null
`actual_decision` is not a low score — it is a case that silently drops out of
its own floor, leaving 13/14. Case identity (unique, non-blank, string) and
terminal expected/actual decisions are now required **before** any floor is
applied. Also covers a non-terminal string like `"mergeable"` and a non-string
id.

## 3 [P1] Non-finite κ

Confirmed — `1e309` loads as `inf` and satisfies any `>=`. Both summary floors
are now bounded on both sides and type-exact: κ finite in `[floor, 1.0]`,
`qualified_origin_cases` a genuine `int` in `[minimum, case_count]`. Tested
against `0.79`, `inf`, `1.5`, `True`, `23.0`, `22`, `57`.

## 4 [P1] Report schema not bound in the dependency-free gate

Confirmed. Rather than restate that one field, `QualificationPolicy` now emits
`as_requirements_payload()` and the sealer compares the artifact's **whole**
declared `requirements` block field-for-field, including exact key set. The
binding test asserts that payload equals
`SafetyQualificationRequirementsV1.model_dump(mode="json")`, so a field added
to the model breaks the build until the stdlib copy restates it — which is a
stronger guarantee than the per-field asserts it replaces.

## 5 [P2] The 1-tuning/1-holdout layout

You offered both options; I took **revise the documentation**, because I think
enforcing it would be the wrong direction and I want to be explicit about that
rather than quiet.

A minimum on tuning cases is a **maximum on holdout**. Holdout evidence is
evidence the engine was never tuned on, so an all-holdout corpus is *stronger*
than a 1/1 one, and a release gate should never reject a corpus for being more
conservative than required. What is enforced is the per-cell holdout floor; the
docs claimed the expected layout as though it were checked. The decision
document, the schema docstring, the corpus runbook and #456 now state the floor
and say why there is deliberately no tuning floor, and
`test_a_corpus_with_more_holdout_than_required_is_accepted` pins the behaviour
so it is a decision rather than an accident.

The "two per cell" rationale is unchanged and still correct — at one case per
cell `ceil(1 × 0.20) = 1` forces that case to be holdout, so no cell could hold
a tuning case at all. It now says two cells leave *room* for the split rather
than that they produce it.

If you'd rather have the tuning floor enforced, say so and I'll add it — but
I'd want the decision recorded as deliberately capping holdout.

## Verification

`ruff check .` clean. Full suite green (`pytest -n auto -m "not perf"
--ignore=tests/test_adapter_static_only.py`, plus `test_adapter_static_only.py`).
Seven new perturbations — unconditional envelope upgrade, sealer skipping the
envelope pairing / terminal decisions / id uniqueness / the requirements block,
κ unbounded, the stdlib report-schema restatement drifting — each fail the
suite. Twenty-four across three rounds.

#341 stays open; the PR does not close it, and the failed-bar rehearsal is
still the last box.

PR #457, opened 2026-08-29
fix(semantics): a reviewed risk tag refines the row it sits in, it does not contradict it (#424)

### LINKED ISSUE #424: The declare_risk_tags repair cannot close the row it is printed on

pengfei-threemoonslab:
## Summary

- Fixes #424: the `declare_risk_tags` repair `declaration_below_inferred_evidence` publishes could not close the row it was printed on. Applying the row's own instruction verbatim replaced a review-tier row with a **blocking** `conflicting_effect_evidence` whose message blamed the reviewer's own manifest.
- Two branches of `_assess_effect` disagreed about what a reviewed `risk_tags` entry is. `claims_above_declared_effect` treats it as **covering**, deliberately and with a docstring saying so — a declared tag is policy-eligible, so it accounts for the observation *and* applies that category's built-in controls. The `contradictory` filter ran first and read the same claim as source evidence outranking the declaration.
- `contradictory` now excludes the two spellings of "declare this category as reviewed": `action_risk_tag_declaration` (the action row) and `risk_hint:manual` (`risk_overrides.tags`). They are named once as `REVIEWED_RISK_TAG_CLAIM_SOURCES` in `core/domain.py` and shared with the one other site that had the pair spelled inline.
- Same class already fixed one branch over in `_source_read_conflict`, for the same reason and with the same wording in its docstring. This branch never got the treatment.

### Blast radius, measured

Every declared effect × every one-, two-, and three-observation combination, applying the repair the engine publishes and re-resolving:

| route | exercised | did not close their row (before) | after |
|---|---:|---:|---:|
| `raise_effect` | 301 | 0 | 0 |
| `declare_risk_tags` | 390 | **281** | 0 |

First-time adoption was never affected — `propose_effect_declaration` always raises the effect or names a covering set. Only the post-declaration repair was broken, and only where the two published rank tables disagree, which is the case the two-route design exists for.

### Deliberately narrow

Not `_is_manifest_owned` wholesale. That predicate also covers `action_scope`, and #417 deliberately made a declared `crm.delete` grant bound the action's effect; excluding it would re-open that fail-open, and worse — the scope claim is policy-eligible, so it would then *cover itself* and raise nothing at all.

A declared tag refines the effect the same person wrote. A declared grant asserts an independent fact that bounds it. Protocol annotations, a source's own scopes, and typed provider facts keep contradicting a weaker declaration exactly as before — `test_a_declared_tag_never_launders_away_evidence_it_does_not_own` is the negative control, and it passes both before and after the fix.

### The gate verdict this moves

`effect: read` + `risk_tags: [destructive]` goes from **blocking** to **pass-eligible**. The issue asked for this to get its own adversarial pass; here is what that pass established:

- `conservative_effect` still resolves to `destructive`, the destructive claim is still policy-eligible, so `_control_effects` still applies the destructive controls.
- The two spellings now reach **identical** gate answers — same decision, same finding set — with and without the controls. Asserted in `test_a_declared_tag_obliges_exactly_what_declaring_the_effect_would`.
- Uncontrolled, `effect: read` + `risk_tags: [financial_write]` is `blocked` on `SHIP-ACTION-FINANCIAL-WRITE-CONTROL-MISSING` rather than on the shape of the manifest — a better answer to the same question.

**One P0 canary changed.** `manual_financial_tag_cannot_downgrade_to_read` encoded the old outcome; the `contradiction_evasion` category asserts `pass_eligible is False` for every case in it, and this case is no longer an evasion. Its slot is filled by the boundary the fix draws — `declared_delete_scope_cannot_downgrade_to_read`, which carries the same risk tag *plus* a declared `billing.delete` grant and stays blocking — so the catalog keeps its planned shape (64 cases, 12 in `contradiction_evasion`). The property the retired canary guarded is asserted in its post-#424 form by `test_a_reviewed_tag_transitions_the_action_instead_of_blocking_it`, in the same file.

### Guards

The issue's suggested guard, plus what it took to make it bite:

- `test_the_published_repair_closes_the_row_it_is_printed_on` — extended to three-observation combinations, now asserts **both** routes are exercised, and requires the whole effect dimension clean *and* pass-eligible rather than the absence of one kind. The old assertion is why the class shipped: the sweep already ran and already passed while the repair was swapping the row for a blocking conflict.
- `test_every_answerable_kind_has_an_answer_that_closes_it` — `_ROUND_TRIP_CASES` now holds a tuple of cases per kind so a kind that publishes two routes gets both walked, and each answer must leave the dimension clean.
- `test_pasting_the_published_tag_block_closes_the_row_end_to_end` — a real scan: raise the row, take the `next_action.declaration_template` it publishes, merge it into the action it names, rescan. Gaps empty, questions answered, the `basis` pin survives, and both categories' controls are what the run now reports.
- `test_the_tag_route_closes_the_row_it_is_printed_on` / `test_the_other_spelling_of_a_reviewed_tag_reads_the_same_way` — the issue's own reproduction and its `risk_overrides.tags` sibling.

Every new or strengthened guard was confirmed to fail with the fix reverted.

## Review round 1

Reviewing this change found the **same missing second route at two more comparisons**, both fixed here. The manifest reaches the effect dimension by two routes — the action row (`DECLARATION_CLAIM_SOURCES`) and `risk_overrides.tags`, which arrives as `risk_hint:manual` carrying a `reviewed_declaration` basis and *no* declaration source. Each of these excluded the first only:

1. **`SHIP-ACTION-EFFECT-DOWNGRADE-DECLARED` quoted the reviewer back at themselves.** It derives "the effect Shipgate inferred" by excluding the manifest's own claims. A reviewed `risk_overrides.tags: [destructive]` beside a source that says only `write` produced `inferred_effect: destructive` and the recommendation *"Set ... effect ... to destructive"* — telling the reviewer to declare the value they had already written. Now reports `write`, which is what the source said. Guard: `test_the_downgrade_row_never_quotes_the_reviewers_own_tag_back_at_them`.

2. **The row said the source agreed with a declaration when only the manifest did.** `declaration_below_inferred_evidence` appends *"; source evidence agrees with the declaration (…)"*, and a `risk_overrides.tags` entry matching the declared effect was named there — the manifest confirming itself, which the comment directly above that filter already forbade. Guard: `test_the_row_never_names_the_manifest_as_the_source_agreeing_with_it`, with a negative control proving `openapi_method` still corroborates.

`_is_manifest_owned` is promoted to `is_manifest_owned_effect_claim` and both sites ask it, so a fourth site cannot spell it a fifth way.

**Trust premise, now pinned.** Excluding `risk_hint:manual` from a source-evidence comparison is only safe because tool-published content cannot produce one. Two things make that true — `_validated_hint_basis` grants the `reviewed_declaration` basis to no other hint source, and no adapter writes that basis or a `"manual"` source — and neither was asserted anywhere, including by `_source_read_conflict`, which has relied on the same argument unstated since it was written. `test_only_the_manifest_can_write_a_reviewed_risk_hint` asserts both.

### Filed rather than folded in

Two findings from the same pass are out of scope and filed:

- **#460** — `effect_readings` counts a `risk_overrides` tag as something the scan *observed*. Same class, but excluding it moves the `basis` pin for every tool a `risk_overrides` selector touches and could weaken `propose_effect_declaration`. Both want their own adversarial pass.
- **#461** — a declared `effect: read` beside a positive risk tag publishes `reversibility: reversible` for a `financial_write` action, because `build_action` maps the policy-eligible `read` claim to a `read_only` tag. Pre-existing and byte-identical on `main` (verified), but this PR moves the wrong claim from reports that were blocked anyway into reports that can pass.

## Review round 2

All three findings reproduced first, then fixed. Both P1s were production defects, not test gaps.

1. **The tag repair still could not close its row on an already-tagged action.** `risk_tags` is one YAML key, so a block naming it replaces it — and the repair named only the newly uncovered category. A row reading `risk_tags: [financial_write]` was asked to delete the tag covering the financial reading, and the next scan reopened the row asking for it back: the #424 defect surviving in the case #424's own repair creates. `EffectRepair.risk_tags` is now the **complete value to write**, with `added_risk_tags` carrying what changed so the instruction and the value come from one place. The declared list is taken *verbatim* from a new `declared_risk_tags` entry on the declared-effect claim — it cannot be rebuilt from the tag claims, because `read_only`, `network_access` and `customer_data` map to no positive effect and produce none, the same lossiness the `declaration_drift` branch already refuses to guess through. Guard: the end-to-end paste test is parametrized over `already_declared` and still merges shallowly, the way a merge instruction is actually pasted.

2. **[#461](https://github.com/ThreeMoonsLab/agents-shipgate/issues/461) folded in rather than deferred.** `build_action` unions a risk tag for every policy-eligible effect claim, and `read` is not a category — it is the assertion that none apply. `effect: read` beside `risk_tags: [financial_write]` synthesized `read_only` on a `financial_write` action, and `derive_side_effect` reads that tag as positive evidence: `reversibility: reversible`, which that helper's own docstring says a declared read must not buy. Deferring was the wrong call, because this branch is what moves the wrong fact into an artifact that can pass. The equivalence guard now compares whole action and capability facts (equal facts are equal against every base — stronger than comparing two diffs) and separately asserts the claim lists still *differ*, so a flattened audit trail cannot pass as agreement.

3. **The contract overstated the equivalence.** A tag *adds* its category to the declared effect; raising the effect replaces it and drops whatever the old value obliged. `effect: external_communication` + a financial tag with approval/audit/idempotency is blocked on the missing confirmation, where `effect: financial_write` with the same controls is not. `docs/passed-verdict-contract.md`, `docs/manifest-v0.1.md` and the changelog now say the tag applies its category's controls *in addition*, with that example; the equivalence claim survives only where it is true — beside `effect: read`, which obliges nothing — and its test is renamed and scoped to that, with `test_a_tag_unions_obligations_with_the_declared_effect` as the negative control.

**Upgrade notes**, both in the changelog: an action spelled `effect: read` + a positive tag loses the `read_only` entry from its published `risk_tags`, so a stored lock shows one `risk_changed` / `narrowed` row on the first scan after upgrading — the safe direction, and correct. Carrying the declared tag list moves the declared-effect claim's `evidence` and `claim_id` for rows declaring both an `effect` and `risk_tags`; an `evidence_hash`-only change is classified `evidence_only` by design and carries no direction.

Round-1 finding [#460](https://github.com/ThreeMoonsLab/agents-shipgate/issues/460) stays filed — it moves the `basis` pin and changes what `propose_effect_declaration` must account for, which is a separate decision.

## Type

- [ ] Input adapter change
- [x] Documentation only

## Verification

Additional local checks run:

- `python -m pytest tests/ -n auto` — full suite green (exit 0).
- `python -m ruff check .`, `python -m compileall -q src tests`, `python scripts/generate_schemas.py --check` — all clean.
- Self-dogfood: `shipgate verify --config shipgate-self.yaml --base origin/main --head HEAD` then `shipgate agent control` → `control_state: complete`, `decision: passed`.
- Each new/strengthened guard re-run with `src/agents_shipgate/core/{semantic_assessment,domain}.py` reverted to `main`, to confirm it fails without the fix.

## Release-readiness notes

- [x] New or changed check IDs are documented in `docs/checks.md` — no check IDs added, removed, or renamed.
- [x] Report/schema changes are additive or documented in `STABILITY.md` — no schema change. No report field, gap kind, or check ID is added or removed; `report_schema_version` is unchanged. What changes is which of two already-published gap kinds a given manifest raises, documented in `CHANGELOG.md`, `docs/manifest-v0.1.md`, and `docs/passed-verdict-contract.md`.

### DIFF
--- a/docs/passed-verdict-contract.md
+++ b/docs/passed-verdict-contract.md
@@ -156,8 +156,13 @@ human review, not an executable Patch.
 A declaration may freely escalate past the evidence. Declaring an effect
 **weaker** than one the scan inferred raises
 `declaration_below_inferred_evidence` (v0.36+) and the action is not
-pass-eligible until a reviewer raises `effect` or acknowledges the difference
-with `actions[].override` (`evidence` + `reason`). An acknowledged override is
+pass-eligible until a reviewer accounts for the observation or acknowledges the
+difference with `actions[].override` (`evidence` + `reason`). Accounting for it
+is one edit, and the row names which: raise `effect` where one value covers
+every uncovered reading, otherwise name the uncovered categories as reviewed
+`actions[].risk_tags` — a declared tag is policy-eligible, so it accounts for
+the reading *and* applies that category's controls, reaching the same gate
+answer as declaring the effect outright. An acknowledged override is
 accepted and the action is pass-eligible again, but it is counted as a semantic
 review concern — like `unscoped` and `ambient` authority, it keeps human review
 mandatory, so a run carrying one is never `passed`.
--- a/docs/passed-verdict-contract.md
+++ b/docs/passed-verdict-contract.md
@@ -161,8 +161,13 @@ difference with `actions[].override` (`evidence` + `reason`). Accounting for it
 is one edit, and the row names which: raise `effect` where one value covers
 every uncovered reading, otherwise name the uncovered categories as reviewed
 `actions[].risk_tags` — a declared tag is policy-eligible, so it accounts for
-the reading *and* applies that category's controls, reaching the same gate
-answer as declaring the effect outright. An acknowledged override is
+the reading *and* applies that category's controls. A tag **adds** its category
+to the effect already declared; it does not replace it, and the obligations of
+both stand. Raising `effect` instead answers the row with a single category and
+drops whatever the previous value obliged, so the two edits are not
+interchangeable: `effect: external_communication` plus a financial tag owes
+confirmation as well as approval, audit, and idempotency, where
+`effect: financial_write` alone does not owe confirmation. An acknowledged override is
 accepted and the action is pass-eligible again, but it is counted as a semantic
 review concern — like `unscoped` and `ambient` authority, it keeps human review
 mandatory, so a run carrying one is never `passed`.
--- a/tests/test_declaration_questionnaire.py
+++ b/tests/test_declaration_questionnaire.py
@@ -655,8 +655,9 @@ def _rendered_gaps(tmp_path: Path, *, tools: list[dict], actions: list[dict]):
     return report, report.release_decision.evidence_coverage.evidence_gaps
 
 
+@pytest.mark.parametrize("already_declared", [[], ["financial_write"]])
 def test_pasting_the_published_tag_block_closes_the_row_end_to_end(
-    tmp_path: Path,
+    tmp_path: Path, already_declared: list[str]
 ) -> None:
     """The §E walk that found #424, as a scan the reader actually performs.
 
@@ -666,14 +667,20 @@ def test_pasting_the_published_tag_block_closes_the_row_end_to_end(
     own repair came back as a *blocking* `conflicting_effect_evidence` — and
     where the pin has to survive it, since a template that re-opens
     `declaration_drift` on the next scan has not closed anything either.
+
+    Parametrized over a row that already carries a covering tag, because
+    `risk_tags` is one YAML key and pasting replaces it. A template naming only
+    the newly uncovered category deleted the tag that was covering the
+    financial reading, and the next scan reopened the row asking for it back —
+    the same defect, in the case the repair itself creates (#424 review).
     """
 
     tools = [
         {
-            "name": "purge_and_notify_customers",
+            "name": "refund_and_purge_customer",
             "description": (
-                "Delete the customer record permanently and email the customer "
-                "about it."
+                "Issue a refund payment to the customer and permanently delete "
+                "their account record."
             ),
             "inputSchema": {
                 "type": "object",
@@ -683,11 +690,13 @@ def test_pasting_the_published_tag_block_closes_the_row_end_to_end(
             },
         }
     ]
-    declared = {
-        "tool": "purge_and_notify_customers",
+    declared: dict = {
+        "tool": "refund_and_purge_customer",
         "effect": "external_communication",
         "authority": {"mode": "none"},
     }
+    if already_declared:
+        declared["risk_tags"] = list(already_declared)
 
     def _scan(actions: list[dict], where: Path):
         where.mkdir()
@@ -708,24 +717,34 @@ def test_pasting_the_published_tag_block_closes_the_row_end_to_end(
         if item.kind == "declaration_below_inferred_evidence"
     )
     template = dict(gap.next_action.declaration_template or {})
-    assert template["risk_tags"] == ["destructive"]
+    # The whole value of the key, not the additions: the same list either way,
+    # so a reviewer who already answered half the question keeps that half.
+    assert template["risk_tags"] == ["financial_write", "destructive"]
     # `override` is the row's *other* route, offered in the same block; a
     # reviewer taking the first one does not also write an acknowledgement.
     template.pop("override", None)
-    merged = {**declared, **{
-        key: value
-        for key, value in template.items()
-        if key not in {"tool_id", "source_id"}
-    }}
+    # Pasted the way a merge instruction is pasted: each key it names replaces
+    # the key it lands on. Unioning here instead would test the reader's
+    # goodwill rather than the block.
+    merged = {
+        **declared,
+        **{
+            key: value
+            for key, value in template.items()
+            if key not in {"tool_id", "source_id"}
+        },
+    }
 
     answered = _scan([merged], tmp_path / "after")
     coverage = answered.release_decision.evidence_coverage
 
     assert [item.kind for item in coverage.evidence_gaps] == []
     questions = coverage.semantic_coverage.declaration_questions
-    assert questions.open == 0 and questions.answered == questions.total
-    # Closed by *applying* both categories' controls, not by silencing either.
+    assert questions.open == 0
+    assert questions.answered == questions.total
+    # Closed by *applying* every category's controls, not by silencing any.
     active = {finding.check_id for finding in answered.findings}
+    assert "SHIP-ACTION-FINANCIAL-WRITE-CONTROL-MISSING" in active
     assert "SHIP-ACTION-DESTRUCTIVE-ROLLBACK-MISSING" in active
     assert "SHIP-ACTION-EXTERNAL-COMMUNICATION-AUDIT-MISSING" in active
 
--- a/tests/test_declaration_questionnaire.py
+++ b/tests/test_declaration_questionnaire.py
@@ -491,52 +491,83 @@ def test_the_answerable_kinds_are_real_gap_kinds_and_belong_to_one_dimension() -
 _PIN = "<CURRENT_PIN>"
 
 
-#: One reachable configuration per answerable kind, with the declaration that
-#: is supposed to close it. `conflicting_effect_evidence` appears twice on
-#: purpose: it is raised about two different surfaces, and only one of them is
-#: a question a declaration can answer.
-_ROUND_TRIP_CASES: dict[str, tuple[dict, dict, dict]] = {
-    # kind: (tool kwargs, declaration that raised it or {}, the answer)
+#: Every reachable configuration whose row a declaration is supposed to close,
+#: grouped by kind. `conflicting_effect_evidence` appears twice on purpose: it
+#: is raised about two different surfaces, and only one of them is a question a
+#: declaration can answer.
+#:
+#: A kind carries more than one case when its row publishes more than one
+#: route out. `declaration_below_inferred_evidence` publishes two — raise the
+#: effect, or name the categories as reviewed `risk_tags` — and only the first
+#: was walked, so the second could swap the row for a *blocking*
+#: `conflicting_effect_evidence` and nothing here noticed (#424).
+_ROUND_TRIP_CASES: dict[str, tuple[tuple[dict, dict, dict], ...]] = {
+    # kind: ((tool kwargs, declaration that raised it or {}, the answer), ...)
     "missing_effect_evidence": (
-        {"source_type": "mcp", "annotations": {"mcp_server": True}},
-        {},
-        {"effect": "write"},
+        (
+            {"source_type": "mcp", "annotations": {"mcp_server": True}},
+            {},
+            {"effect": "write"},
+        ),
     ),
     "inferred_effect_only": (
-        {"risk_hints": "external_communication"},
-        {},
-        {"effect": "external_communication"},
+        (
+            {"risk_hints": "external_communication"},
+            {},
+            {"effect": "external_communication"},
+        ),
     ),
     "declaration_below_inferred_evidence": (
-        {"risk_hints": "external_communication"},
-        {"effect": "read"},
-        {"effect": "external_communication"},
+        # The `raise_effect` route.
+        (
+            {"risk_hints": "external_communication"},
+            {"effect": "read"},
+            {"effect": "external_communication"},
+        ),
+        # The `declare_risk_tags` route. No single effect covers both
+        # `destructive` and `external_communication` under both rank tables, so
+        # the row's own instruction is to name the uncovered category as a
+        # reviewed tag — the exact edit #424 found could not close the row it
+        # was printed on.
+        (
+            {"risk_hints": ("destructive", "external_communication")},
+            {"effect": "external_communication"},
+            {"effect": "external_communication", "risk_tags": ["destructive"]},
+        ),
     ),
     "conflicting_effect_evidence": (
-        {"annotations": {"httpMethod": "POST"}},
-        {"effect": "read"},
-        {"effect": "write"},
+        (
+            {"annotations": {"httpMethod": "POST"}},
+            {"effect": "read"},
+            {"effect": "write"},
+        ),
     ),
     "missing_authority_evidence": (
-        {},
-        {},
-        {"authority": {"mode": "none"}},
+        (
+            {},
+            {},
+            {"authority": {"mode": "none"}},
+        ),
     ),
     "conflicting_authority_evidence": (
-        {"auth": {"type": "oauth2", "scopes": ["a:write"]}},
-        {"authority": {"mode": "none"}},
-        {
-            "scopes": ["a:write"],
-            "authority": {"mode": "scoped", "auth_type": "oauth2"},
-        },
+        (
+            {"auth": {"type": "oauth2", "scopes": ["a:write"]}},
+            {"authority": {"mode": "none"}},
+            {
+                "scopes": ["a:write"],
+                "authority": {"mode": "scoped", "auth_type": "oauth2"},
+            },
+        ),
     ),
     # A pinned declaration whose pin names evidence this scan does not read.
     # The answer is the same declaration re-confirmed against what it reads
     # now — the one-line edit the row asks for (#410 §E).
     "declaration_drift": (
-        {"risk_hints": "external_communication"},
-        {"effect": "external_communication", "basis": "confirmed:0"},
-        {"effect": "external_communication", "basis": _PIN},
+        (
+            {"risk_hints": "external_communication"},
+            {"effect": "external_communication", "basis": "confirmed:0"},
+            {"effect": "external_communication", "basis": _PIN},
+        ),
     ),
 }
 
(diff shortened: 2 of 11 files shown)


tests/test_declaration_questionnaire.py
```diff
@@ -613,6 +655,81 @@ def _rendered_gaps(tmp_path: Path, *, tools: list[dict], actions: list[dict]):
...
+        return report
+
+    raised = _scan([dict(declared)], tmp_path / "before")
+    gap = next(
+        item
+        for item in raised.release_decision.evidence_coverage.evidence_gaps
+        if item.kind == "declaration_below_inferred_evidence"
+    )
+    template = dict(gap.next_action.declaration_template or {})
+    assert template["risk_tags"] == ["destructive"]
```
pengfei-threemoonslab: [P1] Please seed and preserve pre-existing `risk_tags` in this round trip. The production template contains only the newly uncovered tags. With financial + destructive observations and an existing declaration `effect: external_communication`, `risk_tags: [financial_write]`, this assertion yields only `[destructive]`; the shallow merge below replaces `financial_write`, and the next scan reopens `declaration_below_inferred_evidence` asking for `[financial_write]`. Unioning the existing and repaired tags closes the row. As written, the primary #424 failure still exists for already-tagged actions.

tests/test_evidence_backed_pass.py
```diff
@@ -383,13 +395,70 @@ def test_manual_positive_action_tag_cannot_hide_behind_read_effect(
...
+        }
+        return (
+            _gate({**base, "effect": "read", "risk_tags": ["financial_write"]}),
+            _gate({**base, "effect": "financial_write"}),
+        )
+
+    controlled_tag, controlled_effect = _spellings(controls=controls)
+    bare_tag, bare_effect = _spellings()
+
+    assert controlled_tag == controlled_effect
```
pengfei-threemoonslab: [P1] Comparing only the decision and check IDs misses a false public fact introduced into the newly pass-eligible state. In a fully controlled scan, `effect: read` + `risk_tags: [financial_write]` passes but emits action tags `[financial_write, read_only]` and capability `reversibility: reversible`; direct `effect: financial_write` emits only `[financial_write]` and `reversibility: unknown`. The action/capability hashes and diffs diverge too. Please fix the synthetic `read_only` projection (the weaker read claim must not survive beside a positive conservative effect) and make this guard compare action/capability facts and diffs, not only gate IDs.

docs/passed-verdict-contract.md
```diff
@@ -156,8 +156,13 @@ human review, not an executable Patch.
...
 `declaration_below_inferred_evidence` (v0.36+) and the action is not
-pass-eligible until a reviewer raises `effect` or acknowledges the difference
-with `actions[].override` (`evidence` + `reason`). An acknowledged override is
+pass-eligible until a reviewer accounts for the observation or acknowledges the
+difference with `actions[].override` (`evidence` + `reason`). Accounting for it
+is one edit, and the row names which: raise `effect` where one value covers
+every uncovered reading, otherwise name the uncovered categories as reviewed
+`actions[].risk_tags` — a declared tag is policy-eligible, so it accounts for
+the reading *and* applies that category's controls, reaching the same gate
+answer as declaring the effect outright. An acknowledged override is
```
pengfei-threemoonslab: [P2] This overstates the equivalence. A reviewed tag adds its category to the existing declared effect, so both categories' obligations remain; replacing the effect can drop the old category. For example, `effect: external_communication` + a financial tag, with financial controls but no confirmation, is blocked, while direct `effect: financial_write` with the same controls passes. Please say the tag applies the same controls for the named category, not that the whole gate answer is generally identical.

pengfei-threemoonslab (commented):
The resolver-side exclusion itself looks sound: reviewed tags stop self-contradicting, while protocol evidence, source/declaration scopes, and typed facts still block weaker effects. I found two correctness gaps that survive the new guards, though, so I do not think this is ready to land yet.

1. The published tag repair still cannot close its row when the action already has reviewed tags: the template contains only newly uncovered tags, so applying it as the new end-to-end test does replaces an existing covering tag and reopens the same gap on the next scan.
2. The new pass-eligible tag spelling can publish unsupported `read_only` / `reversibility: reversible` facts. A fully controlled run passes, but its action/capability facts and hashes differ from declaring the positive effect directly. This is the deferred #461 behavior, but this PR is what moves it into a passed artifact.

I also left a contract wording comment because a tag unions obligations with the existing effect; it is not generally the same whole gate answer as replacing the effect.

Validation: all changed semantic/questionnaire/action/evidence/P0 suites, Ruff, compileall, schema generation, packaging tests, and both Shipgate verifier routes completed cleanly. The findings below are independent adversarial reproductions that the current suite does not cover.

tests/test_declaration_questionnaire.py
pengfei-threemoonslab: Reproduced, and you are right that the failure is in production rather than only in the test — so I fixed it there and made the test paste the block the way a merge instruction is pasted, rather than unioning on the reader's behalf.

`risk_tags` is one YAML key, so a block naming it replaces it. `EffectRepair.risk_tags` is now the **complete value to write**, and a new `added_risk_tags` carries what changed, so the instruction's "so the X controls apply" clause and the template value come from one place and cannot disagree. The instruction reads `Set …` rather than `Add …` for the same reason.

The declared list is taken **verbatim** from a new `declared_risk_tags` entry on the declared-effect claim, not rebuilt from the tag claims: `read_only`, `network_access` and `customer_data` map to no positive effect and produce no claim, so a rebuilt list would silently delete what a reviewer wrote — the same lossiness the `declaration_drift` branch already refuses to guess through. That claim is the right carrier because its existence is the exact precondition for the row: no declared `effect`, no `declaration_below_inferred_evidence`.

On your fixture the template is now `[financial_write, destructive]` whether or not `financial_write` was already declared, and the `basis` pin is identical in both — so pasting verbatim closes the row and does not re-open `declaration_drift`. The end-to-end test is parametrized over `already_declared` in `[[], ["financial_write"]]` and still does a shallow merge; reverting only the union makes the seeded case fail.

One consequence worth naming: the new claim evidence moves that claim's `claim_id` and the capability `evidence_hash` for rows declaring both an `effect` and `risk_tags`. An `evidence_hash`-only change classifies as `evidence_only` and carries no direction, so it cannot read as a broadening. Noted in the changelog.

tests/test_evidence_backed_pass.py
pengfei-threemoonslab: Agreed, and folded in — you are right that this branch is what moves the wrong fact into an artifact that can pass, so deferring it to #461 was the wrong call.

Fixed at the source: `build_action` unions a risk tag for every policy-eligible effect claim, and `read` is not a category — it is the assertion that none apply, while `semantic_effects` is a *union*, so a row declaring `effect: read` beside `risk_tags: [financial_write]` contributed both. The `read` member is now excluded from that mapping. The conservative effect still contributes `read_only` where the action really does resolve to `read`, and a `read_only` risk *hint* from a source annotation still reaches the row through `_normalized_risk_tags`, so the honest all-read case is untouched.

The guard now compares whole `action_surface_facts` and `capability_facts` between the two spellings, not gate IDs. Facts rather than diffs on purpose: equal facts are equal against every base, which is the stronger statement. The one thing that must still differ is the claim list — that is the audit record of *where* each answer was written — so the test strips `claims` for the equality and then asserts the claim lists are not equal, rather than letting a flattened audit trail pass as agreement. Reverting only the `read` exclusion fails it.

One upgrade note, in the changelog: a stored lock covering such a row shows a single `risk_changed` / `narrowed` entry on the first scan after upgrading. That is the safe direction and it is correct — the action stops claiming to be read-only.

#461 is closed by this.

docs/passed-verdict-contract.md
pengfei-threemoonslab: Correct, and reproduced: `effect: external_communication` + a financial tag with approval/audit/idempotency is **blocked** (`SHIP-POLICY-CONFIRMATION-MISSING`, `SHIP-ACTION-EXTERNAL-COMMUNICATION-AUDIT-MISSING`), while `effect: financial_write` with the same controls is `review_required`. A tag unions obligations; raising the effect drops whatever the previous value obliged.

Rewritten here, in `docs/manifest-v0.1.md`, and in the changelog: the tag applies its category's controls in addition to the declared effect's, and the two edits are explicitly not interchangeable — with that exact example. The equivalence claim survives only where it is true, beside `effect: read`, which obliges nothing, and the test asserting it is renamed `test_a_tag_on_a_read_row_publishes_what_declaring_the_effect_publishes` and scoped to that. `test_a_tag_unions_obligations_with_the_declared_effect` is the new negative control and pins your counterexample, including that replacing the effect is what loses the confirmation.

pengfei-threemoonslab (commented):
Post-merge re-review of final head `28479acfe99512e87e049ecc3aead9f9f3ab5bda`: no remaining findings.

I re-ran the original adversarial cases and confirmed all three prior comments are closed:

- The tag repair now publishes the complete replacement value and preserves pre-existing tags; the exhaustive seeded matrix and the end-to-end shallow-merge rescan both close cleanly.
- `effect: read` plus a positive reviewed tag no longer synthesizes `read_only` or `reversibility: reversible`; action and capability facts now match the direct positive-effect spelling apart from their intentionally distinct claim provenance.
- The contract now describes tags as unioning obligations with the declared effect and scopes equivalence to the `read` case.

Validation completed on the final SHA: focused semantic/questionnaire/action/capability/P0 suites; the full test suite with expected skips; 8/8 isolated packaging tests; Ruff; compileall; generated-schema parity; primary and self-hosted Shipgate verification (`passed`, `mergeable`, no reviewer signals); and all GitHub checks green.

The separately filed #460 remains an intentional follow-up because changing its evidence-pin behavior has a one-time migration effect; I reviewed that boundary and did not treat it as a regression in this PR.

PR #458, opened 2026-08-29
test(samples): a cold-start sample whose declaration questionnaire is committed (#425)

### LINKED ISSUE #425: No sample exercises the declaration questionnaire, so its ordering and rendering are unpinned by any committed artifact
## Summary

The declaration questionnaire is the primary cold-start adoption surface — it is what an adopter at `insufficient_evidence` reads, in order, to reach a verdict. **No shipped sample exercises it.** Every sample in `samples/` answers every declaration question it asks, so the ordering, the counter, the `DeclarationQuestionRow` field shape, and the entire `suggested-declarations.yaml` renderer are covered only by unit tests and never by a committed artifact.

This is not theoretical. [#419](https://github.com/ThreeMoonsLab/agents-shipgate/issues/419) shipped an ordering that put a structurally proven OpenAPI `GET` named `delete_account` at the top of the questionnaire, ahead of a genuinely unknown effect, with a repository-chosen name breaking the tie. It was found in review of [#421](https://github.com/ThreeMoonsLab/agents-shipgate/pull/421), against a fully green suite: all sample goldens, the markdown byte-comparisons, the Conductor field-for-field check, and the in-tree verifier's `passed / mergeable` were all structurally incapable of seeing it.

## Observed

All 15 samples carrying a manifest, scanned at `40552346`:

```
sample                        decision                questions   open
clean_read_only_agent         passed                  0/0         0
openai_agents_sdk_agent       passed                  0/0         0
agent_weakens_gate            passed                  0/0         0
hitl_evidence_covered_agent   passed                  0/0         0
hitl_evidence_agent           review_required         0/0         0
google_adk_agent              review_required         2/2         0
large_multi_framework_agent   review_required         3/3         0
simple_langchain_agent        passed                  0/0         0
simple_crewai_agent           review_required         0/0         0
simple_openai_api_agent       blocked                 4/4         0
simple_anthropic_agent        blocked                 4/4         0
support_refund_agent          blocked                 2/2         0
conductor_agent               insufficient_evidence   2/2         0
ai_generated_refund_pr        insufficient_evidence   1/1         0
```

`open = 0` everywhere — including both samples that *do* reach `insufficient_evidence`, which get there for other reasons. `shipgate-self.yaml`, which the `verify-self` CI job runs, reports `passed` with `0/0`.

## What that leaves uncovered

- **`suggested-declarations.yaml` has no golden at all.** No sample emits the file, so nothing byte-compares the header, the `Question N of M` banners, the reading lines, the proposal annotations, the unfillable-question notes, or the "Also required, and not a per-action question" section. `report.md` and `packet.*` are byte-compared; the file an adopter is actually told to edit is not.
- **`test_sample_expected_report_json_has_no_structural_drift` cannot see `DeclarationQuestionRow`.** It compares field *paths*, and `open_questions` is `[]` in all five goldens, so no `open_questions[].*` member appears in any of them. Adding, removing, or renaming a field on that model is invisible to the one test built to catch exactly that kind of drift.
- **Ordering is unpinned end to end.** `open_questions[]` order, and the `_in_question_order` permutation of `evidence_gaps[]` that follows from it — which selects `primary_evidence_gap`, and so the `Improve evidence:` line, the decision `reason`, and `first_recommended_action` — are exercised only by synthetic in-process fixtures.
- **The progress sentence is only ever rendered at zero.** `Declaration questions: N of M answered; K open (…)` reaches the CLI, the file header, and the PR comment from one rendering, and no committed artifact carries it with a non-zero `K`.

## Why it compounds

The walk this whole surface was built for (`google/adk-samples#1745`) is an **external repository**. Reproducing it needs a clone that is not in this repo and not in CI, so the only end-to-end evidence that the cold-start path works is a local run somebody remembers to do. Every increment of [#410](https://github.com/ThreeMoonsLab/agents-shipgate/issues/410) has been validated that way, and #419 is the first time that gap produced a shipped defect rather than a caught one.

## Suggested direction

One sample that stops at `insufficient_evidence` **with open declaration questions**, carrying `expected/report.json`, `expected/report.md`, and — the point of the exercise — `expected/suggested-declarations.yaml` as a byte-compared golden.

It should carry all three tiers the ordering distinguishes, because a fixture with only one proves nothing about the order:

- an action nothing was read about (no evidence at all);
- an action with only a protocol default standing in for its absence;
- an action with a heuristic reading the scan may not act on;
- an action the scan read as risky, so its block arrives with a proposed answer;
- a source-wide authority question, so the merged-block numbering is rendered too.

Then extend `test_sample_expected_report_json_has_no_structural_drift` to it, so `open_questions[].*` enters the compared path set.

The ADK shape is the one to mirror — `samples/google_adk_agent` already exists and answers everything, so a sibling that deliberately does not is the smallest honest addition. Whether to grow that sample or add a new one is an open call for whoever picks this up.

Not a new public surface under [`CONTRIBUTING.md` § Surface discipline](https://github.com/ThreeMoonsLab/agents-shipgate/blob/main/CONTRIBUTING.md#surface-discipline) — no CLI command, schema version, report block, discovery surface, or adapter. It is test evidence for a surface that already shipped.

## Acceptance

- [ ] At least one sample reports a non-zero `declaration_questions.open`, and its `expected/report.json` carries populated `open_questions[]` rows.
- [ ] That sample's `expected/suggested-declarations.yaml` is committed and byte-compared, the way `report.md` is.
- [ ] `test_sample_expected_report_json_has_no_structural_drift` covers it, so `open_questions[].*` is in the compared path set.
- [ ] Reordering the questionnaire — for example by reverting `_reach` in `core/declaration_questions.py` to rank by `conservative_effect` — fails at least one committed-artifact test, not only a unit test.

---

Found while addressing review feedback on [#421](https://github.com/ThreeMoonsLab/agents-shipgate/pull/421) (#419).

pengfei-threemoonslab:
Closes #425.

## The gap

The declaration questionnaire is the primary cold-start surface — what an adopter at `insufficient_evidence` reads, in order, to reach a verdict — and **no shipped sample exercised it**. Every sample answered every declaration question it was asked, so `open_questions` was `[]` in all five goldens, `suggested-declarations.yaml` had no golden at all, and the progress sentence was only ever rendered at zero.

That is how [#419](https://github.com/ThreeMoonsLab/agents-shipgate/issues/419) shipped an ordering that put a structurally proven OpenAPI `GET` named `delete_account` at the top of the questionnaire, ahead of a genuinely unknown effect, with a repository-chosen name breaking the tie — against a fully green suite.

## The fixture

`samples/google_adk_cold_start_agent` is the sibling of `samples/google_adk_agent` that deliberately stops partway: `insufficient_evidence`, **10 open declaration questions**, `2 of 12 answered`.

| Action | Rung it holds | Asked as |
| --- | --- | --- |
| `update_case_index` | nothing read; name bands as mutating (`+1`) | 2 |
| `assemble_case_timeline` | nothing read; name says nothing (`0`) | 3 |
| `ops.export_case_bundle` | declared authority contradicts the MCP export | 4, then **5 as a note with no block** |
| `ops.queue_backfill` | only the MCP protocol default stands in for evidence (`0`) | 6–7 |
| `list_case_attachments` | nothing read; name bands as retrieving (`-1`) | 8 |
| `issue_goodwill_refund` | read as a financial write, so the block arrives **proposed** | 9 |
| `support.get_update_history` | a structurally proven read whose *name* looks like a write | 10 |
| `ops.append_case_note` | established structurally (`POST`) | never asked |
| `record_case_outcome` | declared, so both its questions are answered | counted, not open |

Both halves of #419's fix are held. The unbounded questions lead the bounded ones; and among the unbounded, the name band orders them **against alphabetical order in both directions** — `update_case_index` leads `assemble_case_timeline`, `list_case_attachments` trails both.

`support.get_update_history` is #419's defect wearing its own hat: proven read-only by a trusted `readOnlyHint`, so it sorts last — and sorts *first* again the moment `_reach` goes back to asking whether a side effect was *measured* rather than *bounded*.

## What is now pinned

- `expected/suggested-declarations.yaml` is byte-compared, the way `report.md` already was — header, `Question N of M` banners, merged-block numbering, reading lines, proposal annotations, and the unfillable-question note in its numbered place between two blocks.
- `open_questions[].*` reaches `test_sample_expected_report_json_has_no_structural_drift`, so a field added to, removed from, or renamed on `DeclarationQuestionRow` is finally visible to the one test built to catch that.
- `_in_question_order`'s permutation is asserted against the questionnaire's own numbering, together with the decision `reason` and `first_recommended_action` that project from it.
- The order itself is restated in `COLD_START_QUESTION_ORDER` beside the reason each row holds its place, so a regeneration cannot absorb a reordering into a YAML diff nobody reads.

## Verified by perturbation

| Change to the engine | Result |
| --- | --- |
| `_reach` ranks by `conservative_effect` | 2 committed-artifact tests fail |
| `_reach` ranks on *measured* rather than *bounded* (the #419 bug) | 2 committed-artifact tests fail |
| `name_shape_band` flattened to `return 0` | 2 committed-artifact tests fail |
| a field added to `DeclarationQuestionRow` | structural-drift fails for **this sample only** |

## Not covered, and why

One rung is missing: an action whose only reading is a keyword-inferred `read`, unbounded because [#357](https://github.com/ThreeMoonsLab/agents-shipgate/issues/357) forbids a heuristic from establishing a read-only action. It is unreachable from an ADK repository — `READ_ONLY_KEYWORDS` are consulted only for `_KEYWORD_GATED_SOURCE_TYPES`, and no ADK source type is one; and any MCP-ish source takes the `protocol_default` branch before the `inferred` one. Bolting on a LangChain/CrewAI source leaves its tools unbound and out of the action surface, and forcing them in with `agent_bindings.declarations` raises `conflicting_binding_evidence`, which takes over the headline the questionnaire is here to own. The sample README says so, and names the two in-process tests that cover it.

## Surface discipline

Test evidence for a surface that already shipped: no CLI command, schema version, report block, agent-discovery surface, or framework adapter.

## Testing

`ruff check .` clean. Full CI split green locally: `pytest -n auto -m "not perf" --ignore=tests/test_adapter_static_only.py`, `pytest tests/test_adapter_static_only.py`, `pytest tests/test_latency_budget.py -m perf`.

## Review rounds

Four review passes over the branch; every finding is fixed on it.

| Round | Found | Notable |
| --- | --- | --- |
| 1 | 6 | Stale counts and a wrong `0 of 0` description of the sibling fixture — plus, while fixing them, an absolute `/var/folders/…/tmp…` path baked into the committed `report.json` by the regeneration recipe. |
| 2 | 1 | The guard added for that leak used `Path(...).is_absolute()`, so on Windows it would have passed on exactly the golden it rejects on Linux. |
| 3 | 2 | The fixture's verdict silently depends on an unrelated check's severity tier; and a missing scaffold surfaced as `FileNotFoundError` rather than as the message. |
| 4 | 1 | The paragraph naming that dependency named one of the two tests it says will fail. |

Two of these changed shared surfaces, so they are worth calling out separately
from the fixture:

- `test_sample_expected_report_json_uses_repo_placeholder_for_manifest_dir` now
  rejects an absolute `generated_reports` path in **any** sample golden, in both
  POSIX and Windows spellings. It was only checking `manifest_dir` and
  `str(Path.cwd())`, so a temp path outside the working directory slipped past
  it for every sample.
- The sample README documents that the fixture's `insufficient_evidence` verdict
  also depends on `SHIP-ADK-EVAL-COVERAGE-MISSING` staying below the high tier,
  and names both tests that would fail if it did not — left in rather than
  silenced with an eval file, because a repository with no declared eval
  coverage is what a cold start actually looks like.

Portability re-verified after the last change: scanned from an unrelated
absolute path, `report.md`, `suggested-declarations.yaml` and the
`declaration_questions` subtree all hash identically to the committed goldens.


samples/google_adk_cold_start_agent/README.md
```diff
@@ -0,0 +1,149 @@
...
+    config_path=sample / "shipgate.yaml",
+    output_dir=Path("expected"),
+    formats=["json", "markdown"],
+    ci_mode="advisory",
+    packet_enabled=False,
+)
+(sample / "expected" / "current-control.json").unlink(missing_ok=True)
+golden = sample / "expected" / "report.json"
+golden.write_text(
+    golden.read_text(encoding="utf-8").replace(os.getcwd(), "<REPO>"),
```
pengfei-threemoonslab: [P2] Make the regeneration recipe platform-independent

On Windows, `report.json` contains JSON-escaped backslashes (`C:\\...`), while `os.getcwd()` is the decoded path (`C:\...`), so this textual replacement is a no-op. The regenerated golden retains an absolute `manifest_dir`, and the new `<REPO>/samples/` guard fails. `generated_reports` also uses platform-native `expected\report.*`, creating cross-platform churn. Please parse the JSON, assign a POSIX `<REPO>/{sample.as_posix()}` manifest path, normalize the generated-report values with `Path(value).as_posix()`, and serialize deterministically.

tests/test_declaration_questionnaire.py
```diff
@@ -1911,3 +1911,360 @@ def test_confirming_one_proposal_reaches_a_verdict(
...
+    """One order, published three times, pinned once.
+
+    ``_in_question_order`` permutes the published gap rows into the order the
+    questionnaire asks them, and ``primary_evidence_gap`` — and so the decision
+    reason and ``first_recommended_action`` — takes the first addressable row
+    of that list. Nothing committed showed the two agreeing, because no sample
+    had two declaration questions to put in an order.
+    """
+
+    golden = _cold_start_golden()
```
pengfei-threemoonslab: [P2] Exercise the live report in this cross-surface guard

This test accepts `tmp_path` but reads only the committed `report.json`, so it proves the golden is internally consistent, not that the current producer still makes `agent_summary.first_recommended_action` agree with the first question. The other fresh checks compare questionnaire coverage and scaffold bytes; `report.md` does not render `first_recommended_action`, and the JSON drift test compares paths only. A producer regression in this field can therefore leave all of those checks green. Load `_cold_start_scan(tmp_path) / "report.json"` here and assert its live gap order, reason, and first action against its live `open_questions`.

tests/test_declaration_questionnaire.py
```diff
@@ -1911,3 +1911,360 @@ def test_confirming_one_proposal_reaches_a_verdict(
...
+    coverage = _cold_start_coverage(golden)
+    gaps = golden["release_decision"]["evidence_coverage"]["evidence_gaps"]
+
+    question_rows = [
```
pengfei-threemoonslab: [P2] Compare question groups rather than gap-row cardinality

`question_rows` has one entry per `EvidenceGap`, while `expected` has one per blank/question. Multiple rows can legitimately ask the same question; a current example is `declaration_drift` plus `declaration_below_inferred_evidence` on one effect, and the list-valued indexing later in this file already accounts for that contract. A valid multi-gap fixture would fail this equality even when its rows are contiguous and correctly ordered. Collapse contiguous equal `(subject_kind, subject_id, dimension)` keys before comparing with `open_questions`, while still rejecting a key that recurs noncontiguously.

pengfei-threemoonslab (commented):
Requesting changes for three reproducibility and test-contract gaps. The fixture itself reproduced cleanly, and the focused, static-adapter, performance, packaging, lint, and Shipgate checks otherwise passed.

samples/google_adk_cold_start_agent/README.md
pengfei-threemoonslab: Fixed in 506dc333 — and confirmed: the replace is a no-op on Windows for exactly the reason you give, so the golden keeps an absolute `manifest_dir` and the machine that produced it fails the `<REPO>/samples/` guard.

The recipe is now structural: `payload["manifest_dir"] = f"<REPO>/{sample.as_posix()}"`, `Path(written).as_posix()` over `generated_reports`, serialized with `json.dumps(payload, indent=2)` — which is exactly what `write_json_report` uses (no `sort_keys`, no trailing newline), so the round trip is byte-identical and nothing but those two fields moves. I checked that the new recipe reproduces the committed `report.json` byte for byte (`118390fb…`) rather than assuming it.

`manifest_dir` is the only field carrying the cwd, so the targeted assignment is equivalent to the old textual replace on POSIX and correct on Windows.

The guard grew the matching half: `test_sample_expected_report_json_uses_repo_placeholder_for_manifest_dir` now rejects a backslash in `manifest_dir` or in any `generated_reports` value, for every sample golden — so a Windows-separated golden is caught rather than merely being off-bytes against everyone else.

tests/test_declaration_questionnaire.py
pengfei-threemoonslab: Fixed in 506dc333. You are right, and the unused `tmp_path` was the tell — the test took the fixture and then never scanned it.

It now reads `_cold_start_scan(tmp_path) / "report.json"` and asserts the live gap order, reason and first action against the live `open_questions`, then separately asserts the committed rows carry that same order (nothing else covers the gap list by value — the drift test compares field paths, and the coverage comparison reads `declaration_questions` rather than the rows beside it).

Demonstrated rather than asserted: making `primary_evidence_gap` return `addressable[-1]` instead of `addressable[0]` — which moves nothing but the reason and `first_recommended_action`, leaving the gap list untouched — fails this test and **no other** cold-start check. Under the old version it failed nothing at all.

tests/test_declaration_questionnaire.py
pengfei-threemoonslab: Fixed in 506dc333. Agreed on the contract, and on the inconsistency — the list-valued indexing in `test_cold_start_questionnaire_covers_every_rung_it_claims_to` already treats a row as not-a-question while this comparison did not.

`_questions_asked_by_gap_rows` now collapses contiguous equal `(subject_kind, subject_id, dimension)` keys. That is sound rather than lenient: `_in_question_order` sorts the question rows by `(question, original position)`, so every row asking one question lands in consecutive question slots — and a key that comes back *after another one intervened* survives the collapse and fails the comparison, because that is the permutation broken rather than a question asked twice.

Both halves are pinned directly in `test_several_rows_asking_one_question_collapse_to_one`, since this fixture raises one row per blank and so could not exercise either. The non-question row (`incomplete_surface`) sitting between two same-key rows is in the contiguous case on purpose — it is dropped, not collapsed into its neighbour. Verified non-vacuous: removing the collapse fails that test.

samples/google_adk_cold_start_agent/README.md
```diff
@@ -0,0 +1,168 @@
...
+(sample / "expected" / "current-control.json").unlink(missing_ok=True)
+
+golden = sample / "expected" / "report.json"
+payload = json.loads(golden.read_text(encoding="utf-8"))
+payload["manifest_dir"] = f"<REPO>/{sample.as_posix()}"
+payload["generated_reports"] = {
+    fmt: Path(written).as_posix()
+    for fmt, written in payload["generated_reports"].items()
+}
+golden.write_text(json.dumps(payload, indent=2), encoding="utf-8")
```
pengfei-threemoonslab: [P2] Force LF for all regenerated goldens

This still is not byte-reproducible on Windows. `Path.write_text(..., encoding="utf-8")` uses platform newline translation, and the scanner's Markdown/questionnaire writers also write text without forcing LF. Because `.gitattributes` marks `samples/**/expected/** -text`, Git preserves those CRLF bytes while the committed goldens are LF. The new tests use `read_text()`, which normalizes CRLF back to LF and therefore misses the churn. Force LF for `report.json`, `report.md`, and `suggested-declarations.yaml` (including this rewrite, e.g. `newline="\n"`) and add a raw-byte CR guard.

tests/test_declaration_questionnaire.py
```diff
@@ -1911,3 +1911,460 @@ def test_confirming_one_proposal_reaches_a_verdict(
...
+    correct.
+
+    A key that comes back *after another one intervened* is the permutation
+    broken, and is deliberately left in place: it survives the collapse and
+    fails the comparison against ``open_questions``.
+    """
+
+    asked: list[tuple[str, str | None, str]] = []
+    for gap in report["release_decision"]["evidence_coverage"]["evidence_gaps"]:
+        dimension = DIMENSION_BY_GAP_KIND.get(gap["kind"])
```
pengfei-threemoonslab, line 2226: [P2] Filter mapped gaps through actual open questions

Membership in `DIMENSION_BY_GAP_KIND` does not always mean a gap is a declaration question. A source-owned `conflicting_effect_evidence` maps to `effect`, but `is_declaration_answerable()` deliberately excludes it; this helper then adds an effect key that is absent from `open_questions`. The production `_in_question_order()` avoids that by filtering keys through the actual open-question rank. Apply the same filter here and add the mapped-but-unanswerable branch as the negative case, otherwise this cross-surface assertion has a false oracle.

pengfei-threemoonslab (commented):
Second pass on `506dc333`: the three findings from my previous review are addressed. I found two remaining P2 issues in the follow-up: Windows regeneration still produces CRLF bytes while the committed `-text` goldens are LF, so the new text-mode tests miss byte churn; and the new gap-to-question oracle counts a mapped but deliberately unanswerable source conflict as a questionnaire row.

The branch is also currently `CONFLICTING` with `origin/main` at `CHANGELOG.md`; please rebase or merge and preserve both Unreleased entries. Focused tests pass on the PR head and the synthetic merged tree. Agents Shipgate returns `review_publishable` / `review_required` because this PR touches a trust root, so human review remains required.

samples/google_adk_cold_start_agent/README.md
pengfei-threemoonslab: Fixed in 7a9bcb57 — and confirmed end to end, including the part that makes this nasty: the blindness is real.

I reproduced it rather than reasoning about it. Writing `report.md` with `newline="\r\n"` (what `write_text(newline=None)` does on Windows) leaves the bytes different on disk, and `test_cold_start_markdown_report_matches_golden` — a byte comparison — **passes**, because `read_text()` applies universal newlines on the way in. So every existing golden comparison in this repo would have shipped the churn silently.

Two changes:

- The recipe forces `newline="\n"` on all three goldens. `report.json` gets it on the rewrite it already did; `report.md` and `suggested-declarations.yaml` are re-written after the scan, since the scanner's own writers use the platform newline.
- `test_sample_expected_goldens_are_committed_with_lf_newlines` reads the **raw bytes** of every `samples/*/expected/*` file and rejects any `\r`, so the guard does not share the blindness it exists to cover. It globs the whole tree rather than this fixture, because `.gitattributes` pins all of `samples/**/expected/**` as `-text` for the same reason.

All committed goldens are LF today, so the guard passes on arrival; verified non-vacuous by making one CRLF, which fails only the new test.

tests/test_declaration_questionnaire.py
pengfei-threemoonslab: Fixed in 7a9bcb57. You are right, and the distinction is exactly the one the module docstring draws and my oracle did not: `DIMENSION_BY_GAP_KIND` answers *which dimension a kind rides on*, which is not *whether a declaration can close it*.

`_questions_asked_by_gap_rows` now filters through the report's own `open_questions` keys, the same way `_in_question_order` filters through the open-question rank rather than through the kind map.

The negative case is pinned directly, since this fixture raises no source-owned conflict and so could not show it — `test_gap_rows_reduce_to_the_questions_they_ask` now covers three things:

- a `conflicting_effect_evidence` **absent** from `open_questions` is dropped rather than counted;
- the same kind **present** in `open_questions` (the branch where the resolver blamed the manifest) is still counted — which is the only thing distinguishing the two branches on the wire, since the published row carries no issue attribution;
- the contiguous collapse, unchanged.

Verified non-vacuous both ways: removing the open-question filter fails that test, and so does removing the collapse.

PR #372, opened 2026-08-11
feat(agent-mode): roll the shared control envelope across the setup commands (#323)

pengfei-threemoonslab:
## Summary

- Closes the horizontal half of #323: `detect --json`, `init --json`, and every `doctor --json` payload carry a `control` field holding the same `shipgate.agent_control/v1` envelope that `verify --format control`, `check --format agent-control-json`, and `agents-shipgate agent control` already emit. One `control_state`, one six-way `permissions` vector, one typed rank-1 `next_action` for the whole adoption walk.
- **This is consolidation, not new surface** (per the [triage note](https://github.com/ThreeMoonsLab/agents-shipgate/issues/323#issuecomment-5211858102)): `derive_agent_control` and the envelope projection already existed and four commands bypassed them. No headline metric is owed under the `CONTRIBUTING.md` gate.
- **No second verdict.** The envelope is a projection of the diagnostics those commands already publish, computed in one module (`cli/setup_control.py`) and emitted through the one existing projection point, so `control.next_action`, `next_actions[0]`, and the agent-mode error line are all projections of one selected route.
- Contract `23 → 24`. `minimum_control_contract_version` **stays at 21**, and the `AgentControl` union is byte-identical to v21.

**`scan` is deliberately not part of this rollout, and #323's scan half stays open.** See below.

The walk, on a fresh repo with an agent and no manifest, reading only `control`:

| step | `control_state` | `decision` | `decision_source` | rank-1 action |
|---|---|---|---|---|
| `detect` | `agent_action_required` | `setup_incomplete` | `setup` | `initialize` → `init` |
| `init --write` | **`human_review_required`** | `setup_incomplete` | `setup` | `shipgate.yaml:13 (agent.declared_purpose)` |
| `doctor` (unresolved) | `human_review_required` | `setup_incomplete` | `setup` | same — derived each run, not remembered |
| `doctor` (resolved) | `agent_action_required` | `setup_complete` | `setup` | `verify` → the gate |
| `verify --format control` | `review_publishable` | `insufficient_evidence` | **`release_decision`** | human review |

### What keeps setup control from being a second gate

**Setup names its own source.** These commands run before a release decision exists, so they report `decision_source: "setup"` and a `decision` from the closed vocabulary `setup_complete | setup_incomplete | setup_not_applicable`. The published schema enforces the pairing both ways, constrains every source to its own engine's vocabulary, and constrains every operation to the engine that decides it.

[...]

Issue #328, opened 2026-08-07
Distinguish product defects from user configuration errors in adopter-facing output


pengfei-threemoonslab:
## Summary

Agents Shipgate does not reliably distinguish **its own defects** from **the user's configuration errors** in adopter-facing output. A first-contact adopter cannot tell "this repository is unusual and needs configuration" from "this tool is broken", and will reasonably assume the former — then spend time fixing something that was never theirs.

Invariant 4 of #327.

## Evidence

Two cases, both observed on real external repositories.

**#307 (now fixed) — a product defect routed as user recovery.** The duplicate `tool_sources[].id` collision was a bug in our ID derivation. In agent mode it surfaced as `internal_error` with a next action of `init --minimal` — a recovery step the user performs, which discards the detection work and hands back a `CHANGE_ME` template. The message pointed the user at their repository; the fault was ours.

**#321 — a product defect typed as an input error.** Sharing one `FunctionTool` across several `LlmAgent`s raises `InputParseError`. That exception name and exit code mean "the input you supplied could not be parsed", but the input is a valid, canonical ADK pattern and the defect is in our catalog identity model. An adopter reading `InputParseError` will look for the malformed file.

The pattern in both: an internal failure inherits an error kind and a next action that imply user fault.

## Why this matters more during first contact

Adoption is exactly when the user has the least basis for judging which side the fault is on, and the most willingness to conclude the tool does not work on their repository and stop. A misattributed error at this point is not a cosmetic problem — it is an abandonment path. Every open bug in this tracker was found by pointing the tool at an external repository, and several presented this way.

## Requested

A taxonomy, applied consistently:

- Classify every adopter-facing failure as **ours** (a defect, a limitation, an unsupported-but-valid shape) or **theirs** (a genuinely malformed or missing input).
- A defect must never emit a user-remediation next action. The honest action is to report the failure, name it as ours, and — where possible — offer a route that does not require the user to fix anything.
- Exit codes and error kinds should follow the classification rather than cutting across it. `InputParseError` should mean the input was bad.
- Unsupported-but-valid is a third state and should say so: the repository is fine, the shape is real, we do not handle it yet. This overlaps the `relevant_but_unsupported` routing in #309 — the vocabulary should be settled once.

## Acceptance criteria

- [ ] A documented classification covering every adopter-facing error kind and exit code.
- [ ] No failure classified as ours emits a next action that asks the user to change their repository.
[...]

PR #459, opened 2026-08-29
feat(agent-mode): finish the control-envelope rollout and prove the adoption walk composes (#323)

### LINKED ISSUE #323: rollout(agent-mode): apply the shared control envelope across commands
> Depends on #333. This issue applies the approved envelope; it does not redesign it.

## Summary

The first-adoption workflow still exposes command-specific machine interfaces. After #333 settles the compact permission/control shape, roll that single vocabulary across `detect`, `init`, `doctor`, `scan`, `verify`, and `check`.

## Semantic boundary

- `scan`, `verify`, and `check` must project the existing authoritative release decision; no renderer may compute a second verdict.
- `detect`, `init`, and `doctor` run before a release decision exists. Their control state derives from setup facts such as detection confidence, manifest validity, and unresolved human-owned placeholders. The payload must identify that source explicitly.
- Command-specific result fields remain available through the documented compatibility window.

## Acceptance criteria

- [ ] All six commands expose the #333 control-state and permission vocabulary in agent mode.
- [ ] Every non-complete state has one typed rank-1 action and optional ranked alternatives.
- [ ] Setup-derived and release-decision-derived control states identify their source and cannot be confused.
- [ ] JSON uses documented streams and is not split into incompatible stdout/stderr shapes.
- [ ] Existing command-specific schemas remain backward compatible through a stated migration window.
- [ ] An end-to-end #327 adoption fixture routes every step using only the shared envelope.
- [ ] The rollout introduces no second verdict and no duplicate envelope schema.

pengfei-threemoonslab:
Closes #323.

#323's envelope rollout landed in #372 (contract v24): `detect`, `init`, and
`doctor` publish `shipgate.agent_control/v1` on `--json`, setup names its own
`decision_source`, and setup authorizes nothing. Three of the seven acceptance
criteria were left open, and this closes them.

## What was actually missing

**Criterion 6 — the end-to-end fixture did not exist.** #327's definition of
done is one fixture that walks an unadopted repository through all three stages
using only the emitted next action. Every defect that epic collects was found by
walking the flow by hand and none by a test, because each command is tested
against its own output rather than against the next command's input.

**Criterion 4 — the two documented streams were two shapes.** `doctor`'s
failure routes carried `control`; `detect`'s and five of `init`'s did not. So
whether a caller that routes on `control` could route at all depended on which
setup command had failed and on which of its failures — and the run that most
needs a route is the one that printed no payload to carry it. `AGENTS.md` still
said error lines carry no `control` object, which had not been true since #372.

**Criterion 1 was asserted, not checked.** Nothing pinned that all six commands
publish the envelope, so "the setup three do" was the only part with a test.

## What the walk found

Writing the fixture found the one place the walk could not compose.
`init --write` over a manifest that already exists published
`next_action.kind: "edit"` on `shipgate.yaml`, `expects: "The manifest reflects
the desired tool sources, agent declared_purpose, and policies"`. That
postcondition was already satisfied every time the route was reached — a
manifest that does *not* load is claimed by the repair route above it. On this
contract `next_action` **is** the step, so the route could not change the
answer: an envelope-only caller opened the file, found nothing to change,
re-ran, and got the identical action back.

It is stage 2 of the walk that hits it. `init --write` writes the manifest and
stops at `human_review_required` for the `declared_purpose` declaration; after a
person supplies it, re-running the command that stopped is the only resume an
envelope-only caller has, and that re-run is this route. The walk could not
leave stage 2.

The route is now the `doctor` invocation for the manifest on disk, which reports
what that manifest still owes — the gate when nothing is wrong, a named repair
when something is. The exit code (2) and the sentence a person acts on are
unchanged; `next_action.kind` moves from `edit` to `command`.

## Changes

- **`setup_failure_routing`** — one place every setup failure builds its
  envelope, rather than one construction per call site. Fail-closed by
  construction: `execution: "failed"`, `decision: "setup_incomplete"`,
  `permissions` all false. `doctor`'s existing helper now delegates to it.
- **`emit_agent_mode_error_routing`** — the sibling of the existing
  `emit_agent_mode_error_action` one level up: it projects `next_action`,
  `next_actions[]`, and `control` from one selected route, so an error line
  cannot publish a control state that authorizes nothing beside a ranked list
  that names a command. All twelve agent-mode error emissions in the three setup
  modules go through it.
- **Two error lines still carry no envelope, and both are documented**: the
  shared `--workspace` refusal, which fires before a workspace exists so there
  is no setup subject for `input_id` to address, and `environment_error`, which
  is emitted before Shipgate is running.
- **`init --minimal` recovery is spelled for its invocation.** That route
  carried a literal `agents-shipgate init --minimal`. The error line normalized
  it on the way out; the envelope publishes `next_action.command` verbatim, so
  one route would have named two entry points.

## Tests

`tests/test_adoption_walk.py` runs the walk as real subprocesses and chooses
every step from the envelope alone — no command-specific result field, no
hand-built command, no look at the workspace to work out which stage it is in.
It asserts one vocabulary, one typed rank-1 action, setup and gate states naming
their source, setup authorizing nothing, the command-specific fields agreeing
with the envelope, and the gate's decision equal to `report.json`'s. It fails a
step that hands back the same action for an unchanged `input_id`, which is the
cycle detector that found the route above.

**Negative control**: reverting only the `_init_advance` change fails the walk
at the expected assertion; nothing else moves.

A second test covers criterion 1 as a table — each of the six commands at the
one entry point it documents. A third sweeps the three setup modules' ASTs and
fails if any of them reaches for a bare emitter, so a seventh error route cannot
be added without the envelope.

`test_errors_json_lists_every_runtime_emitted_kind`'s scanner matched only the
base emitter name, so `config_already_exists` dropped out of it when `init`
moved to the wrapper. It now matches all three spellings — a latent hole:
`emit_agent_mode_error_action`'s call sites were already invisible to it.

## Contract

26 → 27. Both the `AgentControl` union and `shipgate.agent_control/v1` are
byte-identical — the same envelope reaches one more place — so
`minimum_control_contract_version` stays at `21`. Migration note in
`STABILITY.md`.

## The seven criteria, and where each is checked

| # | Criterion | Evidence |
| - | --------- | -------- |
| 1 | All six commands expose the vocabulary in agent mode | `test_every_command_publishes_the_shared_envelope` — one case per command, at the entry point each documents. Table in `AGENTS.md`. |
| 2 | One typed rank-1 action, optional ranked alternatives | `_assert_shared_envelope` on every walk step; the alternatives stay in `next_actions[]`. |
| 3 | Setup and release-decision states name their source and cannot be confused | `_assert_shared_envelope`'s setup branch; the schema rules that enforce the pairing both ways were already pinned in `test_setup_control.py`. |
| 4 | Not split into incompatible stdout/stderr shapes | **This PR.** `test_a_setup_failure_routes_on_the_same_envelope_as_a_setup_answer`, the two fault-path tests, and an AST sweep so a seventh route cannot skip it. |
| 5 | Command-specific schemas stay backward compatible | Nothing removed; `_assert_shared_envelope` checks `next_actions[0]` agrees with `control.next_action` at every step. Migration window stated in `STABILITY.md`. |
| 6 | An end-to-end #327 fixture routes every step on the envelope alone | **This PR.** `test_the_adoption_walk_routes_end_to_end_on_the_envelope_alone`. |
| 7 | No second verdict, no duplicate envelope schema | The walk asserts the gate's `decision` equals `report.json`'s; no schema document changes. |

## Review rounds

Two rounds landed on top of the original commit; the full reasoning is in the
commit messages and in the reply on the review thread.

**Round 1 (`4eeda18c`, self-review)** — eight findings, all in the new code.
The largest: the `--minimal` fallback still dropped `--workspace` / `--write` /
`--json`; the adoption-kit recheck omitted `--agent-instructions-kit`, so it
could not read the file it asked someone to fix; the walk executed whatever
`AGENTS_SHIPGATE_CLI` named; a new test distinguished two entry points by
writing them into command strings, which `NextAction` retargets, so both
collapse under an exported override; and the walk validated the model but not
the published JSON Schema.

**Round 2 (`f3b1339a`, review findings)** — four inline findings, and every one
of them is the same defect at a different depth: *a route that reports it
advanced the run while dropping part of what the caller asked for.*

- A refused write handed the caller onward under the pack the manifest already
  had, so `--control-pack financial-strict` was silently lost. Both
  `skipped_existing` branches now check the delta first — the refresh path had
  it too, and was reporting `setup_complete`.
- Every recovery command was built from the flag list a rerun in a *different*
  workspace may repeat, so `--minimal`, an explicit kit, a raised parse cap and
  the accepted scope boundary were dropped. One builder now feeds all five
  sites.
- `input_id` did not cover `action_kind` or the diagnostic id and severity that
  decide ownership and precedence.
- `STABILITY.md` promised `execution: "failed"` on every setup error envelope,
  which is false for the most common one — `config_already_exists`, the resume
  of the adoption walk itself, reaches an answer and refuses, so it exits 2
  with `"succeeded"`. The contract now states what the schema actually
  enforces and describes `execution` as what it is.

Plus one found while reviewing that fix: `_unapplied_control_pack`'s docstring
claimed a `None` guard the code did not have.

**Round 3 (`62180dab`, second review pass)** — seven findings, and the largest
is a correction rather than a completion.

- **`--control-pack` on argv is not authenticated human provenance.** Round 2
  published the pack reconciliation as a coding-agent `edit` because the value
  came from the command line rather than from inference. A *governed* coding
  agent composes its own argv, so `--control-pack read-only-agent` over a
  `financial-strict` manifest is a weakening it can request for itself. The
  **direction** decides the owner now: strengthening is an agent edit,
  anything that drops an obligation — or a pack this build cannot resolve — is
  a human review naming what it would remove. The comparison moved beside the
  packs it compares and is shared with `control_pack_weakened`.
- **"Which options may a rerun repeat?" has two answers**, and round 2 used one
  list for both. The scope route that reruns the *same* workspace dropped
  `--allow-unresolved-scope` and returned the refusal it was issued to resolve;
  candidate commands dropped a raised parse cap and refused again on a
  truncation the caller had already answered.
- **An adopted candidate lost the request one directory down** — `doctor` there
  advances under the manifest's own pack.
- **An explicit `--control-pack default` is a request**, and the only one that
  can *only* weaken. Read from the parser now — with the trap recorded: typer
  vendors its own click, so an identity check against `click.core.ParameterSource`
  is silently always false.
- Plus a self-contradicting changelog paragraph, a Windows-only test path, and
  a stale `config_already_exists` recovery hint.

Three more came from reviewing that round's own code: the weakening route's
prose was truncated at the clause explaining *why a person owns it*; a second
reader for "which pack governs this manifest" had appeared; and `assert` was
used for narrowing at both reconciliation sites, which `-O` strips.

Rebased onto current `main`.

## Deliberately not in scope

- The `manifest_status == "skipped_existing" and manifest_exit == 0` branch
  beside the one changed here routes an existing *scaffold* manifest to `verify`
  with `setup_complete`. That is a pre-existing fail-open and belongs with #441;
  the branch changed here routes such a manifest to `doctor`, which reports it.
- `internal_error` on a failed manifest render still publishes a user
  remediation for what is our own defect. The envelope projects the route the
  command already published rather than re-deciding it; #328 owns that taxonomy.


src/agents_shipgate/cli/_register_init.py
```diff
@@ -743,17 +759,40 @@ def _init_advance(
...
+        # answer, and it never claims the setup is complete on this run's
+        # behalf. The refusal itself is not softened: the exit code is
+        # unchanged, and the sentence a person acts on is carried verbatim
+        # into `why`.
         return (
             NextAction(
-                kind="edit",
-                path=str(target),
+                kind="command",
+                command=_doctor_command(target),
```
pengfei-threemoonslab, line 960: [P1] Preserve requested manifest configuration before handing off. A loadable existing manifest does not necessarily satisfy this invocation. Repro: start with `policies.control_pack: default`, then run `init --write --control-pack financial-strict --json`. The payload records `selected=default` and `requested=financial-strict`, but this action runs `doctor`; doctor sees a valid on-disk manifest and advances to verification under the weaker `default` pack. An envelope-only caller therefore silently loses the requested policy. Detect requested-vs-on-disk deltas and route to exact reconciliation/edit; use `doctor` only when the existing manifest already matches the requested setup. `test_init_reports_the_choice_even_when_it_writes_nothing` already builds this state and can follow/assert the route.

src/agents_shipgate/cli/_register_init.py
```diff
@@ -970,11 +1009,18 @@ def init(
                     "built-in pack id."
                 ),
             )
-            _emit_agent_mode_error(
+            _emit_agent_mode_error_routing(
                 "config_error",
+                routing=setup_failure_routing(
```
pengfei-threemoonslab, line 1201: [P1] Do not promote a lossy recovery command to the authoritative envelope route. `pack_action` above only replays `_requested_setup_flags`, which omits same-workspace, semantics-bearing options such as `--minimal`, `--agent-instructions-kit`, `--max-python-files`, and `--allow-unresolved-scope`. For example, `init --write --minimal --control-pack bogus --json` emits a next command without `--minimal`, so following it writes an auto-detected/scaffold manifest instead of the explicitly requested legacy template; an explicit instruction kit can likewise become the default kit. Preserve every meaningful original option (while replacing only the invalid value) and add argv-equivalence tests for each failure recovery.

src/agents_shipgate/cli/setup_control.py
```diff
@@ -359,6 +359,93 @@ def setup_control_envelope(
...
+            # and through an absolute path publishes different commands, and
+            # `input_id` is the cache boundary for the answer. Every source a
+            # route can come from is folded in — the recovery, the recheck, the
+            # diagnostics' own rank-1 actions — plus whatever the caller adds.
+            routing_facts=(
+                reason,
+                exit_code,
+                placeholders,
+                recheck_command,
+                action.model_dump(mode="json") if action is not None else None,
```
pengfei-threemoonslab, line 443: [P2] Include every route-selecting fact in `input_id`. `action_kind` is not hashed, and diagnostics are reduced to their actions, dropping IDs/severity that determine ownership and precedence. Two otherwise identical `setup_failure_routing` calls with `action_kind="configure"` versus `"verify"` currently have the same `input_id` while publishing different `next_action.kind` values and different `verify_required`; changing only a diagnostic ID can similarly flip an agent-owned edit to human review. Since this PR defines the ID as the cache boundary for the answer, hash `action_kind` and the full route-selecting diagnostic fields, then pin both cases.

STABILITY.md
```diff
@@ -13,6 +13,54 @@ for reproducible CI.
...
+`control`.** v24 gave those commands the envelope on their `--json` payload and
+left the error stream out; `doctor`'s failure routes picked it up during that
+rollout, and `detect`'s and five of `init`'s did not. So whether a caller
+that routes on `control` could route at all depended on which setup command had
+failed and on which of its failures — and the run that most needs a route is
+the one that printed no payload to carry it. Additive: `next_action` and
+`next_actions[]` are unchanged on those lines and are derived from the same
+selected route as `control.next_action`.
+
+A setup failure envelope is fail-closed by construction, and a consumer may
```
pengfei-threemoonslab: [P1] Align this stated invariant with answered-refusal error lines. The common `init --write`-over-existing-manifest path emits `config_already_exists` on stderr with the ordinary routing envelope, whose `execution` is deliberately `"succeeded"` with exit 2 (the new `setup_failure_routing` doc explicitly distinguishes this case). This contract says consumers may rely on `"failed"` for setup error envelopes and lists no such exception, so a v27 consumer will reject the core resume route. Either change the runtime semantics or carve successful refusals/skipped-target lines out explicitly here and in `docs/errors.json`, `AGENTS.md`, and `docs/agent-contract-current.md`, with a test for the chosen contract.

pengfei-threemoonslab (commented):
Blocking review findings are inline below. GitHub does not allow the PR author to set a formal “Request changes” state on their own pull request, so this is submitted as a comment review.

Additional public-surface cleanup needed in this PR:

- `docs/passed-verdict-contract.md:3` still identifies `0.16.0b7` as contract v26 even though this branch publishes v27 and `.well-known/agents-shipgate.json` links consumers to that page.
- `STABILITY.md:31-33` and `docs/agent-contract-current.md:22-24` say the newly enveloped error lines leave `next_action` / `next_actions[]` unchanged, but the generated-manifest `internal_error` recovery is intentionally changed from bare `init --minimal` to a contextual command. That command-value migration should be stated explicitly.

Validation performed:

- 675 focused adoption/setup/public-contract tests passed locally.
- The broad non-performance suite reached 100%; its only local errors shared the sandbox-blocked isolated wheel-build fixture. `tests/test_packaging.py` passed separately with build-dependency access.
- GitHub CI, Agents Shipgate, and Agents Shipgate Self Dogfood are green on this head.
- Fresh Shipgate verification is `review_publishable` / `review_required`: no static blockers or policy weakening, with human review required because `AGENTS.md` is a protected trust root.

Please address the inline findings and add regression coverage for the emitted routes, not only their individual payload shapes.

pengfei-threemoonslab:
All four inline findings and both public-surface items are addressed in f3b1339a. Each repro reproduced first; each fix verified against it.

**[P1] `_register_init.py` — requested configuration lost on a refused write.** Reproduced exactly as described, and it was worse than the report: the `--agent-instructions` refresh path has the same hole and reported `setup_complete` while dropping the request. Both `skipped_existing` branches are now preceded by a delta check. When this invocation asked for a pack the manifest does not select, the route is an `edit` naming `policies.control_pack` and the exact value; `doctor` is reached only when the manifest already matches.

"Asked for" is `--control-pack` at other than its default — the same rule `_requested_setup_flags` already applies when deciding whether to repeat the flag. Typer cannot distinguish an omitted option from one passed at its default, and two rules for that would disagree at exactly this boundary. Verified that an *omitted* flag over a `financial-strict` manifest does not invent an obligation.

The pack is the only manifest-scoped request a refused write can be compared against, and `_unapplied_control_pack`'s docstring now says why each of the others cannot be: `--minimal` selects a template for a file this run did not write and nothing on disk records which template a manifest came from; `--ci` / `--agent-instructions` / `--claude-code` write other files and carry their own outcomes; `--allow-unresolved-scope` and `--max-python-files` govern how the workspace was read, not what the manifest declares.

`test_init_reports_the_choice_even_when_it_writes_nothing` now asserts the route as well as the payload, and the walk fixture follows the edit and requires the obligation to be gone afterwards.

**[P1] `_register_init.py` — lossy recovery promoted to the authoritative route.** One `_invocation_flags` builder now feeds all five sites (both early-validation routes, the `internal_error` fallback, the adoption-kit recheck, and the dry-run advance). `--minimal`, `--allow-unresolved-scope`, `--agent-instructions-kit`, and a non-default `--max-python-files` ride along; correcting a value means passing the corrected value rather than dropping the flag. Your repro now emits `init --workspace … --write --minimal --json`.

The scoped refusal deliberately does **not** use the list, and that is now stated where it is built: it is a rerun in a *different* workspace, so a workspace-relative kit path, `--allow-unresolved-scope`, and `--json` are all wrong to repeat unchanged. It keeps `_requested_setup_flags` plus `rebased_kit_flags`.

Coverage is per-option and end-to-end: one parametrized case per semantic option so a missing one cannot hide behind the others, plus a test that runs the emitted recovery and inspects the manifest it produced — with the no-`--minimal` case as the control, so the assertion cannot pass vacuously.

**[P2] `setup_control.py` — incomplete `input_id`.** `action_kind` and the whole ranked `Diagnostic` are hashed now. Both cases are pinned behaviourally: one where only `action_kind` moves (different `next_action.kind`, different `verify_required`), and one where the same rank-1 action routes to different actors purely by diagnostic id.

Extended to `detect`'s and `init`'s own `advance_kind`, with the guard for those written **structurally** rather than behaviourally, and the reason recorded: no branch of `_init_advance` or `_detect_advance` varies the kind without varying the action, so a behavioural test there would pass today for a reason unrelated to the fix. `doctor`'s success identity is deliberately untouched — `STABILITY.md` promises it is the same manifest, same `input_id`, on every machine, and folding an entry-point spelling into that id would break a published claim.

**[P1] `STABILITY.md` — the `"failed"` invariant.** Confirmed against the runtime: `config_already_exists` and the unresolved-scope `config_error` both carry `execution: "succeeded"` with exit 2. I kept the runtime semantics — the line is "did the command reach an answer about the workspace", not "what was the exit code" — and corrected the contract in all four surfaces you named.

What every setup error line guarantees is now stated in terms that are actually schema-enforced: `decision_source: "setup"`, a `decision` from the setup vocabulary, every `permissions` field false, and never `control_state: "complete"`. And `execution` is documented as *not* a marker for "this is an error line" — the `error` field is — with both values tabulated and their causes. Pinned by a test that asserts both shapes occur, and by one asserting the invariant set across both.

**Public-surface cleanup.** `docs/passed-verdict-contract.md` now stamps v27. Both command-value migrations — the recovery flags, and the `internal_error` fallback moving from a bare `agents-shipgate init --minimal` to this invocation with `--minimal` added — are stated explicitly under a "Routing behaviour" heading rather than left under "nothing is removed".

**One more, found reviewing the fix.** `_unapplied_control_pack`'s docstring claimed it guarded a `None` on-disk pack; the code did not. A manifest that loads while its pack does not resolve would have published an edit whose `why` read "the manifest still selects None". No such manifest is reachable today — an unknown pack id fails schema validation, so the defect route claims it first — but that is `_manifest_defect` and `_manifest_control_pack` happening to agree while running different amounts of the loader. Guarded, so a disagreement falls through to `doctor` instead of becoming a route.

**Validation:** full suite, `test_adapter_static_only`, the perf job, and `ruff` all green locally on f3b1339a.

src/agents_shipgate/cli/_register_init.py
```diff
@@ -722,6 +848,42 @@ def _init_advance(
...
+        # configuration it does not carry. Both branches below hand the caller
+        # onward — one to the gate, one to `doctor` — and both would drop the
+        # request silently: `doctor` reads a manifest that loads and advances to
+        # verification under the pack that is *there*, so an envelope-only
+        # caller loses the pack it asked for without anything ever saying so.
+        # A recovery that completes with less than the caller requested reports
+        # success for work it did not do, and on this contract it does it as the
+        # authoritative route.
+        #
+        # This is an agent-owned edit, not a human review, because the value
```
pengfei-threemoonslab: [P1] Do not treat agent-process argv as authenticated human authorization. A governed coding agent can choose `--control-pack` itself. Starting from an existing `financial-strict` manifest, `init --write --control-pack read-only-agent --json` now emits a coding-agent-owned edit even though that transition removes `safeguards.idempotency` for `write` and `safeguards.rollback` for `production_operation`—the documented `control_pack_weakened` direction. Existing-manifest pack changes need a human route unless the implementation can prove monotonic strengthening or authenticated human provenance.

pengfei-threemoonslab (commented):
Second-pass review of `f3b1339a`. The direct paths behind the previous four comments were improved, but executing the newly emitted recovery routes exposed three P1 issues plus several contract/test gaps. Please treat the P1 comments as blocking. GitHub does not permit `REQUEST_CHANGES` on a self-authored PR, so I am submitting this as `COMMENT`.

Additional findings that cannot be anchored cleanly:

- `docs/errors.json` still tells `config_already_exists` consumers to hand-edit or delete the manifest. Contract v27 now publishes conditional authoritative routes (`doctor`, an exact pack route, or a skipped-target edit), and `AGENTS.md` tells agents to prefetch this catalog. The recovery hint should tell consumers to follow `next_actions[0]` and describe those cases instead of suggesting deletion.
- The PR currently conflicts with current `main` in `CHANGELOG.md`, `STABILITY.md`, and `docs/passed-verdict-contract.md`; it needs a rebase/conflict resolution before it can merge.

Validation performed: focused adoption/control/public-contract suites passed; the broad non-perf suite reached 100% with only the sandboxed packaging-build fixture failing, and all 8 packaging tests then passed with build-dependency access; Ruff passed on every changed Python file; all three current-head GitHub workflows are green. A fresh Agents Shipgate verifier run returned `review_publishable`, `merge_verdict: human_review_required`, because the PR changes a release trust root.

Please add end-to-end regression tests that execute the emitted commands for the same-workspace capped retry, a large scoped candidate, an already-adopted candidate with a requested pack, and explicit `--control-pack default`; helper-only assertions do not cover these call-site losses.

src/agents_shipgate/cli/_register_init.py
```diff
@@ -225,6 +228,85 @@ def _requested_setup_flags(
...
+    agent_instructions_kit: Path | None,
+    max_python_files: int,
+) -> list[str]:
+    """Every semantics-bearing option this invocation carried, as flags.
+
+    The list a *same-workspace* rerun must repeat. `_requested_setup_flags` is
+    the smaller answer to a different question — what a rerun in a **different**
+    workspace may repeat — and it deliberately omits
+    ``--agent-instructions-kit`` because a kit path is only meaningful relative
+    to a workspace. Every recovery this command publishes reruns the same
```
pengfei-threemoonslab, line 322: [P1] Preserve the full invocation on the same-workspace capped retry. The actual `setup_command` for `_unresolved_scope_actions` is still built from `_requested_setup_flags`, so it drops `--allow-unresolved-scope` and `--agent-instructions-kit`. In a two-project workspace, `init --write --allow-unresolved-scope --max-python-files 1 --json` emits a same-root retry without `--allow-unresolved-scope`; following it exactly returns `refused_unresolved_scope` and creates no manifest. A custom kit likewise falls back to the bundled kit. Route that retry through this full-invocation builder, replacing only the parse cap, and execute both variants in regression tests.

src/agents_shipgate/cli/_register_init.py
```diff
@@ -1461,6 +1730,11 @@ def init(
             ),
             tool_surface_origin=tool_surface_origin,
             scaffold_summary=scaffold_summary,
+            # What this invocation asked the manifest to select, and what the
+            # manifest on disk actually selects. A refused write leaves the two
+            # able to disagree, and the onward routes cannot see it.
+            requested_control_pack=control_pack,
```
pengfei-threemoonslab: [P1] Reconcile the requested pack before publishing scoped candidate routes. `_init_advance` returns `scope_actions[0]` before it reaches the pack-mismatch branch, and those candidate actions do not inspect an adopted candidate's manifest. Repro: an ambiguous root contains a candidate with an existing `default` manifest; root `init --write --control-pack financial-strict --json` emits `doctor` for that candidate. Following it reports `setup_complete` and advances to verify while the manifest remains `default`, silently losing the request. Candidate routing must compare/reconcile the on-disk pack and preserve the correct ownership boundary.

src/agents_shipgate/cli/_register_init.py
```diff
@@ -641,6 +723,48 @@ def _init_reason(
...
+    the routes for both are decided above. Guarded here rather than left to
+    those earlier branches: they answer ``_manifest_defect``, which runs the
+    loader, while this reads ``_manifest_control_pack``, which runs the loader
+    *and* ``resolve_control_pack``. Two functions with two failure surfaces
+    happen to agree on every manifest reachable today, and "happen to agree" is
+    not a thing to route on — without the guard, a manifest that loads while
+    its pack does not resolve publishes an edit whose ``why`` reads "the
+    manifest still selects None".
+
+    "Asked for" is ``requested != DEFAULT_CONTROL_PACK_ID``, the same rule
```
pengfei-threemoonslab: [P2] Preserve an explicit request for the default pack. Against an existing `financial-strict` manifest, both an omitted option and explicit `--control-pack default` currently emit the same `doctor` route and the same identity, leaving `financial-strict` selected. The current Click/Typer stack can distinguish these with `Context.get_parameter_source(...)=COMMANDLINE` versus `DEFAULT`; carry that provenance into request detection and `input_id`.

src/agents_shipgate/cli/_register_init.py
```diff
@@ -225,6 +228,85 @@ def _requested_setup_flags(
...
+            agent_instructions=agent_instructions,
+            control_pack=control_pack,
+        ),
+        *(["--allow-unresolved-scope"] if allow_unresolved_scope else []),
+        *(
+            ["--agent-instructions-kit", str(agent_instructions_kit)]
+            if agent_instructions_kit is not None
+            else []
+        ),
+        # A raised parse cap is why `--write` did not refuse on a large
```
pengfei-threemoonslab: [P2] Apply this invariant to scoped candidate commands too. A nondefault `--max-python-files` is omitted because candidate routes use the smaller setup flag list, but the cap remains semantics-bearing for the selected child. With a complete root parse at 2000 and a candidate containing 1002 Python files, following the emitted candidate command re-truncates, exits 2 with `refused_unresolved_scope`, and writes no manifest despite its `expects` claiming creation. Carry the raised cap while rebasing only genuinely workspace-relative values.

CHANGELOG.md
```diff
@@ -2,6 +2,74 @@
...
+  `detect`, `init`, and `doctor` on stdout and left the error stream out.
+  `doctor`'s failure routes picked it up during that rollout; `detect`'s and
+  five of `init`'s did not. So whether a caller that routes on `control` could
+  route at all depended on which setup command had failed and on which of its
+  failures — and the run that most needs a route is the one that printed no
+  payload to carry it. Every agent-mode error line from those three commands
+  now carries the same envelope its `--json` payload would, projected from the
+  same selected route as `next_action` and `next_actions[]`, so the three
+  cannot disagree. A failure envelope is fail-closed by construction:
+  `execution: "failed"`, `decision: "setup_incomplete"`, and every field of
```
pengfei-threemoonslab: [P2] Remove this contradictory invariant. These lines say every failure envelope has `execution: "failed"` and `decision: "setup_incomplete"`, while lines 52–59 of the same new entry correctly document `config_already_exists` as `execution: "succeeded"` and allow the full setup decision vocabulary. State only the schema-enforced invariants here so consumers are not given two incompatible contracts.

tests/test_setup_control.py
```diff
@@ -2041,3 +2051,596 @@ def test_a_manifest_type_mismatch_routes_to_the_editor_not_the_tracker(
...
+    ),
+    (
+        "control_pack",
+        {"control_pack": "financial-strict"},
+        ["--control-pack=financial-strict"],
+    ),
+    (
+        "agent_instructions_kit",
+        {"agent_instructions_kit": Path("/ws/.agents-shipgate/kit.yaml")},
+        ["--agent-instructions-kit", "/ws/.agents-shipgate/kit.yaml"],
```
pengfei-threemoonslab: [P3] Make this expected path platform-neutral. `Path("/ws/.agents-shipgate/kit.yaml")` stringifies with backslashes on Windows, while this expected argv hardcodes POSIX separators, so the new parametrized test fails there. Derive the expected token with `str(kit_path)` (or use `tmp_path`).

pengfei-threemoonslab:
All seven inline findings and both unanchored items are addressed in 62180dab, rebased onto current `main`. Each repro reproduced first.

**[P1] `--control-pack` on argv is not authenticated human provenance.** You are right, and this is the one I got wrong rather than merely incomplete — I considered the objection in the first round and talked myself out of it on the grounds that the value came from the command line rather than from inference. That reasoning does not survive the case that matters: a *governed* coding agent composes its own argv.

The **direction** decides the owner now. Confirmed your numbers before changing anything: `financial-strict -> read-only-agent` weakens `write` and `production_operation`. A transition that keeps at least every obligation the manifest has today is a coding-agent `edit`; one that drops any obligation is `human_review_required` with no command, naming the effects and controls it would remove; a pack this build cannot resolve is a human route too, because "cannot compare" is not "nothing changed".

`_weakened_pack_rules` moved to `core.control_packs.weakened_pack_obligations` beside the packs it compares, and `verify_policy` delegates to it. Two implementations of "did this get weaker?" is how one of them stops seeing a downgrade (#410 §F), and the setup route needed the same answer.

Both directions are pinned, so "always human" cannot pass the pair.

**[P1] The same-workspace capped retry.** Reproduced: following it returned the same `refused_unresolved_scope` and wrote nothing. The root cause was that "which options may a rerun repeat?" has *two* answers and I had used one list for both. There are two now, with one derivation: `_portable_setup_flags` is what transfers to another workspace, and `_invocation_flags` is that plus `--allow-unresolved-scope` and the workspace-relative kit. The retry replaces the cap and nothing else. The test follows it and requires the manifest it promised to exist.

**[P2] Candidate commands and the parse cap.** Same list, same fix: a cap bounds how much is read, not which directory, so it is in the portable one. Pinned by asserting every emitted candidate `init` carries the bound the root run was given.

**[P1] `scope_actions[0]` returned before the pack check.** `scope_candidate_actions` now reconciles an adopted candidate's on-disk pack through the same shared route and the same ownership rule, before it can hand back `doctor`. The test requires that no route for that candidate is a `doctor` handoff while a request is outstanding.

**[P2] Explicit `--control-pack default`.** Read from the parser now, and carried into the route and the identity. There is a trap worth recording: **Typer ships a vendored Click**, so the value from `ctx.get_parameter_source(...)` is a `typer._click.core.ParameterSource` member while `click.core.ParameterSource.COMMANDLINE` is a different enum class — `is` and `==` are both false, silently, in the direction where every option reads as unasked and the route simply never fires. It is compared by member name, with that written down at the call site, and pinned by a test whose negative control (restoring the identity spelling) fails with `assert None == 'financial-strict'`.

**[P2] The contradictory changelog invariant** is gone; that paragraph states only the schema-enforced ones and points at the `execution` split below it.

**[P3] The Windows path** is derived from the `Path` rather than written out.

**`docs/errors.json`.** The `config_already_exists` hint now says to follow `next_actions[0]` and describes the three cases it can carry — `doctor`, the pack reconciliation with its two owners, and an unresolved human-owned declaration outranking both — and says deleting the manifest discards the declarations in it, which is a person's decision rather than a recovery. `control` is listed in that kind's `additional_fields`; the kinds shared with non-setup commands keep the general rule in `stderr_format`, because listing it per kind there would promise it on lines that cannot carry one.

**Rebase.** Done — both conflicts were additive top-of-section blocks in `CHANGELOG.md` and `STABILITY.md`; the landed entries keep their position and this branch's follow. `docs/passed-verdict-contract.md` resolved cleanly at v27.

**Three more found while fixing, all in this round's own code:**

- The weakening route's prose ran past the 400-byte envelope cap, and the clause being truncated was the one saying *why a person owns the decision* — leaving a route that read like a formatting quirk. The reason leads now and the effect list is fitted to what is left, the shape `_placeholder_review_why` already uses. On a 130-character path the sentence lands at 396 bytes with the detail dropped, which is the right end to lose.
- I had introduced a *second* reader for "which pack governs this manifest" — one taking bytes, one taking a path. One `manifest_control_pack` now serves both.
- Both reconciliation call sites used `assert` to narrow `str | None`. `assert` is stripped under `-O`, and what it narrows is the difference between naming a pack and printing `None`; both branches test the operands directly.

**Validation:** full suite, `test_adapter_static_only`, the perf job, and Ruff all green locally on 62180dab.

PR #464, opened 2026-08-30
feat(mcp): report contradictory client annotations

### LINKED ISSUE #462: `readOnlyHint: true` on a tool with destructive evidence is refused internally and reported nowhere — the readers who obey the hint are MCP clients
## Summary

An MCP tool can self-assert `readOnlyHint: true` while every independent signal says destructive. The engine handles this correctly for **its own** verdict — the permission lattice keeps `effect='destructive', risk_level='critical'` and the pass gate is not fooled — and emits **no finding of the contradiction anywhere**.

The reader who needs that finding is the one who cannot run our lattice: downstream MCP clients read `readOnlyHint` / `destructiveHint` raw and use them to calibrate confirmation prompts. A false `readOnlyHint: true` means real under-prompting in every client that honors it. Shipgate computes the proof that the hint misleads those clients, uses it to protect its own conclusion, and discards it.

This is the #403 shape one audience further out: a stage computed the right signal, connected it to *our* decision, and did not connect it to the reader whose behavior the signal governs.

## Reproduction

At `71254cb4`, one MCP tool named `delete_account`, enumerated at `high` extraction confidence, no declaration in play:

```python
tool = Tool.model_validate({
    "id": "t", "name": "delete_account", "source_type": "mcp", "source_id": "s",
    "annotations": {"readOnlyHint": True},
})
[t2] = risk_hints.enrich_tools_with_risk_hints(manifest, [tool])
assess_tool_semantics(t2, None)
classify_tool_permission(t2)
```

With `readOnlyHint: true`:

```
hints:            [('read_only', 'high', 'mcp_annotation'), ('destructive', 'medium', 'keyword')]
effect:           status='inferred', confidence='medium'
issues:           ['inferred_effect_only']
permission_class: classes=('write','destructive','unknown'), effect='destructive',
                  risk_level='critical', reasons=('inferred_effect_only','mcp_annotation','risk_hint:keyword')
```

Same tool, no annotations:

```
hints:            [('destructive', 'medium', 'keyword')]
effect:           status='protocol_default', confidence='low'
issues:           ['missing_effect_evidence']
permission_class: classes=('write','destructive','unknown'), effect='destructive',
                  risk_level='critical', reasons=('missing_effect_evidence','mcp_protocol_default','risk_hint:keyword')
```

Three observations:

1. **The contradiction exists and is held in one place.** Both hints sit side by side on the tool; both reasons (`mcp_annotation`, `risk_hint:keyword`) sit side by side in the profile. Nothing joins them into a finding.
2. **The verdict layers are fail-closed.** Effect resolves `destructive` either way; `is_effectively_read_only` is `False`; the #409 resolver already refuses to let a declaration lean on the hint, with a docstring naming this exact threat (`core/semantic_assessment.py:544` — *"an MCP server can assert `readOnlyHint: true` about itself"*). The distrust is implemented; only the *report* of what was distrusted is missing.
3. **Asserting the false hint is strictly better for the server than silence.** `inferred_effect_only` at `medium` beats `missing_effect_evidence` at `low`. Defensible as resolution mechanics — but it means the misleading state outranks the honest-ignorance state, and neither is reported.

## The delta form

Both states above resolve to the same classes / effect / risk level, so a PR that **only flips the hint** — adds `readOnlyHint: true`, or sets `destructiveHint: false` — changes what every downstream client will do and changes nothing the capability delta is built from. No `SHIP-MCP-*` check fires on the flip itself: the family (`checks/mcp_permissions.py`) covers structural expansion (`SHIP-MCP-PERMISSION-EXPANDED`), read-only server addition (`SHIP-MCP-READONLY-SERVER-ADDED`), env-secret passthrough, auto-approve, unknown schema — none of their predicates is the annotation.

That flip is the assertion-only narrowing attack: the capability stays what it was; the *label clients trust* narrows.

## Where each half already lives

- `core/risk_hints.py:314` — `readOnlyHint: true` → `read_only` hint at `high`; `:323` — `destructiveHint: true` → `destructive` + `write` at `high`; keyword-derived `destructive` from the same function.
- `core/semantic_assessment.py:357` / `:370` — the hint becomes a `high` static-declaration effect claim; `:346` already validates the boolean *type* (`invalid_semantic_annotation`) — the value's honesty is never assessed.
- `core/capability_lattice.py:359` — the profile conditions on the annotations and carries both reasons.
- `checks/mcp_permissions.py` — the check family this belongs to.

Everything the finding needs is computed on the object it would be filed against.

## Proposed check

One new check id in the existing family (working name `SHIP-MCP-ANNOTATION-CONTRADICTION`), review-tier — the spec calls these hints advisory, so this is a review question, not a block:

- **Static form:** an explicit `readOnlyHint: true` (or explicit `destructiveHint: false`) on a tool whose independent evidence — structural, or keyword at its stated basis — classifies write/destructive, produces one finding per tool naming: the hint, the contradicting evidence *with its basis honestly stated* (a keyword inference is named as one, per the #357 vocabulary line), and the client consequence in one sentence.
- **Delta form (with `--base`):** an annotation changed on a tool whose derived evidence class did not change → one finding naming the flip and the unchanged evidence.
- **Spec defaults respected:** absence is never a contradiction. `destructiveHint` defaults to *true* in the spec — only an explicit `false` asserts narrowing. Only explicit values are challenged.
- **`tool_sources[].trust` does not silence the static form.** Trust governs whether *we* believe the hint for resolution; the finding is about what the server publishes to *clients*, which trust config cannot make true. Whether trust downgrades severity is an implementation decision to document with a test either way.

## Surface discipline

Per [CONTRIBUTING.md § Surface discipline](../blob/main/CONTRIBUTING.md#surface-discipline):

1. **Headline metric:** material-true-finding rate on capability-touching MCP PRs — the observer-mode activation metric. This is the flagship cold-repo finding class: it needs zero manifest, zero declarations, and it is checkable by the maintainer in under a minute against their own source.
2. **Existing surface:** extends `checks/mcp_permissions.py`; joins two signals the engine already computes; no new engine, no schema bump, one additive check id (additive per STABILITY).
3. **Non-goals respected:** no runtime execution, no new adapter, no new command.

## Acceptance

- [ ] An MCP source with `delete_account` + `readOnlyHint: true` yields the finding with both sides' evidence and the tool's source pointer, with **no manifest declarations required**.
- [ ] The same tool without annotations yields no finding (honest ignorance is #396/#410's territory, not this check's).
- [ ] Explicit `destructiveHint: false` beside structural destructive evidence yields the finding; an *absent* `destructiveHint` never does.
- [ ] Delta: a PR that only flips a hint on an evidence-unchanged tool yields the delta-form finding naming the flip.
- [ ] The finding's text passes the adopter-vocabulary guard (#329) and states inferred bases as inferred (#357).
- [ ] Adversarial: a tool with `readOnlyHint: true` and genuinely read-only evidence yields nothing — corroboration is not contradiction.

## Related

#431 (the route that makes MCP servers visible at all — this check is what runs once they are), #440 (the reader-boundary epic; this is the same failure one audience further out), #409 / PR #268 (where the distrust of self-assertion was built), #425 (the golden that should pin this finding's rendering), #357 (vocabulary for inferred bases).

Found while building the MCP-maintainer evaluation plan: this is the finding class that is true, novel to the maintainer, and verifiable in one minute — the engine already knows it and does not say it.

pengfei-threemoonslab:
## Summary

- add the default-on review-tier SHIP-MCP-ANNOTATION-CONTRADICTION check for explicit narrowing MCP hints that conflict with independent structural or inferred evidence
- detect hint-only annotation flips against unchanged base evidence without changing the report schema
- preserve the trust boundary: source trust does not silence contradictions, while reviewed declaration overrides suppress only the exact evidence they answer
- document the check and cover static, delta, trust, override, absent-hint, and genuine-read-only cases

## Verification

- full non-performance suite passed (excluding packaging in the sandboxed run)
- packaging suite passed separately: 8 passed
- Ruff, compileall, generated-schema checks, and diff whitespace checks passed
- Agents Shipgate base-to-head verifier: control_state=complete, decision=passed

Fixes #462

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,26 @@
 
   No engine, schema or check-ID change: goldens, registration and tests only.
 
+- **MCP clients can now see when a server's reassuring annotation contradicts
+  the evidence beside it.** (#462) A tool that publishes `readOnlyHint: true`
+  while independent structural or explicitly labeled inferred evidence says it
+  can write, or `destructiveHint: false` while evidence says it can destroy,
+  now raises `SHIP-MCP-ANNOTATION-CONTRADICTION`. The finding is review-tier:
+  it does not weaken or replace the conservative permission verdict the engine
+  already computes, but it names the hint, the contradicting evidence and its
+  basis, the tool's source pointer, and the under-prompting consequence for MCP
+  clients that consume the annotation directly.
+
+  Absence still follows the MCP defaults and never fires. A source's
+  `tool_sources[].trust` setting cannot make a published contradiction true,
+  while a reviewed action override suppresses only the exact inferred claim
+  IDs it answers. With a base report, the delta form proves a hint-only flip by
+  reconstructing the possible prior boolean states and matching them against
+  the base annotation hash; no raw annotation field or report-schema change is
+  required. Static, hint-only-delta, default, corroboration, trust, reviewed
+  override, source-provenance, catalog, and adopter-vocabulary regressions pin
+  the behavior.
+
 - **The declaration questionnaire is now pinned by something committed.**
   (#425) It is the primary cold-start surface — what an adopter at
   `insufficient_evidence` reads, in order, to reach a verdict — and no shipped
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,12 +47,22 @@
   Absence still follows the MCP defaults and never fires. A source's
   `tool_sources[].trust` setting cannot make a published contradiction true,
   while a reviewed action override suppresses only the exact inferred claim
-  IDs it answers. With a base report, the delta form proves a hint-only flip by
-  reconstructing the possible prior boolean states and matching them against
-  the base annotation hash; no raw annotation field or report-schema change is
-  required. Static, hint-only-delta, default, corroboration, trust, reviewed
-  override, source-provenance, catalog, and adopter-vocabulary regressions pin
-  the behavior.
+  IDs it answers. Policy-eligible structural evidence now corroborates a
+  narrowing hint against weaker keyword or pattern noise, while structural
+  side-effect evidence still triggers review. Reviewed effect declarations and
+  mutually conflicting hints from the same source stay on their existing
+  semantic-evidence routes instead of being counted as independent evidence.
+
+  Inferred-only instances remain detailed, informational findings with
+  `support.policy_eligible: false`; the scan also emits their evidence gap, but
+  release accounting excludes them from blockers and named review items. With
+  a base report, the delta form separately proves that independent evidence is
+  unchanged, detects any exact client-visible annotation-map change, and names
+  a hint flip only when reconstruction matches the base hash. The hash no
+  longer uses finding-fingerprint normalization, so otherwise ignored keys and
+  list order cannot produce a false proof. Static, exact-delta, co-change,
+  corroboration, trust, reviewed-override, source-provenance, end-to-end scan
+  publication, and adopter-vocabulary regressions pin the behavior.
 
 - **The declaration questionnaire is now pinned by something committed.**
   (#425) It is the primary cold-start surface — what an adopter at
--- a/docs/checks.json
+++ b/docs/checks.json
@@ -2446,6 +2446,32 @@
     {
       "autofix_safe": false,
       "category": "mcp_permissions",
+      "default_severity": "high",
+      "description": "MCP narrowing annotation contradicts side-effect evidence.",
+      "docs_url": "https://github.com/ThreeMoonsLab/agents-shipgate/blob/main/docs/checks.md#ship-mcp-annotation-contradiction",
+      "dynamic_default": false,
+      "evidence_fields": [
+        "form",
+        "published_annotations",
+        "contradictions",
+        "independent_evidence",
+        "annotation_changes",
+        "independent_evidence_unchanged",
+        "client_consequence"
+      ],
+      "fires_when": "An explicit readOnlyHint true or destructiveHint false contradicts independent structural or clearly labeled inferred evidence on the same MCP tool.",
+      "floor_severity": "medium",
+      "id": "SHIP-MCP-ANNOTATION-CONTRADICTION",
+      "mvp_tier": "lifecycle",
+      "rationale": "MCP clients can use read-only and destructive annotations to reduce confirmation prompts even when independent evidence indicates the call can write or destroy data.",
+      "recommendation": "Review or correct the MCP annotation before clients use it to reduce confirmation prompts.",
+      "requires_human_review": true,
+      "requires_human_review_regardless_of_patch": false,
+      "suggested_patch_kind": "manual"
+    },
+    {
+      "autofix_safe": false,
+      "category": "mcp_permissions",
       "default_severity": "critical",
       "description": "MCP side-effecting tool is auto-approved.",
       "docs_url": "https://github.com/ThreeMoonsLab/agents-shipgate/blob/main/docs/checks.md#ship-mcp-auto-approve-side-effect",
--- a/docs/checks.json
+++ b/docs/checks.json
@@ -2456,6 +2456,7 @@
         "contradictions",
         "independent_evidence",
         "annotation_changes",
+        "annotation_surface_changed",
         "independent_evidence_unchanged",
         "client_consequence"
       ],
--- a/docs/checks.md
+++ b/docs/checks.md
@@ -1115,13 +1115,34 @@ can let an agent act outside the review boundary.
 Fires when an explicit MCP narrowing annotation conflicts with independent
 side-effect evidence on the same tool: `readOnlyHint: true` beside write or
 destructive evidence, or `destructiveHint: false` beside destructive evidence.
-Absent annotations follow the MCP defaults and never trigger this check. The
-evidence names structural support as structural and keyword or pattern evidence
-as inferred. When a diff proves only the hint changed while that independent
-evidence stayed the same, the finding records the annotation flip explicitly.
-The finding routes to review because MCP clients may use these advisory hints to
-reduce confirmation prompts; it does not change the engine's conservative
-permission verdict.
+Absent annotations follow the MCP defaults and never trigger this check.
+Policy-eligible protocol, provider, or scope evidence outranks a conflicting
+keyword or pattern inference: for example, an OpenAPI `GET` does not become a
+contradiction merely because its description mentions a later delete. Structural
+side-effect evidence still triggers the check.
+
+Only source observations independent of the published MCP annotation contribute.
+Reviewed `action_surface.actions[].effect` declarations are deliberately excluded:
+they are human trust-root assertions, not a second observation of the server.
+Likewise, a source that publishes both `readOnlyHint: true` and
+`destructiveHint: true` remains a `conflicting_effect_evidence` semantic gap; this
+check does not relabel that same-source contradiction as independent evidence.
+
+Supported structural instances route to review because MCP clients may use these
+advisory hints to reduce confirmation prompts. Inferred-only instances remain
+visible in `report.findings` with `support.policy_eligible: false` and
+`agent_action: informational`; they also produce an evidence gap, but are excluded
+from blockers, named review items, and release contribution. The check never
+changes the engine's conservative permission verdict.
+
+With a base report, `annotation_surface_changed` compares the exact complete
+annotation map (including otherwise unrecognized keys and list order), while
+`independent_evidence_unchanged` compares the independent semantic evidence. The
+finding uses `form: delta` and fills `annotation_changes` only when reconstruction
+proves the exact narrowing-hint flip against the base hash; unrelated co-changes
+remain the static form. This semantic check runs in the normal
+`scan`/`check`/`verify` path. It is intentionally not duplicated in the narrower
+`mcp audit` host/config-diff rule table.
 
 ### SHIP-MCP-UNKNOWN-TOOL-SCHEMA
 
--- a/docs/checks.md
+++ b/docs/checks.md
@@ -1110,6 +1110,19 @@ financial, or production is configured with approval mode `approve`. This is
 an explicit release blocker because auto-approved side-effecting MCP tools
 can let an agent act outside the review boundary.
 
+### SHIP-MCP-ANNOTATION-CONTRADICTION
+
+Fires when an explicit MCP narrowing annotation conflicts with independent
+side-effect evidence on the same tool: `readOnlyHint: true` beside write or
+destructive evidence, or `destructiveHint: false` beside destructive evidence.
+Absent annotations follow the MCP defaults and never trigger this check. The
+evidence names structural support as structural and keyword or pattern evidence
+as inferred. When a diff proves only the hint changed while that independent
+evidence stayed the same, the finding records the annotation flip explicitly.
+The finding routes to review because MCP clients may use these advisory hints to
+reduce confirmation prompts; it does not change the engine's conservative
+permission verdict.
+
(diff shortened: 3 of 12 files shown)


pengfei-threemoonslab (commented):
## Review — `feat(mcp): report contradictory client annotations` (#464 / closes #462)

Reviewed at `ed5f000d` against merge-base `0581f3db`. Full non-perf suite green locally (`pytest -n 8 -k "not perf"`, exit 0), Ruff clean, `explain` / `list-checks` / `docs/checks.json` all consistent.

The design is right and the delta mechanism is genuinely nice: reconstructing the two candidate prior boolean states and matching them against `tool_surface_facts.hashes.annotations` proves a hint flip without a schema bump, and reusing `acknowledged_effect_claim_ids` (the `ed5f000d` fixup) instead of re-deriving override ids is the correct call. `_INDEPENDENT_EFFECT_BASES` correctly excludes `protocol_default` and `reviewed_declaration`, and MCP's `destructiveHint` default is respected.

But I ran the issue's own reproduction end to end through `scan`, and **the headline case produces no finding at all**. Details below, plus three smaller defects I could reproduce. I'd hold the merge on #1 and #2.

---

### 1. (blocking) #462's acceptance criterion 1 fails end to end: the flagship finding never reaches `report.findings`

Exactly the repro from the issue — cold MCP repo, no declarations:

```json
{"tools": [{"name": "delete_account",
            "description": "Remove a customer account.",
            "annotations": {"readOnlyHint": true}}]}
```

`scan` output:

```
Counts: critical=0, high=0, medium=0, low=0, suppressed=0
Top findings:
- none
```

The check *does* fire — I instrumented `mcp_permissions._annotation_contradiction_findings` inside a real `scan` and it returns `['SHIP-MCP-ANNOTATION-CONTRADICTION']`. It is then deleted by `cli/scan/decision.py:177-199`:

```python
support = finding.support
if support is None or support.policy_eligible:
    policy_eligible_findings.append(finding)
    continue
...
findings = policy_eligible_findings
```

`tool_finding` derives `support` from `provenance_kind` (`checks/base.py:37`, `_heuristic_support`), and on this tool the only independent claim is `('destructive', 'medium', 'inferred_keyword', 'risk_hint:keyword')`, so `_finding_provenance` returns `keyword_heuristic` → `policy_eligible=False` → dropped. All that survives is an anonymous gap row:

```json
{"kind": "inferred_policy_applicability",
 "why": "SHIP-MCP-ANNOTATION-CONTRADICTION: Policy applicability is supported only by heuristic evidence.",
 "next_action": {"path": "/tools/0", "expects": "Provide reviewed or structural evidence for every indeterminate predicate."}}
```

which names neither `readOnlyHint`, nor the destructive evidence, nor the client consequence — i.e. exactly the information #462 was filed to publish.

Add `"auth": {"scopes": ["accounts:delete"]}` to the same tool and it works perfectly (`high SHIP-MCP-ANNOTATION-CONTRADICTION`, correct evidence). So the check is fine for structural evidence and invisible for the keyword-only case — which is the one the issue calls "the flagship cold-repo finding class: it needs zero manifest, zero declarations."

The drop rule itself is pre-existing and deliberate ("heuristic evidence cannot create a blocking policy finding"). What's new is that this check's stated primary audience is *not* the release gate — it's an MCP client that reads the raw hint — so being non-policy-eligible shouldn't mean being unpublished. Two plausible resolutions, your call:

- keep the finding out of gating but out of the drop as well (it never sets `blocks_release`), or
- accept the drop and make the gap row carry the finding's content, so the maintainer still sees the hint, the evidence, and the consequence.

Either way this needs an **end-to-end test through `scan`**, not through `mcp_permissions.run()`. Every one of the 10 new tests calls `run()` directly, which is why nothing caught this.

### 2. (blocking) `independent_evidence_unchanged` is computed from the wrong thing and reports `false` when the evidence is provably unchanged

`checks/mcp_permissions.py:205-211`:

```python
"form": "delta" if annotation_changes else "static",
...
"independent_evidence_unchanged": bool(annotation_changes),
```

The field is set from whether the flip was *reconstructible*, not from the signature comparison whose name it borrows. Reproduced end to end — base has no hints, head adds `readOnlyHint: true` **and** a benign `"title": "Delete account"` in the same PR:

```
form: static | annotation_changes: [] | independent_evidence_unchanged: False
```

and yet, on the two real reports:

```
base independent: [('destructive','structural_scope','auth_scope'), ('destructive','inferred_keyword','risk_hint:keyword')]
head independent: [('destructive','structural_scope','auth_scope'), ('destructive','inferred_keyword','risk_hint:keyword')]
_claim_signature(base) == _claim_signature(head) -> True
```

Two consequences:

- a field published in `evidence_fields` (so it's contract surface) states the opposite of the fact it names; and
- the delta form — the assertion-only narrowing attack the issue is built around — is defeated by touching *any* second annotation key in the same commit. `test_delta_requires_only_the_contradicting_hint_to_change` pins this as expected behavior, but its scenario (flipping `readOnlyHint` *and* `destructiveHint` on a write tool) is the attack, not a corner case.

The narrowing to only the contradicted annotation in `ed5f000d` is right — reconstructing more keys isn't feasible. The bug is the reporting of the degraded case. Cheap fix: derive `independent_evidence_unchanged` from `_claim_signature(base_claims) == _claim_signature(current_claims)`, and add a separate bit for "annotations changed but the exact flip could not be reconstructed" (you already have `tool_annotation_hash(tool.annotations) != base_tool.hashes.annotations` in hand). Then a co-changed annotation degrades honestly instead of silently.

### 3. The check never weighs contradicting evidence against corroborating structural read evidence — reproducible false positive on a GET endpoint

`_independent_effect_claims` keeps every non-`read` claim and discards every `read` claim, with no comparison of strength. So:

```python
Tool(name="get_account",
     annotations={"readOnlyHint": True, "httpMethod": "GET"},
     description="Fetch an account record before you delete it.")
```

→ **fires**, `severity=high`, on this claim set:

```
read         high   protocol_structure   mcp_annotation
read         high   protocol_structure   openapi_method     <- ignored
destructive  medium inferred_keyword     risk_hint:keyword  <- the whole basis of the finding
```

A second one: `search_audit_log`, `readOnlyHint: true`, description *"Search the audit log for remove events."* → fires, on the word "remove" in prose.

That is #462's last acceptance bullet inverted — "a tool with `readOnlyHint: true` and genuinely read-only evidence yields nothing; corroboration is not contradiction." Here the *only* structural evidence on the tool (HTTP `GET`) corroborates the annotation and is not consulted, while a `medium` keyword hit on the description is enough to publish a `high`-severity claim that the server is misleading its clients. The engine's own resolver is more careful about this pair (`_source_read_conflict` only calls it a conflict when both sides are high-confidence and non-manifest-owned).

Suggestion: require the contradicting claim to at least tie the strongest corroborating `read` claim, or suppress when every non-`read` claim is `inferred_*` and a policy-eligible `read` claim exists. Note this FP is currently masked by #1 (keyword-only findings are dropped), so fixing #1 without fixing this will surface it.

### 4. The trust test pins nothing

`tests/test_mcp_permissions.py:199` `test_internal_source_trust_does_not_silence_published_contradiction` sets `manifest.tool_sources[0].trust = "internal"` and asserts the finding still fires. It fires for every value:

```
trust=None          finding=True   configured_tool_source=None
trust='internal'    finding=True   configured_tool_source=None
trust='external'    finding=True   configured_tool_source=None
trust='untrusted'   finding=True   configured_tool_source=None
```

Three independent reasons it cannot fail:

1. `_mcp_tool()` mints `source_id="mcp-tools"` while the fixture manifest configures `id: "codex"`, so `configured_tool_source(...)` is `None`;
2. `_assessed` (`tests/test_mcp_permissions.py:92`) calls `assess_tool_semantics(enriched, declaration)` and never passes `tool_source=`, so no source config reaches the assessment at all; and
3. `tool_sources[].trust` has **no runtime consumer anywhere in `src/`** — `grep -rn trust src/agents_shipgate` returns only the field declaration and two code comments (`semantic_assessment.py:568`, `:929`).

The PR body and the CHANGELOG both present this as a preserved trust boundary ("A source's `tool_sources[].trust` setting cannot make a published contradiction true"). That's true, but vacuously, and the test that's meant to pin it would keep passing if the check started honoring trust tomorrow. Either wire the fixture through `configured_tool_source` + `assess_tool_semantics(..., tool_source=...)` so the assertion has something to bite on, or drop the test and say in the docs that trust is not consulted because nothing consults it.

---

### Smaller things

**`structural_scope` is labelled "declared scope evidence" (`mcp_permissions.py:54`).** In the working run the scope came from the MCP server's own `tools.json` `auth.scopes`, and the adopter is told: *"Review readOnlyHint: true against inferred keyword evidence and declared scope evidence"* — which reads as "you declared it" for evidence the server published. `structural_scope` covers both `auth_scope` (server) and `action_scope` (manifest), and one label can't serve both. It also contradicts the sentence you added to `docs/checks.md`: "The evidence names structural support as structural." Separately, `bases = sorted(...)` is alphabetical, so the weakest basis is named first when there are two.

**A reviewed declaration is not independent evidence.** `reviewed_declaration` is excluded from `_INDEPENDENT_EFFECT_BASES`, so a manifest declaring `effect: destructive` (or `risk_tags: [destructive]`) for a tool the server publishes as `readOnlyHint: true` yields nothing when no keyword or scope evidence exists — verified. Consistent with #424's "a reviewed tag is not source evidence", but it's also the strongest available statement that the hint misleads clients, and it's the state an adopter reaches *after* answering the questionnaire. Worth a sentence in `docs/checks.md` saying the exclusion is deliberate.

**`readOnlyHint: true` + `destructiveHint: true` is quiet.** The blanket `claim.source != "mcp_annotation"` filter drops the server's own `destructive` assertion, so a server telling clients "read-only" (per spec, `destructiveHint` is ignored when `readOnlyHint` is true) while asserting destructive beside it produces nothing from this check — it only reaches the reviewer as a `conflicting_effect_evidence` gap. Probably acceptable, but it falls out of a blunt filter rather than a considered rule.

**The family is now split across three enumerations.** `policies/mcp-permissions.shipgate.yaml` and `cli/mcp.py:_RULES` each list the five other `SHIP-MCP-*` ids; neither gained the sixth. If that's deliberate because `mcp audit` stays scoped to Codex-config diffs, a line saying so would stop the next person reading it as an oversight — this is the recurring "second implementation" shape from #322.

**CHANGELOG overclaims slightly.** It lists a "catalog" regression among the tests that pin the behavior; I don't see one in `tests/test_mcp_permissions.py` beyond the generic `docs/checks.json` roundtrip that every check gets.

pengfei-threemoonslab:
Addressed the full review in 95090ec3. I also validated the exact integration shape with a clean synthetic merge of that head into origin/main at 0310c4399, without adding branch churn.

- The keyword-only flagship case now survives the real scan path as a detailed informational finding while support remains policy-ineligible, the evidence gap remains, and release accounting excludes it. A run_scan regression pins the full contract.
- Delta evidence now reports annotation-surface change, independent-evidence equality, and exact reconstructed hint flips separately. Annotation hashing preserves every client-visible key and list order.
- Policy-eligible structural read evidence now suppresses weaker keyword noise for both narrowing hints; both reported prose false positives and structural side-effect counterexamples are covered.
- Source trust is exercised through configured full-scan fixtures for unset, internal, external, and untrusted values.
- Evidence labels and ordering are corrected. Docs now explain reviewed-declaration exclusion, same-source semantic conflicts, and why this full-scan check is intentionally separate from mcp audit. The CHANGELOG overclaim is removed.

Final validation on the combined latest-main tree: Ruff clean; git diff --check clean; 7,015 non-performance tests passed with 7 skipped; 8 wheel-packaging tests passed. All required PR checks pass (with the expected release-tag-consistency skip). Agents Shipgate control is complete with decision passed against the same latest main revision, and GitHub reports the PR mergeable and clean.

Issue #460, opened 2026-08-29
effect_readings counts a risk_overrides tag as something the scan observed


pengfei-threemoonslab:
## What happens

`effect_readings` is documented as "what the scan saw" — the readings the questionnaire prints under *"what this scan read this action's effect as"*, and the set `effect_derivation_id` digests into the `basis` pin. Its filter excludes `DECLARATION_CLAIM_SOURCES`, which is the **action row** only. A reviewed `risk_overrides.tags` entry reaches the effect dimension by the other route — `risk_hint:manual`, basis `reviewed_declaration` — and is not excluded.

```python
from agents_shipgate.core.domain import Tool, ToolRiskHint
from agents_shipgate.core.semantic_assessment import assess_tool_semantics, effect_readings

tool = Tool.model_validate({
    "id": "t", "name": "send_email", "source_type": "google_adk", "source_id": "s",
    "extraction_confidence": "high", "extraction": {"surface": "enumerated"},
    "risk_hints": [
        # exactly what `_apply_manual_override` writes for `risk_overrides.tags`
        ToolRiskHint(tag="external_write", source="manual",
                     confidence="high", basis="reviewed_declaration"),
    ],
})
print(effect_readings(assess_tool_semantics(tool, None).effect))
# [EffectReading(effect='external_communication', sources=('risk_hint:manual',),
#                observed=True, policy_eligible=True)]
```

`observed=True` on a line a human typed into `shipgate.yaml`.

## Why it matters

Three consumers read this set, and each reads it as evidence:

- **The questionnaire** prints it under "what this scan read this action's effect as". It did not read it; someone wrote it.
- **`effect_derivation_id`** digests it into `action_surface.actions[].basis`. Its docstring's second property — *"it does not move when the answer arrives... because declaration claims are excluded by `effect_readings`"* — does not hold for this route: adding a `risk_overrides.tags` entry moves the pin and raises `declaration_drift` on every pinned action the selector touches.
- **`propose_effect_declaration`** derives a first-answer proposal from it, so a proposal can be partly a restatement of the manifest.

This is the same class as #424 — the manifest reaching the effect dimension by two routes and a comparison excluding only the first. Two other instances of it were closed in [#457](https://github.com/ThreeMoonsLab/agents-shipgate/pull/457) by promoting `_is_manifest_owned` to `is_manifest_owned_effect_claim` and asking it at both sites.

## Why it is filed rather than folded into #457

[...]

PR #465, opened 2026-08-30
test(samples): pin the #424 repair loop with a committed artifact (#424)

pengfei-threemoonslab:
## Summary

Closes the last open acceptance item on #424:

> The repair loop is exercised by a committed artifact, not only unit tests.

#457 fixed the engine and guarded it thoroughly, but **every guard on the
`declare_risk_tags` route runs against tools built in-test** — the same blind
spot that let the class ship. `test_the_published_repair_closes_the_row_it_is_printed_on`
sat inside that blind spot and passed for the entire time the published repair
was swapping its row for a blocking conflict.

`samples/google_adk_cold_start_agent` (#425) could not close the gap: a cold
start has no declaration to challenge, so `declaration_below_inferred_evidence`
is the one questionnaire shape it cannot render — and a challenged row is the
whole subject of #424.

### The fixture

`samples/declaration_repair_agent` is the step *after* a cold start. Every
action is declared, controlled and owned, so the two challenged rows are the
only thing standing between it and a verdict:

| | verdict | open declaration questions |
| --- | --- | --- |
| as committed | `insufficient_evidence` | 2 |
| after pasting both published blocks | `passed` | 0 |

Both halves are committed artifacts. `expected/suggested-declarations.yaml` is
the questionnaire an adopter reads here;
`test_pasting_the_committed_repair_blocks_reaches_passed` takes the blocks out
of *that file*, merges each into the action it names (the `override:`
alternative deleted, as the block's own comments instruct), rescans, and
requires `passed`.

Both actions are `source_type: mcp` with keyword-inferred readings — the shape
the issue calls out as first-contact MCP evaluation.

### Two rows, because the second is the case #424's own repair creates

`billing.cancel_invoice_email` already carries a reviewed
`risk_tags: [financial_write]`. `risk_tags` is one YAML key, so a block naming
it **replaces** it: the first draft of the repair published only the newly
uncovered category, which told that reviewer to delete their own tag and let
the next scan reopen the row asking for it back. The committed golden publishes
`[financial_write, destructive]`, and
`test_the_published_repair_keeps_the_tag_the_reviewer_already_wrote` reads both
the declared list and the published list out of committed files rather than
restating either.

### The guards close over both routes a regression can take

Each confirmed by perturbation rather than asserted:

| perturbation | what fails |
| --- | --- |
| revert `semantic_assessment.py` / `domain.py` / `action_surface.py` to pre-#424 | `test_repair_scaffold_matches_its_golden` (bytes), and the paste test with **both** rows reopened |
| make the repair publish only the added tags | `test_repair_scaffold_matches_its_golden` |
| ...then regenerate the golden to make that pass — the natural next move | `test_pasting_the_committed_repair_blocks_reaches_passed`, naming the row it reopened, and `test_the_published_repair_keeps_the_tag_the_reviewer_already_wrote` |

That last row is the point. A committed golden alone is not a guard, because
the reflex when a byte comparison fails is to regenerate it; the paste test is
what makes regeneration fail too.

### Verification of the acceptance list this rides on

Re-verified independently at `0581f3db` before writing anything, so this PR is
not resting on the merged PR's own numbers:

- the issue's reproduction closes its row;
- **1656** repair applications (every declared effect × every 1–3 observation
  combination, bare and already-tagged, applied by *replacing* `risk_tags`)
  leave **0** rows open;
- **92** first-answer proposals, **0** regressions;
- `#417` declared scopes still contradict — `crm.delete` + `effect: read` still
  raises `conflicting_effect_evidence`;
- a non-covering tag still raises `declaration_below_inferred_evidence`.

### Fixture hygiene

Every control the fixture declares was checked by dropping it and rescanning;
each one blocks or holds back the **repaired** state. Two spellings that would
have been redundant are deliberately absent —
`policies.require_approval_for_tools` (already satisfied by `approval.required`
on each row) and `safeguards.idempotency` on the row (already satisfied by
`policies.require_idempotency_for_tools`, which is what
`SHIP-SIDEFX-IDEMPOTENCY-MISSING` actually reads) — and the README says which
and why. `report.md` gets a byte comparison so the sample ships no committed
artifact that nothing reads.

The owners are set through `risk_overrides.tools[].owner` with no `tags:`
entry, deliberately: `risk_overrides.tags` is the manifest's *other* route into
the effect dimension, so keeping it out leaves this golden clear of #460.

## Review rounds

Three passes over the fixture itself. Two found real gaps in what it pinned.

**1. The fixture pinned the row but not the *route*.** `declaration_below_inferred_evidence`
publishes two, and only `declare_risk_tags` was broken by #424. A change that
answered these rows by raising `effect:` instead would still close them, so the
paste test would stay green while the one committed artifact for the tag route
was silently repurposed. Confirmed by forcing the route to `raise_effect` and
regenerating the golden: `test_the_repair_golden_still_renders_a_challenged_row`
is what fails, naming the block that switched.

**2. Two control lines were dead, and the README over-claimed.** The first draft
said the manifest declares every control the actions owe. Dropping each line and
rescanning found `policies.require_approval_for_tools` already satisfied by
`approval.required` on each row, and the row's `safeguards.idempotency` already
satisfied by `policies.require_idempotency_for_tools` — which is what
`SHIP-SIDEFX-IDEMPOTENCY-MISSING` actually reads. Both removed; every remaining
line is now load-bearing against the *repaired* state, which is the honest
claim and is the one the README makes. Note that a one-at-a-time sweep cannot
tell "dead" from "redundant with its pair" — the two approval spellings each
looked droppable until both were dropped together and the run blocked.

**3. Wording, twice.** The README used "two routes" for both `effect_repair`'s
kinds and the block's account-or-override choice; and `authority: {mode: none}`
on tools that delete correspondence read as a modelling recommendation when it
is a fixture convenience — publishing a credential instead would add an
`auth_scope` reading, and the readings here are deliberately keyword-only.

Also added in review: a byte comparison for the sample's `report.md`, which
would otherwise have been a committed artifact nothing reads.

## Maintainer review round (`ca831f0f`)

Five findings, each reproduced here before being addressed.

**The one adopter-facing string #424 is about was committed and unread.** #424 reached a reader as a *sentence*, and this sample commits it in `evidence_gaps[].next_action.expects` and `agent_summary.first_recommended_action.why` — neither covered, since `report.json` is compared only for schema version, decision, and the set of field *paths*. It matters here specifically: `EffectRepair.instruction` interpolates the whole list and the newly added part separately, and `billing.cancel_invoice_email` is the first row in the repo where they can differ (the only other test pinning `expects` for this route uses a row with no pre-existing tags, so `tags == added` by construction). Confirmed by rewriting the f-string to interpolate `added`: all four fixture tests stayed green while the adopter would read `[destructive]` on the release surface and `[financial_write, destructive]` in the block. Now joined, with a count assertion so a golden that stops emitting one surface fails rather than agreeing vacuously.

**Test 3's docstring understated it.** The separating perturbation is *rewriting*, not deletion: canonicalize `financial_write` → `financial_action` and regenerate, and the row still closes, the sample still reaches `passed`, the sentence still agrees with the block — and `test_the_published_repair_keeps_the_tag_the_reviewer_already_wrote` is the sole failure. It is the only guard on "never rewrite what the reviewer wrote". Docstring corrected.

**The regeneration recipe dropped every normalization its sibling exists to record** — textual `<REPO>` replace (a silent no-op on Windows), no `generated_reports` normalization, no forced `newline="\n"` — which made the repo's *designated recovery path* the counterexample to `test_sample_expected_goldens_are_committed_with_lf_newlines`'s own docstring. Replaced with the sibling's runnable block; verified by extracting it back out of the committed README and running it verbatim against dirtied goldens.

**Two false claims.** `destructive` does not introduce `confirmation.required` (`external_communication` already obliged it on both rows), and approval is new only on `support.delete_case_message`; `safeguards.rollback` is the only control new to both. And the cold-start contrast was much too broad — that sibling declares two actions, pre-fills `effect: financial_write`, and prints a challenged *authority* note. The narrow true claim is that its golden holds **zero** `risk_tags:` blocks, so a declaration challenged in the effect dimension appears in no committed artifact before this one. Corrected in all four places it appeared.

**Cross-link** is now bidirectional, and #424's acceptance checklist is ticked — all six, since none had been, with the deviation from the issue's proposed location recorded there.

## Not in scope

#460 stays filed. It moves the `basis` pin for every tool a `risk_overrides`
selector touches and changes what `propose_effect_declaration` must account
for; this fixture is deliberately built so that change cannot churn its golden.

## Type

- [ ] Input adapter change
- [x] Documentation only
- [x] Tests and sample fixtures

## Verification

CI is authoritative. Run locally:

- `python -m pytest tests/ -n 8` — full suite green (exit 0), before and after.
- `python -m ruff check .`, `python -m compileall -q src tests`,
  `python scripts/generate_schemas.py --check` — all clean.
- Every new guard re-run under each perturbation in the table above.

## Release-readiness notes

- [x] No user-code import added to default scan paths — the sample's tools are
      read from a JSON inventory; nothing is executed.
- [x] No check IDs added, removed, or renamed
- [x] No schema change; `report_schema_version` unchanged. No engine source
      file is touched by this PR.


pengfei-threemoonslab (commented):
## Review

Reviewed at `63de60ab`, merged onto `main` (`342e9722`) locally first — the PR base `0581f3db` is one commit behind and `#459` moves nothing in this sample's goldens.

**Verdict: approve.** The central claim holds under independent perturbation: this really is a committed artifact that the regenerate-the-golden reflex cannot turn into a rubber stamp. Two follow-ups below, neither blocking; the second is the one I'd actually like folded in.

---

### What I reproduced rather than took on trust

Each row of the PR body's perturbation table, re-run here (source backed up to `/tmp` and restored with `cp`, never `git checkout --`):

| perturbation | what failed |
| --- | --- |
| `tags = list(added)` in `_effect_repair` — the round-2 defect | `test_repair_scaffold_matches_its_golden`, on bytes |
| …then regenerate the golden with this README's recipe | `test_pasting_the_committed_repair_blocks_reaches_passed`, naming `billing.cancel_invoice_email`, **and** `test_the_published_repair_keeps_the_tag_the_reviewer_already_wrote` |
| revert the `REVIEWED_RISK_TAG_CLAIM_SOURCES` exclusion in `_assess_effect` | byte comparison + paste test, with **both** rows reopened |

The second row is the whole argument of the PR and it is genuinely closed.

One perturbation that isn't in the table, and which I think is the better justification for the third test than the one the docstring gives: canonicalize the reviewer's spelling (`financial_write` → `financial_action`, one category under two names — the thing the source comment warns against) and regenerate. The paste test stays **green** — the row still closes, the sample still reaches `passed` — and `test_the_published_repair_keeps_the_tag_the_reviewer_already_wrote` is the only failure. So it isn't a restatement of the byte comparison; it is the sole guard on the "never rewrite what the reviewer wrote" half of the round-2 fix. Worth saying so in its docstring.

Other claims checked and found accurate:

- **Fixture minimality.** Dropped each declared control from the *repaired* manifest and rescanned. Every one moves the endpoint off `passed`: all six row controls → `blocked`; either policy list, and each individual entry of `require_confirmation_for_tools` → `blocked`; either owner → `review_required`; `tool_sources[].binding` → `insufficient_evidence`. `risk_overrides.tools[].reason` and `agent.declared_purpose` are schema-required, so they aren't droppable at all. The README's claim, caveat included, is honest.
- **`SHIP-SIDEFX-IDEMPOTENCY-MISSING` reads the policy list, not the row.** True — `checks/side_effects.py:16` reads `context.manifest.policies.idempotency_tools()` and never consults `safeguards.idempotency`. The manifest comment justifying that omission is correct.
- **The `#460` clearance.** `risk_hint:manual` appears **zero** times in the committed golden, so the fixture really does stay off `risk_overrides`' route into the effect dimension.
- **The repaired state is clean, not merely `passed`** — zero evidence gaps and zero findings — so the paste test's verdict assertion isn't concealing a traded-down gap.
- **The paste is a real paste.** `ActionDeclarationConfig` is `extra="forbid"`, so `tool_id` / `source_id` / `basis` in a published block are load-bearing fields rather than silently-ignored noise, which is what makes `action.update(block)` proof that the block is pasteable verbatim. And pasting *without* deleting `override:` fails closed with a good message (`<REVIEW_REQUIRED> is an unfilled scaffold placeholder and is not reviewed evidence: …`), so the README's "delete the override stanza" is guidance rather than a load-bearing secret.
- **The structural-drift registration earns its place, by more than the comment claims.** Against the union of the other five samples already in that list, this one contributes 9 new field paths — `declaration_template.risk_tags` *and* `declaration_template.override{,.evidence,.reason}`, `…semantic_assessment.effect.claims[].evidence.declared_risk_tags` across four fact surfaces, and `reason_counts.declaration_below_inferred_evidence`.
- Full suite green after merging `main`; `ruff check` clean; the four new tests cost 0.18 s in total.

---

### 1. The regeneration recipe drops every normalization its sibling README exists to record — medium, docs only

`samples/declaration_repair_agent/README.md` § Regenerating runs `run_scan`, then says in prose: *"replace the absolute repository path in `expected/report.json` with `<REPO>`, and delete the `expected/current-control.json`."*

`samples/google_adk_cold_start_agent/README.md` — landed one PR earlier, in `#458` — documents four things there "that look like details and are not". This recipe reproduces three of them:

- **The textual `<REPO>` replace is a Windows no-op.** `json.dumps` escapes the separators, so the file holds `C:\\repo\\samples\…` while `os.getcwd()` is `C:\repo\samples\…`; the two never match, `manifest_dir` stays absolute, and `test_sample_expected_report_json_uses_repo_placeholder_for_manifest_dir` fails *on the machine that produced the golden*. The sibling assigns the field a POSIX value structurally, for exactly this reason.
- **`generated_reports` is never normalized.** On Windows the relative `output_dir` records `expected\report.json`, tripping the `"\\" not in written` assertion in that same test.
- **Newlines are inherited, not forced.** All three writers open text mode with `newline=None`, and `.gitattributes` pins `samples/**/expected/** -text` so Git preserves whatever was written. A Windows regeneration commits CRLF and fails `test_sample_expected_goldens_are_committed_with_lf_newlines` — whose own docstring asserts that *"the regeneration recipes force `newline="\n"`"*. This recipe is now the counterexample to that sentence.

This matters more than an unreferenced doc would, because it is the designated recovery path: `test_repair_scaffold_matches_its_golden`'s failure message is *"Regenerate it with a real scan — see that sample's README."* It is also prose rather than runnable code for two of the three steps, so it isn't copy-pasteable even on POSIX.

Suggested fix: lift the sibling's `python - <<'PY'` block with the sample path swapped. I confirmed the committed goldens are byte-reproducible from it.

Nit in the same block: `from agents_shipgate.cli.scan.orchestrator import run_scan`, where the sibling README and ~20 test modules use the package-level `from agents_shipgate.cli.scan import run_scan`.

---

### 2. The one adopter-facing string #424 is about is committed here, and nothing reads it — the PR's own standard, one level down

`#424` reached an adopter through a *sentence*, not a YAML template — the issue names `ci/release_decision.py:1824` templating it into the guidance an adopter reads. This sample commits that sentence three times:

- `release_decision.evidence_coverage.evidence_gaps[0..1].next_action.expects`
- `agent_summary.first_recommended_action.why`

and zero times in `report.md` (grep: 0). `report.json` is compared only for `report_schema_version`, the decision, and the set of field *paths* — so none of those bytes are pinned. `next_action.accepted_values` carries the complete list as a fourth uncompared copy.

That would be a shrug elsewhere, except this fixture is the first place in the repo where the sentence and the value *can* disagree. `EffectRepair.instruction` interpolates `tags` and `added` separately, and the only test that pins `expects` for this route — `tests/test_declaration_monotonicity.py:330` — uses a row with **no pre-existing tags**, so `tags == added` there and the two spellings coincide by construction. `billing.cancel_invoice_email` is the first case where they differ (`[financial_write, destructive]` vs `destructive`).

Verified, not reasoned: rewriting the f-string to interpolate `added` instead of `tags` — the round-2 defect confined to the sentence, `declaration_template` left complete — leaves **all four new tests green**, along with `test_effect_coverage`, `test_reports`, `test_evidence_gap_ranking`, `test_adopter_vocabulary`, `test_declaration_monotonicity` and `test_fix_task_contract`. The adopter would read `risk_tags: [destructive]` on the release surface and `[financial_write, destructive]` in the block, and the fixture built to catch precisely that says nothing.

Closing it is a value-join between two artifacts the sample already ships, in `test_the_repair_golden_still_renders_a_challenged_row`:

```python
for gap in golden["release_decision"]["evidence_coverage"]["evidence_gaps"]:
    if gap["kind"] != "declaration_below_inferred_evidence":
        continue
    action = gap["next_action"]
    published = blocks[action["declaration_template"]["tool"]]["risk_tags"]
    # The sentence an adopter reads first must name the list the block pastes.
    assert f"risk_tags: [{', '.join(published)}]" in action["expects"]
    assert action["accepted_values"] == published
```

Both assertions hold against the committed golden as it stands (checked), and both fail under the perturbation above. It is the same "publish the whole list" property the PR already proves for the YAML half, asserted on the half an adopter reads first.

---

### Small notes

- `#424` is already closed, so nothing here auto-closes it. Worth ticking its last acceptance box so the checklist doesn't read as abandoned — and noting there that the box was deliberately *not* satisfied where the issue proposed (the `#425` sample), with the reason. The PR body has that argument; the issue doesn't.
- The cross-link is one-directional: this README points back at `google_adk_cold_start_agent`, but not the reverse. One line in the sibling would close the pair.

samples/declaration_repair_agent/README.md
```diff
@@ -0,0 +1,128 @@
+# Declaration repair agent
+
+The repair sibling of [`google_adk_cold_start_agent`](../google_adk_cold_start_agent/).
+That fixture is a cold start — nothing is declared, so every question it
```
pengfei-threemoonslab: [P2] Narrow the cold-start contrast to the challenged effect route

This is not true of the sibling fixture: its manifest declares `ops.export_case_bundle.authority.mode: none` (rendered as a `conflicting_authority_evidence` note) and `record_case_outcome` effect/authority (counted as answered), while its questionnaire also pre-fills `effect: financial_write`. What this sample uniquely adds is the `declaration_below_inferred_evidence` / `declare_risk_tags` path, not declarations or nonblank questions in general. Please correct this claim and the mirrored wording in `samples/README.md` and this sample's `shipgate.yaml`.

samples/declaration_repair_agent/README.md
```diff
@@ -0,0 +1,128 @@
...
+scan would reopen the row asking for it back — the #424 defect surviving inside
+the case #424's own repair creates. The committed golden must publish the
+complete value, and
+`test_the_published_repair_keeps_the_tag_the_reviewer_already_wrote` reads it
+out of the golden and says so.
+
+## Why everything else is already answered
+
+The manifest declares every control both actions owe *after* the repair as
+well as before — `destructive` adds approval, confirmation and rollback to
```
pengfei-threemoonslab: [P2] Do not attribute confirmation to the repair

`BUILTIN_EFFECT_OBLIGATIONS` already assigns `confirmation.required` to `external_communication`, which both actions declare before repair. Adding `destructive` makes approval and rollback newly required for `support.delete_case_message`, and only rollback newly required for `billing.cancel_invoice_email` (approval already comes from `financial_write`). Please also correct the later “most ... only after” explanation so the fixture rationale matches the executable control semantics.

samples/declaration_repair_agent/README.md
```diff
@@ -0,0 +1,128 @@
...
+run_scan(
+    config_path=sample / "shipgate.yaml",
+    output_dir=Path("expected"),          # resolved under the manifest dir
+    formats=["json", "markdown"],
+    ci_mode="advisory",
+    packet_enabled=False,
+)
+```
+
+Then replace the absolute repository path in `expected/report.json` with
```
pengfei-threemoonslab: [P2] Make golden regeneration platform-neutral

Following this recipe on Windows produces artifacts rejected by this PR’s own golden guards. The report, Markdown, and questionnaire writers use platform newlines, while `.gitattributes` preserves those CRLF bytes; textual path replacement also does not safely handle JSON-escaped backslashes or normalize `generated_reports` entries such as `expected\\report.json`. Please mirror the cold-start recipe: structurally set `manifest_dir`, normalize `generated_reports` with `.as_posix()`, and rewrite all three goldens with `newline="\n"`.

pengfei-threemoonslab (commented):
Reviewed the full PR at `63de60ab` across implementation behavior, fixture integrity, documentation, and release-contract evidence.

The fixture mechanics themselves are sound: the committed scaffold and Markdown regenerate byte-for-byte; normalized `report.json` matches a fresh scan; pasting the committed blocks reaches `passed` with zero open questions; both touched test modules pass in full (117 tests); Ruff, compileall, schema generation, and `git diff --check` are clean; and the required GitHub checks are green (five succeeded, one intentionally skipped).

I found three P2 documentation correctness issues below. Please address them before treating these sample instructions as canonical. Shipgate independently reports `review_publishable` / `human_review_required` because the new sample `shipgate.yaml` is a protected trust-root surface; its verification receipt reproduced successfully.

samples/declaration_repair_agent/README.md
pengfei-threemoonslab: Confirmed and corrected. Checked the sibling rather than reasoning about it: it declares two actions, pre-fills `effect: financial_write`, and renders one "No block is offered" note for the challenged authority row — so "nothing is declared / every question is a blank" was wrong on three counts.

The narrow claim it can support is that its questionnaire holds **zero** `risk_tags:` blocks, so a declaration challenged in the *effect* dimension appears in no committed artifact before this one. That is now the wording in all four places it appeared: this README (opening and § What it pins), `samples/README.md`, the sample `shipgate.yaml` header, and the framing comment plus failure message in `tests/test_declaration_questionnaire.py`.

Fixed in `ca831f0f`.

samples/declaration_repair_agent/README.md
pengfei-threemoonslab: Confirmed and corrected — I had it wrong in two ways, not one. Verified against `BUILTIN_EFFECT_OBLIGATIONS`:

| | owed before | newly owed by `destructive` |
| --- | --- | --- |
| `support.delete_case_message` | `confirmation.required`, `safeguards.audit_log` | `approval.required`, `safeguards.rollback` |
| `billing.cancel_invoice_email` | `approval.required`, `confirmation.required`, `safeguards.audit_log`, `safeguards.idempotency` | `safeguards.rollback` |

So `confirmation.required` is never introduced by the repair, approval is new on the first row only, and `safeguards.rollback` is the one control new to both. The "most ... only after" sentence was the same error generalized — three of the five are owed before — and is gone. The README now carries the table above, and the manifest comment states the per-action split.

Fixed in `ca831f0f`.

samples/declaration_repair_agent/README.md
pengfei-threemoonslab: Confirmed and corrected. Replaced with the sibling recipe, path swapped, package-level import included — structural `manifest_dir`, `generated_reports` through `.as_posix()`, and all three goldens rewritten with `newline="\n"`.

Verified rather than assumed: extracted the fenced block back out of the committed README and ran it verbatim against deliberately dirtied goldens, and it restores them byte-for-byte.

The point about this being the *designated* recovery path is what made it worth more than a doc nit — `test_repair_scaffold_matches_its_golden` names the README in its failure message, so the recipe was the counterexample to `test_sample_expected_goldens_are_committed_with_lf_newlines`'s own docstring. The README now says that in as many words.

Fixed in `ca831f0f`.

pengfei-threemoonslab:
Both follow-ups from the review body are in, plus the three P2s (replied on their threads). All addressed in `ca831f0f`, on top of a merge of `main` at `342e9722`.

## 2 — the sentence nothing read

This was the one worth folding in, and I reproduced it before touching anything rather than taking the walkthrough on trust. Rewriting the `EffectRepair.instruction` f-string to interpolate `added` instead of `tags` — the round-2 defect confined to the sentence, `declaration_template` left complete — and regenerating the golden leaves **all four** of the fixture's tests green, with the adopter reading `risk_tags: [destructive]` on the release surface and `[financial_write, destructive]` in the block.

Took the suggested join, in `test_the_repair_golden_still_renders_a_challenged_row`, with one addition: a count assertion that every challenged row contributed a join. Without it the loop is satisfied by a golden that stopped emitting one of the two surfaces — the agreement would be vacuous rather than false, which is the failure mode a value-join is most likely to rot into.

It fails under the perturbation naming both surfaces, and the message says *"An adopter reads the sentence first."*

## The canonicalization argument for test 3

You're right, and it is a better justification than the one I gave. Reproduced: canonicalize `financial_write` → `financial_action` in the declared list and regenerate, and the row still closes, the sample still reaches `passed`, the sentence still agrees with the block, and `test_the_published_repair_keeps_the_tag_the_reviewer_already_wrote` is the **sole** failure.

So my docstring was wrong where it mattered — it said the test "does not bite on a source regression on its own". It does, on the one regression that matters to it, and it is the only guard on *never rewrite what the reviewer wrote*. Rewritten to say that, with the perturbation named. Note the separating perturbation is **rewriting**, not deletion; deletion is caught by three other guards.

## 1 — the regeneration recipe

Replaced wholesale with the sibling's block, path swapped, `from agents_shipgate.cli.scan import run_scan` included. Verified by extracting the fenced block back out of the committed README and running it verbatim against deliberately dirtied goldens — it restores all three byte-for-byte. The README now names why each normalization is load-bearing and records that the earlier draft was the counterexample to `test_sample_expected_goldens_are_committed_with_lf_newlines`'s docstring.

## Small notes

- **#424's checklist.** Ticked — all six, not just the last: the first five were closed by #457 and none had been ticked, so the list read as abandoned end to end. Added a note attributing five to #457 and the sixth to this PR, and recording that the sixth was deliberately *not* satisfied where the issue proposed, with the reason (a cold start has no declared effect to challenge, and adding one moves that fixture off the property four files assert).
- **Cross-link.** Now bidirectional; the sibling's opening points here and says what it has no example of.

## What I did not change

The `passed` assertion in the paste test stays. It couples the fixture to the whole gate, and I take that as the point rather than a cost — "the repair reaches a verdict" is the claim, your check that the repaired state is *clean* (zero gaps, zero findings) is what rules out a traded-down gap, and the assertion message explains the coupling to whoever trips it.

Thank you for the perturbations — the sentence/value split and the canonicalization case are both things the fixture should have pinned from the start, and neither was reachable by reading the diff.

Issue #463, opened 2026-08-29
Cold-reader artifacts lead with the verdict, and the cold verdict is an abstention by construction — render the surface first when no manifest exists


pengfei-threemoonslab:
## Summary

Every human-facing artifact opens with the verdict. The console summary leads with `Decision:` (`cli/_helpers.py:460`), the GitHub step summary leads with ``Decision: `…` `` then `Summary:` (`ci/github_summary.py:31`), and the scan block quoted in #396 has the same shape: `Decision → Summary → Evidence coverage`.

For an adopted repository in CI, verdict-first is correct — the reader's question is "may this merge?".

For the **cold reader** — a maintainer seeing shipgate output on their repository for the first time, with no committed manifest — the verdict is `insufficient_evidence` almost by construction: there is no trust root yet, so abstention is the *honest and expected* answer (#393 measured it as the default verdict for an entire framework class; adoption walk 1 in #410 ran seven commands without ever reaching a verdict). So the first line of first contact is an abstention, and the content that would earn a second command — the enumerated tool surface, the destructive list, the capability delta, the findings — renders below the fold of a headline that says "I don't know".

The cold reader's question is not "may this merge?". It is **"what can this thing do, and what did this PR change?"** — a question the artifact already answers, in the wrong order.

## Constraint stated up front: this is a reordering, not a second verdict

The #323 rollout rule stands: no renderer computes a verdict. The decision object is rendered **once, verbatim, unchanged** — this issue moves *where*, never *what*. Machine JSON (`report.json`, envelopes, contracts) is untouched.

## Proposal

When the run is a cold one — derivable from facts the engine already holds (no committed manifest / the first-adoption carve-out of PR #297, or the #326 local-review mode; **not** a new flag) — the human artifacts (console summary, `report.md` lead, PR/step summary) lead in this order:

1. **The surface.** N tools from M sources; effect breakdown; the destructive/write names spelled out.
2. **The delta**, when a base is given — subject-grouped (#437/#439's fix, not the change-count rendering).
3. **The findings**, subject-grouped (#364).
4. **The gate status, last:** `insufficient_evidence` rendered as what it is for this reader — *"the gate is not configured for this repository; answering N questions would establish it"* — with the #410 questionnaire as the pointer.

Adopted repositories (committed manifest) keep verdict-first ordering unchanged.

## Why ordering is the defect and not a cosmetic preference

[...]

Issue #326, opened 2026-08-06
feat(init/verify): add a side-effect-free local-review mode for external PRs


pengfei-threemoonslab:
## Problem

A common evaluation workflow is to run Agents Shipgate against an external PR before that repository has adopted Shipgate. Today `init --write` creates `shipgate.yaml` and also creates or edits the repository's tracked `.gitignore` so reports are ignored.

That behavior is appropriate for durable adoption, but it is intrusive for a temporary local review: the act of configuring the scanner changes the target worktree and can become mixed into the PR being assessed.

## Observed workflow

While evaluating [google/adk-samples#1745](https://github.com/google/adk-samples/pull/1745), preview routed to:

```text
shipgate init --workspace /path/to/checkout --write --json
```

Init then reported:

```json
{
  "path": "/path/to/checkout/shipgate.yaml",
  "created": true,
  "gitignore": {
    "status": "created",
    "path": "/path/to/checkout/.gitignore"
  }
}
```

The side effects were transparent, which is good, but there was no first-class non-adoption mode.

## Proposal

Add an explicit local-review mode, for example `init --local-review`, that:

- creates a clearly local/ephemeral manifest or accepts an external manifest bound to the workspace;
- keeps generated reports out of Git using `.git/info/exclude` or an external output directory;
- does not modify tracked project files by default;
- records manifest provenance in verifier artifacts;
- cannot be mistaken for a committed repository trust root or durable release policy.

The existing `init --write` adoption behavior should remain unchanged.

## Security boundary

A local-review manifest is useful for static assessment but is not equivalent to a repository-reviewed policy. Verification receipts and handoff output should mark it as ephemeral/uncommitted and route trust-root-sensitive conclusions accordingly. This feature must not make local assertions look like reviewed release evidence.

## Acceptance criteria

- [ ] Local-review setup creates no tracked-file changes in a clean checkout.
- [ ] Generated reports do not appear as untracked files.
- [ ] JSON output lists every local side effect and its recovery/removal path.
- [ ] Verifier artifacts identify the manifest as local/ephemeral.
- [ ] Durable `init --write` retains its current managed `.gitignore` behavior.
- [ ] Documentation distinguishes temporary assessment from repository adoption.
- [ ] A local/ephemeral manifest cannot produce release-authoritative `passed`, mergeable, or human-authorization evidence merely because it is structurally valid.
- [ ] `verify` clearly identifies which conclusions are provisional local-review output and which require a committed trust root.

[...]

PR #466, opened 2026-08-30
Render cold-reader artifacts surface first

pengfei-threemoonslab:
Addresses #463

Summary:
- derive a presentation-only cold context from committed-manifest and first-adoption facts without changing report or packet schemas
- lead console, report, GitHub summary, PR comment, and packet Markdown/HTML/PDF with surface and subject-grouped findings/delta, then the unchanged release decision, while retaining the existing tool-surface and diff detail below it
- preserve verdict-first bytes for adopted repositories and force verdict-first whenever cold output carries block-tier content
- share one bounded findings projection, cap and display-encode repository-derived action names, and reserve the PR-comment verdict/control block even for very large or hostile tool surfaces
- reuse the scan-derived repository context in console and verifier output rather than re-probing Git
- add a committed cold-mode golden plus empty-repo, archived-verify, exact-once cold decision, destructive-action, machine-output, truncation, and adversarial regressions

Scope:
- this change covers report-backed scan and verify artifacts
- `verify --preview` remains routing-only and emits no readiness report to reorder
- the local-review surface described in #326 remains a follow-up

Validation:
- `python -m ruff check .`
- post-rebase focused regression suite: 296 passed
- full non-performance suite: all non-packaging tests passed; the isolated packaging build was rerun with dependency access
- `pytest tests/test_packaging.py -q` (8 passed)
- static-only adapter suite (442 passed) and P0 safety/binding/agent-boundary suites
- post-rebase Agents Shipgate verify: `control_state=complete`, `decision=passed`, `merge_verdict=mergeable`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,15 +10,31 @@
   multi-manifest console rows, `report.md`, GitHub step summaries, PR comments,
   and packet Markdown/HTML/PDF share that ordering. A cold run carrying any
   active block-tier content remains verdict-first, and committed-manifest
-  repositories retain their prior bytes and verdict-first order. Machine JSON,
-  schemas, contracts, and decision computation are unchanged.
+  repositories retain their prior bytes and verdict-first order. Cold reports
+  render the primary decision exactly once; adopted reports preserve the
+  pre-existing capability-consequence projection as part of their byte-stable
+  layout. Machine JSON, schemas, contracts, and decision computation are
+  unchanged.
+
+  The ordering is lossless: tool-surface and action/tool-diff detail still
+  appears below the decision. Compact packet finding lists share one bounded
+  truncation projection, write/destructive action names are capped with an
+  explicit omission count, and repository-owned names are display-encoded so
+  they cannot forge a decision line. PR comments consume the scan's exact
+  ephemeral repository context, including manifest introduction, rather than
+  re-probing or depending on report-evidence fallback.
 
   `samples/google_adk_cold_start_agent/expected/cold-report.md` is the committed
   first-contact golden: its test creates a Git repository whose agent sources
   are committed while `shipgate.yaml` is not, then byte-compares the rendered
   artifact. The same suite pins an empty repository, first-adoption verifier
-  output, packet ordering, exact-once decision rendering, the block-tier
+  output, packet ordering, cold exact-once decision rendering, the block-tier
   override, and adopted-sample byte stability.
+
+  This release covers report-backed scan and verify artifacts. `verify
+  --preview` remains a routing-only operation and produces no readiness report
+  to reorder; the local-review surface described in #326 is not implemented by
+  this change.
 
 - **The #424 repair loop is now pinned by a committed artifact.** (#424)
   `declaration_below_inferred_evidence` publishes a `declare_risk_tags` route,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,24 @@
 
 ## Unreleased
 
+- **Cold-reader artifacts now lead with what the agent can do.** (#463) On a
+  repository with no committed Shipgate manifest, human output starts with the
+  root-reachable tool surface, its effect breakdown and write/destructive
+  action names; then shows a subject-grouped capability delta when one exists,
+  subject-grouped findings, and the unchanged release decision. Console,
+  multi-manifest console rows, `report.md`, GitHub step summaries, PR comments,
+  and packet Markdown/HTML/PDF share that ordering. A cold run carrying any
+  active block-tier content remains verdict-first, and committed-manifest
+  repositories retain their prior bytes and verdict-first order. Machine JSON,
+  schemas, contracts, and decision computation are unchanged.
+
+  `samples/google_adk_cold_start_agent/expected/cold-report.md` is the committed
+  first-contact golden: its test creates a Git repository whose agent sources
+  are committed while `shipgate.yaml` is not, then byte-compares the rendered
+  artifact. The same suite pins an empty repository, first-adoption verifier
+  output, packet ordering, exact-once decision rendering, the block-tier
+  override, and adopted-sample byte stability.
+
 - **The #424 repair loop is now pinned by a committed artifact.** (#424)
   `declaration_below_inferred_evidence` publishes a `declare_risk_tags` route,
   and every guard on it ran against tools built in-test — the same blind spot
--- a/samples/README.md
+++ b/samples/README.md
@@ -31,7 +31,7 @@ running a scan first:
 | [`simple_openai_api_agent`](simple_openai_api_agent/) | [`report.md`](simple_openai_api_agent/expected/report.md) | [`report.json`](simple_openai_api_agent/expected/report.json) |
 | [`simple_langchain_agent`](simple_langchain_agent/) | [`report.md`](simple_langchain_agent/expected/report.md) | [`report.json`](simple_langchain_agent/expected/report.json) |
 | [`conductor_agent`](conductor_agent/) | [`report.md`](conductor_agent/expected/report.md) | [`report.json`](conductor_agent/expected/report.json) |
-| [`google_adk_cold_start_agent`](google_adk_cold_start_agent/) | [`report.md`](google_adk_cold_start_agent/expected/report.md) | [`report.json`](google_adk_cold_start_agent/expected/report.json) |
+| [`google_adk_cold_start_agent`](google_adk_cold_start_agent/) | [`report.md`](google_adk_cold_start_agent/expected/report.md) ([cold first contact](google_adk_cold_start_agent/expected/cold-report.md)) | [`report.json`](google_adk_cold_start_agent/expected/report.json) |
 | [`declaration_repair_agent`](declaration_repair_agent/) | [`report.md`](declaration_repair_agent/expected/report.md) | [`report.json`](declaration_repair_agent/expected/report.json) |
 
 Two samples stop with **open declaration questions**, so they are the two that
@@ -43,7 +43,8 @@ edit — as a byte-compared golden. They render the two halves of that file:
   challenged authority row printed as a note. The fixture exists to pin the
   order the questions are asked in. Read the diff before regenerating it: a
   change in block *order* is a change in which question this tool asks a human
-  first.
+  first. Its own [README](google_adk_cold_start_agent/README.md) documents the
+  [cold-report regeneration procedure](google_adk_cold_start_agent/README.md#regenerating-the-cold-report).
 - [`declaration_repair_agent`](declaration_repair_agent/expected/suggested-declarations.yaml)
   is the step after, and adds the one shape the cold start has no example of: a
   declaration challenged in the **effect** dimension, whose repair is a
--- a/samples/google_adk_cold_start_agent/README.md
+++ b/samples/google_adk_cold_start_agent/README.md
@@ -6,7 +6,12 @@ nothing open — so a scan of it writes no `suggested-declarations.yaml` at all
 and no committed artifact beside it ever renders one. This fixture stops
 partway on purpose: it reaches `insufficient_evidence` with **ten open
 declaration questions**, and ships the questionnaire an adopter actually reads
-as a byte-compared golden.
+as a byte-compared golden. It also ships
+[`expected/cold-report.md`](expected/cold-report.md), rendered from a temporary
+Git repository where the agent sources are committed and `shipgate.yaml` is
+not, to pin the surface-first first-contact order from #463. The ordinary
+[`expected/report.md`](expected/report.md) remains the committed-manifest,
+verdict-first golden.
 
 For the step after this one, see
 [`declaration_repair_agent`](../declaration_repair_agent/): a manifest that has
--- a/samples/google_adk_cold_start_agent/README.md
+++ b/samples/google_adk_cold_start_agent/README.md
@@ -156,6 +156,57 @@ for name in ("report.md", "suggested-declarations.yaml"):
 PY
 ```
 
+### Regenerating the cold report
+
+`expected/cold-report.md` comes from a different repository state than the
+ordinary goldens: the agent sources are committed, while `shipgate.yaml`
+exists only in the worktree. Recreate that state in a temporary repository;
+running the ordinary recipe above cannot produce the cold-reader order.
+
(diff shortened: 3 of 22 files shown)


pengfei-threemoonslab (commented):
## Review — cold-reader ordering (#463)

Reviewed at `e0c29be` against `origin/main` (base is one commit behind `342e972`; no conflict). Ran `ruff check` (clean), the full non-packaging suite (green, exit 0), and reproduced the cold artifacts by hand from `samples/google_adk_cold_start_agent`. The suite being green is part of the report below: none of the defects here are caught by an existing test.

The shape of the change is right, and the parts that are hard to get right are gotten right: machine surfaces are untouched and *asserted* untouched; the fail-closed guard covers both `decision.blockers` and unsuppressed `blocks_release` findings, including the deliberately inconsistent packet-vs-report case; the adopted goldens are pinned byte-identical in both directions; and `_commit_present_at_ref` correctly separates "unborn HEAD" (`False` → cold) from "Git failed" (`None` → verdict-first) while reusing verify's audited, no-shell Git boundary instead of opening a second process-execution surface in scan code. That last decision is the right one and the docstring earns it.

But the reordering itself has two defects that fire on the *first-contact artifact this PR exists to produce*, and one scope question. Blocking on 1 and 2.

---

### 1 · P0 — every cold packet renders negative truncation counts

`packet/markdown.py:235-240` and `packet/html.py:219-226` compute the elision counts as `len(...) - 5` / `len(...) - 8` and gate them with `if hidden:` / `if hidden_subjects:`. A negative int is truthy, so the "… and N more" line renders whenever there are *fewer* items than the cap — i.e. essentially always.

Reproduced end to end with the committed sample (sources committed, `shipgate.yaml` not), `packet.md`:

```
## Findings by subject

- `adk-ops-agent \(agent-wide\) \(at shipgate.yaml\)` — review \(1 medium\)
  - medium SHIP-ADK-EVAL-COVERAGE-MISSING — Google ADK eval coverage is not declared
  - … and -4 more findings
- … and -7 more subjects
```

`packet.html` is the same, and `packet.pdf` inherits it. The HTML empty branch is worse: it does not return after `<li>none</li>`, so a cold packet with zero findings renders `<li>none</li><li>… and -8 more subjects</li>`.

`if hidden > 0:` fixes it, but the better fix is to not re-implement it: `top_findings_block` already does truncate-with-count, plural agreement included, and its module docstring in `core/findings/subject_rollup.py` is explicit that one function exists so surfaces "cannot disagree about how many of them a reader gets to see". `_append_cold_findings` / `_render_cold_findings` are a fourth copy of that logic with different limits (8/5 vs the console's `_CLI_SUBJECT_LIMIT`/`_CLI_ROW_LIMIT`) and a different plural helper.

The reason this shipped is worth fixing too: `test_unadopted_human_artifacts_lead_with_the_surface` asserts only `.index(...)` orderings on the packet — never a byte of its body. Either byte-compare a cold `packet.md` the way `cold-report.md` is compared, or at minimum assert `"more finding" not in packet_markdown` for the one-finding fixture.

### 2 · P0 — the cold PR comment can truncate the verdict out of the artifact entirely

`SurfaceLead.text_lines()` (`report/human_order.py:48-64`) enumerates **every** `write` / `financial_write` / `destructive` tool name on one unbounded line. `_cold_reader_lines` puts that line above the verdict in both comment styles, and both are hard-capped at 6000 chars (`_truncate` for `findings`, `_join_with_preserved_agent_block` for `capability-review`).

With 300 write actions on the sample report, both styles produce 6000 characters of tool names and nothing else:

```
comment chars: 6000
'- Merge verdict:' present: False
'Release gate' present: False
'- Agent control state:' present: False
'- Static-verdict boundary:' present: False
```

The verdict, the control state, `must_stop`, the permissions, the findings and the artifact links are all gone. That is the acceptance criterion "the decision appears exactly once per artifact" failing to *zero*, and it is a regression introduced by the reorder: before this change the verdict was above the truncation point, so truncation only cost detail.

This is not a synthetic worry — the issue's own motivating case is external MCP-server PRs, and `github-mcp-server` (#403) is ~50–100 tools. Sixty write tools at ~35 chars each, before markdown escaping doubles the punctuation, is already more than half the budget.

Suggested fix: bound the enumerated names the way every other truncating surface here does (show N, then "… and K more"), and add a test that renders a cold comment over a large write surface and asserts the verdict line survives. A cap also improves the console and step-summary output, where the line is currently a single unwrapped wall of text.

### 3 · P1 — cold mode *removes* sections rather than reordering them

`report/markdown.py:175-181` skips `_append_tool_surface`, `_append_action_surface_diff` and `_append_tool_surface_diff` under `surface_first`; `cli/_helpers.py` and `ci/github_summary.py` do the same for their diff lines. The committed golden shows the effect — `expected/cold-report.md` has no `## Tool Surface Summary`, no `## Action Surface Diff`, no `## Tool Surface Diff`, while `expected/report.md` has all three.

So the cold reader loses, relative to the adopted reader: high-risk tool count, wildcard tools, missing descriptions, the per-source breakdown, the diff's `new_high_risk_effects` / `policy_drift_items` / `new_findings` / `resolved_findings` counts, and the `disabled — <reason>` note that says *why* there is no delta.

The replacement is not equivalent: `capability_delta_by_subject` returns `[]` whenever `report.capability_change.enabled` is `False`, which is exactly the no-base case. Verified on a cold `verify --base HEAD~1`: `capability_change.enabled=False`, `tool_surface_diff.enabled=False` — and the rendered `report.md` carries no delta section and no explanation of its absence. A reader who asked for a base comparison is told nothing about it.

The issue is explicit that this "moves *where*, never *what*". Move those three sections below the decision rather than dropping them.

### 4 · P1 — `include_decision=not surface_first` makes "exactly once" a mode-dependent rule

`_append_capability_intent_diff(..., include_decision=not surface_first)` suppresses the `- Decision: …` line inside `## Capability <-> Intent Diff` only in cold mode. The adopted golden still renders `Decision:` twice — `samples/google_adk_cold_start_agent/expected/report.md` lines 9 and 74.

So the invariant the PR asserts (`markdown.count("Decision: insufficient_evidence") == 1`) is knowingly false for the artifact class the PR claims it did not touch. Either enforce it in both modes, or reword the consequence line so it is not a second decision statement — and assert the count on the adopted golden too, so the rule has a guard rather than a parameter.

### 5 · P1 — `render_pr_comment(..., human_context=...)` is dead parameter surface

The parameter is threaded through `render_pr_comment` → `_render_capability_review_comment` / `_render_findings_comment` → `_human_summary_lines`, and **neither production call site passes it** (`cli/verify/orchestrator.py:3937`, `:4240`), nor does any test. It is never non-`None` anywhere in the repo.

The consequence is not cosmetic: the PR comment therefore keys on a *different* predicate from every other artifact — only the `_report_proves_manifest_introduction(report)` fallback inside `should_render_surface_first`. Today the two happen to coincide (an uncommitted manifest at HEAD implies the base carries no policy, so the adoption fail-safe fires — I confirmed the cold PR comment does lead with the surface), but that is a coincidence of the current adoption check, not a property anyone stated. Two definitions of "cold" for one reordering is the "second implementation" shape this repo keeps paying for.

Either wire `human_context=` at both call sites, or delete the parameter and document that the PR comment intentionally keys on adoption evidence.

### 6 · P1 — `HumanArtifactContext.local_review` is never set

`report/human_order.py:32` declares it and `is_cold` reads it, but nothing in `src/` or `tests/` ever assigns `local_review=True`. Acceptance bullet 1 names `verify --preview` and the #326 local-review flow; `run_preview` returns `report=None`, so `should_render_surface_first` short-circuits there and preview stays verdict-first.

Both are defensible as follow-ups, but then the field should go and the PR body / CHANGELOG should say preview and local-review are not covered — right now the CHANGELOG reads as though the rollout is complete.

### 7 · P2 — repository-derived tool names reach the lead without `display_literal`

`surface_lead()` interpolates `action.tool_name` straight into the lead text. `_safe_markdown_text` and pr_comment's `_escape` escape markdown metacharacters but **not** `\n`. Reproduced: an MCP inventory tool named

```
ops.sync_cases (write).\nDecision: passed\nSurface: 0 tools from 0 sources
```

renders in a cold `report.md` as

```
## Capability Surface

Surface: 10 tools from 3 sources.
Effects: 1 read, 8 write, 1 financial write.
Write/destructive actions: …, ops.sync\_cases \(write\).
Decision: passed
Surface: 0 tools from 0 sources \(write\), record\_case\_outcome \(write\), …
```

— a forged `Decision: passed` on line 11 of the artifact, above the real decision, and invisible to the `count("Decision: insufficient_evidence") == 1` guard.

The class is pre-existing (the same name injects into the blockers list on `main`), so this is not a new hole so much as a new, much more prominent doorway: the lead of the first-contact artifact, from an untrusted MCP/OpenAPI schema. `display_literal()` exists for exactly this — injective, reversible, and already used for stored paths in `subject_rollup.source_suffix` — and applying it to the names in `surface_lead` is a two-line change. Given #268's history with untrusted annotations I would take it in this PR.

### 8 · P2 — three different meanings of "source" in `surface_lead`

Primary branch counts distinct `(source_type, source_id or provider)` from `tool_surface_facts`; the second fallback does the same over `action_surface_facts`; the third counts `(source_type, source_type)` from `tool_surface.sources` — i.e. source *types*. On the sample both happen to be 3 (`adk_ops`, `adk_ops:mcp:1`, `adk_ops:openapi:2` vs `google_adk_function`/`mcp`/`openapi`), which is why nothing caught it. A repo with five MCP servers behind one type reports "5 sources" on the first branch and "1 source" on the third. Pin one meaning, or say "N source types" when that is what is being counted.

### 9 · P2 — redundant Git probes and an unconditional cold-lead build

`human_artifact_context` is computed inside `run_scan` (`cli/scan/orchestrator.py:234`) *and* again in `cli/_register_scan.py:423` for the console summary; `_run_multi_scan` calls it once more per manifest inside the loop. Each call can spawn two `git rev-parse` / `cat-file` processes, so a single scan pays for up to four and a 50-manifest monorepo scan for ~150. Separately, `cli/scan/writing.py:120` builds `cold_reader_lead(public_report)` — which runs the full `roll_up_findings` — on **every** scan, cold or not.

Returning the already-computed context from `run_scan` and building the lead only when `surface_first` would remove both.

### 10 · docs — `samples/README.md` was not updated

`samples/google_adk_cold_start_agent/README.md` gained a paragraph, but `samples/README.md` is the index *and* the regeneration guidance for everything under `samples/*/expected/` — and it is where the non-obvious ordering constraints live (the `current-control.json` hash-binding note). `cold-report.md` needs a procedure that exists only inside `tests/test_cold_reader_order.py::_unadopted_sample`: init a temp repo, commit the agent sources but not `shipgate.yaml`, scan, compare. After the next `report_schema_version` bump, whoever regenerates the goldens has a documented recipe for six of them and none for this one. Please add it beside the existing note.

---

### Summary

- **1** and **2** are blocking: both produce visibly broken output in the artifact the PR exists to improve, and **2** loses the verdict entirely on a realistically sized MCP surface.
- **3** and **4** are the scope question — the issue says reorder, and the diff also removes. Worth settling before the golden is pinned, because the golden makes the removal the contract.
- **5**–**10** are cleanups; **7** I would take now rather than later.

Nothing here touches the decision engine or a machine surface, and I could not find a path where the reorder softens a block: the guard on `decision.blockers` + unsuppressed `blocks_release` findings holds in every probe I tried, including the packet's independent copy.

pengfei-threemoonslab:
Addressed in `56f1b03b` (rebased onto current `main` at `44dc024c`). Thank you for the thorough probes; they exposed two first-contact failures that ordinary ordering assertions did not.

1. **Negative packet truncation counts:** packet Markdown/HTML now consume the shared subject-rollup truncation projection used by the other human surfaces. Empty, below-cap, at-cap, and omitted-count cases are covered; negative counts can no longer render.
2. **PR verdict lost to truncation:** write/destructive names are capped with an explicit omission count, and the complete cold-reader lead has its own 2,000-character budget inside the 6,000-character comment budget. Both comment styles retain verdict, control state, boundary text, and links under a 300-action surface and a single hostile multi-kilobyte identity.
3. **Sections removed rather than reordered:** restored Tool Surface Summary, Action Surface Diff, and Tool Surface Diff below the release decision in report Markdown, console, GitHub summary, and both PR-comment styles. Findings already promoted into the lead are not duplicated.
4. **Exact-once ambiguity:** the invariant is now explicitly cold-mode-only. Cold reports render the primary decision exactly once; adopted reports preserve the existing capability-consequence line and byte identity. Tests pin both contracts, and the CHANGELOG no longer overstates the adopted behavior.
5. **Dead PR-comment context:** `run_scan` now exposes the exact ephemeral context through a callback while preserving its public return shape. Single scan, multi-scan, and both verifier PR-comment writes reuse that context instead of relying on a separate fallback definition of cold.
6. **Unused local-review flag:** removed `local_review`. The CHANGELOG and PR now state that `verify --preview` emits no readiness report and that #326's local-review surface is deferred; the PR uses `Addresses #463` rather than claiming to close the uncovered follow-up.
7. **Repository-derived name injection:** tool names and capability-delta identities pass through `display_literal`, so newlines and invisible characters remain visible data rather than forged verdict/surface lines. Sorting and deduplication still use raw identities.
8. **Source-count semantics:** identity-backed branches report `sources`; the type-only fallback now reports `source type(s)` so the unit matches what was counted.
9. **Redundant Git probes / unconditional rollup:** scan and verify propagate the already-derived context; the cold lead is built only for cold human packet formats, not adopted or JSON-only output.
10. **Golden regeneration docs:** the sample index links the cold first-contact golden, and the sample README now gives the reproducible temporary-Git-repository recipe (agent sources committed, manifest uncommitted, expected output excluded, LF bytes).

The rebase conflict was additive documentation only. I retained `main`'s new `declaration_repair_agent` row, two-sample questionnaire guidance, and #424 changelog entry, then composed the cold-report links and scope note around them.

Post-rebase evidence:
- 296 focused cold-reader, rollup, packet, report, first-adoption, and declaration-questionnaire tests passed
- Ruff, docs-link, and documentation checks passed
- the full non-performance suite passed apart from the sandbox-blocked packaging download; `tests/test_packaging.py` then passed 8/8 with dependency access
- static-only adapter suite passed 442 tests; P0 safety/binding/agent-boundary suites passed
- `git diff --check` is clean
- Agents Shipgate: `control_state=complete`, `decision=passed`, `merge_verdict=mergeable`

Issue #456, opened 2026-08-29
Corpus delivery: current 38-case pre-1.0 and 80-case beta qualification tracks


pengfei-threemoonslab:
## Current delivery contract — 2026-09-08

This remains the cross-release **data/evidence parent**, not an additional release gate. #572 owns the v1.0 go/no-go and sequencing. The policy approved in #519 is implemented; this issue owns delivery against it, not a new policy.

| Track | Live requirement | Delivery |
|---|---|---|
| Optional 0.x | 38 cases: 14 passed / 14 review-required / 10 blocked, 21 strata, at least 16 qualifying origins | Verify any #508 frozen-data handoff; #509 produces/validates exact receipts and promotion evidence |
| 1.0, including prerelease tags | 80-case beta: 30 passed / 20 review-required / 30 blocked, 21 weighted strata, at least 32 qualifying origins | #512 human primary labels and final-wheel evidence → #509 independent signing/promotion → #510 exact-candidate rehearsal |

Both keep κ≥0.80, per-stratum holdout≥20%, zero unsafe auto-passes and the approved exact-outcome floors. Actual insufficient-evidence remains a coverage miss; it is not a ground-truth class. Current qualification results are v6, corpus/index v4; #569 coordinates the deliberate report 1.0 compatibility change before final v1.0 receipts.

**Beta sourcing starts now, in parallel with engineering.** Neither a v0.16 publication nor completed pre-1.0 qualification is a prerequisite for 1.0. Do not promote agent-produced pre-1.0 labels into beta human labels, or count the old miner labels as an adjudicated corpus.

## Remaining work

- [ ] #512 delivers the separately human-labeled beta corpus, accountable freeze and exact final-wheel receipts/index/unsigned qualifying result.
- [ ] #509 validates the applicable handoff, establishes the independent signer and promotes the exact signed bytes/locations.
- [ ] #510 records meaningful negative rejection and a positive final-candidate rehearsal without publication.
- [ ] The approved active pre-1.0 retirement is coordinated with #569 **before** final source/wheel freeze; historical reader compatibility is explicit.
- [ ] #511 remains separately tracked, non-gating historical participant validation. It cannot be used to hold a 1.0 tag that otherwise meets its requirements.
- [ ] #566 reconciles current operator instructions and delivery children, preserving visibly historical counts.

[...]

Issue #312, opened 2026-08-05
benchmark: re-run the labeled corpus on the v0.18 candidate and compare with W27


pengfei-threemoonslab:
> v0.18 release-exit measurement. Run after the behavior-changing adoption work so the comparison represents the candidate, not an intermediate state. The 1.0 must-block and needs-human bars make this P1 release evidence, not optional analysis.

## Summary

Re-run the labeled corpus and report the movement. This is the last remaining measurement obligation from #292, split out because it is a measurement task with a pass/fail bar, not a design question.

## Why now

The most recent re-evaluation on disk is `benchmark/miner/results/2026-W27-reeval.csv` (2026-07-08, run against v0.15.0). Both #296 and #297 landed after it — 2026-07-28 and 2026-07-29 — and both changed cold-start behavior. No numbers have been produced since.

## What to measure

Re-run the 19 labeled decided PRs at fixed SHAs against current `main`, and report against the W27 baseline:

- `must_block_caught` and `needs_human_caught` must hold at **1.0**. These are hard bars; a regression here blocks the release regardless of what else improves.
- `benign_escalation_rate` — W27 measured 0.286 and attributed most of it to a cold-start whole-repo-surface artifact. That attribution now needs re-testing: the #292 work turned out not to be about that artifact, so the explanation may not survive.
- `blocked_recall` — 0.0 on real history at W27 (the gate abstained rather than blocked on both real `must_block` cases). Report whether it moved.

## Usable outcomes and coverage

The W27 files have 19 case rows, 17 verifier results and two baseline failures. Report all attempted cases and their denominators; an abstention can satisfy a non-passing safety catch while still failing to give the user a useful review.

Alongside the existing four metrics:

- Report counts and denominators for terminal allowed/complete outcomes, named human-review outcomes, blocked outcomes, `insufficient_evidence`/coverage misses, input/base failures and product errors. Keep execution success separate from operational completion.
- For benign cases, report how many actually reach the allowed terminal outcome, how many are unnecessarily escalated, and how many receive no answer. Do not discard IE/error cases to improve a rate.
- For human-review cases, report whether the output names the changed capability and an actionable request; a raw escalation is not proof that a reviewer can finish the workflow.
- Preserve W27's original labels and fixed SHAs for comparison. If #520 introduces a different scoring/ground-truth model, publish an explicit separate mapping/view; do not silently relabel the baseline or change the approved 1.0 safety catch bars.
[...]

PR #467, opened 2026-08-30
Fix reviewed risk override evidence handling

### LINKED ISSUE #460: effect_readings counts a risk_overrides tag as something the scan observed

pengfei-threemoonslab:
## Summary

- exclude manifest-owned `risk_overrides.tags` claims from `effect_readings` and source-only declaration pins
- keep reviewed risk tags as separately audited proposal constraints without letting them seed proposals or authorize themselves
- preserve exact action-row tag spelling/order, route material reviewed constraints to human ownership, and retain machine authoring for genuinely evidence-only proposals
- add real scan, patch-application, pin-stability, alias/unmapped-tag, explicit-empty, and exhaustive proposal regressions

## Related issue audit

- #456 requires real receipt-bound cases, independent human labels/adjudication, a signed qualification artifact, protected repository variables, and a promotion rehearsal. Those external evidence prerequisites are absent, so this PR does not fabricate them or claim the issue complete.
- #312 is explicitly scoped to the v0.18 candidate and the repository is currently `0.16.0b7`. The fixed W27 corpus was integrity-checked, but publishing a v0.18 rerun now would be premature, so this PR does not claim the issue complete.

## Validation

- full non-performance suite excluding separately-run static and packaging groups
- 8/8 isolated packaging tests
- static adapter, P0 safety/binding, and agent-boundary suites
- Ruff, compileall, generated-schema drift, and `git diff --check`
- two independent review loops with no remaining findings
- committed `origin/main..HEAD` Shipgate verification: `control.state=complete`, `decision=passed`, `merge_verdict=mergeable`

Fixes #460


pengfei-threemoonslab (commented):
I reviewed commit `7bae490a7f1a155cab642e8f8b819bb43b072f8d` against `origin/main` and found two reproducible P2 issues.

### [P2] Preserve reviewed-field replacement provenance in the questionnaire

At `src/agents_shipgate/ci/release_decision.py:1832`, the proposal audit marker is emitted only when `reviewed_constraints` is non-empty. For an existing `risk_tags: []` (and likewise for an existing tag list that does not change the evidence-only proposal), `proposal_replaces_declared_risk_tags` is correctly true and the action is correctly human-owned, but `constraints` remains empty. Consequently, `src/agents_shipgate/cli/scan/declarations.py:_proposal_audit()` finds no marker and `suggested-declarations.yaml` describes the new non-empty tag list as merely “proposed from the evidence above”; it never tells the reviewer that accepting it replaces an explicit reviewed field.

That makes the human-facing trust surface disagree with the structured `expects`/authorship decision this PR just tightened. Please emit replacement/audit provenance whenever `proposal_replaces_declared_risk_tags` is true, including the original `risk_tags: []`, and add a scaffold assertion for the explicit-empty case.

### [P2] Bind explicit risk-tag ownership into the scan run identity

At `src/agents_shipgate/core/domain.py:249`, `risk_tags_declared` is excluded from serialization even though it now changes a published operational decision. With identical tool evidence, omitting the key yields `next_action.authorable_by="coding_agent"`, while an explicit `risk_tags: []` yields `"human"`; however, both serialize to identical action facts and both produce `run_id=agents_shipgate_a86663f81fa5c2ea` in a minimal reproduction.

That conflicts with the documented contract in `docs/agent-recipes.md:400-402` that an unchanged `run_id` means nothing semantic about the manifest/tool surface changed, and it lets plain scan/report/packet consumers deduplicate opposite write-authority routes. The verifier request/current-control identity is still safe because it binds the exact config and Git/worktree input, so this is P2 rather than P1. Please keep the bit off the public wire if desired, but include its key-presence in the scan run-identity payload and extend the existing absent-vs-explicit-empty regression to assert distinct run IDs.

Validation performed:

- Focused declaration-authoring, drift, questionnaire, schema, privacy, and action-surface tests passed.
- Ruff, generated-schema verification, `compileall`, and `git diff --check` passed.
- The full suite’s non-packaging tests passed; the eight packaging cases shared a sandbox-only build-backend download failure, then all 8 passed when rerun with declared build-dependency access.
- All current GitHub checks pass.
- Agents Shipgate verification reports `control.state=complete`, `decision=passed`, and `merge_verdict=mergeable` for the static gate. These two review findings are outside its current checks.

I am posting a comment review because the authenticated GitHub account owns this PR and therefore cannot approve it or submit “request changes.” I recommend addressing both findings before merge.

pengfei-threemoonslab (commented):
Follow-up review of `f582b855cb6645700ea5bac091a7a79e4d77673b`: no further findings.

The two P2 findings from my prior review are resolved:

- Human-facing proposal provenance now includes the existing reviewed value and an explicit replacement sentence whenever a proposal changes `risk_tags`, including the `risk_tags: []` case that previously lost the audit marker. Unconstrained proposals retain their evidence-only wording.
- Scan identity now binds explicit reviewed `risk_tags` key-presence without widening the public action-facts schema. The regression proves the absent and explicit-empty cases still serialize to identical public action facts while producing distinct `run_id` values and opposite authorship routes.

I also re-reviewed the combined PR diff for replacement semantics, selector determinism, backward run-ID stability, and human/coding-agent authorship boundaries. I found no additional actionable issue.

Validation:

- Full local test suite passed, including all 8 isolated packaging tests.
- Focused declaration and run-identity suites passed.
- Ruff, generated-schema verification, `compileall`, and `git diff --check` passed.
- The verification receipt reproduced successfully.
- Agents Shipgate reports `control.state=complete`, `decision=passed`, `merge_verdict=mergeable`, with no blockers or review items; this is a static-only verdict.
- All five active GitHub checks pass at this commit; the release-tag-only check is skipped as expected. GitHub reports the PR merge state as clean.

This is a comment review because the authenticated account owns the PR and GitHub does not permit self-approval. From this review pass, the PR is good to go.

Issue #396, opened 2026-08-20
`insufficient_evidence` headline claims the tool surface "could not be fully enumerated" when coverage reports 12/12 reachable


pengfei-threemoonslab:
## Summary

The `insufficient_evidence` headline says the tool surface **could not be fully enumerated**, directly above coverage numbers stating that all 12 catalog tools were reached with zero unbound tools, zero identity gaps and zero source warnings. The real condition is low extraction *confidence*, not incomplete *enumeration*, and the message picks the wrong half of its own disjunction.

## Evidence

Scan output, adk-samples#1745 / `python/agents/smart_closer`, CLI `0.16.0b7`:

```
Decision: insufficient_evidence
Summary: Insufficient evidence: the tool surface could not be fully enumerated
         (create_salesforce_quote [adk_agent]).
Evidence coverage: mixed (12 low-confidence tool(s); 12/12 catalog tools reachable;
                   36 semantic evidence gap(s); 0/12 actions pass-eligible)
```

```json
"binding_coverage":  {"total_catalog_tools": 12, "reachable_tools": 12,
                      "unbound_tools": 0, "possible_tools": 0,
                      "pass_eligible": true, "gap_count": 0},
"identity_coverage": {"canonical_tools": 12, "pass_eligible_tools": 12,
                      "ambiguous_name_count": 0, "gap_count": 0},
"source_warning_count": 0
```

The `incomplete_surface` gap body states the condition accurately as a disjunction:

```json
"why": "tool surface enumeration is incomplete or not high confidence"
```

Only the second disjunct is true here. The headline reports the first.

## Why this matters

- It contradicts the numbers printed on the following line, which is the fastest way to make a deterministic tool look broken.
- It sends the user to fix enumeration when nothing is missing.
- It obscures what is actually being asked. The gap's own fields say so plainly: `accepted_values: ["reviewed_explicit_inventory"]`, `requires_human_review: true`. This is a **human-attestation gate**, not an evidence gap.

Naming it honestly — "no reviewed inventory has attested this tool surface" — is both truer and more actionable, and it is the only framing under which the transcription work the remedy asks for is obviously worth doing.

## Suggested direction

Split the two conditions in the reason text and select on which one fired:

- enumeration genuinely incomplete (unbound/possible tools, source warnings, unresolved names) → "the tool surface could not be fully enumerated"
- surface complete but not attested → "no reviewed tool inventory attests this surface (extraction_confidence=medium)"

[...]

PR #478, opened 2026-08-31
Publish the determinism boundary: a generated coverage matrix of what shipgate can prove (#473)

### LINKED ISSUE #473: Publish the determinism boundary: a generated coverage report of what shipgate can prove, per source type and shape
## Summary

`insufficient_evidence` outside the determinism boundary reads as *"this tool can't test my repository"* — unless the boundary itself is a published specification. Generate — **from the code, not by hand** — the coverage matrix:

> source type × declaration shape (literal registration / dynamic construction / factory / export artifact) × extraction-confidence ceiling × what that ceiling means for verdicts

and publish it with every release. The pitch inverts the weakness: *inside the boundary, every word is proven; here is the boundary, drawn precisely.* Deterministic tools are allowed — required — to know exactly what they cannot see; this makes that knowledge a public artifact instead of a per-run disappointment.

## Why generated, not written

Every input to the matrix already exists in code: measured extraction confidence (#393 / PR #400), the `_AST_ONLY_SOURCE_TYPES` ceiling, the exclusion-ledger reasons (#403), per-source-type shapes. A hand-written table would be the route-table class (#433: never hand-maintain a table for a function that owns the routes) — it would drift within two releases and then be a published false claim, which is worse than no page.

## Scope

- A generator that walks the registered source types / shapes / ceilings and emits the matrix (markdown page + a machine-readable JSON beside it).
- **Fail-closed enumeration**: a source type added without coverage metadata fails the generation — the report cannot silently omit a route.
- A drift guard in CI: the committed page equals the regenerated page (the sample-golden pattern, #425).
- The `insufficient_evidence` headline links here — #396 fixed what the headline *says*; this gives its reader the map of *why*.

## Acceptance

- [ ] Matrix generated from code; committed; regenerated-equals-committed enforced in CI.
- [ ] Adding a source type without coverage metadata breaks the build (negative control committed).
- [ ] Every cell's confidence ceiling matches the engine's actual behavior on a sample (spot-checked by tests for at least the ADK `medium` ceiling and the MCP export `high` path).
- [ ] IE output links to the published page.
- [ ] Published with the release artifacts (site sync includes it).

## Surface discipline

A published artifact projected from the engine — no second judgment, no new command beyond the internal generator. Headline metric: activation / time-to-first-value — the boundary spec is what turns an honest abstention from a dead end into a scoping answer.

## Related

#393 / #400 (measured confidence) · #403 (exclusion ledger) · #396 (the headline this page backs) · #440 (reader-boundary epic) · the OWASP mapping (risk coverage; this page is repository coverage).

pengfei-threemoonslab:
Closes #473.

## What this adds

`docs/determinism-boundary.md` and `docs/determinism-boundary.json`
(`shipgate.determinism_boundary/v1`): for every built-in input and each of the
four declaration shapes — export artifact, literal registration, factory,
dynamic construction — what the scan reads, which `Tool.source_type` it
produces, the extraction-confidence ceiling that route reaches, and what that
ceiling means for a release verdict.

The page inverts the weakness the issue names. It publishes, among other
things, that:

- the **only** source-code route in any input that reaches `high` is a fully
  resolved ADK module — LangChain, CrewAI, and the OpenAI Agents SDK top out at
  `medium` and can never be pass-eligible on static extraction alone;
- a wildcard MCP export reaches a `high` ceiling and still cannot be
  pass-eligible, because the surface it belongs to was never established;
- ADK's own ceiling is a property of the *module*, not of the tool read first:
  the same `def` is `high` in a resolved module and `medium` beside one
  unresolved expression.

## Why it is generated, not written

Per #433, a hand-maintained table beside the function that owns the routes
drifts, and a drifted published boundary is a false claim with a URL. So:

- **Each adapter declares its own coverage** (`coverage: ClassVar[SourceCoverage]`),
  beside the code that mints its confidences. It declares only what the adapter
  *does*: the emitted source type, the ceiling, the `extraction["surface"]`
  evidence, the annotation flags.
- **The consequence column is derived.** For each route the generator builds a
  probe tool from those declared facts and asks the engine's own predicates —
  `extraction_is_complete()` and `surface_is_complete()` — what they amount to.
  The five outcome sentences are keyed by that derived answer, so the page
  cannot describe a state the predicates do not reach. This is what catches the
  wildcard-MCP cell: a table that restated "high means pass-eligible" would have
  got it wrong.
- **`extraction_is_complete()` is now the single definition** of "the adapter
  read this tool's contract with full confidence", shared by the semantic
  resolver, `low_confidence_tool_count`, and the generator.

## Fail-closed enumeration

Generation raises rather than omitting a route, in both directions:

- an adapter in the registry with no coverage → `BoundaryCoverageError`;
- a source type added to `AST_ONLY_SOURCE_TYPES` or `MCP_SOURCE_TYPES` that no
  cell emits → `BoundaryCoverageError`;
- a coverage that leaves a declaration shape unanswered, or gives one shape two
  routes without naming them, → refused at construction;
- a route claiming `high` for an AST-only source type without proving the
  surface, or claiming `extraction["surface"]` for a contract source type no
  adapter writes it for, → refused at construction.

All four are committed negative controls in
`tests/test_determinism_boundary.py`.

## Drift guard

`scripts/generate_schemas.py` emits both files through the existing `_emit`
path, so `--check` (already run by `ci.yml` and `release-verify.yml`) fails on
any difference between the committed page and the regenerated one. The suite
asserts the same thing independently.

## The cells are true of the engine

The bookkeeping above would all pass with wrong cells. `tests/test_determinism_boundary.py`
runs the **real adapters over real inputs** and asserts the measured
`(source_type, extraction_confidence, extraction["surface"], annotations)` is
exactly what the page publishes for that route, including both predicates:

- MCP export → `high` / permits pass (the acceptance item)
- MCP wildcard → `high` and still `set_unproven`
- ADK resolved module → `high`; ADK unresolved expression → `medium` (the
  acceptance item); ADK agent config → `low`
- OpenAPI → `high`; LangChain `@tool` → `medium`; CrewAI prebuilt → `low`
- Codex host config, both routes: schema present → `high`, server naming no
  tools → `medium` + wildcard/unknown-schema flags

## `insufficient_evidence` links to it

`scan` stdout, the GitHub step summary, and `report.md` print one line on an
`insufficient_evidence` verdict — spelled once in `schemas/disclaimers.py` so
the three surfaces cannot word it differently. #396 fixed what the headline
says; this gives its reader the map of why. Sample goldens regenerated
(`conductor_agent`, `declaration_repair_agent`, `google_adk_cold_start_agent`,
including its `cold-report.md`), and the conductor control pointer rebound to
the new `report.md` bytes.

## Publication

`.well-known/agents-shipgate.json` gains `determinism_boundary_url`,
`determinism_boundary_schema_version`, and a `schemas.determinism_boundary`
entry, so the file the site pins at each release carries both URLs. Also listed
in `docs/INDEX.md` and `llms.txt`. `docs/framework-adapter-checklist.md` gains
the coverage step (the fail-closed check enforces it either way).

## Surface discipline

Headline metric: **activation / time-to-first-value**. An honest abstention
outside the boundary currently reads as "this tool cannot analyse my
repository"; with the boundary published it reads as a scoping answer. No new
CLI command, no second decision engine, no new verdict — the page is a
projection of the existing extraction predicates, emitted by the internal
generator that already produces `docs/checks.json`.


pengfei-threemoonslab (commented):
Expert review result: **changes needed before merge**.

I reproduced cases where the generated boundary disagrees with real loader/check behavior. This is behavioral coverage, not generated-file drift: I reviewed exact head `489ce69`, ran focused probes/tests, `scripts/generate_schemas.py --check`, the full suite (including packaging with network), and Agents Shipgate; all pass, as do GitHub checks.

GitHub does not let the PR author submit `REQUEST_CHANGES`, so this is a `COMMENT` review. Please treat the `[P1]` findings as blocking for this published contract.

src/agents_shipgate/inputs/coverage.py
```diff
@@ -0,0 +1,635 @@
...
+#: of :func:`~agents_shipgate.ci.release_decision.evidence_below_ie_threshold`.
+#: That clause is real and it is never the one that fires first: every action
+#: below ``high`` also raises an ``incomplete_surface`` semantic issue, and the
+#: predicate's *first* clause treats ``semantic_coverage.gap_count > 0`` as
+#: sufficient on its own. Publishing the ratio would have told an adopter with
+#: one medium action in ten that they were under the bar.
+CELL_OUTCOME_VERDICTS: dict[CellOutcome, str] = {
+    "not_applicable": "No such declaration exists for this input.",
+    "not_extracted": (
+        "Nothing enters the catalog, so no check runs on it and no verdict "
```
pengfei-threemoonslab: **[P1] Preserve artifact-level verdicts for `not_extracted` routes**

This is not true globally. A dynamic Conductor `CALL_MCP_TOOL` emits no `Tool`, but populates `dynamic_tool_surfaces`; the Conductor check then raises HIGH `SHIP-CONDUCTOR-DYNAMIC-TOOL-SURFACE-NOT-ENUMERABLE`, so a check runs and the release decision is withheld. `test_conductor_dynamic_surfaces_have_precise_sources` reproduces it. Please support per-cell consequences or make the global wording neutral, with a real-scan assertion.

src/agents_shipgate/inputs/langchain.py
```diff
@@ -361,6 +362,58 @@ class LangChainAdapter:
...
+            BoundaryCell(
+                shape="export_artifact",
+                status="extracted",
+                reads=(
+                    "A reviewed tool inventory in MCP export form. It is read as a "
+                    "published contract, which is why it is the only LangChain route "
+                    "that reaches `high`."
+                ),
+                emits=("langchain_inventory",),
+                ceiling="high",
```
pengfei-threemoonslab, line 384: **[P1] Keep wildcard inventories out of `proven`**

A LangChain inventory containing `{"wildcard": true}` loads at `high`, but carries `wildcard_tools=true` and `surface_complete=false`. With no `surface_flags`, this cell publishes `proven`. CrewAI, Google ADK, n8n, and Codex plugin inventories share this wrapper behavior. Reject wildcard inventories for the reviewed route or model wildcard variants with a partial outcome, backed by real-loader tests.

src/agents_shipgate/inputs/google_adk.py
```diff
@@ -2198,6 +2199,97 @@ class GoogleADKAdapter:
...
+            BoundaryCell(
+                shape="dynamic_construction",
+                status="extracted",
+                reads=(
+                    "An unresolved tools expression, `Agent(**config)`, a rebound "
+                    "tool variable, a star-import shadow, or any extractor warning "
+                    "the module could not classify records a surface gap and caps "
+                    "the module at `medium`."
+                ),
+                emits=("google_adk_function",),
```
pengfei-threemoonslab, line 2317: **[P1] Include resolved-toolset actions in module-wide downgrades**

The prose says the whole module drops to `medium`, but these cells emit only `google_adk_function`. `_resolve_extraction_evidence()` also lowers `mcp` and `openapi` actions from resolved toolsets; `test_module_gaps_reach_tools_a_resolved_toolset_contributed` proves the MCP result is `medium`. Add medium/partial variants for those emitted types, or otherwise model this cross-cell downgrade.

src/agents_shipgate/inputs/n8n/_adapter.py
```diff
@@ -242,6 +243,67 @@ class N8nAdapter:
...
+            ),
+            BoundaryCell(
+                shape="dynamic_construction",
+                status="extracted",
+                reads=(
+                    "An MCP client node whose tool selection is `all`, `all_except`, "
+                    "or unreadable: one `<node>.*` action stands in for a selection "
+                    "the workflow does not enumerate."
+                ),
+                emits=("n8n_mcp_client_tool",),
```
pengfei-threemoonslab, line 331: **[P1] Model the dynamic workflow-tool route too**

A workflow-tool node with an expression-backed `toolName` still emits `n8n_workflow_tool` at `medium` while recording dynamic facts; a direct loader probe with `toolName: "={{ $json.tool_name }}"` produced that result. This cell only covers the MCP-client wildcard route, so the table omits a real catalog path and consequence. Add the workflow-tool variant and a real workflow fixture.

src/agents_shipgate/inputs/codex_plugin.py
```diff
@@ -966,6 +967,54 @@ class CodexPluginAdapter:
...
     artifact_class: ClassVar[type | None] = CodexPluginArtifacts
 
+    coverage: ClassVar[SourceCoverage] = SourceCoverage(
+        adapter="codex_plugin",
+        label="Codex plugin package / marketplace",
+        reads=(
+            "Plugin manifests, marketplace indexes, and the skills, apps, hooks, "
+            "MCP server stubs, and MCP inventories a plugin ships."
+        ),
+        manifest_section="codex_plugins",
```
pengfei-threemoonslab, line 977: **[P1] Do not advertise `codex_plugins:` as standalone configuration**

`load_codex_plugin_artifacts()` returns no plugin artifacts without a `tool_sources[]` row of type `codex_plugin`; top-level `codex_plugins:` only supplements discovered sources. The generator turns `manifest_section` into an alternative and publishes a route that cannot activate this adapter. Distinguish primary from supplemental configuration (or align the loader), and test through the real dispatcher.

src/agents_shipgate/inputs/conductor.py
```diff
@@ -966,6 +967,51 @@ class ConductorAdapter:
...
+            BoundaryCell(
+                shape="factory",
+                status="not_applicable",
+                reads="A workflow task is declared, never constructed by the file.",
+            ),
+            BoundaryCell(
+                shape="dynamic_construction",
+                status="not_extracted",
+                reads=(
+                    "A `CALL_MCP_TOOL` whose `method` or server is a workflow "
```
pengfei-threemoonslab: **[P1] Split dynamic method from dynamic server behavior**

The loader records both as dynamic facts, but only a non-literal `method` returns `None`. A literal method plus expression-backed `mcpServer` still creates `conductor_mcp_call` at `medium`. This cell says either field adds no action, so one advertised case is false. Split the variants and test both outcomes.

scripts/generate_schemas.py
```diff
@@ -1930,6 +1930,211 @@ def build_host_grants_drift_schema() -> tuple[Path, str]:
...
+route is therefore already a scan that cannot reach `passed` — the point is not
+how many you have.
+
+A ceiling is a ceiling, not a promise — an individual action can land lower, and
+effect, authority, identity, and binding evidence are judged separately and can
+withhold a pass from a route marked `proven` here.
+"""
+
+_BOUNDARY_REMEDY = """\
+Every row below `proven` has the same route out, and it is deliberately the only
```
pengfei-threemoonslab: **[P2] Generate remedies only where the adapter accepts them**

This is not implementable for many rows. `inventory_manifest_key()` has no inventory route for `sdk_function`, `conductor_mcp_call`, or `codex_config_mcp`, and `not_extracted`/`not_applicable` rows cannot generally become `proven` via inventory; the next paragraph also contradicts this claim. Generate source/status-specific remediation or limit the promise to adapters that support inventories.

src/agents_shipgate/schemas/disclaimers.py
```diff
@@ -11,3 +11,24 @@
...
+#: declaration shape.
+#:
+#: An ``insufficient_evidence`` verdict is only actionable if its reader can
+#: tell "this tool cannot analyse my repository" from "this repository declares
+#: its tools in a shape nothing static can read". #396 fixed what the headline
+#: says; this is where its reader goes for why (#473). The page is generated
+#: from the adapter registry and regenerated in CI, so the link cannot outlive
+#: the boundary it describes.
+DETERMINISM_BOUNDARY_URL = (
+    "https://github.com/ThreeMoonsLab/agents-shipgate/blob/main/docs/"
```
pengfei-threemoonslab, line 25: **[P2] Pin this evidence reference to the same release**

Released reports embed this constant, but `blob/main` changes as adapters evolve while the matrix schema version only versions shape. An archived report can later link to a matrix describing routes its scanner did not implement. Use an immutable tag/commit URL or a package-versioned documentation endpoint.

Issue #474, opened 2026-08-31
RFC: prove the MCP configuration review gap before adding committed capability state


pengfei-threemoonslab:
## First advisory release scope — 2026-09-11

Owner decision: **adoption > completeness; make v1.0 ready with repository-independent work, then expand adoption.** The state-model RFC remains a hypothesis to validate after the existing host workflow is usable; it is not a prerequisite for first diff.

Removed from the selected v1.0 milestone; kept open and deferred, not resolved. #643/#644/#645/#646 define current product acceptance; #572 preserves the historical qualified no-go and #648 owns the actual publication contract. The prior evidence and acceptance below remain a record of the deferred work, not newly satisfied conditions.

---

## Problem

Agents Shipgate already evaluates supported Codex, Claude Code, Cursor and VS Code MCP configuration changes; see [host-boundary support](https://github.com/ThreeMoonsLab/agents-shipgate/blob/main/docs/host-boundary-support.md). Local attestations, organization policy pins and the append-only registry also exist. This RFC must not reimplement them or describe their inputs as entirely ungated.

The remaining question is narrower: **for a real MCP-consuming repository whose useful review is blocked by a configuration or inventory gap, what reviewed inventory—if any—is needed to explain the change?**

A server URL, executable or remote binding is itself nameable evidence even when its full downstream tool surface cannot be enumerated. The report should retain that known binding and distinguish its downstream coverage limit (#520).

## First increment — establish the missing review

1. Identify one real repository/PR where an existing supported host/config route cannot answer a concrete server/allowlist change. Record the exact unsupported file shape or missing inventory and the reviewer's current alternative.
2. Reuse the existing host/config reader, subject reconciliation, capability delta and frozen payload. Demonstrate a server or allowlist transition naming what was added/removed, source evidence, and what the reader cannot establish about downstream tools.
3. Compare the existing derived snapshot/baseline/input identities with an **optional** committed reviewed inventory. Add a lock-style artifact only if the reproduced workflow needs durable reviewed state that those primitives cannot carry.
4. If an artifact is justified, specify identity, regeneration and staleness before making it part of a gate. Derived facts and human-reviewed declarations remain distinct. Generating a snapshot cannot invent effect, authority or deployment-binding assertions.

This is a conditional follow-on to #521's observed workflow, not a prerequisite for installing the CLI or using the existing host check. A useful result may be that no new state artifact is needed.

## What is deferred or removed

[...]

Issue #470, opened 2026-08-31
Publish the capability delta as a standalone attestation: an in-toto predicate any consumer can verify


pengfei-threemoonslab:
## Summary

`verify` already computes the capability delta and binds it to receipts. Its only consumers today are our own renderers. Publish it as a **standalone, versioned attestation — an in-toto predicate** — so external tools (runtime gateways, policy engines, dashboards, other CI) can consume *"what can the agent do after this change"* without running shipgate.

This is the standards move of the cycle: the merge-time verifier's most legible fact becomes portable. It is also how the merge layer and the runtime layer become **complementary instead of competing** — a runtime policy engine that ingests the delta attestation gets a reviewed baseline it cannot compute itself, because by the time it runs, the change is already merged.

## Scope

- **Predicate type + payload**: the payload is the frozen capability schema from #469, wrapped in an in-toto attestation predicate (versioned predicate URI). No second payload shape.
- **Emitter**: `verify` writes the attestation beside the existing artifacts. One projection — the delta in the attestation and the delta in the PR comment come from the same computation (the #433 rule: never two renderings of one value from two code paths).
- **Reference consumer**: a small standalone verifier (script or minimal lib) that validates the attestation and prints the delta — the proof that consuming it requires nothing of ours at runtime.
- **Docs**: a worked example from a shipped sample, and an integration page addressed to consumers ("how to read a capability delta"), not to adopters.
- **Binding**: attested subject = the reviewed commit/tree, chained to the existing receipt identity (`input_set_id`); no new cryptography in this increment.

## Dependencies

- #469 (payload schema) blocks this.
- #437/#439 (subject-based delta) blocks this: the first published delta must count subjects, not changes — an attestation that says `+2` for one added tool is dead on arrival with exactly the audience this format is for.

## Acceptance

- [ ] Attestation emitted for the shipped samples; committed as sample goldens (the #425 pattern).
- [ ] Reference verifier validates a good attestation and rejects a tampered payload and a mismatched subject.
- [ ] Value-equality test: delta in the attestation ≡ delta in the PR comment for the same run (one projection, proven).
- [ ] A tool added but unbound (the #437 class) appears in the attestation per the subject-based rules — never silently absent.
- [ ] Docs page with the predicate URI, schema version, and a copy-paste consumer example.
- [ ] Emitted on a tagged release install, not only from a source checkout.

## Surface discipline

[...]

Issue #439, opened 2026-08-27
`capability_review` counts changes, not subjects: one added tool reports `+2` and prints the same name twice


pengfei-threemoonslab:
## Summary

`capability_review` counts and lists **changes**, not **subjects**. One tool added to a source produces two rows and reports `+2`, and the PR comment prints the same name twice under a heading that reads as a summary:

```
- Capability delta: +2, 0 modified, -0
- Top capability changes:
  - `update_incident`: review required; Capability added. (tools.json)
  - `update_incident`: review required; tool added (tools.json)
```

One tool was added. A reviewer reads two capabilities.

This is the surface adjacent to #364, which fixed exactly this reasoning for findings — *"a scan of four money-moving tools presented one fact as seventeen rows"* — and introduced `core.findings.subject_rollup` so human output groups by subject. `capability_review` was not part of that change, so the same report now renders findings grouped by subject and capability changes ungrouped, a few lines apart.

## Reproduction

[grafana/mcp-grafana#1080](https://github.com/grafana/mcp-grafana/pull/1080) — adds one tool, `update_incident`. Tool surface extracted to a `type: mcp` source, 99 tools at base, 100 at head. CLI `0.16.0b7` at `fc96bad9`.

Ground truth:

```
base 99   head 100   added: ['update_incident']   removed: []
```

`capability_review`:

```json
{"added": 2, "modified": 0, "removed": 0,
 "top_changes": [
   {"change_type": "action_added", "subject_kind": "action", "subject": "update_incident",
    "impact": "review_required", "rationale": "Capability added."},
   {"change_type": "tool_added",   "subject_kind": "tool",   "subject": "update_incident",
    "impact": "review_required", "rationale": "tool added"}]}
```

Both rows are individually correct — the tool entered the catalog *and* the action surface, which are genuinely different dimensions. The defect is that the count and the human-facing list are keyed on the change rather than on the thing that changed.

## Mechanism

`cli/verify/capability_review.py:82` — the counts are lengths of the change lists:

```python
added=len(capability_change.added),
modified=len(capability_change.broadened) + len(capability_change.narrowed),
removed=len(capability_change.removed),
```

and `top_changes` (`:65-75`) sorts the same per-member changes by impact, direction, kind, subject, id, then truncates. `subject` is a sort key; it is never a grouping key. `report/pr_comment.py:164` renders that list as-is.

[...]

Issue #437, opened 2026-08-27
Capability delta reports `+0` on a PR that adds a tool: the diff is computed over the analysed surface, which is what the new tool is missing from


pengfei-threemoonslab:
## Summary

The PR comment reports `Capability delta: +0, 0 modified, -0` on a pull request whose entire content is one added tool. The number is self-consistent — the tool is unbound, so it is not *bound* capability — but it sits two lines above a gap tally and reads, to a reviewer, as "this PR changed no capability."

Raised in [#433](https://github.com/ThreeMoonsLab/agents-shipgate/issues/433) as "worth a separate look — whether an added-but-unbound tool should appear in the delta at all", and deliberately left out of [#434](https://github.com/ThreeMoonsLab/agents-shipgate/pull/434): what counts as a delta is a decision change, not a rendering one.

## Reproduction

Two commits in an MCP repository. Base declares `list_issues` and `delete_repository` and binds both; head adds `find_duplicate` to `tools.json` and does not declare it — updating `shipgate.yaml` is a trust-root edit a coding agent cannot self-approve, so this is the normal shape, not a contrived one. CLI at `16261d99`.

```
capability_review added/modified/removed:  0 0 0

tool_surface_diff.enabled:  True
tool_surface_diff.tools:    []

tool_catalog names:            ['find_duplicate', 'list_issues', 'delete_repository']
tool_surface.total_tools:      2
binding_surface_facts.unbound: ['tool_v2_7cbbbd23…']
```

The catalog has three tools and the surface has two. The diff ran, found nothing, and is not wrong about what it looked at.

## Where the number comes from

`capability_review.added` is `len(capability_change.added)` (`cli/verify/capability_review.py`), and the tool half of `capability_change.added` is `report.tool_surface_diff.tools` filtered to `kind == "added"` (`core/findings/verifier_blocks.py::build_capability_change`).

`tool_surface_diff` is `compute_tool_surface_diff(public_tool_surface_facts, base_facts, …)` (`cli/scan/sanitization.py`), and `public_tool_surface_facts` is built from `public_tools` — the **analysed** set, which the binding graph has already narrowed to `reachable_tool_ids`. An unbound tool is exactly the tool that is not in it, on either side. So `+0` is structural: the delta cannot see the class of change the exclusion ledger exists to report.

## Why the two lines together are the problem

`pr-comment.md` renders them adjacent:

```
- Capability delta: +0, 0 modified, -0
[...]

PR #479, opened 2026-08-31
Freeze one capability schema for the exported delta and the committed state (#469)

### LINKED ISSUE #469: One capability schema for the exported delta and the committed state — freeze it before either surface ships
## Summary

Two planned public surfaces will serialize the same internal truth: the **exported capability delta** (a standalone attestation external tools can verify) and the **committed capability state** (the lock-file RFC). Both are projections of the capability-fact layer (`CapabilityFactV1`, the delta rows, the permission profile). If each surface defines its own serialization, we ship two divergent schemas of one structure — the recurring "second implementation" class (#322 documented it for commands; #232 lived it for report schemas).

This issue freezes **one versioned capability schema** before either surface exists, so both consume it and neither invents it.

## Scope

- Inventory the internal shapes the schema must carry: capability facts (`schemas/capabilities.py`), subject-based delta rows (`core/capability_delta.py` after the #437/#439 fix), permission profile fields, tool identity (identity is taken at the adapter read boundary — the schema records that identity, it never re-derives one).
- Define the frozen, versioned public schema: field names, identity keys, required/optional split, and the evolution rules (additive within a minor, deprecation over a minor cycle, per [STABILITY.md](../blob/main/STABILITY.md)).
- Committed as a JSON Schema + a prose spec page. No emitter ships from this issue.

## Explicitly out of scope

The attestation envelope (predicate type, signing) and the state-file semantics (location, regeneration discipline) belong to their own issues. This issue owns only the shared payload.

## Acceptance

- [ ] JSON Schema + spec page committed, with a version field and stated evolution policy.
- [ ] The delta-attestation and state-file designs both reference this schema by version; neither defines a second payload shape.
- [ ] Negative control: a field that exists internally but is not in the schema does not appear in either surface's output (the frozen-union lesson from #323: prove the schema is load-bearing, not decorative).
- [ ] Subject identity in the schema follows the subject-based counting rules (#437/#439): one subject, one row — the schema makes the `+2`-for-one-tool class structurally impossible.
- [ ] Round-trip test: internal facts → schema → parse → equal facts, on the shipped samples.

## Surface discipline

Per [CONTRIBUTING.md § Surface discipline](../blob/main/CONTRIBUTING.md#surface-discipline): this is **consolidation, not new surface** — it exists to prevent two parallel schemas of one structure. Headline metric: it is the prerequisite for the delta-attestation adoption target (external consumers of the format). It adds no command, no check, no engine path.

## Related

The delta attestation and lock/state RFC (filed alongside; this issue blocks both) · #432 (source binding — the provenance the schema must carry) · #437/#439 (subject-based delta counting) · #364/#433 (subject naming rules) · STABILITY.md.

pengfei-threemoonslab:
Closes #469.

## What this does

Two planned public surfaces will serialize the same internal truth: the **exported capability delta** published as a standalone attestation (#470) and the **committed capability state** (#474). If each defines its own serialization we ship two divergent schemas of one structure — the "second implementation" class #322 documented for commands and #232 lived through for report schemas.

This freezes **one versioned capability payload** before either surface exists, so both consume it and neither invents one.

| | |
|---|---|
| Schema version | `shipgate.capability_payload/v1` |
| JSON Schema | [`docs/capability-payload-schema.v1.json`](docs/capability-payload-schema.v1.json) — generated, drift-gated |
| Prose spec | [`docs/capability-payload.md`](docs/capability-payload.md) |
| Models | `agents_shipgate.schemas.capability_payload` |
| The one projection | `agents_shipgate.core.capability_payload` |

One document, two views discriminated on `view`: `state` for a point in time, `delta` for the movement between two. Both validated by the same schema file.

## Surface discipline

Per [CONTRIBUTING.md § Surface discipline](CONTRIBUTING.md#surface-discipline), answering the gate head-on as the issue does:

1. **Which headline metric?** This is the prerequisite for #470's adoption target (independent external consumers of the delta format). It is **consolidation, not new surface** — it exists to prevent two parallel schemas of one structure.
2. **Can an existing surface carry it?** It extends the one decision engine: the delta comes from `diff_capability_fact_sets`, the same engine the capability lock diff and the reviewer surfaces already consume. No parallel judgment, no second verdict.
3. **Non-goals respected?** No command, no check, no engine path, no adapter, no network. `release_decision.decision` remains the only release gate.

**Nothing emits the payload yet, by design** — the issue says "No emitter ships from this issue." No artifact is written, and `contract_version`, `report_schema_version`, the runtime contract JSON, `.well-known/agents-shipgate.json`, the capability lock, and every existing schema are unchanged.

## Acceptance

- [x] **JSON Schema + spec page committed, with a version field and stated evolution policy.** The spec states additive-within-`v1`, `/v2` for anything narrowing, deprecation over a minor cycle, and consumers must ignore unknown fields. `STABILITY.md` carries the same guarantees.
- [x] **Both designs reference this schema by version; neither defines a second payload shape.** The spec has a *For the two consuming surfaces* section naming #470 (`view: "delta"` as its predicate payload) and #474 (`view: "state"`), and a test fails if either stops being named.
- [x] **Negative control: an internal field not in the schema does not appear in the output.** Every model sets `extra="forbid"`; each excluded internal field is recorded with its reason in `UNPUBLISHED_FACT_FIELDS` / `UNPUBLISHED_LOCK_FIELDS`; and a **frozen-union test** (the #323 lesson) asserts those maps together cover every `CapabilityFactV1` field — so a new internal field cannot reach either surface, *or be silently dropped from both*, without a written decision. Two sweeps prove it on output: no excluded key appears in any shipped sample's payload, and an excluded field's *value* (a marker planted in `source_location`) appears nowhere in the serialized bytes.
- [x] **Subject identity makes the `+2`-for-one-tool class structurally impossible.** `subject.key` is derived from `{agent, provider, tool_id}` and deliberately **not** from the subject kind, so a tool and its action are one row carrying two changes. `subject.key` is unique across `subjects[]`; `capability_id` is unique across *both sides* of every transition within a row; and `summary` and each subject's `transition` are **recomputed on parse** — a payload whose counts or rollups disagree with its rows is rejected, not repaired, so a tampered or hand-edited attestation fails.
- [x] **Round-trip test on the shipped samples.** For all 16 samples: facts → payload → JSON → parse → equal, compared **field by field through `PUBLISHED_FACT_FIELDS`** rather than by spot check, plus total coverage (every fact published exactly once) and schema validation.

## One thing I added beyond the issue text

The frozen payload had no way to express the **#437 class** — an added-but-unbound tool produces no capability fact, so it would have been silently absent. #470's acceptance explicitly forbids that ("never silently absent"), which means the first surface to need it would have had to invent a second payload shape. So `analysis_coverage` is in `v1`:

- `status` is `not_requested | unavailable | complete`, and **neither of the first two means zero** — reading either as "nothing was left out" re-creates the `+0`-on-an-added-tool defect.
- Only `complete` may name subjects: *"we did not look"* cannot be written in the shape of *"we looked and found none"*.
- Subjects are **named**, not only counted (#433).
- Deliberately **not** disjoint from `subjects[]`: a tool that lost its binding is both removed from analysed capability and newly outside analysis, and both are true.

It is a caller input on both projection functions, because it cannot be derived from capability facts — the subjects it names are exactly the ones that produced no fact.

## Design notes worth a reviewer's attention

- **One classifier, not two.** `classify_semantic_permission` is split out of `core.capability_lattice` so the payload's `permission` block and `mcp audit` cannot disagree for the same evidence. The lattice's `risk_score` / `risk_level` / `reasons` are **not** published — a heuristic tuned for the audit surface is not a capability fact, and an external consumer gating on it would be gating on our tuning. An unmeasured profile is fail-closed: `status: "unavailable"` with `side_effect_unknown: true`, and the schema rejects the combination that would read as "measured and harmless".
- **Two digests, two questions.** `capability_set_digest` covers semantic content (evidence and `evidence_hash` stripped); `evidence_set_digest` covers provenance. Both are computed over the *published* rows, so a consumer recomputes them from the payload alone. A test proves an evidence-only move leaves the first digest still and moves the second.
- **A delta's `base`/`head` refs are equal to the state payloads' own refs** — tested — so a delta and a state can be proven to describe the same state.
- **No wall clock.** Two projections of the same static inputs are byte-identical; a test builds each sample twice with the fact list reversed and compares bytes.
- **Worked examples are generated, not written**, from `samples/ai_generated_refund_pr` (whose PR is one added tool) by `scripts/generate_schemas.py`, under the same `--check` drift gate as the schemas — the #425 pattern.

## Verification

- Full suite green (`pytest -n auto -m "not perf"`), `ruff check .` clean, `compileall` clean.
- `python scripts/generate_schemas.py --check` clean — no existing schema moved.
- 88 new tests in `tests/test_capability_payload.py`, plus schema/example drift tests in `tests/test_schema_roundtrip.py`.

---

## Review rounds

Four self-review passes ran on this branch after the first commit. Each found something the frozen format would otherwise have had to break compatibility to fix, so they are listed rather than squashed away.

**Round 1 — five findings** (`bc06c602`)

- `_merge_subject_refs` returned early when the key *and* the display name matched, so the identity guard below it never ran on that path: a `subject.key` collision between two same-named tools would have merged them silently. The identity check is now first and unconditional.
- A delta with zero subject rows could declare `base` and `head` digests that differ — an attestation saying "nothing changed" between two provably different states. No rows means both digests are equal by construction, so the invariant is free to enforce, and now is.
- The payload re-spells two closed vocabularies it cannot import without inverting the schemas/core layering (`PermissionClass`, the `reversibility` literal). Nothing pinned them, so widening one internally would have left CI green and raised a `ValidationError` inside the projection on the first adopter repo that produced the value. Now pinned — and, more generally, **every** published field's annotation is pinned against the internal field it projects.
- `canonical_payload_json` passed `default=str`, so a digest an external consumer verifies could be computed over a lossy `str()` rendering. It raises now.
- The subject-key test compared the helper against itself, so a divergence from the recipe published for external consumers would have passed. It recomputes from stdlib now.

**Round 2 — two findings** (`127d9a1f`)

- `evidence_set_digest` is keyed by `capability_id`, but uniqueness was enforced only *within* a row. A repeat across rows silently dropped one capability's provenance from the digest. `capability_id` is now unique across the whole payload.
- A `state` payload carries every row its digests are taken over and the spec promises they are recomputable — but nothing checked them. It now verifies its own digests on parse, and the recipe moved from the projection into the frozen format so the two cannot disagree.

**Round 3 — one finding, the significant one** (`c9054b0f`)

`subject_transition` derived `added`/`removed`/`modified` from the *kinds of a subject's changes*. A delta row carries only the capabilities that moved, so a tool with two operations that loses one produced `transition: "removed"` and `summary.removed_subjects: 1` — for a tool still fully present in head (reproduced against the shipped sample). The mirror case reported a tool that gained a second operation as an added subject.

That is #439's defect in the other direction, and the row shape could not express the truth. `CapabilityDeltaSubject` now carries `present_in_base` / `present_in_head`, read off the two fact sets and never off the changes; `transition` is recomputed from that pair on parse, and presence bounds the changes in the other direction too.

**Round 4 — two, both about what the spec promises** (`eaa2a866`)

The digest recipe was documented as recomputable but only checked against itself; it is now recomputed from the serialized payload with stdlib only. And the payload's absence of any release impact / severity / verdict is now stated with its reason: publishing a per-subject impact in an interchange format invites a consumer to gate on it, which is a second verdict by another name.


src/agents_shipgate/core/capability_lattice.py
```diff
@@ -171,17 +186,42 @@ def classify_tool_permission(tool: Tool) -> CapabilityPermissionProfile:
         classes.add("unknown")
 
     normalized = _normalize_classes(classes)
-    score = _risk_score(tool, normalized, side_effect_unknown=side_effect_unknown)
-    return CapabilityPermissionProfile(
+    return SemanticPermissionClassification(
         classes=tuple(sorted(normalized, key=lambda item: PERMISSION_CLASS_RANK[item])),
```
pengfei-threemoonslab: [P1] Make permission-class ordering total

`normalized` is a `set`, but `financial` and `production` both have rank 3. Because the sort key compares only that rank, their relative order inherits hash-randomized set iteration. The same assessment produced `('production', 'financial')` under `PYTHONHASHSEED=1` and `('financial', 'production')` under seed 4, which changed the public payload bytes and `capability_set_digest`. Sort with a total key such as `(PERMISSION_CLASS_RANK[item], item)` and add a cross-process regression test.

pengfei-threemoonslab (commented):
Blocking review: I found ten contract-level issues that should be resolved before `shipgate.capability_payload/v1` is frozen. The highest-risk failures are nondeterministic state identities, a published permission expansion mislabeled as `evidence_only`, and a delta that cannot represent a newly added unbound subject.

Verification was otherwise clean: focused payload/schema tests passed (111), targeted Ruff and schema-drift checks passed, the PR checks are green, and the source-checkout Agents Shipgate verifier reported `passed` / `mergeable`. Those checks do not exercise the cross-process, cross-language, adversarial-parser, or base/head-coverage cases in the inline findings.

Additional P2 follow-ups found during the review:
- the generated schema description still says a subject transition is rolled up from changes, although the implementation now derives it from base/head presence;
- a tool rename can lose the head spelling because `_merge_subject_refs` publishes the lexical minimum name;
- the permission and state-ref models should reject classifier-impossible tuples and negative/inconsistent counts.

This is submitted as a comment review because the authenticated GitHub account is also the PR author; please treat the inline P1 findings as changes requested.

src/agents_shipgate/core/capability_payload.py
```diff
@@ -0,0 +1,538 @@
...
+    """
+
+    assessment = fact.semantic_assessment
+    if assessment is None:
+        return CapabilityPermissionFacts(
+            status="unavailable",
+            classes=(),
+            side_effect_unknown=True,
+        )
+    classification = classify_semantic_permission(assessment)
```
pengfei-threemoonslab, line 597: [P1] Treat assessment-driven permission changes as semantic

The fact layer places the full `semantic_assessment` only in `evidence_hash`, while this new projection derives published `permission` fields from it. In a reproduced structural-write to inferred-write change, the permission moved from `('write',), side_effect_unknown=False` to `('write', 'unknown'), True`, and the capability-set digest changed, but the existing diff saw only `evidence_hash` and emitted `semantic_direction='evidence_only'`. The portable delta therefore labels a published permission expansion as provenance-only. Make the projected permission participate in semantic hashing/classification and emit the corresponding semantic change.

src/agents_shipgate/core/capability_payload.py
```diff
@@ -0,0 +1,538 @@
...
+    )
+
+
+def project_capability_delta(
+    base_facts: list[CapabilityFactV1],
+    head_facts: list[CapabilityFactV1],
+    *,
+    base_ref: str | None = None,
+    head_ref: str | None = None,
+    analysis_coverage: CapabilityAnalysisCoverage | None = None,
```
pengfei-threemoonslab: [P1] Carry base and head analysis coverage in a delta

One coverage snapshot cannot distinguish a newly added unbound subject from one that was already unbound in the base. With `base_facts=[]`, `head_facts=[]`, and head coverage naming a new subject, this function emits empty `subjects`, `added_subjects=0`, and equal base/head refs; the exact same payload represents the unchanged-preexisting case. That leaves the #437/#470 case unrepresentable in the frozen delta. Accept base/head coverage (or a directional coverage delta), bind each side into its state identity, and name/count the coverage transition.

src/agents_shipgate/schemas/capability_payload.py
```diff
@@ -0,0 +1,921 @@
...
+                f"({self.capability_changes}) cannot be fewer than subjects "
+                f"({self.subjects}): every subject row carries at least one change"
+            )
+        return self
+
+
+class _CapabilityPayloadBase(BaseModel):
+    model_config = ConfigDict(frozen=True, extra="forbid")
+
+    capability_payload_schema_version: Literal["shipgate.capability_payload/v1"] = (
```
pengfei-threemoonslab: [P1] Make promised wire fields actually required

Giving wire fields defaults removes them from the generated JSON Schema's `required` arrays. The committed schema accepts a state payload with `capability_payload_schema_version`, `analysis_coverage`, and `view` omitted, while the direct model silently fills them; the same pattern affects `subjects` and many row fields that the spec calls always present. A version/discriminator should never be repaired by a consumer. Make the public wire models require these keys and have the producer pass the explicit defaults, with negative omission tests against both validators.

src/agents_shipgate/schemas/capability_payload.py
```diff
@@ -0,0 +1,921 @@
...
+    rows and a ``+2`` (#439).
+
+    ``name`` is the adopter-facing spelling and is not identity: two providers
+    may ship the same name. Consumers that render a name must qualify it with
+    ``provider`` when two rows share one ``name``.
+    """
+
+    model_config = ConfigDict(frozen=True, extra="forbid")
+
+    key: str
```
pengfei-threemoonslab: [P1] Validate the subject key against its identity material

Validation checks only key uniqueness, not `key == subject_key(agent, provider, tool_id)`. Both JSON Schema and Pydantic accepted an arbitrary replacement key; splitting two capabilities with the same agent/provider/tool ID across two rows under different invented keys also validates after recomputing the state digest. Thus the schema can still express `+2` for one logical tool despite the acceptance criterion. Add a `CapabilitySubjectRef` validator that recomputes the published recipe (including coverage refs).

src/agents_shipgate/schemas/capability_payload.py
```diff
@@ -0,0 +1,921 @@
...
+    """Canonical JSON for digesting — sorted keys, no insignificant space.
+
+    Deliberately without a ``default=`` fallback. A digest an external consumer
+    verifies must not be computable over a lossy ``str()`` rendering: two
+    distinct values whose text form coincides would digest identically, and the
+    consumer would read two different states as one. A value this cannot
+    serialize is a bug in the caller, and raising says so.
+    """
+
+    return json.dumps(payload, sort_keys=True, separators=(",", ":"))
```
pengfei-threemoonslab: [P1] Specify interoperable canonical JSON bytes

`json.dumps` defaults to ASCII escaping, but the spec states only sorted keys and compact separators. For a valid non-ASCII identity, Python hashes an escaped `caf\u00e9` string while a normal JavaScript implementation hashes UTF-8 `café`; the subject key and state digests differ. Since the format is expressly for independent external consumers, adopt a normative canonicalization scheme (for example RFC 8785/JCS) or fully specify escaping/key ordering and publish cross-language Unicode test vectors before freezing v1.

docs/capability-payload.md
```diff
@@ -0,0 +1,341 @@
...
+`UNPUBLISHED_LOCK_FIELDS`; the load-bearing one is `cli_version`, which would
+change every consumer's digest on a release that changed nothing about what the
+agent can do.
+
+## Evolution policy
+
+`shipgate.capability_payload/v1` is **frozen**. It follows the compatibility
+rules in [`../STABILITY.md`](../STABILITY.md):
+
+- **Additive within the version.** A new optional field with a default may be
```
pengfei-threemoonslab: [P1] Reconcile additive-v1 evolution with the shipped validators

This compatibility promise is the opposite of the implementation: every model uses `extra='forbid'`, the generated object schemas use `additionalProperties: false`, and enums are closed `Literal`s. An existing v1 Pydantic or JSON Schema consumer therefore rejects both a future optional field and a future enum value instead of ignoring/falling back. Either ship a lenient reader/schema contract for v1 consumers, or require `/v2` for every field/vocabulary addition; the current evolution guarantee cannot be met.

src/agents_shipgate/schemas/capability_payload.py
```diff
@@ -0,0 +1,921 @@
...
+            capability_id = record["capability_id"]
+            if capability_id in evidence:
+                # Keying provenance by capability id silently drops a collision,
+                # which would make two states with different provenance digest
+                # alike. One capability is one row; say so rather than hash a
+                # set that is quietly one entry short.
+                raise ValueError(
+                    f"duplicate capability_id across subjects: {capability_id!r}"
+                )
+            evidence[capability_id] = record["evidence"]
```
pengfei-threemoonslab: [P1] Bind all published state into the state reference

`_semantic_only` removes `digests.evidence_hash`, while this evidence map contains only the `evidence` block; `analysis_coverage` is also outside both digests. Mutating the published evidence hash, or changing coverage from `not_requested` to `complete` with an omitted dangerous subject, produces a different valid payload with the same `CapabilityStateRef`. Matching refs therefore cannot prove the #470 delta and #474 state describe the same complete published state. Include these fields in the identity (or add a full-payload/coverage digest), or stop publishing the unbound field.

scripts/generate_schemas.py
```diff
@@ -1709,6 +1717,100 @@ def write_capability_lock_diff_schema(
...
 
+def build_capability_payload_schema() -> tuple[Path, str]:
+    """Generate the frozen shared capability payload schema (#469)."""
+
+    from agents_shipgate.schemas.capability_payload import (
+        CAPABILITY_PAYLOAD_SCHEMA_VERSION,
+        CapabilityPayloadV1,
+    )
+
+    schema = CapabilityPayloadV1.model_json_schema()
```
pengfei-threemoonslab, line 1957: [P1] Do not present this generated schema as the semantic validator

Pydantic `model_validator` rules are absent from `model_json_schema()`. For example, changing the committed delta example's `summary.subjects` by 100 is accepted by `jsonschema.validate` but rejected by `CapabilityDeltaPayloadV1`; uniqueness, presence rollups, digest verification, and permission invariants have the same split. An external tool told to validate the one schema file therefore does not receive the documented guarantees. Encode every expressible constraint in JSON Schema and publish a small language-neutral semantic validator/explicit second-stage contract for the rest, with negative tests against the artifact external consumers actually use.

src/agents_shipgate/schemas/capability_payload.py
```diff
@@ -0,0 +1,921 @@
...
+    model_config = ConfigDict(frozen=True, extra="forbid")
+
+    transition: CapabilityRecordTransition
+    changed_dimensions: tuple[CapabilityHashName, ...] = Field(default_factory=tuple)
+    semantic_direction: CapabilitySemanticDirection = "unknown"
+    semantic_changes: tuple[CapabilitySemanticChange, ...] = Field(default_factory=tuple)
+    before: CapabilityRecord | None = None
+    after: CapabilityRecord | None = None
+
+    @model_validator(mode="after")
```
pengfei-threemoonslab, line 881: [P1] Validate transition claims against the carried records

This validator checks only which sides are present and whether `changed_dimensions` is empty. It still accepts `transition='changed'` with identical before/after records, `changed` with different capability IDs, arbitrary dimension names that do not match the differing digests, and `added` with `semantic_direction='removed'`. Those are contradictory attestations that the reference parser should reject. Enforce ID equality/difference by transition, derive the exact changed dimensions from the two records, and require membership directions to match the transition.

src/agents_shipgate/core/capability_lattice.py
pengfei-threemoonslab: Reproduced and fixed. Confirmed the seed dependence directly: `('production', 'financial')` under `PYTHONHASHSEED=1` and `('financial', 'production')` under 4 — the stable sort preserves input order on the tied rank, and the input is a `set`.

`classify_semantic_permission` now sorts by `(PERMISSION_CLASS_RANK[item], item)`. `test_permission_classes_are_ordered_the_same_in_every_process` runs the classification in subprocesses under seeds 0/1/4/7/11/random and asserts one result.

The rank table is also restated in `schemas.capability_payload` (schemas cannot import core without inverting the layering), so `test_the_payload_permission_ranks_match_the_lattice` pins the ranks alongside the existing vocabulary parity test — a tie-break only works if both spellings agree on the ranks.

src/agents_shipgate/core/capability_payload.py
pengfei-threemoonslab: Reproduced and fixed — and the payload was contradicting *itself*, not just the fact layer: on the structural→inferred case the published permission went from `('financial',) / side_effect_unknown=False` to `('financial', 'unknown') / True`, `capability_set_digest` moved, and the same payload said `semantic_direction='evidence_only'`.

Two changes rather than one, because the projection and the parser both needed it:

- `_published_direction` in the projection: whenever the engine says `evidence_only` but the published records are not semantically equal, reclassify from the published rows and attach a `permission_changed` semantic change. Gained classes or newly-unknown side effects broaden; the reverse narrows; both is `mixed`; a `measured`↔`unavailable` status move is `unknown`, because losing a measurement is not a narrowing.
- `CapabilityRecordTransitionEntry` now *rejects* `evidence_only` whenever the two published records differ semantically. That generalises past `permission` — any future published field derived from the assessment is covered — and it holds for a payload this projection did not produce.

`records_semantically_equal` and the state digest share one definition of "semantic content of a record" (`record_semantic_projection`), so they cannot disagree about what counts. Tests: `test_an_assessment_only_change_that_moves_permission_is_not_evidence_only` (broadened, narrowed on the reverse, and a genuine file move still reading as `evidence_only`) and `test_evidence_only_cannot_be_claimed_over_differing_semantics`.

src/agents_shipgate/core/capability_payload.py
pengfei-threemoonslab: Reproduced and fixed. Confirmed the two cases produced byte-identical payloads.

`analysis_coverage` is no longer on the shared base. A `state` carries one `CapabilityAnalysisCoverage`; a `delta` carries `CapabilityCoverageDelta` with `base`, `head`, a derived `status`, and the recomputed `newly_outside_analysis` / `no_longer_outside_analysis`. `project_capability_delta` takes `base_analysis_coverage` and `head_analysis_coverage`.

Two rules worth calling out:

- **A comparison is only as established as its weaker side.** `coverage_delta_status` returns `complete` only when both sides are, and `unavailable` outranks `not_requested` — the fail-open shape stays distinguishable from never having asked. When the status is not `complete` both directional lists are empty: head may still name subjects, but "newly" is not something the payload knows.
- **Each side is bound into its own state identity**, per your other finding: `CapabilityStateRef.analysis_coverage_digest`, which the delta checks against the coverage it carries.

Tests: `test_a_delta_distinguishes_newly_unbound_from_already_unbound` (three distinct payloads for newly/already/resolved), `test_a_coverage_comparison_is_only_as_established_as_its_weaker_side`, `test_a_coverage_transition_cannot_be_asserted`.

src/agents_shipgate/schemas/capability_payload.py
pengfei-threemoonslab: Reproduced and fixed. Confirmed the committed schema accepted a payload with `capability_payload_schema_version`, `view`, `analysis_coverage` and `subjects` all omitted.

Every field of every model is now required — no schema defaults anywhere in the payload — and the projection passes the constants explicitly, including `view` and the version literal. Nullable stays nullable: `str | None` with no default is *required and may be null*, which is the distinction the spec wanted ("null means not stated by the evidence").

The one place a default was doing real work is now a named constructor instead: `CapabilityAnalysisCoverage.not_requested()`. A producer that did not look publishes `{"status": "not_requested", "subjects_outside_analysis": []}` rather than omitting the block, so a consumer never has to decide what an absent coverage block would have meant.

`test_the_committed_schema_rejects_stage_one_violations` covers the omissions against the committed artifact.

src/agents_shipgate/schemas/capability_payload.py
pengfei-threemoonslab: Reproduced and fixed — including the split you describe: two rows with invented keys and a recomputed state digest validated cleanly, so `+2` for one logical tool was still expressible.

`CapabilitySubjectRef` now recomputes `subject_key(agent, provider, tool_id)` on parse and rejects any key the published recipe does not produce, and `key` carries a `^capsubj_[0-9a-f]{16}$` pattern so stage-one validation catches malformed ones. That turns the documented derivation into an enforced one — a consumer following the published recipe reaches the same key or the payload is refused. Coverage refs go through the same model, so `subjects_outside_analysis` is covered too.

The test that builds a duplicate now has to construct a genuinely different, self-consistent identity to get past the key validator, which is the right shape for it.

src/agents_shipgate/schemas/capability_payload.py
pengfei-threemoonslab: Reproduced and fixed. `canonical_payload_json({"n": "café"})` was `{"n":"caf\\u00e9"}`.

Now `ensure_ascii=False` and `allow_nan=False`, and the digest takes the UTF-8 bytes. The spec states the scheme rather than gesturing at it, and says why each part is safe to rely on across languages:

- UTF-8, never escaped.
- Object keys sorted — and **every object key in this payload is ASCII** (schema field names, plus `cap_`-prefixed hex capability ids), so code-point and UTF-16 code-unit ordering coincide and the sort is unambiguous. String *values* may be non-ASCII.
- Integers only, no floats and no non-finite numbers, so the genuinely language-dependent part of JSON canonicalization does not arise.
- No `default=` fallback.

Within those constraints it agrees with RFC 8785. Test vectors: `test_canonical_json_is_utf8_and_unescaped`, `test_subject_key_survives_a_non_ascii_identity`, and the independent digest-recipe test now runs over a sample with a `café_tool` subject.

docs/capability-payload.md
pengfei-threemoonslab: Agreed, and taken the second route: **`v1` is closed, and that is the whole policy.**

You are right that the promise and the implementation were opposite, and I think the implementation is the correct half. The two schemas make different promises for different reasons: `report.json` is open by construction and tells consumers to ignore what they do not know; this payload is closed by construction — `additionalProperties: false` everywhere, closed enums everywhere — precisely because a frozen interchange format an external tool verifies must not accept content that tool cannot account for. A closed schema cannot also be additive, so the spec now says so instead of promising both.

The rules, in the spec and in `STABILITY.md`: any addition, removal, or change of meaning is `/v2`, published beside `v1`; `v1` stays committed and readable for at least one minor cycle after `v2` lands (the deprecation rule applied to a whole version, which is the unit here); a producer migrates by emitting both for a cycle rather than bending `v1`.

The paragraph also now names the mechanism that decides when a `v2` is due — the type-parity test comparing every published field's annotation against the internal field it projects — so a widened `Literal` fails CI rather than raising on an adopter's machine.

src/agents_shipgate/schemas/capability_payload.py
pengfei-threemoonslab: Reproduced and fixed. Confirmed that mutating `digests.evidence_hash` left the whole `CapabilityStateRef` unchanged.

Three digests now, and together they bind every field a state publishes:

- `capability_set_digest` — semantics (`evidence` and `digests.evidence_hash` stripped, unchanged).
- `evidence_set_digest` — the `capability_id →` provenance map, whose value is now `{"evidence": ..., "evidence_hash": ...}`. The record's own evidence digest belongs on the provenance side, which is where it was falling through.
- `analysis_coverage_digest` — new, so coverage is bound too. A `delta` carries `base`/`head` coverage explicitly and checks each side's digest against it, which is also what makes the per-side coverage in the other finding verifiable rather than merely present.

`state_digests` moved into `schemas.capability_payload` alongside the format, and a `state` payload verifies all three on parse — the spec's "recomputable from the payload alone" is now enforced, not just documented. Tests: `test_coverage_is_bound_into_each_state_identity`, `test_a_state_ref_that_disagrees_with_its_coverage_is_rejected`, and the independent stdlib recomputation of all three.

scripts/generate_schemas.py
pengfei-threemoonslab: Agreed, and fixed on both halves rather than only the second.

**Pushed into the file everything JSON Schema can express** (via a `_postprocess_capability_payload` pass, following the existing `_postprocess_*` pattern): every field required; closed enums; the `view` discriminator; `^capsubj_[0-9a-f]{16}$` and `^[0-9a-f]{64}$` patterns; `minimum: 0` on counts; `minItems: 1` on `capabilities[]` and `changes[]`; `uniqueItems` where identical rows are the cheap half; `if/then` for the transition ↔ sides ↔ direction coupling; `if/then` for the presence ↔ transition coupling plus "present on at least one side"; and the coverage naming rule. Fourteen negative cases now run **against the committed artifact**, not the models — `test_the_committed_schema_rejects_stage_one_violations`.

**Named the residue as stage two.** The rules that need a recomputation genuinely cannot be expressed in JSON Schema, so the schema's own `description` now says validation is two stages, says stage one is not sufficient, and points at the spec; and the spec has a *Validating a payload: two stages* section with a table of the nine stage-two rules and what each costs (one sha256, a count, a set difference). `CapabilityPayloadV1` is named as the reference implementation of both stages, and `test_the_schema_says_it_is_only_stage_one` fails if the description or the spec stops saying it.

The standalone language-neutral verifier stays #470's deliverable, which is where its acceptance criteria already live — but it now has an explicit contract to implement rather than having to be reverse-engineered from Pydantic.

src/agents_shipgate/schemas/capability_payload.py
pengfei-threemoonslab: Reproduced all four and fixed. The record publishes every per-dimension digest, so most of this is derivable rather than merely checkable.

- **`changed_dimensions` is derived**, not annotated: it must equal exactly the digests that differ between the two records (`changed_record_dimensions`). Naming a dimension that did not move, or omitting one that did, is rejected.
- **Two identical records are not a change** — an empty derived set is refused for `changed`/`reidentified`.
- **Identity distinguishes the two paired transitions**: `changed` requires the same `capability_id` *and* `identity_hash` on both sides; `reidentified` requires both to differ. I check `identity_hash` as well as the id because it is the meaningful statement and does not depend on the `cap_<identity_hash>` id convention holding for a legacy fact.
- **Membership directions must match**: `added` carries `semantic_direction: "added"`, `removed` carries `"removed"`, and a paired transition cannot claim either.
- Plus the `evidence_only` rule from the permission finding: it requires the two published records to be semantically equal.

Tests: `test_changed_dimensions_are_derived_from_the_records`, `test_changed_and_reidentified_are_distinguished_by_identity`, `test_two_identical_records_are_not_a_change`, `test_a_membership_change_cannot_claim_the_other_direction`. The first three couplings are also encoded in the published JSON Schema as `if/then`, so a stage-one consumer gets them too.

pengfei-threemoonslab:
## Review addressed — all ten P1 findings and the three P2 follow-ups

Every finding reproduced first, then fixed; each inline thread has the reproduction and the resolution. Pushed as `1c6c3303`.

**The three you flagged as highest risk**

| | Was | Now |
|---|---|---|
| Nondeterministic state identity | `('production','financial')` under `PYTHONHASHSEED=1`, the reverse under 4 — published bytes and `capability_set_digest` moved between runs | Total sort key; cross-process test over six seeds |
| Permission expansion labelled `evidence_only` | `('financial',)` → `('financial','unknown')`, side effects newly unknown, reported as provenance-only — while the same payload's `capability_set_digest` *had* moved | Projection classifies what it publishes and says why (`permission_changed`); the model rejects `evidence_only` over semantically differing records |
| Newly-unbound subject unrepresentable | Newly-unbound and already-unbound produced byte-identical payloads | Per-side coverage plus recomputed `newly_outside_analysis` / `no_longer_outside_analysis`, each side bound into its own state ref |

**A note on two of the fixes, since they went further than the finding asked**

- On the `evidence_only` mislabel: rather than special-casing `permission`, the parser now rejects `evidence_only` whenever the two published records differ *semantically*, sharing one definition of "semantic content of a record" with the state digest. That covers any future published field derived from the assessment, and holds for payloads this projection did not produce.
- On the JSON Schema: I did both halves. Everything JSON Schema can express is now in the published file — required keys, patterns, non-negative counts, non-empty lists, and `if/then` for the transition ↔ sides ↔ direction and presence ↔ transition couplings — pinned by fourteen negative tests **against the artifact**, not the models. The residue is genuinely unexpressible (it needs recomputation), so it is named as stage two in the schema's own `description` and enumerated in a spec table with what each rule costs. The standalone verifier stays #470's deliverable, but it now has a contract to implement rather than Pydantic to reverse-engineer.

**On the evolution policy** I took your second option. `v1` is closed: any addition is `/v2`. Promising additivity while shipping `additionalProperties: false` and closed enums was promising something a `v1` consumer does not do, and of the two halves the closed one is the right one for a format an external tool verifies. `v1` stays published and readable for a minor cycle after `v2`; producers migrate by emitting both.

**P2 follow-ups**: the stale "rolled up from changes" wording is gone from the generated schema; a rename now publishes the **head** spelling instead of the lexical minimum (which was as likely to be the name no longer in the tree); and the permission and state-ref models reject classifier-impossible tuples and inconsistent counts.

**Verification**: full suite green, `ruff` clean, `generate_schemas.py --check` clean, `compileall` clean. 137 tests in `tests/test_capability_payload.py`; new modules at 98%/97%, with only unreachable fail-closed raises uncovered. The committed examples were regenerated and still validate.

One thing worth your explicit sign-off, since it is a scope judgment rather than a defect: the payload still does not participate in `contract_version` or `.well-known`. Nothing emits it, so advertising it would be a contract bump for a schema no artifact carries — but the natural moment to add it is when #470 ships an emitter, and I would rather you confirm that than assume it.

src/agents_shipgate/core/capability_payload.py
```diff
@@ -0,0 +1,689 @@
...
+    function only groups the engine's per-capability rows onto their subjects.
+
+    Coverage is supplied **per side**, because one snapshot cannot answer the
+    #437 question: "a tool was added and is unbound" and "a tool has been
+    unbound since before this change" are different facts, and only the first is
+    something a reviewer of this diff must act on. The payload names the
+    transition between the two sides rather than making a consumer infer it.
+    """
+
+    diff = diff_capability_fact_sets(base_facts, head_facts)
```
pengfei-threemoonslab: [P1] Snapshot both sides before this multi-pass projection. `base_facts` and `head_facts` are caller-owned mutable lists, but this function walks them in the diff, subject-ref, presence, and state-ref passes; `_state_subjects` walks each side twice again. A `list` subclass returning fact F on its first iteration and G later produces a valid payload whose delta row names F while `head` exactly matches the state of G (`head` was iterated five times). An ordinary list mutated by another thread has the same mixed-revision failure. Convert each side to an immutable tuple once at API entry and use only those snapshots throughout (including `project_capability_state`) so the promised “one computation” is an actual consistency boundary.

src/agents_shipgate/schemas/capability_payload.py
```diff
@@ -0,0 +1,1277 @@
...
+    """The canonical serialization every digest in this payload is taken over.
+
+    Fully specified, because the format exists for consumers that are not this
+    program and may not be this language:
+
+    * **UTF-8, never escaped.** ``json.dumps`` escapes non-ASCII by default, so
+      a Python implementation would hash ``caf\\u00e9`` where a JavaScript one
+      hashes ``café`` — the same identity, two digests. ``ensure_ascii=False``,
+      and the digest takes the UTF-8 bytes.
+    * **Object keys sorted, no insignificant whitespace.** Every object key in
```
pengfei-threemoonslab, line 205: [P1] Enforce the canonicalization domain instead of asserting it. `CapabilityRecord.capability_id` is unrestricted and becomes a dynamic key in the evidence map; a schema-valid state with IDs `cap_\ue000` and `cap_😀` orders those keys differently under Python code-point sorting and RFC 8785/JavaScript UTF-16 sorting, producing different evidence digests. Schema-valid integers such as `source_start_line=9007199254740993` also round in JavaScript. Constrain dynamic keys to the producer's actual ASCII `cap_[0-9a-f]{16}` form and digest inputs to I-JSON-safe integers (and constrain semantic-change JSON values), or use a real JCS implementation. Add cross-language astral-key and safe-integer vectors to the normative spec; that page currently states only sorted keys/compact separators, not the Unicode rules claimed here.

src/agents_shipgate/schemas/capability_payload.py
```diff
@@ -0,0 +1,1277 @@
...
+    whether the direction is honest — is derived from the records themselves, so
+    a contradictory attestation is rejected rather than carried.
+    """
+
+    model_config = ConfigDict(frozen=True, extra="forbid")
+
+    transition: CapabilityRecordTransition
+    changed_dimensions: tuple[CapabilityHashName, ...]
+    semantic_direction: CapabilitySemanticDirection
+    semantic_changes: tuple[CapabilitySemanticChange, ...]
```
pengfei-threemoonslab: [P1] Do not reuse the open internal `CapabilitySemanticChange` model as a frozen wire type. Its `before`/`after` fields still default to `None` and are typed `Any`: removing both from the generated delta is accepted by JSON Schema and Pydantic, which silently repairs them to null, despite the “every field required” contract. It also admits values outside the documented integer-only canonical domain, and a future field added to the internal model would mutate v1 without the payload's type-parity guard noticing. Define a payload-specific model with required-but-nullable, JSON-safe fields, and add an exhaustive test that every object in `$defs` has `required == properties` (this is currently the only miss).

src/agents_shipgate/schemas/capability_payload.py
```diff
@@ -0,0 +1,1277 @@
...
+                "transition 'reidentified' is an identity moving: it requires "
+                "different capability_id and identity_hash on the two sides (got "
+                f"{before.capability_id} / {after.capability_id})"
+            )
+        # Evidence-only means *published* semantics did not move. The fact layer
+        # folds the semantic assessment into evidence_hash alone, and this
+        # payload publishes a permission block derived from it — so inheriting
+        # the fact layer's classification unchanged would label a published
+        # permission expansion as provenance-only.
+        if self.semantic_direction == "evidence_only" and not records_semantically_equal(
```
pengfei-threemoonslab: [P1] `semantic_direction` and `semantic_changes` are still unchecked producer assertions, although the model says the direction is derived/honest. On the checked-in changed binding row, replacing `unknown` with `broadened`, `narrowed`, or `mixed` validates; a pure evidence-only record pair also validates when relabeled `broadened` and its explanations are deleted. At minimum enforce semantic equality iff the direction is `evidence_only`, and recompute the directions/explanations that are visible from carried scope/resource/permission data. For dimensions that cannot be derived from the payload, document the fields as signed producer assertions and remove the contradiction-free/reference-verifier guarantee rather than implying stage two proves them.

src/agents_shipgate/schemas/capability_payload.py
```diff
@@ -0,0 +1,1277 @@
...
+        _require_sorted(self.subjects, _delta_subject_sort_key, what="subjects")
+        _require_unique(
+            [
+                capability_id
+                for entry in self.subjects
+                for capability_id in _entry_capability_ids(entry.changes)
+            ],
+            what="capability_id across subjects",
+        )
+        expected = delta_summary(self.subjects)
```
pengfei-threemoonslab, line 1476: [P1] Reconcile the state-ref counts with the membership rows. The valid example says base 1/1, head 2/2, one added subject, and one net added capability; changing the head counts to 100/100 still passes both schema and reference validation. Enforce `head.subject_count - base.subject_count == added_subjects - removed_subjects`, plus the analogous capability equation computed from added/removed record transitions. This also makes an empty delta require equal capability/subject counts. Local non-negative and `capability_count >= subject_count` checks do not bind the two refs to this delta.

src/agents_shipgate/schemas/capability_payload.py
```diff
@@ -0,0 +1,1277 @@
...
+    )
+
+
+CapabilityPayload = Annotated[
+    CapabilityStatePayloadV1 | CapabilityDeltaPayloadV1,
+    Field(discriminator="view"),
+]
+
+
+class CapabilityPayloadV1(RootModel[CapabilityPayload]):
```
pengfei-threemoonslab, line 1576: [P1] Make the advertised reference parser schema-strict. `CapabilityPayloadV1.model_validate_json` currently coerces wire-invalid values: `summary.subjects="2"`, `present_in_base="false"`, and `permission.side_effect_unknown="false"` are all accepted even though the published JSON Schema rejects them. Since the spec says this entrypoint implements both validation stages, either validate stage one before Pydantic or expose a strict entrypoint/configuration that applies recursively while still accepting JSON arrays for tuple fields. Add parity tests for stringified integers/booleans so the reference cannot accept a larger language than the public schema.

pengfei-threemoonslab (commented):
Follow-up infrastructure/contract review at `1c6c3303`. The remediation materially fixes the highest-risk issues from the first pass: permission ordering is process-independent, published permission moves are no longer mislabeled `evidence_only`, coverage is per-side and bound into state refs, subject keys are validated, all three state axes are digested, transition identities/dimensions are checked, and the v1 evolution policy now matches a closed validator.

I still found eight release-contract gaps that should be resolved before `shipgate.capability_payload/v1` is frozen. The common infrastructure concern is trust-boundary drift: the JSON Schema, the reference parser, the producer, and the normative spec still accept or describe different sets of payloads. The inline P1s include executable repros and bounded fixes.

Verification evidence: the focused payload/schema suite passed (145 tests); targeted Ruff, compileall, schema drift, and `git diff --check` passed; all current GitHub checks are green; and the source-checkout Agents Shipgate verifier reports `passed` / `mergeable` with current control `complete`. A local full non-perf run reached 100%, with only eight packaging-fixture setup errors because the sandboxed isolated wheel build could not download `hatchling`; GitHub's packaging-capable CI is green. These checks do not cover the adversarial type, snapshot, cross-language, and contradictory-ref cases below.

Infra lifecycle sign-off: deferring `contract_version` / `.well-known` registration until an emitter exists is reasonable; #470 should add discovery and the emitted artifact atomically. This review is submitted as `COMMENT` because the authenticated account is also the PR author; please treat the inline P1s as changes requested.

scripts/generate_schemas.py
```diff
@@ -1709,6 +1717,231 @@ def write_capability_lock_diff_schema(
...
+    ] + [
+        {
+            "not": {
+                "allOf": [
+                    when("present_in_base", False),
+                    when("present_in_head", False),
+                ]
+            }
+        }
+    ]
```
pengfei-threemoonslab, line 1892: [P1] Stage one still omits cross-field rules that JSON Schema can express. If the example's added subject is changed to carry a structurally valid removed capability, external `jsonschema.validate` accepts it while Pydantic correctly rejects it because a subject absent from base may contain only additions. The same artifact accepts classifier-impossible permissions such as `status=unavailable`, `classes=[read]`, `side_effect_unknown=false`; the reference parser itself also misses the reciprocal `unknown in classes => side_effect_unknown`. Add the subject-presence/change and permission conditionals (plus `uniqueItems` where applicable) here and negative artifact tests. The stage-two checklist does not list these rules, so leaving them out gives standalone consumers no correct validation route.

docs/capability-payload.md
```diff
@@ -0,0 +1,401 @@
...
+`tests/test_capability_payload.py` able to compare published rows against the
+facts that produced them.
+
+`capability_id` is unique across the whole payload, not only within a row —
+provenance is keyed by it, so a repeat would quietly drop one capability from
+`evidence_set_digest`.
+
+### State digests
+
+`CapabilityStateRef` carries two digests over the **published** rows, so a
```
pengfei-threemoonslab: [P1] Update the normative document to the contract the code now implements. This section still says two digests and defines provenance as `capability_id → evidence`, omitting both `analysis_coverage_digest` and the now-bound per-record `evidence_hash`. Line 98 also presents the single-state coverage shape as applying to both views, while a delta requires `{base, head, status, newly_outside_analysis, no_longer_outside_analysis}`. Finally, lines 170/253 (and STABILITY) say an empty delta requires all state digests to agree, but the primary #437 coverage-only delta correctly has zero subject rows and different coverage digests. Specify all three exact preimages, show both coverage shapes, and narrow the empty-delta equality rule to capability/evidence digests; otherwise an independent implementation will compute the wrong ref or reject the intended coverage-only payload.

src/agents_shipgate/core/capability_payload.py
pengfei-threemoonslab: Reproduced and fixed. A `list` subclass returning F on the first iteration and G afterwards produced a payload whose delta row named `cap_62f686271fe07f21` while `head.capability_set_digest` matched the state of G exactly — the mixed-revision failure you describe.

Both entrypoints now take one tuple snapshot per side, before anything reads them, and every later pass uses only the snapshots. `project_capability_state` does the same. The signatures take `Sequence` and never retain the caller's object, so `_state_subjects`, `_subject_keys`, `_delta_subject_refs` and the two `state_ref` calls all read the same frozen input.

That is what makes “one computation” an actual consistency boundary rather than a description of intent.

src/agents_shipgate/schemas/capability_payload.py
pengfei-threemoonslab: Reproduced and fixed. Confirmed `""` and `"😀"` sort opposite ways under code-point and UTF-16 ordering, and that `9007199254740993` survived the encoder.

Rather than document around it, the domain is now enforced:

- `capability_id` is constrained to `^cap_[0-9a-f]{16}$` — the producer's actual form. It is the **one dynamic object key** in any digest preimage, so restricting it makes every key in the payload ASCII and the sort unambiguous. `subject.key` and the state digests already had patterns; per-record digest tokens are now bounded to an ASCII token pattern too.
- Every published integer is `strict` and bounded to `|n| ≤ 9007199254740991`, with `MAX_SAFE_INTEGER` named and the reason recorded — a JavaScript consumer reads anything past it rounded, so an unbounded line number would digest differently there.
- `CapabilityChangeFact.before`/`after` are typed to the same domain (`str | int | bool | tuple[str, …] | None`), which is the other half of your point.

The spec now has a *Canonical bytes* section stating all five rules normatively — including the code-point-vs-UTF-16 disagreement and why constraining the key removes it — plus three cross-language vectors with their exact digests. `test_the_published_canonicalization_vectors_hold` asserts the table against the implementation, and `test_the_spec_states_the_unicode_and_integer_rules` fails if the page stops stating them.

src/agents_shipgate/schemas/capability_payload.py
pengfei-threemoonslab: Agreed and fixed. Confirmed both halves: removing `before`/`after` from every semantic change in the committed delta validated against the schema *and* was silently repaired to null by Pydantic.

`CapabilityChangeFact` is now payload-owned: a closed `kind` vocabulary, a required `direction` per dimension, `before`/`after` required-but-nullable and typed to the canonical domain, and a required `rationale`. Nothing about it can be widened by an edit to the internal report model.

Your last suggestion was the valuable one, so I implemented it as a general guard rather than a one-off: `test_every_published_object_requires_every_property` asserts `required == properties` for **every** object in `$defs`, and the generator raises during schema generation if any object publishes an optional property. You were right that this was the only miss — and it would not have stayed the only one.

src/agents_shipgate/schemas/capability_payload.py
pengfei-threemoonslab: Agreed, and I took the stronger of the two routes you offered rather than documenting the fields as assertions.

`semantic_direction` and `semantic_changes` are now **derived** from the two carried records and recomputed on parse. `published_semantic_shift` walks the published dimensions — `scope`, `resource`, `authority.scopes`, `broad_scope`, `risk_tags` (gain widens, loss narrows); `effect` and `reversibility` by rank; effect flags; controls, where losing a *proven* control widens; and `permission` — and rolls them up. Dimensions with no ordering (auth type, credential mode, evidence owner, operation) are reported as changes with direction `unknown`.

Consequences worth naming:

- Both of your repros now fail: relabelling the checked-in binding row `broadened`/`narrowed`/`mixed` is rejected, and so is deleting the explanations of a row that has them.
- `evidence_only` is not a claim at all any more — it *is* "the records are equal apart from provenance".
- The direction now means the direction of *what this payload publishes*, which is narrower than the fact layer's and is precisely the gap that produced the earlier `evidence_only` mislabel. The spec says so under *Direction and explanations are derived, not asserted*.
- This replaced the projection's local classifier, so there is one implementation rather than two.

One bug the change surfaced and I fixed with it: a dimension whose own direction was `mixed` contributed neither widening nor narrowing to the rollup, so a scope that gained one entry and lost another reported the whole row as `unknown`.

src/agents_shipgate/schemas/capability_payload.py
pengfei-threemoonslab: Reproduced and fixed — head counts of 100/100 on the one-added-tool example validated against both the schema and the parser.

`CapabilityDeltaPayloadV1` now enforces both equations:

- `head.subject_count - base.subject_count == added_subjects - removed_subjects`
- `head.capability_count - base.capability_count == (added record transitions) - (removed record transitions)`

`reidentified` and `changed` are net zero on the second, which is what makes it exact rather than approximate. As you note, it also makes an empty delta require equal counts on both axes — that now fires *before* the digest comparison, so the test for the empty-delta rule sets the counts straight first and then checks the digests, which is a better-shaped test than it was.

Both are in the stage-two table, since each needs a subtraction over the rows.

src/agents_shipgate/schemas/capability_payload.py
pengfei-threemoonslab: Reproduced and fixed. `summary.subjects="2"`, `present_in_base="false"` and `permission.side_effect_unknown="false"` were all coerced by the parser and all rejected by the published schema.

Every scalar in the payload is now strict — `StrictText`, `StrictFlag`, and the bounded integer aliases — applied at field level rather than model level, deliberately: `ConfigDict(strict=True)` would also stop `model_validate` accepting JSON arrays for tuple fields, which is the shape a consumer gets from `json.loads`. Field-level strictness closes the language gap without closing that door, and it leaves the generated JSON Schema unchanged.

`test_the_reference_parser_accepts_no_more_than_the_schema` asserts the parity directly: for each of stringified integer, stringified boolean, stringified boolean nested in `permission`, and a fractional count, both `jsonschema.validate` and `CapabilityPayloadV1.model_validate_json` must refuse.

scripts/generate_schemas.py
pengfei-threemoonslab: Agreed on both counts, and the reciprocal you spotted in the parser was a real hole — thank you.

Added to stage one, all verified against the committed artifact:

- **Presence bounds the changes**, not just the transition label: `present_in_base: false` constrains every entry's `transition` to `added`, and the mirror for head.
- **The permission shapes the classifier can produce**: `unavailable` implies empty classes and unknown side effects; `measured` implies at least one class; `read` is the whole profile or absent; `destructive` carries `write`; and `side_effect_unknown` ⇔ the `unknown` class **in both directions**. The reference parser was enforcing only one direction, so it is fixed there too — the reciprocal is the half that would mislead a consumer asking "is this safe?".
- A membership change carries no explanations.
- `uniqueItems` on every list.

The rejection table now runs twenty-six mutations against the file rather than the models. Two rules stay stage two because they are genuinely computations — the state-ref count reconciliation and the derived direction/explanations — and both are now listed in the spec's stage-two table, so a standalone consumer has a complete route rather than an unstated gap.

docs/capability-payload.md
pengfei-threemoonslab: All three corrections made — and the third was a genuine defect in the prose rather than the code, which is the more dangerous kind.

- **Three digests, with exact preimages.** The section now gives a table: `capability_set_digest` over the `subjects[]` array as `{"subject": …, "capabilities": [ … ]}` with `evidence` and `digests.evidence_hash` removed from each record; `evidence_set_digest` over the `capability_id → {"evidence": …, "evidence_hash": …}` map; `analysis_coverage_digest` over the whole coverage object. An independent implementation can rebuild all three from the page.
- **Both coverage shapes are shown**, with the delta's `{base, head, status, newly_outside_analysis, no_longer_outside_analysis}` and the rule that the directional lists are empty unless both sides are `complete`.
- **The empty-delta rule is narrowed to the capability and evidence digests**, with the reason stated: a change that only moves what could *not* be analysed has no subject rows by construction, and that coverage-only payload is exactly the #437 case this schema exists to make expressible. The code was already correct here; the spec and `STABILITY.md` would have led an independent implementation to reject the intended payload. Both are fixed.

The page also gained the *Canonical bytes* section and *Direction and explanations are derived* from the other findings, and the stage-one/stage-two lists are updated to match what the artifact now enforces.

pengfei-threemoonslab:
## Second review addressed — all eight P1s

Reproduced each first, then fixed. Pushed as `73498d93`; every inline thread has the repro and the resolution.

You named the common thread exactly right — the schema, the parser, the producer and the spec each accepted or described a different set of payloads — so the fixes are grouped that way:

**Producer ≠ itself.** `project_capability_delta` read each caller-owned list five times. A `list` subclass returning F first and G later produced a payload whose delta row named `cap_62f686271fe07f21` while `head`'s digest matched G. Both entrypoints now snapshot each side to a tuple at entry and read nothing else.

**Parser ≠ schema.** Strict scalars everywhere, so `"2"`, `"false"` and a fractional count are refused exactly where the published schema refuses them. Applied at field level rather than model level on purpose: `ConfigDict(strict=True)` would also stop `model_validate` accepting JSON arrays for tuple fields.

**Schema ≠ parser.** Presence now bounds the *changes* a subject may carry, not just its transition label; the classifier-impossible permission tuples are rejected by the artifact; membership changes carry no explanations. The reciprocal `unknown in classes ⇒ side_effect_unknown` you spotted was missing from the parser too — fixed there as well, since that direction is the one that misleads a consumer asking "is this safe?". Twenty-six negative cases now run against the file.

**Spec ≠ code.** Three digests with exact preimages, both coverage shapes, and — the one that was a real defect in the prose — the empty-delta rule narrowed to the capability and evidence digests. As written it would have led an independent implementation to *reject* the coverage-only #437 delta the code correctly allows.

**Two I took the stronger of the routes you offered:**

- `semantic_direction` and `semantic_changes` are now **derived** from the two carried records and recomputed on parse, rather than documented as signed assertions. Both your repros fail now, `evidence_only` stops being a claim at all, and the projection's local classifier is gone — one implementation. Surfaced and fixed a bug in the process: a dimension whose own direction was `mixed` contributed neither widening nor narrowing to the rollup.
- On the change model, your suggested exhaustive `required == properties` test is in — and the generator now *raises* if any object publishes an optional property, so it can't become a one-off. You were right that `CapabilitySemanticChange` was the only miss; it would not have stayed the only one.

`capability_id` is constrained to `^cap_[0-9a-f]{16}$` for the reason you gave: it is the one dynamic object key, so bounding it makes every key ASCII and removes the code-point-vs-UTF-16 disagreement rather than documenting around it. Every published integer is bounded to the I-JSON safe range. The spec publishes three cross-language vectors with exact digests, pinned by a test so the table can't drift.

**Verification**: full suite green, `ruff`, `compileall`, `generate_schemas.py --check` and `git diff --check` clean. 156 tests in the payload file; new modules at 99%/98% with only defensive raises uncovered.

Thanks for the infra sign-off on deferring `contract_version` / `.well-known` — agreed that #470 should add discovery and the emitted artifact atomically, and I'll note that on #470 so it isn't rediscovered there.

PR #481, opened 2026-08-31
Add replayable incident fixture suite

### LINKED ISSUE #471: Three replayable incident fixtures: the gate catching the failure shapes that already happened in the wild
## Summary

The strongest argument for a merge-time capability gate is that its failure class has **already happened in public**. Package three replayable fixtures, each mapping a publicly documented incident *shape* (not a reproduction of any vendor's bug) to a shipgate verdict a reader can re-run in one command:

1. **An agent weakens its own gate.** An agent-authored PR relaxes the repository's own guard configuration. Public shape: [claude-code #40117](https://github.com/anthropics/claude-code/issues/40117) (an agent bypassed explicit deny rules to land failing commits; closed not-planned), and GitHub hardcoding protection of `.github/agents/` for its own agent ([community discussion](https://github.com/orgs/community/discussions/187679)) — the platform's stated reason: otherwise "the agent can change its own execution rules". We already ship the `agent_weakens_gate` sample; this packages it as a narrated, replayable demo.
2. **The governed edits the governance file.** A PR touching agent-instruction/configuration files routes to human review — the `.github/agents/` class, generalized and repository-configurable instead of hardcoded per vendor.
3. **A capability change rides a routine release.** A PR that lands a malicious prompt/instruction change beside ordinary changes (the shape of the Amazon Q VS Code extension incident, 2025-07): the diff surfaces the prompt-surface change as a capability-relevant finding rather than noise.

Plus one non-fixture deliverable: a **pre-written incident-response article template** — when the next agent-changes-its-own-permissions incident lands, the detection write-up and fixture link publish within 24–48h instead of being written under deadline.

## Shape of each fixture

Minimal repo state + the PR diff + the expected verdict + a one-command replay + a ≤1-page write-up citing only public sources. Replays must work from a clean checkout of a **tagged release** (install-path realism), not only from `main`.

## Honest gap handling

Fixture 2 requires governance files as gated inputs at the *path* level. If that surface has not shipped when this lands, fixture 2 is committed as an **expected-fail** documenting the gap and linking the lock/state RFC — a fixture that fails honestly is itself evidence; a fixture quietly skipped is not. The decision to add that surface lives in the RFC, not here.

## Acceptance

- [ ] Three fixtures, each replayable in one command from a tagged-release install.
- [ ] Each write-up ≤1 page, public citations only, no reproduction of any vendor's actual vulnerable code.
- [ ] Fixture 1 reuses `agent_weakens_gate` (no second sample of the same fact).
- [ ] Fixture 2 lands as expected-fail if its surface is unshipped, with the gap named.
- [ ] The response-article template exists in the repo (or the marketing workspace) with fixture links and a filled example.
- [ ] Verdicts shown are the engine's real output — no mocked or hand-edited results anywhere in the demos.

## Surface discipline

Samples and documentation only; no new engine surface. Fixture 2 may *motivate* surface — that ask is made in the lock/state RFC with its own metric, not smuggled in here. Headline metric: activation (these are the demos a first contact replays).

## Related

`samples/agent_weakens_gate` · #425 (committed-artifact pinning pattern) · the lock/state RFC (fixture 2's surface) · #466 (the cold-reader artifact these demos produce).

pengfei-threemoonslab:
## Summary

- package three replayable incident shapes while reusing the reviewed `agent_weakens_gate` sample
- keep `governed_edits_governance` as an explicit expected-fail for the unshipped `.github/agents/**` path-level surface tracked in #474
- add `prompt_change_rides_release`, which changes a synthetic prompt beside package and changelog updates and routes to human review
- fail the expected-fail contract closed when no report exists, and detect gap resolution by the governed path named in finding evidence rather than one hard-coded check ID
- add one-page public-source write-ups, an incident-response article template, and a filled example with v0.18.0 package guidance plus a source-checkout fallback before release
- prove all three commands run from an installed wheel rather than only from a source checkout

**Headline metric: activation.** Each incident shape now has a one-command deterministic verifier result, an evidence-disciplined explanation, and a reusable response template. This work does not change the detection engine, schemas, check IDs, or release-decision rules.

## Real fixture output

- `agent_weakens_gate`: `blocked`, `can_merge_without_human: false`
- `governed_edits_governance`: desired `human_review_required`; observed `passed` / `mergeable`, `can_merge_without_human: true`, with the known gap printed and pinned
- `prompt_change_rides_release`: `review_required` / `human_review_required`, with both `SHIP-AGENT-BOUNDARY-PROTECTED-SURFACE-UNCLASSIFIED` and `SHIP-VERIFY-TRUST-ROOT-TOUCHED`; the trigger artifact proves `TRIGGER-PROMPTS-OR-POLICIES` routed the prompt while `TRIGGER-DOCS-ONLY-NEGATIVE` did not match

No verdict is mocked or checked in as edited output; each replay materializes a base/head Git history and invokes the normal verifier.

## Review follow-up

Commit `8daa340c` addresses the full review at `0d353fc`: version/install guidance, complete output documentation, the expected-fail merge permission, changelog coverage, typed fail-closed report handling, deletion-capable replay metadata, registry privacy and collision checks, replay-specific copy/list behavior, path-based gap resolution, metadata invariants, citation qualification, prompt-focused naming, trigger proof, and activation framing.

## Validation

- `python -m ruff check` on all changed Python files
- focused fixture, static-surface, documentation-link, and public-surface suites: passed
- installed-wheel packaging suite: 9 passed
- repository-wide `pytest -n auto -m "not perf"`: 7,059 passed, 7 skipped; its only 9 setup errors were the sandboxed isolated wheel build's dependency download, whose complete suite passed separately with build access
- all three incident replay commands from the source checkout: passed with the documented outputs
- committed PR-mode Agents Shipgate verify: `passed` / `mergeable` / control `complete`

Closes #471


pengfei-threemoonslab (commented):
## Review: replayable incident fixture suite

Reviewed at `0d353fc` against `main`. I ran all three replays, re-derived every
verdict the write-ups claim, checked all four public citations against their
sources, and simulated the #474 surface shipping to confirm the expected-fail
actually fails loudly when the gap closes.

**Verdict: sound engine work, honest framing, and the load-bearing guard works.
The blocking problems are all in the published prose, not the code** — most
importantly, every `uvx` command this PR documents fails today.

---

### What I verified (all confirmed)

| Claim | Result |
| --- | --- |
| `agent_weakens_gate` → `blocked` | ✅ (existing test) |
| `governed_edits_governance` → `mergeable` / `passed`, 0 findings, exit 0 | ✅ ran it |
| `capability_change_rides_release` → `human_review_required` / `review_required` | ✅ ran it |
| `SHIP-VERIFY-TRUST-ROOT-TOUCHED` names `prompts/release.md`, `matched_glob: **/prompts/**` | ✅ in `report.json` |
| The expected-fail is a *real* gap, not an unread diff | ✅ `trigger.input_status: complete`, `changed_files: ['.github/agents/release-reviewer.agent.md']`, `findings: []` |
| `ruff check` / `ruff format --check` | ✅ clean |
| Full suite (`pytest -n auto -m "not perf"`) | ✅ exit 0 |
| `tests/test_packaging.py` cost | ✅ **+4.2s** locally (3.1s call + 1.1s setup) — the CI-budget worry in `0d353fc` is unfounded; no 15-min-cap risk |

I also patched `**/.github/agents/**` into `TRUST_ROOT_SURFACES` to simulate
#474 increment 2 landing. The replay flipped to
`human_review_required` and exited **20** with
`Expected-fail resolved: … update this fixture's contract.` The
single most important property of this PR — that the honest gap cannot rot
into a silent lie — genuinely holds.

**Citations** — all four resolve and support the prose; I read each source:

- `anthropics/claude-code#40117` — real, titled *"Agent bypasses git pre-commit
  hooks using `--no-verify`, stash, and quiet flags despite explicit deny rules"*.
- `community/discussions/187679` — real, titled *"Why is Copilot now prohibited
  from modifying files under .github/agents?"*.
- GHSA-7g7f-ff96-5gcw — advisory says verbatim *"inappropriately scoped GitHub
  token"*, *"automatically included in a release"*, *"unsuccessful in executing
  due to a syntax error"*, *"removing 1.84.0 from distribution channels"*.
  `filled-example.md` paraphrases all four accurately.
- JFrog — confirms *"utilizing a malicious prompt. The malicious prompt asks the
  AI agent in use to produce commands for wiping the machine…"*, which supports
  "carried a prompt directed at the coding assistant".

That is a better sourcing job than most vendor security blogs. Good.

---

## Blocking

### 1. Every documented `uvx` command fails today

PyPI's latest `agents-shipgate` is **0.15.0**. This branch is `0.16.0b7`, #471
is milestoned **v0.18.0**, and both new fixture names exist in no published
release. So after this merges, a cold reader who follows `docs/incidents/README.md`
gets:

```
uvx agents-shipgate fixture run governed_edits_governance
# → Fixture 'governed_edits_governance' not found
```

Affects `docs/incidents/README.md` (the "Run all three from an installed
release" block), all three write-ups, `filled-example.md`,
`response-article-template.md`, and `samples/README.md`. The existing
`uvx … agent_weakens_gate` line in the README works only because that sample
shipped in v0.15.0.

This is also the acceptance criterion #471 wrote down — *"replayable in one
command from a **tagged-release** install"* — and the headline metric is
activation, so a first-contact reader hitting `not found` is the worst possible
failure. Either pin the floor (`uvx agents-shipgate@0.18.0 fixture run …`) or
state it in one line ("available from v0.18.0; before that, run from a checkout
with `./shipgate fixture run …`").

### 2. The write-ups under-report the engine's own output

`capability_change_rides_release` emits **two** review items, not one:

```
SHIP-AGENT-BOUNDARY-PROTECTED-SURFACE-UNCLASSIFIED  medium  prompts/release.md
SHIP-VERIFY-TRUST-ROOT-TOUCHED                      medium  prompts/release.md
```

`capability-change-rides-release.md` and `filled-example.md` both name only
`SHIP-VERIFY-TRUST-ROOT-TOUCHED`. The `review_required` decision is not
attributable to that check alone. `response-article-template.md` sets the
standard this violates — *"Copy verdicts and check IDs from fresh JSON
artifacts"*, *"check IDs and changed paths: `[values]`"* (plural) — and the
filled example is supposed to model that discipline.

### 3. The expected-fail write-up omits the number that quantifies the gap

`agent-weakens-gate.md` and `capability-change-rides-release.md` both state
`can_merge_without_human` is `false`. `governed-edits-governance.md` lists
desired verdict, decision, verdict, and missing check — but **not**
`can_merge_without_human`, which is `true`. That is the fact that says the
governance edit would merge with no human at all. In a suite whose premise is
"a fixture that fails honestly is itself evidence", the one omission being in
the failing write-up reads badly. Add the line.

### 4. No `CHANGELOG.md` entry

`## Unreleased` is untouched, but this adds two user-facing `fixture` names
and a new docs tree. The section becomes the release body, so this silently
drops out of the v0.18.0 notes.

---

## Should fix

### 5. `_report_replay_expectation` reads its inputs through `getattr` on `object`

[`cli/fixture.py:437-448`](src/agents_shipgate/cli/fixture.py#L437-L448):

```python
def _report_replay_expectation(replay, *, verifier: object, report: object) -> int:
    ...
    findings = getattr(report, "findings", [])
```

`run_verify` is typed `-> tuple[VerifierArtifact, ReadinessReport | None, int]`,
and it returns `(verifier, None, 0)` on the `not trigger["run_shipgate"]` path
(`cli/verify/orchestrator.py:798`). With `report is None`, `findings` silently
becomes `[]`, `absent_check_ids` is vacuously satisfied, and the expected-fail
reports "expected-fail" forever — the exact failure this fixture exists to
prevent.

It is **latent, not live**: today `TRIGGER-EXISTING-MANIFEST-PRESENT` force-runs
both replays, and the `observed_decision` pin would catch `decision=None`. But
the check-ID half of the contract fails open, the two `SimpleNamespace` unit
tests could never catch it, and the repo has been bitten by this shape before.
Type the parameters (`VerifierArtifact`, `ReadinessReport | None`) and assert
`report is not None` before reading findings. There is no mypy gate here, so
the annotation is the only guard — make it a real one.

### 6. `ReplayFixture` cannot express a file deletion

[`fixtures.py:36-37`](src/agents_shipgate/fixtures.py#L36-L37):

```python
base_files: tuple[tuple[str, str], ...]
head_files: tuple[tuple[str, str], ...]
```

but `_run_verify_pr_fixture` takes `Callable[[Path], dict[str, str | None]]`,
and `_apply_fixture_files` treats `None` as "delete". So the new data-driven
mechanism is strictly *less* expressive than the two hand-written fixtures it
generalizes — and the canonical incident shape, `agent_weakens_gate`, **is** a
deletion (`{".github/workflows/agents-shipgate.yml": None}`). The next shape
that removes a CODEOWNERS rule or a policy pack can't use this registry.
Widen both fields to `tuple[tuple[str, str | None], ...]`; `agent_weakens_gate`
then collapses into the registry and one of the three `if name == …` branches
in `fixture_run` goes away.

### 7. A replay name colliding with a sample directory silently shadows it

`fixture_path` rewrites `name = replay.base_fixture` before touching the
filesystem, and `list_fixtures` appends replay rows unconditionally. Demonstrated:

```python
REPLAY_FIXTURES["support_refund_agent"] = <any replay>
# duplicate count for support_refund_agent: 2
# fixture_path resolves to: agent_weakens_gate
```

Not reachable today, but nothing prevents it and `REPLAY_FIXTURES` is a public
mutable dict. One module-level assertion (`no replay name is a bundled sample
directory`) closes it permanently.

---

## Polish

**8. `fixture copy <replay>` produces a mislabeled tree.** Ran it:

```
governed_edits_governance/README.md         → "# agent_weakens_gate … deleting the repo's Shipgate CI gate"
governed_edits_governance/shipgate.yaml     → project.name: agent-weakens-gate
governed_edits_governance/INCIDENT-FIXTURE.md
```

Since [`cli/fixture.py:180`](src/agents_shipgate/cli/fixture.py#L180) is already
writing files into the copy, also overwrite `README.md` from
`replay.description` — otherwise anyone staging this for a walkthrough reads the
wrong demo.

**9. The resolution detector is keyed to one check ID.** #474 increment 2 says
"a change to a declared governance path routes to human review, with the file
named" — that may well ship a *new* check ID rather than
`SHIP-VERIFY-TRUST-ROOT-TOUCHED`. The exit code stays correct (20 via the
verdict mismatch — `test_expected_fail_is_not_resolved_by_an_unrelated_review_route`
pins this, good), but the operator sees `merge_verdict='human_review_required',
expected 'mergeable'`, which reads as a regression rather than "your gap closed,
convert this fixture". Consider keying resolution on **the gap's path appearing
in any finding's evidence** rather than on a check ID, and echoing `known_gap` +
the conversion instruction in the divergence branch too.

**10. Duplicated intersection.** `resolved_check_ids` (line 450) and
`unexpectedly_present` (line 466) are the same expression computed twice.

**11. Unguarded field invariants.** Nothing enforces that `known_gap` and
`desired_merge_verdict` are set together, or that `absent_check_ids` is
non-empty whenever `desired_merge_verdict` is set — with an empty
`absent_check_ids` the entire "Expected-fail resolved" branch is unreachable.
A `__post_init__` is three lines.

**12. `fixture list --json` path is misleading for replays.** Three rows now
report `samples/agent_weakens_gate`, and that directory's contents are not what
the replay runs.

**13. README reading order.** The new paragraph lands between the two demos and
`**…and here's the failure mode.** These two cases are constructed fixtures…`,
so "These two cases" now trails a paragraph advertising three more. Either move
it below the failure-mode paragraph or reword to "Both of those cases".

**14. Both GitHub-thread citations are weaker than the prose implies.**
`community/discussions/187679` is **Unanswered** — the "prevent agents from
modifying their own instructions" rationale comes from four community
commenters, not from GitHub, and at least one reads as speculation ("Maybe
GitHub Copilot was operating under…"). And `claude-code#40117` is **closed as
`not_planned`** (#471 noted this; the write-up drops it). The prose is already
carefully hedged ("participants describing", "the reporter documented") — better
than #471's framing, which attributed the rationale to "the platform". Finish
the job with one clause each: *"an unanswered community thread; GitHub has
published no official rationale"* and *"the report was closed as not-planned"*.
Fixture 2's entire premise rests on the weaker of the two.

**15. `capability_change_rides_release` is named for something the engine
reports as zero.** The generated PR comment says
`Capability delta (analysed surface): 0 subjects across 0 changes` — the
`review_required` verdict comes from a path-level trust-root touch, not from
capability analysis. The write-up hedges this correctly ("The deterministic
claim is narrower"), but the *name* is what gets quoted, and renaming a shipped
fixture ID is a deprecation cycle under `STABILITY.md`. Worth settling now —
`prompt_change_rides_release` says exactly what it does.

**16. One free win for the "not noise" claim.** The write-up says release
metadata "cannot make the change disappear as noise". The trigger artifact
actually proves it: `matched_rules` shows `TRIGGER-PROMPTS-OR-POLICIES →
run_shipgate`, and unlike the `governed_edits_governance` run,
`TRIGGER-DOCS-ONLY-NEGATIVE` does **not** match. Cite that and the claim stops
being an assertion.

---

## Surface discipline

`ReplayFixture`, `REPLAY_FIXTURES`, and `replay_fixture()` are new public names
in `agents_shipgate.fixtures`, and the two fixture IDs become surface subject to
`STABILITY.md` deprecation rules once tagged. `CONTRIBUTING.md § Surface
discipline` asks for the headline metric — #471 names it (activation), but the
PR body and CHANGELOG don't. Also consider `_REPLAY_FIXTURES` + the existing
`replay_fixture()` accessor; `tests/test_fixture.py` is the only importer of the
dict and can use the accessor instead.

## Credit where due

- The wheel-install test is the right test. `--target` + `PYTHONPATH` proves the
  bundled `samples/` resolve through `importlib.resources` from a non-source
  location — it would catch a packaging regression that the source-tree tests
  structurally cannot. (It does not exercise the console-script path, but CI's
  *"A released install reports itself as one"* step already covers that shape.)
- Reusing `agent_weakens_gate` rather than committing a third near-duplicate
  sample is the right call, and the `ReplayFixture` docstring explains why.
- `test_expected_fail_is_not_resolved_by_an_unrelated_review_route` is exactly
  the test I would have asked for: the desired verdict alone must not be allowed
  to close a named gap.
- The static-analysis boundary language in every write-up ("does not prove why
  the author removed it", "does not replace credential scoping, signed releases,
  extension-store review") is the discipline this project's positioning depends
  on, and it is applied consistently.

pengfei-threemoonslab:
Addressed the full review in `dd65ff64`.

- Pinned installed replay guidance to v0.18.0 and added the pre-release `./shipgate` checkout fallback everywhere the new fixtures are documented.
- Renamed the release demo to `prompt_change_rides_release`; its contract and write-ups now record both review findings and the trigger proof, and the governance expected-fail now records `can_merge_without_human: true`.
- Added the Unreleased changelog entry and the activation headline metric here and in the PR body.
- Typed replay evaluation as `VerifierArtifact` + `ReadinessReport | None`, fail closed on a missing report, widened replay files to `str | None`, moved `agent_weakens_gate` into the registry, made the registry/class private, and added collision and metadata-invariant guards.
- Gap resolution now keys on the governed path in finding evidence, accepts a future new check ID, and reports the known gap/conversion instruction on divergence.
- Replay list/copy output now distinguishes the replay identity from its backing sample and writes the requested incident README.
- Qualified the GitHub/Claude citation status, fixed the README transition, and added tests for every new guard.

Fresh validation includes all three live replays, focused fixture/static/docs/public-contract suites, the installed-wheel packaging suite (9 passed), and committed-ref Shipgate verification (`passed`, `mergeable`, control `complete`). The repository-wide non-perf run had 7,059 passing and 7 skipped tests; its only setup errors were the sandboxed wheel builder's dependency download, and that entire packaging suite passed separately with build access.

PR #482, opened 2026-08-31
Clarify verifier causes, evidence gaps, and preview routing

### LINKED ISSUE #396: `insufficient_evidence` headline claims the tool surface "could not be fully enumerated" when coverage reports 12/12 reachable

### LINKED ISSUE #414: Two `next_action` fields in one preview payload, different commands, no precedence marker — and the non-authoritative one comes first
## Summary

A single `verify --preview --json` payload contains two fields named `next_action`, carrying different commands, with no marker of which is authoritative. The non-authoritative one appears first when reading the document top to bottom, and it instructs the reader to run the command they just ran.

Reported by a first-time user during an adoption walk, whose question was verbatim: *"what is the difference between the current `agents-shipgate verify --preview --json` and the next action `... verify --preview --json`?"*

## Observed

`agents-shipgate verify --preview --json --workspace <adk-samples> --base <sha> --head pr-1745`, CLI `0.16.0b7` at `0a744364`, verifier schema 0.10, unconfigured workspace.

`trigger.next_action` — nested inside `trigger`, so it is encountered **first**:

```json
{
  "kind": "command",
  "command": ".../shipgate verify --preview --json",
  "why": "This change looks agent-related; start with verify preview and adopt Shipgate if the preview routes setup."
}
```

`control.next_action` — appears later in the document:

```json
{
  "actor": "coding_agent",
  "kind": "initialize",
  "command": ".../shipgate init --workspace <repo>/python/agents/smart_closer --write --json",
  "why": "This unconfigured workspace looks agent-related; initialize the local Shipgate contract ..."
}
```

`control.allowed_next_commands` contains only the `init` command. The trigger's suggestion is not in the allowlist, which is the only signal in the payload distinguishing the two — and it requires the reader to already know that `allowed_next_commands` is the arbiter.

## Mechanism

`src/agents_shipgate/triggers.py:319-329` builds the trigger's `next_action` from the matched rule's `command` (here `TRIGGER-GOOGLE-ADK-AGENT-TOOLS-CHANGED`, whose command is `agents-shipgate verify --preview --json`) or a default, with no parameter for the calling context. The same value is produced whether the trigger was evaluated standalone or embedded inside a `verify --preview` run.

Standalone, the advice is correct and useful — the trigger catalog is also consumed from the `AGENTS.md` rule table, from `shipgate check`, and as a CI pre-filter, where "you matched a capability rule, go run verify preview" is exactly the right next step. Embedded in a preview run, that precondition is already satisfied.

## Impact

- The documented contract is `control` (quickstart: *"Switch on `control.state`; follow `control.next_action`, `control.allowed_next_commands` … never infer control from prose"*), so a reader who follows the docs is fine. A reader who scans the JSON top-down is not, and `trigger` precedes `control` in the document.
- The misleading field's `command` is self-referential, which reads as a loop rather than as advice scoped to a different caller.
- Same class as #322 and #394: one question ("what should I run next?") answered twice by two subsystems with different context, and the payload does not say which answer governs.

## Reproduction

```bash
agents-shipgate verify --preview --json --workspace <any unconfigured agent repo> --base <base-sha> --head <head-ref> \
  | python3 -c "import json,sys; v=json.load(sys.stdin); print('trigger:', v['trigger']['next_action']['command']); print('control:', v['control']['next_action']['command']); print('allowed:', v['control']['allowed_next_commands'])"
```

Both fields print; only the second is in `allowed`.

---

Found 2026-08-23 during a fourth adoption walk of google/adk-samples#1745, by the user reading the payload rather than the docs.

### LINKED ISSUE #420: An internal policy id is printed as the prefix of an adopter-facing gap explanation (`mixed_policy_evidence`)
## Summary

An internal policy identifier is printed verbatim as the prefix of a user-facing evidence-gap explanation:

```
mixed_policy_evidence | builtin-effect-control-applicability: Policy applicability mixes authoritative and heuristic evidence.
```

`builtin-effect-control-applicability` is a rule id from `core/lenses/action_surface.py:404`. To a reader it looks like a log line that escaped into the report, and it appears at the moment they are least equipped to absorb it — immediately after a declaration they wrote was challenged.

## Not covered by #329, and deliberately so

I checked before filing, and this is **outside** the guard #329 established rather than a hole in it. `internal_vocabulary()` (`core/adopter_text.py:189`) detects *internal identity terms* — derived ids and identity-model vocabulary — which is exactly what #329 set out to remove. Run against the real string it returns clean:

```python
>>> internal_vocabulary('builtin-effect-control-applicability: Policy applicability mixes authoritative and heuristic evidence.')
()
```

So no existing rule is failing. The observation is that **internal rule/policy identifiers are a second category of internal string that reaches the same adopter surfaces**, and the mechanism #329 built for the first category has no equivalent for it.

## Reproduction

`google/adk-samples#1745`, `python/agents/smart_closer`, CLI `0.16.0b7` at `aa6bd618`. Manifest declaring `send_email` with `effect: read` (below its inferred `external_communication`) — i.e. any run that trips the #409 monotone rule:

```bash
agents-shipgate scan -c <manifest> --format json
python3 -c "
import json
r=json.load(open('agents-shipgate-reports/report.json'))
for g in r['release_decision']['evidence_coverage']['evidence_gaps']:
    if g['kind']=='mixed_policy_evidence': print(g['why'])
"
```

Both gaps a reader sees at that moment, side by side:

```
declaration_below_inferred_evidence  declared effect 'read' is weaker than inferred 'external_communication' evidence (risk_hint:keyword)
mixed_policy_evidence                builtin-effect-control-applicability: Policy applicability mixes authoritative and heuristic evidence.
```

The first names the tool, both values, and the source of the inference. The second names an internal rule and restates its own gap kind in abstract terms — no tool, no values, no action.

## Suggested direction

Two independent parts, either useful alone:

1. Drop the `policy_id` prefix from adopter-facing text, or move it to a structured field beside `why` where a machine reader can still use it.
2. Give `mixed_policy_evidence` the shape its sibling already has — name the subject, the two kinds of evidence in tension, and what closes it. "Policy applicability mixes authoritative and heuristic evidence" is a statement about the resolver, not about the reader's manifest.

Worth considering whether the #329 guard should grow a second term class for rule and policy identifiers, so this category cannot reappear the way the identity terms did.

---

Found 2026-08-24 on the fifth adoption walk of `adk-samples#1745`, at the point where the #409 monotone rule fires. Walk context: https://github.com/ThreeMoonsLab/agents-shipgate/issues/410#issuecomment-5401529032

### LINKED ISSUE #436: The headline names the blocking cause only when a governance notice is also in play; an ordinary blocked PR still gets a bare count
## Summary

[#365](https://github.com/ThreeMoonsLab/agents-shipgate/issues/365) taught the verifier headline to name the worst blocker rather than only count it. Its own commit message states why:

> That leading line also names the worst blocker instead of only counting it — a count reads the same whether the agent is missing a docstring or can move funds with no enforced control.

That leading line is `_report_primary_headline`, and `_verifier_headline` reaches it from **two branches only** — the newly-introduced-manifest branch and the self-approval branch. Every other run returns `report.agent_summary.headline` directly:

```python
# cli/verify/orchestrator.py::_verifier_headline
if manifest_introduced and not pure_adoption_review and report is not None:
    ...
    return _compose_with_reserved_suffix(_report_lead(report, suffix), suffix)   # <- named
note = _self_approval_note(...)
if note is not None:
    if report is not None and _blockers_outrank_governance(report.release_decision):
        return _compose_with_reserved_suffix(_report_lead(report, note), note)   # <- named
    return _compose_with_reserved_suffix(..., note)
if report is not None and report.agent_summary is not None:
    return _lead(_one_clause(report.agent_summary.headline))                     # <- counted
```

So the case #365 was written for — a PR that blocks release on critical findings — gets the named cause **only if it also touches the trust root or introduces a manifest**. An ordinary blocked PR, which is the common one, still gets the count.

## Reproduction

Two commits in an MCP repository whose manifest declares `delete_repository`; the head commit adds one unrelated tool and touches nothing else. `verify --base HEAD~1`, CLI at `16261d99`.

The blockers are there, both **critical**, and the picker finds them:

```
blockers: [('critical', 'SHIP-POLICY-APPROVAL-MISSING',
            'delete_repository lacks a declared approval policy'),
           ('critical', 'SHIP-ACTION-DESTRUCTIVE-ROLLBACK-MISSING',
            'delete_repository has destructive capability without required controls')]

_worst_blocker(...).title
  -> 'delete_repository has destructive capability without required controls'

_report_primary_headline(report)
  -> '2 active finding(s) block release; 3 review item(s) accepted as debt.
      Most severe: delete_repository has destructive capability without required controls.'
```

And the line that actually ships:

```
verifier.headline
  -> '2 active finding(s) block release; 3 review item(s) accepted as debt. ...'

"Most severe" in verifier.headline  ->  False
```

`_report_primary_headline` computed exactly the right sentence and nothing published it. Two critical blockers, and the one line that reaches the PR comment, `control.reason`, `control.next_action.why` and any triage list names neither.

## Why this is worth fixing rather than leaving

The ranking half of #365 shipped and works: when a governance notice competes with a critical blocker, the blocker leads. The *naming* half shipped only on the routes where that competition exists. A reviewer's exposure to the improvement is therefore inversely related to how ordinary their PR is — the repository that has already adopted, changes no policy, and simply blocks on a destructive tool is the one that still reads a count.

It also makes the surrounding line harder to interpret. On [#433](https://github.com/ThreeMoonsLab/agents-shipgate/issues/433)'s reproduction the headline now carries `Not fully analysed: find_duplicate [github_mcp] — …` next to `14 active finding(s) block release`, and a reader has no way to tell whether the 14 are about the named subject or about something else entirely. `Most severe:` is exactly the clause that would answer that.

## Suggested direction

Route the plain branch through `_report_primary_headline` as well — it already falls back to `agent_summary.headline` when there is no blocker, so the no-blocker case is unchanged by construction:

```python
if report is not None and report.agent_summary is not None:
    return _lead(_report_primary_headline(report))
```

Two things to settle deliberately rather than by accident:

- **Budget.** `_lead` fits its context with `_fit_sentences` against `MAX_ENVELOPE_PROSE_BYTES - len(primary)`, and `primary` grows by up to `_HEADLINE_TITLE_MAX_CHARS` (120) here. On the plain branch there is no reserved suffix, so nothing is at risk of being cut — but the evidence-gap context will yield more often. That is the documented priority order (verdict, then cause, then context), so it is probably right; it should be stated rather than discovered.
- **Blast radius.** Every blocked run's `verifier.headline`, `control.reason` and `control.next_action.why` gains a clause. It is prose, not a documented identifier, but it is worth a `## Migration Note` and a check of the PR-comment goldens.

## Acceptance

- [ ] The reproduction above puts `Most severe: delete_repository has destructive capability without required controls.` in `verifier.headline`, `control.reason` and `control.next_action.why`.
- [ ] A blocked run that touches no trust root and introduces no manifest names its worst blocker — the case with no existing test.
- [ ] A run with no blockers gains no clause (`_report_primary_headline` already returns the summary unchanged; the point is to pin it).
- [ ] The picker stays deterministic across the branches: severity, then check id, then title, same row on the plain path as on the governance paths.

---

Found 2026-08-26 while reviewing [#434](https://github.com/ThreeMoonsLab/agents-shipgate/pull/434) (#433). Noted there as out of scope: it is a severity-attribution question rather than a naming one, and it moves every blocked headline.

pengfei-threemoonslab:
Summary
- name the deterministic worst blocker in ordinary blocked verifier headlines and propagated control prose
- distinguish fully enumerated lower-confidence surfaces from genuinely incomplete enumeration
- consume embedded trigger advice so verifier control is the only operational next action
- replace engine-owned policy IDs with adopter-facing mixed-evidence guidance while preserving public check labels
- document the projection changes and update generated contracts and sample goldens

Testing
- PYTHONPATH=src python -m pytest -q --ignore=tests/test_packaging.py
- PYTHONPATH=src pytest tests/test_packaging.py -q
- python -m ruff check .
- git diff --check
- ref-bound Agents Shipgate verify: complete / passed / mergeable

Closes #436
Closes #396
Closes #414
Closes #420


pengfei-threemoonslab (commented):
## Review

Reviewed at `c58e684a` against `origin/main` `004523e0`. I ran the suite, `ruff check`, and reproduced each finding against the branch rather than reading the diff alone.

The `#436` half is right and lands cleanly: the picker is genuinely shared across all three branches now, the no-blocker case is unchanged by construction, and on `#433`'s own reproduction the line reads the way the issue wanted:

```
14 active finding(s) block release; 3 review item(s) accepted as debt. Most severe: delete_repository has
destructive capability without required controls. Not fully analysed: find_duplicate [github_mcp] — …
```

The other three have real problems. `#396` and `#420` each fix the sentence the issue quoted and leave the *paired* sentence saying the opposite, and `#414` neuters one of the two self-referential commands in the payload. Details below, blocking items first.

---

### B1 — Stale base; the conflict is semantic, not just textual

Merge base is `98286e12`, three commits behind `main`. `git merge origin/main` conflicts in two files, and neither resolves by picking a side:

**`src/agents_shipgate/core/semantic_assessment.py`** — `main` (`c7deae26`, #473) renamed `_surface_is_complete` → `surface_is_complete` and factored out `extraction_is_complete(tool)`, whose docstring states it is *"the sole extraction input to pass eligibility here"* and *"named once"*. This PR's hunk calls `_surface_is_complete(tool)` and re-spells `tool.extraction_confidence != "high"` twice in the same block. Keeping the PR's lines gives a `NameError`; keeping main's names still leaves a third spelling of the predicate #473 just consolidated. Use `surface_is_complete()` / `extraction_is_complete()`.

**`samples/conductor_agent/expected/current-control.json`** — both sides touched it. `main` rebound `report_markdown` (the determinism-boundary line added to `report.md`); this PR rebinds `report`. Both moved `current_control_id`. After the merge *both* artifact digests differ and `current_control_id` has to be recomputed over the merged refs — regenerate rather than hand-merge. (`test_sample_current_control_pointers_bind_the_committed_artifacts` will reject a hand-edit, but only after you've spent the time.)

The green CI on this PR is a green run against the stale base.

---

### B2 — One evidence-gap row now says the surface is fully enumerated *and* asks the reader to make it enumerable

The `why` is chosen from `surface_is_complete(tool)` in `semantic_assessment.py:853`; the `next_action` is chosen independently from `inventory_manifest_key(tool.source_type)` in `ci/release_decision.py:1720`. Nothing joins them. From the PR's own regenerated golden, `samples/conductor_agent/expected/report.json`:

```
why      : "tool surface is fully enumerated, but no reviewed tool inventory attests it
            (extraction_confidence=medium)"
action   : provide_complete_inventory
           "The complete statically-bound tool surface must be enumerable."
expects  : "Provide a complete MCP export, OpenAPI spec, or reviewed explicit tool inventory,
            then rerun verification."
```

The other branch is no better. `inventory_manifest_key("google_adk_function")` → `declare_tool_inventory`, whose `action_why` is *"…an explicit local tool inventory bound to this source is the supported way to **make the full surface enumerable**"* — so `#396`'s own reproduction still contradicts itself after this fix.

The new test masks it. `tests/test_evidence_gap_ranking.py:293` hand-writes a coherent action (`kind="declare_tool_inventory"`, `why="A reviewed inventory attests the static extraction."`) that the engine never emits. Build the gap through `build_release_decision` on a real tool and the fixture stops being writable that way.

Both halves need to come off the same discriminator — `#396`'s `why` and its `next_action` are one rendering, not two.

---

### B3 — "fully enumerated" is a positive claim built on a default-true predicate

`surface_is_complete()` returns `True` for every source type outside `AST_ONLY_SOURCE_TYPES` that carries no wildcard annotation — the bottom `return True`. `main`'s own #473 comment names the direction: *"inert here means 'the surface is complete', which is the fail-open direction."* The old disjunction was honest under either reading; the new sentence asserts enumeration for surfaces nobody measured:

```
mcp                  conf=low     -> tool surface is fully enumerated, but no reviewed tool inventory attests it (…=low)
openapi              conf=low     -> tool surface is fully enumerated, but no reviewed tool inventory attests it (…=low)
conductor_mcp_call   conf=medium  -> tool surface is fully enumerated, but no reviewed tool inventory attests it (…=medium)
```

Those are bare `Tool`s with no extraction data at all. And in the conductor golden the claim now sits in a report that also carries two HIGH `SHIP-CONDUCTOR-DYNAMIC-TOOL-SURFACE-NOT-ENUMERABLE` findings — "contradicts the numbers printed on the following line" is the sentence `#396` was filed with.

`#396` asked for *"no reviewed inventory has attested this tool surface (extraction_confidence=medium)"*, which is what `evidence_gap_headline` now produces and is true either way. The "fully enumerated" preamble is the added part and the only unearned one. Make the positive claim only where the adapter measured it (`tool.extraction.get("surface") == SURFACE_ENUMERATED`); otherwise state the attestation gap alone.

---

### H1 — The `#396` fix is a byte-exact string join across two modules, and its failure mode is a silent revert to the reported bug

`core/evidence_actions.py:293` re-parses the sentence written at `core/semantic_assessment.py:856` with a `fullmatch`. On a miss it falls through to `_GAP_PHRASE["incomplete_surface"]` — `"the tool surface could not be fully enumerated"`, i.e. exactly the string this issue exists to remove.

Nothing crosses the join: `tests/test_semantic_assessment.py:351` pins the producer, `tests/test_evidence_gap_ranking.py:282` hand-writes the same literal into a consumer-side fixture. Reword the producer, update its own test, and the headline regresses with the suite green.

(The `(low|medium)` alternation is exhaustive — `Confidence = Literal["low","medium","high"]` and `high` is unreachable in that branch — so that part is fine.)

Carry the distinction structurally (a separate `SemanticIssue.kind`, or one format/parse pair exported from `semantic_assessment` and imported here), and add one test that runs tool → `assess_tool_semantics` → `build_release_decision` → `evidence_gap_headline`.

---

### H2 — `mixed_policy_evidence`'s remedy sentence is hard-coded and wrong for three of its four authoritative bases

The tension clause is derived; the remedy is a fixed string:

```
('reviewed_declaration', 'inferred_keyword')
  Policy applicability combines a reviewed declaration with keyword inference.
  Confirm the reviewed declaration or provide structural evidence, then rerun verification.   ✅

('protocol_structure', 'inferred_regex')
  Policy applicability combines protocol structure with regular-expression inference.
  Confirm the reviewed declaration or provide structural evidence, then rerun verification.   ❌

('typed_provider_fact','structural_scope','inferred_keyword','inferred_regex')
  Policy applicability combines structural scope evidence and typed provider facts with keyword
  inference and regular-expression inference.
  Confirm the reviewed declaration or provide structural evidence, then rerun verification.   ❌
```

In the second and third there is no reviewed declaration to confirm, and structural evidence is already what's present. `#420` asked for "the subject, the two kinds of evidence in tension, **and what closes it**" — the third part is the one still describing the resolver rather than the reader's manifest. Derive the remedy from `authoritative` the same way the tension clause is derived.

Related and pre-existing, but now inconsistent with the improved `why`: `_GAP_PHRASE["mixed_policy_evidence"]` = *"policy evidence mixes declared and inferred sources"* is also wrong when nothing is declared.

---

### H3 — Dropping `policy_id` leaves no structured replacement

`EvidenceGap` has no `policy_id` field, so for an engine-owned policy the id is now unrecoverable from the report. `#420` offered two options — *"Drop the `policy_id` prefix from adopter-facing text, **or** move it to a structured field beside `why`"* — and this takes the first without the second.

The cost shows up in the PR's own canary: `tests/test_p0_policy_evidence_canaries.py:375` had to weaken its selector from `"builtin-effect-control-applicability" in gap.why` to `gap.kind == "mixed_policy_evidence"`, which will also pick up any future policy producing that kind on the same fixture. `next_action.path` disambiguates today only because there is exactly one engine-owned policy that emits gaps.

---

### M1 — The plain headline path is now the only route that can exceed `MAX_ENVELOPE_PROSE_BYTES`

`_lead()` never bounds `primary`; the two governance branches do, through `_compose_with_reserved_suffix`'s `_bounded_bytes`. `_HEADLINE_TITLE_MAX_CHARS` is a **codepoint** cap, so a multibyte blocker title overflows the byte budget:

```
plain branch      : 409 bytes
governance branch : 399 bytes   (same report, trust-root touched)
```

409 bytes reaches `verifier.headline` and `control.reason` intact and is then tail-truncated by `truncate_prose` in the compact envelope — so the two published surfaces disagree, and the cut lands mid tool name, which is the thing `_fit_sentences`' own docstring says never to do to a subject.

`_report_primary_headline` already has the parameter for this and no caller uses it: pass `title_byte_budget` from the plain branch, or wrap in `_bounded_bytes(…, MAX_ENVELOPE_PROSE_BYTES)`.

`#436` also asked that the budget effect be *"stated rather than discovered"*. It isn't in either migration note: on #433's reproduction the context note goes from 3 sentences kept to 2.

---

### M2 — `_embedded_trigger` flattens every kind to `"none"`, overloading an existing value

Standalone, `kind: "none"` means "no action is warranted" (`triggers.py:348`, the skip case). Embedded it now also means "the diff could not be read" (`input_required`), "stop conditions fired" (`stop`), and "let the scan decide" (`unclassified`) — three states where an action *is* warranted. A consumer switching on `kind` and skipping on `"none"` is now wrong in the embedded case; `authoritative: false` only helps a reader who already knows to look for it.

`#414` asked for two things: say which answer governs, and stop publishing a self-referential command. Both are reachable without collapsing the kind — keep `kind` as evaluated, set `command: null`, add the two markers. That also keeps the `docs/agent-contract-current.md` line about `not_evaluated` → `input_required` true instead of having to delete it.

(Schema-wise the two new fields are safe: `trigger` is `additionalProperties: true` in the verifier, agent-result, agent-boundary and codex-boundary schemas, and the trigger block's `schema_version` mirrors the *catalog*, not the result shape. No bump is owed.)

---

### M3 — The self-referential command is still in the embedded trigger, one field earlier

On this branch, `verify --preview --json` against an unconfigured ADK repo:

```
trigger.matched_rules[0].command : …/agents_shipgate verify --preview --json     ← still there
trigger.next_action              : {"kind":"none","command":null, …}
control.next_action.command      : …/agents_shipgate init --workspace … --write --json
```

`matched_rules` precedes `next_action` in the document, so the top-down reader `#414` describes meets an executable, self-referential command *before* reaching either `next_action`. "Verifier control is the only operational next action" isn't yet true of the payload.

---

### M4 — `core/policy_evidence.py` now authors adopter prose but is not in the vocabulary sweep

`tests/test_adopter_vocabulary.py:858` `ADOPTER_FACING_MODULES` lists neither `core/policy_evidence.py` nor `core/semantic_assessment.py`; this PR turns both into prose emit sites. `#420` closed with *"worth considering whether the #329 guard should grow a second term class for rule and policy identifiers, so this category cannot reappear"* — the term class was added, but the module that produces the prose is still unswept.

`test_an_engine_policy_id_is_internal_adopter_vocabulary` only asserts the detector detects its own input; it proves nothing about any renderer. The one real assertion is the P0 canary, covering one gap on one fixture.

---

### M5 — The regenerated goldens gained a trailing newline the product does not write

`report/json_report.py:49` writes `json.dumps(payload, indent=2)` — no trailing newline. Regenerating the conductor golden on this branch:

```
fresh report.json     : 50989 bytes, no trailing \n
committed golden      : 50990 bytes, trailing \n
current-control.json  : sha256:cf431c90… / 50990   ← binds the post-processed bytes
```

The guard stays green because it hashes the committed bytes, but the shipped sample is once more "a digest for a file the product never writes" — the class #446 closed. Regenerate rather than post-process, and check the trailing byte before rebinding. Both `samples/*/expected/report.json` files in this diff are affected.

---

### Low

- **No CHANGELOG entry.** `CONTRIBUTING.md` § *Changing the CHANGELOG*: entries go under `## Unreleased`, and that section becomes the GitHub Release body. Two migration notes exist, but the release body won't mention behavior every blocked run now sees. #478 and #479 both added entries.
- **`_build_verifier`'s embedded trigger is untested.** The one new test (`tests/test_verify.py:3568`) exercises `run_preview` only. The `_build_verifier` call site at `orchestrator.py:3633` — `verifier.json` and `verify-run.json`, the two surfaces the migration note names *first* — has no coverage. Verified: reverting `_embedded_trigger(trigger)` → `trigger` at that line and re-running `test_verify.py test_verify_orchestrator.py test_diff_input_status.py test_headline_ranking.py test_verify_run.py test_verifier_control_contract.py test_verifier_scenarios.py test_verifier_blocks.py test_public_surface_contract.py` passes clean.
- **`AGENTS.md:544`** (mirrored at `llms-full.txt:569`) still says unqualified that `evaluation_status: not_evaluated` carries `next_action.kind: input_required`. True of the standalone command it describes, but it is the surface a coding agent reads *before* parsing `verifier.json`, and `docs/agent-contract-current.md` was updated for exactly this distinction.
- **Nits.** `_join_evidence_labels`' `"static evidence"` fallback is unreachable — both call sites pass a non-empty intersection. `_POLICY_EVIDENCE_LABELS[basis]` will `KeyError` if a new authoritative/heuristic `EvidenceBasis` is added; `.get(basis, basis.replace("_", " "))` is cheaper than the incident. `builtin-high-impact-approval` in `INTERNAL_POLICY_IDS` is inert — it never reaches `policy_evidence_gap` — while `docs/checks.md:264` deliberately teaches it to adopters as a value of `evidence.policy_id`; worth a line saying the ban is on prose, not on the structured field.
- CI `test` ran 14m47s against the 15m cap. Not this PR's doing, but the rebase re-run has no margin.

---

### What's right

- `#436`'s change is correct, deterministic across all three branches, and the no-blocker case is pinned rather than assumed.
- Both migration notes are pitched at the right altitude, and the `docs/agent-contract-current.md` → `llms-full.txt` mirror is consistent.
- The P0 canary was strengthened, not retired: it now asserts the rendered `why` *and* headline are free of internal vocabulary.
- `evidence_gap_headline`'s new phrase is exactly the wording `#396` asked for, and `internal_vocabulary()` returns clean on all five new sentences.
- The full suite passes locally on the branch (`pytest -q --ignore=tests/test_packaging.py`, exit 0), and `ruff check` is clean.

pengfei-threemoonslab:
Addressed in `765b7d97` after rebasing onto `004523e0`.

- **B1:** rebased onto current `main`, resolved the `surface_is_complete` / `extraction_is_complete` integration, regenerated the Conductor artifacts, and rebound both report hashes plus `current_control_id`.
- **B2/B3/H1:** replaced the prose-to-regex join with a structural `unattested_surface` kind. `incomplete_surface` now means enumeration actually failed; lower-confidence extraction without an enumeration defect means attestation is missing. Only exact adapter evidence `surface: enumerated` earns positive enumeration wording. The remediation is selected from that same kind, and a real Tool → semantic assessment → release decision → headline test covers the complete path.
- **H2:** mixed-policy explanations and remedies now derive from the authoritative bases actually present; the headline no longer assumes a declaration exists. All four authoritative bases are tested.
- **H3:** added optional structured `EvidenceGap.policy_id`. Engine-owned ids stay machine-readable while remaining absent from adopter prose; current report, packet, verifier, and capability schemas were regenerated.
- **M1:** blocker titles are now bounded by UTF-8 bytes before headline composition. Generated verdict/cause prose remains whole; context yields only by complete sentence. A hostile multibyte-title regression test and migration documentation cover the 400-byte behavior.
- **M2/M3:** embedded triggers preserve the evaluated `next_action.kind`, clear its command, retain the non-authoritative pointer to `control.next_action`, and clear commands from `matched_rules[]`. Both preview and `_build_verifier` paths are tested.
- **M4:** added `core/policy_evidence.py` and `core/semantic_assessment.py` to the adopter-vocabulary sweep.
- **M5:** regenerated the affected JSON goldens through the product writer with no trailing newline, added a terminator guard for all sample report goldens, and rebound the Conductor pointer after normalization.
- Added the Unreleased CHANGELOG entry, regenerated `llms-full.txt`, removed the unreachable label fallback, made new evidence labels forward-compatible, and documented that the `builtin-*` prohibition applies to prose rather than structured `policy_id`.

The one low-priority `AGENTS.md` wording change is intentionally not in this coding-agent commit: Shipgate preflight classifies `AGENTS.md` as a protected trust root and returned `human_review_required` when it was included in the edit plan. `docs/agent-contract-current.md` now states the standalone-vs-embedded distinction. A maintainer can apply the equivalent clarification to `AGENTS.md` in a separately reviewed trust-root edit.

Validation on the new head:

- complete test suite excluding packaging: passed
- isolated packaging suite: `8 passed`
- `ruff check .`: passed
- `scripts/generate_schemas.py --check`: passed
- committed-ref Shipgate verify: `control_state=complete`, `decision=passed`, `merge_verdict=mergeable`; current control id `sha256:4355eb97fdb9d9c361bd23f37c9f046f57cf97461ffcdd8032a041a9fb9d85bc`

Please re-review the new head when CI settles.

PR #483, opened 2026-09-01
Read an MCP server's tool surface out of its own source (#431)

### LINKED ISSUE #431: MCP servers whose tool surface exists only as code have no route: `detect` reports the official MongoDB and Grafana servers as not agent projects

pengfei-threemoonslab:
Closes #431.

`detect` reported the official MongoDB and Grafana MCP servers as **not an agent
project**. Both publish dozens of tools — `drop-database`, `delete-many`,
`update_incident` — and neither commits an export, so every route discovery had
was filename-shaped and found nothing. The discriminator was never the language
(two of the three servers walked are Go); it was whether the repository happens
to *emit* its tool list.

Reproduced before the change, on all three:

```
is_agent_project : false
suggested_sources: []
next_action      : "Stop: … no framework imports and no parseable tool artifact"
```

## What this adds

`tool_sources[].type: mcp_server_source` reads the tool name where every one of
those servers writes it: as a string literal at a registration site. It runs no
code, evaluates no schema library, infers no type, and reads no annotation the
source declares about itself.

Which sites count comes from a **built-in, versioned registry of named
registration idioms** (`agents_shipgate.inputs.mcp_idioms`), never from a
configured pattern — at `detect` time there is no manifest, and post-adoption a
configured regex would hand the definition of evidence to configuration
shipping in the same pull request as the tool it could hide (#268). Reading is
done over a **masked** copy of the source in which comments and string bodies
are overwritten, so a registration can never be found inside a comment or
inside another string, and a name is a name only when the masking pass recorded
a real literal at that offset.

## Scope, by measurement

The issue required a survey before choosing the idiom set. Thirty MCP servers
and the three official SDKs were cloned and scanned; the result is published at
[`docs/mcp-registration-idioms.md`](docs/mcp-registration-idioms.md).

| idiom | repos | largest | shipped |
|---|---|---|---|
| Python `@mcp.tool` (FastMCP) | 5 | `awslabs/mcp` (488) | **no — follow-up** |
| Go `Tool{Name: "…"}` | 3 | `github/github-mcp-server` (135) | yes |
| TS `.tool(` / `.registerTool("…"` | 3 | `cloudflare/mcp-server-cloudflare` (85) | yes |
| Go `NewTool("…"` | 2 | `hashicorp/terraform-mcp-server` (63) | yes |
| Go `MustTool("…"` | 1 | `grafana/mcp-grafana` (132) | yes |
| TS `static toolName = "…"` | 1 | `mongodb-js/mongodb-mcp-server` (77) | yes |

Python FastMCP is the largest measured shape and is deliberately deferred: it
is a different extraction mechanism (real AST, name from the function, schema
from the signature, provenance from an import binding), and #393 requires each
mechanism to bring its own adversarial probe list. The lexical reader here had
**eight fail-open constructs in its first draft**, all found by writing that
list. Filed as a follow-up with the measurement.

One shape was measured and rejected: a `{name: …, description: …}` object
literal, in 14 of the surveyed repositories. It matches most configuration in a
TypeScript codebase; a reader resting on it would invent tools rather than find
them.

## Verified against the live repositories

| acceptance (#431) | result |
|---|---|
| `detect` on `mongodb-mcp-server` → `is_agent_project: true`, named source | `mcp_server_source` at `packages`, confidence `medium` |
| `scan` enumerates the tools, `drop-database` + `delete-many` present | 61 tools, both present, each with its own file and line |
| extraction confidence measured, not assumed (#393) | `medium` + per-file `extraction.surface` `enumerated`/`partial` |
| a runtime-built name is unenumerated, and the ledger records it (#403) | 3 rows, `adapter_parse` / `name_not_literal` / `evidence_gap` |
| `mcp-grafana` #1080 delta names `update_incident` | 107 → 108, added `update_incident`, nothing removed |
| `github-mcp-server` regresses nothing | export route stays preferred and `high`; source route withheld and named |

The full chain was run on the real MongoDB checkout: `detect` → `init --write`
→ `scan` produces a manifest naming the route and a report with 61 tools.
Grafana's delta was measured by loading `pull/1080/head` and its parent: the
counts differ from the issue's 99 → 100 (a different counting method, and the
reader excludes test files), the **delta** is exactly the one the issue names.

Six repositories beyond the beachhead were checked end to end and all resolve a
route: cloudflare (80 tools), terraform (61), `modelcontextprotocol/servers`
(24 + 18 unenumerated), context7 (2), plus the three vendors.

## What it refuses to guess

- A name that is not a literal — a constant, a template substitution, a
  concatenation — is reported **unenumerated**, never dropped: a typed
  `SourceSurfaceOmission` the exclusion ledger accounts for by subject, holding
  its own file's surface at `partial` (per-file, as #393 settled).
- A file whose masking pass cannot complete (an unterminated string or block
  comment) is reported unreadable rather than read as though the rest were
  code, and holds the **whole source** at `partial` — as does a file skipped
  for size and the walk cap.
- `.tool("issues")` with one argument is a lookup, not a registration.
- `static toolName = "backup" + SUFFIX` resolves to a literal this reader can
  see; reading it as the name would publish a tool the server registers under
  another one, so it is unresolved instead.
- `NewTool(meta, mcp.Tool{Name: "issue_read"})` reports **one** site, not a
  tool and an omission for the wrapper that named it one argument later.

## Where an export still wins

An export is the server's own published contract, carries the input schemas
this route deliberately does not read, and is `high` against this route's
`medium`. `detect` withholds the source route wherever a parseable MCP export
covers the workspace and records the withheld route in `excluded_sources` with
the export that displaced it — named and visible, never silently dropped. With
both configured, observations are never joined by name (#386), so the export's
actions keep their own identity and their `high` evidence.

## Effect evidence

MongoDB writes `static operationType = "delete"` beside `drop-database`. That
arrives as a **low-confidence `inferred_keyword` hint**: it can contradict a
reviewer who declares the tool read-only and can never close an effect gap or
make an action pass-eligible on its own. `read` and `metadata` are deliberately
unmapped — a tool server asserting its own harmlessness is the one claim an
untrusted source has an incentive to make.

## Routing

`TRIGGER-MCP-TOOL-REGISTRATION-SOURCE` routes such a diff, on tokens derived
from the same registry and pinned against it by test. They are deliberately
narrower than what the reader matches: a matched capability rule overrides the
workspace stop condition, so a router firing on `.tool(` would turn "not an
agent project" into "run" for any diff containing that substring. Verified
against the real `mongodb-mcp-server#1417` diff — the walk that produced this
issue, and which previously matched no rule at all.

The trigger catalog's `schema_version` is **unchanged**: the rule uses existing
predicates and adds no field, so no consumer has to change to keep reading it.
`report_schema_version` is unchanged too — `generate_schemas.py --check` moves
only the determinism-boundary artifacts, which gain the new input's four rows.

## Surface discipline

- **Headline metric:** activation rate and time-to-first-verdict for MCP
  servers whose surface is only in code — currently *undefined* rather than
  poor, since `detect` says Stop.
- **Existing surface:** extended, not paralleled — one adapter in the existing
  registry, feeding the one decision engine and the existing questionnaire.
- **Non-goal:** filed as the evidence the "a real workflow needs one" clause
  asks for. Three vendor servers measured, two invisible; thirty surveyed.

## Also in this change

- `checks/mcp_permissions.MCP_SOURCE_TYPES` was a second copy of the constant in
  `core.semantic_assessment`; it now imports the one, and a test pins that.
- `_collect_dir_hits`'s framework list is named `CONVENTIONAL_DIR_FRAMEWORKS`
  with the reason it differs from the score sheet, instead of being a second
  copy inline.
- `tools/shipgate-detect.py` (zero-install) does **not** gain this detection:
  porting the reader means a second implementation of the load-bearing matcher,
  which deserves its own increment and a shared conformance corpus. The
  divergence is named in the script's own "intentional simplifications" list and
  pinned by a test, and it is filed as a follow-up.

## Review notes

Two things were implemented, measured, and reverted; both are recorded in code
comments and in the survey doc so the next person does not re-derive them:

1. **Splitting a monorepo by which package declares the MCP dependency.** More
   correct for `modelcontextprotocol/servers` and cloudflare; on MongoDB the SDK
   is declared by an eval-harness package and by three dev-dependency packages
   supporting the one server, so the split returned four scopes and withheld the
   route from the repository this issue is about. The single route is
   over-broad for a multi-server monorepo, visible in the manifest, and narrowed
   in one line.
2. **A pre-filter at the call site keyed off the trigger's diff tokens.**
   `static toolName` does not appear in `public static readonly toolName:
   string = "…"`, so four MongoDB tools were reported by `scan` and never by
   `detect`. The pre-filter now lives inside `scan_source`, where no caller can
   implement a weaker copy of it, and a test pins that.

### DIFF
--- a/src/agents_shipgate/cli/discovery/mcp_source.py
+++ b/src/agents_shipgate/cli/discovery/mcp_source.py
@@ -41,6 +41,7 @@ from pathlib import Path, PurePosixPath
 from agents_shipgate.inputs.mcp_idioms import (
     GO_FRAMEWORK_MODULES,
     MAX_SOURCE_FILE_BYTES,
+    SKIP_DIRECTORY_NAMES,
     TYPESCRIPT_FRAMEWORK_PACKAGES,
     SourceLanguage,
     is_scannable_path,
@@ -86,6 +87,11 @@ class McpSourceDiscovery:
     unresolved_count: int = 0
     #: Human-readable evidence for ``FrameworkDetection.evidence``.
     evidence: tuple[str, ...] = ()
+    #: The declared-dependency reasons behind the language gate, on their own.
+    #: Scoring reads this rather than indexing into ``evidence``: the rendered
+    #: lines are ordered for a human and gain conditional entries at the end,
+    #: so "the dependency reason" was a list position rather than a fact.
+    framework_evidence: tuple[str, ...] = ()
     #: Workspace-relative files that carried a resolved registration.
     candidate_files: tuple[str, ...] = ()
     #: ``{type, path, reason}`` rows for a route discovery found and withheld.
@@ -136,12 +142,17 @@ def discover_mcp_server_source(
         if pair[1] is not None
         and _language_in_scope(pair[1], framework.languages)
     ]
+    # Sorted before the cap, so which files are read is a property of the
+    # workspace and not of the walk order — and capped where the flag is set,
+    # rather than beside it: reporting `truncated` while reading every file
+    # published "this count is a lower bound" for a count that was exact.
+    scannable.sort(key=lambda pair: pair[1])
     truncated = len(scannable) > max_source_files
     names: set[str] = set()
     languages: set[SourceLanguage] = set()
     unresolved_by_file: dict[str, int] = {}
     candidate_files: list[str] = []
-    for path, relative in sorted(scannable, key=lambda item: item[1]):
+    for path, relative in scannable[:max_source_files]:
         language = language_for_path(path)
         if language is None:  # pragma: no cover - filtered above
             continue
@@ -205,6 +216,7 @@ def discover_mcp_server_source(
         tool_names=tuple(sorted(names)),
         unresolved_count=unresolved,
         evidence=evidence,
+        framework_evidence=tuple(sorted(framework.reasons)),
         candidate_files=tuple(sorted(candidate_files)),
         truncated=truncated,
     )
@@ -221,8 +233,13 @@ def _framework_evidence(
         relative = _relative(path, workspace)
         if relative is None:
             continue
+        # The same skip set the source walk uses, not a narrower one of its
+        # own: a manifest under `coverage/` or `out/` is no more this
+        # repository's declaration than one under `node_modules/`, and two
+        # lists would let a directory be skipped for registrations while still
+        # granting the language gate that admits them.
         parts = PurePosixPath(relative).parts
-        if any(part in {"node_modules", "vendor", "dist", "build"} for part in parts):
+        if any(part in SKIP_DIRECTORY_NAMES for part in parts):
             continue
         try:
             text = path.read_text(encoding="utf-8", errors="replace")
--- a/src/agents_shipgate/cli/discovery/mcp_source.py
+++ b/src/agents_shipgate/cli/discovery/mcp_source.py
@@ -1,0 +1,354 @@
+"""Discovery for MCP servers whose tool surface exists only as code (#431).
+
+``detect`` classified the official MongoDB and Grafana MCP servers as "not an
+agent project". Both publish dozens of tools — ``drop-database``,
+``delete-many``, ``update_incident`` — and neither commits an export, so every
+route discovery had was filename-shaped and found nothing. This module supplies
+the missing route: it asks whether a workspace *is* an MCP server written in
+TypeScript or Go, and if so which directory holds the registrations.
+
+Two facts have to hold together before the route is offered, and the pairing is
+the whole design.
+
+**Declared framework evidence.** A repository that declares no MCP dependency
+is not an MCP server just because a class of its own happens to spell a field
+``toolName``. #393's lesson is that a proof resting on a spelling is the
+fail-open shape, so the dependency — ``@modelcontextprotocol/*`` in a
+``package.json``, an MCP module in a ``go.mod`` — is what turns a spelling into
+provenance. It is also the cheap half: a workspace with no such dependency is
+answered without reading a single source file.
+
+**A resolved registration.** Dependency evidence alone says the repository
+*uses* MCP, which every client does too. At least one tool name has to be
+readable at a registration site before this claims a tool surface.
+
+Where both hold and the workspace *also* has a parseable MCP export, the export
+wins: it is the server's own published contract, it carries the input schemas
+this route deliberately does not read, and it is ``high`` confidence against
+``medium``. The source route is then withheld and recorded in
+``excluded_sources`` — named and visible, never silently dropped, because a
+route that vanishes without a reason is indistinguishable from one nobody
+implemented.
+"""
+
+from __future__ import annotations
+
+import json
+from collections.abc import Iterable, Sequence
+from dataclasses import dataclass, field
+from pathlib import Path, PurePosixPath
+
+from agents_shipgate.inputs.mcp_idioms import (
+    GO_FRAMEWORK_MODULES,
+    MAX_SOURCE_FILE_BYTES,
+    TYPESCRIPT_FRAMEWORK_PACKAGES,
+    SourceLanguage,
+    is_scannable_path,
+    language_for_path,
+    scan_source,
+)
+from agents_shipgate.inputs.mcp_server_source import SOURCE_TYPE
+
+#: How many source files discovery reads before it stops. Discovery runs on an
+#: unknown workspace with no manifest and has to stay bounded; the scan-time
+#: adapter has its own, larger cap. Truncation is reported, never silent.
+DEFAULT_MAX_SOURCE_FILES = 1500
+
+#: How many tool names the detection evidence names before it summarises. The
+#: evidence line is read by a human deciding whether to adopt, and 110 names is
+#: not evidence, it is a wall.
+_EVIDENCE_NAME_LIMIT = 5
+
+#: Dependency sections of a ``package.json`` that count. ``devDependencies`` is
+#: included on purpose: a server that bundles the SDK declares it there as
+#: often as not, and the question here is "was this repository written against
+#: MCP", not "does it ship the SDK at runtime".
+_PACKAGE_JSON_DEPENDENCY_KEYS = (
+    "dependencies",
+    "devDependencies",
+    "peerDependencies",
+    "optionalDependencies",
+)
+
+
+@dataclass(frozen=True)
+class McpSourceDiscovery:
+    """What discovery concluded about the workspace's own registration sites."""
+
+    #: Workspace-relative directory to point ``tool_sources[].path`` at, or
+    #: ``None`` when no route was found.
+    path: str | None = None
+    #: Languages whose idioms resolved at least one name.
+    languages: tuple[SourceLanguage, ...] = ()
+    #: Distinct resolved tool names, sorted.
+    tool_names: tuple[str, ...] = ()
+    #: Registration sites whose name could not be resolved to a literal.
+    unresolved_count: int = 0
+    #: Human-readable evidence for ``FrameworkDetection.evidence``.
+    evidence: tuple[str, ...] = ()
+    #: Workspace-relative files that carried a resolved registration.
+    candidate_files: tuple[str, ...] = ()
+    #: ``{type, path, reason}`` rows for a route discovery found and withheld.
+    excluded: tuple[dict[str, str], ...] = ()
+    #: Whether the file cap was reached before the walk finished.
+    truncated: bool = False
+
+    @property
+    def detected(self) -> bool:
+        return self.path is not None
+
+
+@dataclass
+class _FrameworkEvidence:
+    """Declared MCP dependencies, per language.
+
+    """
+
+    languages: set[SourceLanguage] = field(default_factory=set)
+    reasons: list[str] = field(default_factory=list)
+
+
+def discover_mcp_server_source(
+    workspace: Path,
+    *,
+    files: Sequence[Path],
+    exported_source_paths: Iterable[str] = (),
+    max_source_files: int = DEFAULT_MAX_SOURCE_FILES,
+) -> McpSourceDiscovery:
+    """Decide whether this workspace registers MCP tools in its own source.
+
+    ``exported_source_paths`` are the workspace-relative paths of MCP exports
+    discovery already accepted. Their presence is what withholds this route.
+    """
+
+    workspace = workspace.resolve()
+    framework = _framework_evidence(workspace, files)
+    if not framework.languages:
+        return McpSourceDiscovery()
+
+    # Paired with the workspace-relative path once, here: a file that is not
+    # under the workspace at all (a symlink out of the tree) has no relative
+    # form, and inventing one from its basename would put it in the wrong
+    # directory for both the skip rules and the route's common ancestor.
+    scannable = [
+        pair
+        for pair in ((path, _relative(path, workspace)) for path in files)
+        if pair[1] is not None
+        and _language_in_scope(pair[1], framework.languages)
+    ]
+    truncated = len(scannable) > max_source_files
+    names: set[str] = set()
+    languages: set[SourceLanguage] = set()
+    unresolved_by_file: dict[str, int] = {}
+    candidate_files: list[str] = []
+    for path, relative in sorted(scannable, key=lambda item: item[1]):
+        language = language_for_path(path)
+        if language is None:  # pragma: no cover - filtered above
+            continue
+        try:
+            if path.stat().st_size > MAX_SOURCE_FILE_BYTES:
+                continue
+            text = path.read_text(encoding="utf-8", errors="replace")
+        except OSError:
+            continue
+        result = scan_source(text, language)
+        resolved = [site.name for site in result.sites if site.name is not None]
+        opaque = sum(1 for site in result.sites if site.name is None)
+        if opaque:
+            unresolved_by_file[relative] = opaque
+        if not resolved:
+            continue
+        languages.add(language)
+        names.update(resolved)
+        candidate_files.append(relative)
+
+    if not names:
+        return McpSourceDiscovery(
+            unresolved_count=sum(unresolved_by_file.values()), truncated=truncated
+        )
+
+    root = _common_directory(candidate_files)
+    # Counted over the directory the route actually points at, so the number
+    # discovery publishes is the number the adapter will report once the route
+    # is configured. A whole-workspace count would name registrations `scan`
+    # never reaches.
+    unresolved = sum(
+        count
+        for relative, count in unresolved_by_file.items()
+        if root == "." or PurePosixPath(relative).is_relative_to(root)
+    )
+    evidence = _evidence_lines(
+        framework, languages, names, root, truncated, unresolved
+    )
+    covering_export = _preferred_export(exported_source_paths)
+    if covering_export is not None:
+        return McpSourceDiscovery(
+            unresolved_count=unresolved,
+            truncated=truncated,
+            excluded=(
+                {
+                    "type": SOURCE_TYPE,
+                    "path": root,
+                    "reason": (
+                        f"An MCP tool export ({covering_export}) already names "
+                        "this server's surface, and an export is read at high "
+                        "confidence with its input schemas; reading the "
+                        "registrations in source would restate it at medium."
+                    ),
+                },
+            ),
+        )
+
+    return McpSourceDiscovery(
+        path=root,
+        languages=tuple(sorted(languages)),
+        tool_names=tuple(sorted(names)),
+        unresolved_count=unresolved,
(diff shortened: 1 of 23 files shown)


pengfei-threemoonslab (commented):
Blocking review: GitHub does not allow the PR author to submit a formal `REQUEST_CHANGES` review on their own PR, so this is posted as a comment review. I recommend resolving the four inline findings before merge.

I reviewed the new lexical reader, discovery route, adapter, generated boundary artifacts, and focused tests. The core approach is well documented and the existing checks are green, but I found two cases where the gate evaluates the wrong or incomplete action surface, one lexer false positive, and one detect/scan contract mismatch. Each has a minimal reproduction in the inline comments.

Validation performed:
- focused MCP-source suite: passed
- exact CI suite: all non-packaging tests passed; the eight packaging setup errors were caused by sandboxed build-isolation downloads, and `pytest tests/test_packaging.py -q` passed when rerun with network access
- static-only adapter suite: passed
- Ruff and `scripts/generate_schemas.py --check`: passed
- Agents Shipgate verifier: `control_state=review_publishable`, `release_decision=review_required`, four protected-surface review items, no deterministic blocker

Please add adversarial regression cases for the four comments before merging.

src/agents_shipgate/inputs/mcp_idioms.py
```diff
@@ -0,0 +1,1150 @@
...
+            width = 4 if marker == "u" else 2
+            digits = body[index + 2 : index + 2 + width]
+            if len(digits) == width:
+                try:
+                    out.append(chr(int(digits, 16)))
+                    index += 2 + width
+                    continue
+                except ValueError:
+                    pass
+        out.append(_SIMPLE_ESCAPES.get(marker, marker))
```
pengfei-threemoonslab: [P1] Decode Go string literals with Go's grammar. This decoder is shared with TypeScript and falls through on Go octal escapes. For example, \`mcp.MustTool("delete\137all", desc, handler)\` registers the name \`delete_all\` (octal 137 is \`_\`), but \`scan_source\` returns \`delete137all\`. The real action is therefore absent while a nonexistent action ID is evaluated. Please split decoding by language, or reject every escape form that cannot be decoded exactly, and add this to the adversarial sweep.

src/agents_shipgate/inputs/mcp_idioms.py
```diff
@@ -0,0 +1,1150 @@
...
+    return masker.result()
+
+
+def _opens_regex(out: list[str], text: str, index: int) -> bool:
+    previous = _previous_significant(out, index - 1)
+    if previous == "" or previous in _REGEX_PRECEDING_CHARS:
+        return True
+    if previous.isalnum() or previous in "_$":
+        return _preceding_word(text, index - 1) in _REGEX_PRECEDING_WORDS
+    return False
```
pengfei-threemoonslab, line 762: [P2] Mask regex literals that begin a control-statement body. This heuristic returns false after \`)\`, but a regex can validly start the following statement: \`if (ok) /\.registerTool("fake", handler)/.test(value);\` is valid JavaScript, and the current scanner reports a resolved \`fake\` tool from it. That violates the module's guarantee that regex/string bodies cannot create registrations. Please track control-statement context or fail closed on this ambiguous slash, with a regression test.

src/agents_shipgate/cli/discovery/mcp_source.py
```diff
@@ -0,0 +1,371 @@
...
+    # never reaches.
+    unresolved = sum(
+        count
+        for relative, count in unresolved_by_file.items()
+        if root == "." or PurePosixPath(relative).is_relative_to(root)
+    )
+    evidence = _evidence_lines(
+        framework, languages, names, root, truncated, unresolved
+    )
+    covering_export = _preferred_export(exported_source_paths)
```
pengfei-threemoonslab: [P1] Do not let an unrelated or partial export erase the source route. \`_preferred_export\` returns the first MCP export anywhere in the workspace. This PR also documents that multi-server monorepos are collapsed to one source route, so an export for server A suppresses every source-only registration under server B; similarly, a parseable export containing only part of one server's surface suppresses the remaining registrations. Please bind the export to the same complete surface (or retain both and surface the identity ambiguity) before withholding this route, and add a two-server regression.

src/agents_shipgate/cli/discovery/mcp_source.py
```diff
@@ -0,0 +1,371 @@
...
+    unresolved_by_file: dict[str, int] = {}
+    candidate_files: list[str] = []
+    for path, relative in scannable[:max_source_files]:
+        language = language_for_path(path)
+        if language is None:  # pragma: no cover - filtered above
+            continue
+        try:
+            if path.stat().st_size > MAX_SOURCE_FILE_BYTES:
+                continue
+            text = path.read_text(encoding="utf-8", errors="replace")
```
pengfei-threemoonslab: [P2] Keep discovery and the scan-time adapter on the same decoding contract. \`errors="replace"\` can let \`detect\` resolve a registration from a non-UTF-8 file and write an \`mcp_server_source\` route, while \`load_mcp_server_source\` uses strict \`load_text_file\`, records \`unreadable_file\`, and may enumerate zero tools. That breaks the promised \`detect -> init -> scan\` agreement. Please decode strictly here and handle the failure consistently with the adapter, with an invalid-UTF-8 regression.

src/agents_shipgate/inputs/mcp_idioms.py
pengfei-threemoonslab: Fixed in 34804c5d — you were right, and the failure is worse than a wrong string: the real action goes missing while an action id nobody serves takes its place in the catalog.

Split by language rather than patched. `decode_literal(body, language)` dispatches to `_decode_go` (`\a`, `\nnn` as exactly three octal digits ≤ 255, `\xHH`, `\uHHHH`, `\UHHHHHHHH`) and `_decode_typescript` (`\0` only when not followed by a digit, `\xHH`, `\uHHHH`, `\u{…}` bounded at 10FFFF, line continuation, `NonEscapeCharacter`). Both take your second option too: **anything the grammar does not define returns `None`**, which reaches `_resolve_name` as `name_not_literal`, so the site becomes a recorded omission rather than a guessed name. TypeScript legacy octal is deliberately in that bucket — whether `\137` is octal depends on a strictness mode this reader does not track, so it refuses instead of picking.

`_Masker` now carries its language so the same contract applies to Go interpreted strings, Go runes, and TS template literals alike. Go raw strings are untouched, since they process no escapes at all.

Adversarial sweep gains five cases, including your exact one:

- `MustTool("delete\137all", …)` → `delete_all`
- `MustTool("\x61bc", …)` → `abc`
- `MustTool("delete\qall", …)` → omission
- `MustTool("delete\13", …)` (truncated octal) → omission
- the same `\137` in TypeScript → omission

plus a direct `decode_literal` table over eight undecodable forms. A perturbation that points Go at the JavaScript decoder, and one that guesses instead of refusing, are both caught.

src/agents_shipgate/inputs/mcp_idioms.py
pengfei-threemoonslab: Fixed in 34804c5d. Confirmed your example first: `if (ok) /\.registerTool("fake", handler)/.test(value);` did report a resolved `fake` tool, so the module was violating the guarantee in its own docstring.

Took the "track control-statement context" option rather than failing closed. Failing closed on every ambiguous slash was not affordable — `)` before `/` is overwhelmingly `foo(a) / 2`, and treating those as regex would blank real code on any line with a second slash, trading an invented tool for a lost one.

`_opens_regex` now resolves the ambiguity instead of assuming: when the previous significant character is `)`, it walks back to the matching `(` and reads the word in front of it. `if`, `for`, `while`, `switch`, `catch`, `with` → the slash begins the body, so it is a regex. Anything else → division. `_previous_significant` returns the position alongside the character so the matching `(` can be found in the already-masked buffer, where parens inside strings and comments are gone.

Three sweep cases: a regex opening an `if` body, one opening a `while` body, and `const ratio = total(a) / scale(b) / 2;` to pin that ordinary division is untouched. A perturbation that restores the always-division reading is caught.

src/agents_shipgate/cli/discovery/mcp_source.py
pengfei-threemoonslab: Fixed in 34804c5d. This one I had flagged to myself as a trade-off and got wrong: I reasoned that "detect refuses multi-project scopes, so an export in this workspace is an export for this server" — but this PR *also* collapses a multi-server monorepo into one route, so the two facts together are exactly the hole you describe. Both of your cases lose real actions.

Took the first option: bind the export to the surface it displaces. `_covering_export` now loads each accepted export through the real MCP adapter and withholds the source route only when the exported names **contain every tool the source route resolved**. Location is no longer the test, and neither is existence.

Where an export exists but does not contain the source surface, both routes are suggested and the shortfall is named in the detection evidence — `An MCP tool export is also present and does not name N of these registrations (…); both routes are suggested, and a reviewed tool_identity binding is what joins the two surfaces`. That is your second option for the residue, and it is the right one: two sources describing one server are reconciled by a reviewed binding (#386), never by dropping one of them.

A wildcard export claims a surface without enumerating it, so it can never be shown to contain anything and never displaces the source route.

Four regressions:

- **two servers** — `servers/alpha` commits an export, `servers/beta` is source-only; `beta_write` survives, `excluded` is empty, and the evidence names the shortfall
- **partial export** — one server, export names `update_incident`, source also registers `delete_incident`; the remainder survives
- **complete export** — unchanged, still withheld, reason now reads `names every one of these N registrations`
- **wildcard export** — never displaces

Perturbations restoring either "any export wins" or "no export wins" are both caught.

src/agents_shipgate/cli/discovery/mcp_source.py
pengfei-threemoonslab: Fixed in 34804c5d. Discovery now reads through the adapter's own `load_text_file` and treats `InputParseError` the way the adapter does — the file contributes no names. Sharing `is_scannable_path` between the two was supposed to guarantee the `detect → init → scan` agreement, and leaving the *read* unshared undid it one layer down.

Worth recording how the regression test had to be built, because my first attempt passed for the wrong reason. Putting the invalid bytes *inside the tool name* cannot see the difference: with `errors="replace"` the name becomes replacement characters, fails `TOOL_NAME_RE`, and is unresolved — the same outcome as a strict refusal, so both readers agree and the test is vacuous. The bytes have to sit in a **comment** with an ordinary registration beside them:

```go
package tools
// \xff\xfe
var T = mcpgrafana.MustTool("broken_file", d, h)
```

Lenient decoding resolves `broken_file`; strict refuses the file and enumerates nothing from it. The test asserts the two tool sets are equal, that `broken_file` is in neither, and that the adapter records `unreadable_file`. A perturbation restoring `errors="replace"` is caught — it was not, before the fixture was rebuilt this way.

PR #486, opened 2026-09-01
Add side-effect-contained local review flow

### LINKED ISSUE #326: feat(init/verify): add a side-effect-free local-review mode for external PRs

pengfei-threemoonslab:
## Summary

- add `init --local-review` for temporary external-repository assessment without tracked worktree changes
- keep the ephemeral manifest and reports private through scoped `.git/info/exclude` entries, with structured side-effect and recovery metadata
- bind Git-proven manifest provenance into verifier artifacts and fail closed through the existing release-decision engine
- document temporary review separately from durable `init --write` adoption

Closes #326

## Type

- [ ] Input adapter change
- [ ] Documentation only

## Verification

Additional local checks run:

- `PYTHONPATH=src pytest -q --ignore=tests/test_packaging.py`
- `PYTHONPATH=src pytest tests/test_packaging.py -q`
- `PYTHONPATH=src ruff check .`
- `python -m compileall -q src tests`
- `PYTHONPATH=src python scripts/generate_schemas.py`
- `PYTHONPATH=src python scripts/build-llms-full.py`
- committed `origin/main...HEAD` Shipgate verification: `control_state=complete`, `decision=passed`, `merge_verdict=mergeable`

## Release-readiness notes

- [x] New or changed check IDs are documented in `docs/checks.md`

## Surface discipline

- Primary metric: reduce time-to-first-verdict and increase activation for external PR reviewers.
- Existing surface reused: this extends `init` and `verify`; `SHIP-VERIFY-LOCAL-REVIEW-PROVISIONAL` is an ordinary finding in the single `release_decision` path, not a second verdict.
- Non-goals preserved: no agent execution, no tool calls, no LLM calls, no runtime claims, no network access in static scanning, and no automatic human-authority declarations.


pengfei-threemoonslab (commented):
## Review: `Add side-effect-contained local review flow` (#326)

Reviewed at `52fa9ec`, base `c516437` (one commit behind `main` @ `6f4252d`; no file overlap, so no rebase needed for correctness). Everything below was reproduced locally against this branch; the full suite is green (`pytest -q --ignore=tests/test_packaging.py` → exit 0), `ruff check` passes, and `generate_schemas.py` / `build-llms-full.py` produce no diff.

**Verdict: changes requested.** The security core is genuinely sound and I could not break it (details in *What holds up* at the end). What I can't sign off on is the **routing change that rides along with it**: `verify --preview` now sends *every* unconfigured repository to `init --local-review`, and the resulting state is a closed loop that no envelope-following agent can exit. Plus four fail-open edges where the provenance signal is defaulted, hardcoded, or absent.

---

## Blocking

### B1 — `verify --preview` routes every unconfigured repo to `--local-review`, and preview cannot see that it ran

[`orchestrator.py:4986`](https://github.com/ThreeMoonsLab/agents-shipgate/blob/52fa9ec/src/agents_shipgate/cli/verify/orchestrator.py#L4986) hardcodes `--local-review` in `_preview_init_command`, with no condition. Every preview test flipped accordingly (`test_discovery_scope.py` ×4, `test_verify.py` ×1), and `test_preview_routes_unconfigured_repo_to_local_review` asserts `"--write" not in command`.

Preview's "already adopted" probe looks for `config_path.name` (`shipgate.yaml`). `.agents-shipgate-local-review.yaml` is invisible to it, so preview never observes that its own recommended step completed:

```
$ shipgate init --workspace $R --local-review --json     # writes the manifest, exit 0
$ shipgate verify --preview --workspace $R --format control
{"control_state": "agent_action_required",
 "next_action": {"kind": "initialize",
   "command": "shipgate init --workspace $R --local-review --json",
   "why": "This unconfigured workspace looks agent-related; create a side-effect-contained local-review manifest before verifying."}}
```

Re-running that command yields `manifest_status: skipped_existing`. The route's `expects` can never be satisfied by the route itself — the mirror image of the #323 class ("a route whose `expects` already holds cannot advance"), and it lands on the default first-run path for every new user.

This is also a product change well past #326's scope: `docs/use-cases/ai-generated-agent-prs.md` — the flagship AI-generated-PR recipe — replaced `init --write` with `init --local-review`, so the headline documented workflow now produces a run that can never return `passed`/`mergeable`. #326 asked for an *explicit* non-adoption mode, not for adoption to stop being the default route.

**Ask:** keep `--write` as preview's route and reach `--local-review` from an explicit signal (a `--local-review` flag on preview, or an "external repository" determination). If preview does route there, it must recognise an existing local-review manifest as configured.

### B2 — the documented `--local-review` + `--base/--head` recipe cannot work

`docs/use-cases/ai-generated-agent-prs.md` now prescribes:

```bash
agents-shipgate init --workspace . --local-review --json
agents-shipgate verify --config .agents-shipgate-local-review.yaml \
  --base origin/main --head HEAD --json
```

The second command always fails:

```
Config error: Head tree 'HEAD' does not contain .agents-shipgate-local-review.yaml.
```

The manifest is by design never committed, and archive/`--head` mode reads the manifest out of the head tree. So `--local-review` is confined to `--no-base` worktree mode — which produces no base-vs-head capability delta, i.e. exactly the evidence that reviewing someone else's PR needs. On my fixture that path returns `insufficient_evidence` with `Capability lock diff unavailable: no --base ref was provided.`

Worse, the error's own `next_action` is `verify --preview`, which (per B1) routes back to `init --local-review`, which is already done. The loop closes.

**Ask:** either let the head archive be scanned under a worktree-resident manifest, or state plainly in the docs that local-review supports worktree comparison only — and then say what evidence it can actually produce.

### B3 — `scan --changed-files` with an ephemeral manifest still returns `passed`

```
$ shipgate scan -c .agents-shipgate-local-review.yaml --changed-files changed.txt
$ jq '.release_decision.decision, [.findings[].check_id]' agents-shipgate-reports/report.json
"passed"
[]
```

[`_register_scan.py:88`](https://github.com/ThreeMoonsLab/agents-shipgate/blob/52fa9ec/src/agents_shipgate/cli/_register_scan.py#L88) builds `VerificationContext(changed_files=paths, diff_text_available=False)` and the new field defaults to `"repository"`, so `verify_local_review.run` returns `[]` on the very path `docs/checks.md:971` documents as supplying a verification context to the `SHIP-VERIFY-*` family. `report.json` is the machine contract (per CLAUDE.md § Output handling), and it is clean.

This contradicts acceptance criterion 7 ("a local/ephemeral manifest cannot produce release-authoritative `passed` … merely because it is structurally valid").

Root cause is two fail-open defaults, both worth changing on principle:

- `VerificationContext.manifest_provenance: Literal[...] = "repository"` — any construction site that forgets the field gets release authority.
- `manifest_provenance(path, *, committed_at_head: bool | None = True)` — the default asserts the strongest claim the function can make.

Both should default to `"unknown"` / be required, so a forgotten call site fails closed. That inverts the polarity of this whole feature.

### B4 — `local_review.reports.git_visibility` is a hardcoded string and is false on the refusal path

`local_review_side_effects()` emits `"git_visibility": "excluded"` as a constant, never derived from whether an exclude block was written. On the ambiguous-monorepo refusal — which is *the* external-repository case — nothing is written and the claim is false:

```
manifest_status:      refused_unresolved_scope
manifest on disk:     False
exclude block written: False
local_review.reports: {"git_visibility": "excluded", "created_by_init": false, ...}
side_effect kinds:    ["manifest"]        # no git_private_exclude entry at all
```

The payload is self-inconsistent: it claims exclusion while listing no exclusion effect. A caller that trusts it will scan and leave `agents-shipgate-reports/` as untracked files in someone else's PR — the exact harm #326 exists to prevent, and acceptance criteria 2 and 3.

**Ask:** derive `git_visibility` from the outcome, and omit (or explicitly negate) the `local_review` block when setup refused.

### B5 — `--local-review` failures emit no structured error at all

Both `ensure_local_review_excludes` call sites do `except Exception as exc: raise typer.BadParameter(str(exc))`, which bypasses `_emit_agent_mode_error_routing`:

```
$ AGENTS_SHIPGATE_AGENT_MODE=1 shipgate init --workspace $NOGIT --local-review --json
Usage: shipgate init [OPTIONS]
╭─ Error ─────────────────────────────────────────────────╮
│ Invalid value: --local-review requires a Git checkout   │
╰─────────────────────────────────────────────────────────╯
exit 2, zero JSON
```

Every sibling `init` error path emits `{"error", "message", "next_action", "next_actions", "control"}` (verified with `--control-pack nope`). This is the one route with no machine exit — and it is exactly where preview lands an agent on a non-Git checkout (tarball, vendored source, `pip download`). The bare `except Exception` also converts genuine internal errors into usage errors.

### B6 — the control envelope has no route to durable adoption

`--write` appears nowhere in any control envelope this feature produces. After `init --local-review`, `control.next_action` is a human review of the placeholders and `allowed_next_commands` is `null`. After `verify`:

```json
"next_action": {"actor": "human", "kind": "review",
                "why": "1 finding(s) require human review."},
"allowed_next_commands": ["… verify --config .agents-shipgate-local-review.yaml --no-base --json"]
```

The only permitted next command is a re-run of the identical verify, which reproduces `SHIP-VERIFY-LOCAL-REVIEW-PROVISIONAL` forever. `fix_task.allowed_repairs` names adoption only in prose (`command: null`). The escape hatch exists solely as `local_review.durable_adoption.command` in the *init* payload — invisible to an envelope-only caller, which is the documented contract.

This is the "every published next step must be able to change the answer" rule from the PR #399 review, and the `insufficient_evidence` human-wall from the 2026-07-26 adoption review, reintroduced.

### B7 — the agent-facing instruction surfaces now contradict the CLI

- `CLAUDE.md` § Permissions still reads: *"`scan`, `preflight`, `init`, … are **read-only** with respect to user code; safe to run without confirmation"*, with `init --write` called out separately as the writing mode. `init --local-review` writes a manifest **and** mutates `.git/info/exclude` without `--write`. An agent following the checked-in rules will run it unconfirmed on an unfamiliar repo.
- `AGENTS.md` (the canonical agent instructions per CLAUDE.md), the three byte-identical `SKILL.md` bundles (`skills/`, `plugins/claude-code/skills/`, `adoption-kits/claude-code-skill/`), `.agents/skills/agents-shipgate/`, and `decide-shipgate-relevance.md` ×3 all still say `init --write` — while the CLI now emits `--local-review`. `grep -rn "local-review" AGENTS.md .claude/ .agents/` is empty.

Per the bundled-prompt checklist this also means bumping `EXPECTED_*_RENDER_SHA256` and appending the outgoing hash to `prior_render_sha256`. Either update these surfaces or (preferably, per B1) don't change the routing.

---

## Should fix

### S1 — linked worktrees share one `.git/info/exclude`

`git_path(workspace, "info/exclude")` runs `git rev-parse --git-path info/exclude`, which resolves to the **common** dir in a linked worktree, while the anchored `/…` patterns apply relative to *each* worktree's top level. Reproduced:

```
$ git worktree add -b feature $LINKED     # from $MAIN
$ shipgate init --workspace $LINKED --local-review --json
   exclude path: $MAIN/.git/info/exclude
   entries:      ["/.agents-shipgate-local-review.yaml", "/agents-shipgate-reports/"]
$ touch $MAIN/.agents-shipgate-local-review.yaml && git -C $MAIN status --porcelain -uall
   (empty — a real file in the main checkout is now invisible)
```

And because `block_id = sha256(workspace_relative)[:16]` is computed against the *worktree's* root, a second run from `$MAIN` collides with the first and reports `status: "unchanged", changed: false` — so the block is owned by nobody, and whichever worktree cleans up first blinds the other. Worth folding the worktree identity (or `$GIT_COMMON_DIR` vs `$GIT_DIR`) into `block_id`, and saying in the JSON that the scope is repository-wide, not worktree-local. Relevant here specifically because this project develops in worktrees.

### S2 — an absent manifest is reported as `uncommitted`, and the note displaces the real cause

```
$ shipgate verify --workspace $R --config missing.yaml --no-base --json
"base_notes": ["Manifest provenance: uncommitted. Conclusions are provisional
   static-assessment output; Git must prove a committed, human-reviewed
   repository trust root before this run can carry release authority."]
"diff_status": {"detail": "verify stopped at the missing manifest.", ...}
```

The *only* `base_notes` entry is a provenance claim about a file that does not exist. `_build_verifier` attaches the note before the missing-manifest and refs-missing early exits, and `path_committed_at_head` returns `False` for "absent" and "present but untracked" alike. Absence reported as a defect of the thing is the class closed in #384/#387/#389. Give `ManifestProvenance` a distinct `absent` state, and skip the note on the early-exit paths.

### S3 — the exclude-block version is baked into the pattern that searches for it

`_upsert_block` builds its regex from `f"# agents-shipgate:local-review:start v={LOCAL_REVIEW_EXCLUDE_VERSION} id={block_id}"`. When the version becomes `2`, the `v=1` block will not be found: a second block is appended and the first is orphaned forever, invisible to `git status`. The version field defeats its own purpose.

The existing, hardened module gets this right — `gitignore_block.py:56` is `rb"^# agents-shipgate:start v=(\d+)\r?$"`, with explicit version handling and an "Upgrade the CLI" refusal. More broadly, `local_review.py` is a second hand-rolled managed-block writer (markers, CRLF detection, ambiguity refusal, `to_json()`) beside one that already exists and is tested. CONTRIBUTING.md § Surface discipline argues for extending `gitignore_block` with a target path + entry list rather than adding a parallel implementation; the recurring "second implementation" bug class from #322 is exactly this shape.

### S4 — provenance is derived three times from three different path spellings

- `run_verify` → `config_path=config_relative`, `committed_snapshot=archive_head` (drives the finding)
- `_build_verifier` → `config_path=config_path` (absolute), `committed_snapshot=not worktree` (drives the note)
- `_write_artifacts` → `config_path=Path(logical_config)`, `committed_snapshot=archived_head` (drives `plan.inputs.options.manifest_provenance`)

`logical_config` falls back to `_display_path(...)` when the config sits outside `resolved_input_root`, and `_configured_manifest_provenance` maps an unrelatable absolute path to `"unknown"`. The plan option is the sole gate on `_evaluate_authorization_overlay` (`authorization_requires_committed_repository_manifest`), so a disagreement would silently strip authorization capability with nothing in the report explaining it. I probed a scoped subdirectory layout and all three agreed (`repository`), so I have **no reproducer** — but this is the value-join-across-surfaces shape from the #403 review. Derive it once and thread it through.

### S5 — `init --local-review` silently shadows a committed trust root

On a repo with a committed `shipgate.yaml`, `init --local-review` writes `.agents-shipgate-local-review.yaml` beside it with no warning that a reviewed policy already exists. The ephemeral manifest can declare a weaker policy and produce a report that looks like a run of that repository. `SHIP-VERIFY-LOCAL-REVIEW-PROVISIONAL` marks it, but the operator was never told there was a real trust root to use instead. Refuse, or route to the committed manifest.

### S6 — non-atomic write to a file Shipgate does not own

`exclude_path.write_bytes(updated)` truncates in place. Interrupted, the user's personal exclude file is left corrupt and the only backup is `LocalReviewExcludeOutcome._previous_bytes` in memory. Write to a temp file in the same directory and `os.replace`. Separately, `_validate_exclude_target`'s `is_symlink()` check followed by `write_bytes()` is TOCTOU; `os.open(..., O_NOFOLLOW)` closes it.

### S7 — no executable cleanup

The JSON gives recovery *actions* and *paths* but no command, and nothing retires the block or the manifest. After the reviewer moves on — or adopts durably — a stale ephemeral policy and a `/agents-shipgate-reports/` exclusion persist per-user, invisible to `git status`. An `init --local-review --undo` (or having `init --write` retire the block) would close it; this repo's own idiom is to publish exact commands.

---

## Nits

- `_literal_gitignore_path`: the `startswith(("#", "!"))` and trailing-space branches are unreachable — every value starts with `/` and ends with the manifest name or `/`.
- `rollback_local_review_excludes` reads `outcome._previous_bytes` / `._previous_existed` from module scope. If they're part of the contract, they aren't private.
- The human-readable `init --local-review` output names no removal path and no block id; only `--json` carries them.
- **~450 of the ~1,527 changed lines are unrelated `ruff format` reflow.** `ruff format --diff` on the merge-base versions of the five touched files reports 452 changed lines, and CI runs `ruff check` only — not `ruff format --check`. It buries the security-relevant hunks (`orchestrator.py`: 312 changed lines, ~22 feature-related) and will conflict with anything else touching that file. Please split it out; if the team wants formatting enforced, that's its own PR plus a CI step.

---

## What holds up

Worth saying explicitly, because the hard part is right:

- **Suppression immunity and the severity floor both hold.** `checks.ignore` on `SHIP-VERIFY-LOCAL-REVIEW-PROVISIONAL` from inside the ephemeral manifest leaves it unsuppressed (`suppressed: false`, still `review_required`), and `severity_overrides: low` is refused at load with the floor message. An attacker-controlled ephemeral policy cannot switch off its own marker.
- **Reserved-name precedence survives force-adding.** A committed `.agents-shipgate-local-review.yaml` still classifies as `local_review`; only the rename to `shipgate.yaml` (the explicit adoption gesture) changes it.
- **Archive mode cannot be fed an untracked manifest** — `Head tree 'X' does not contain …` — so the `committed_snapshot=True` short-circuit in `_configured_manifest_provenance` is safe.
- **The downgrade is real and measured.** Same sample, committed manifest → `passed` / `mergeable` / `can_merge_without_human: true`; renamed to the reserved path → `review_required` / `human_review_required` / `false`.
- **The worktree really does stay clean** through `init --local-review` + `verify` (`git status --porcelain -uall` empty), including generated reports.
- `_evaluate_authorization_overlay` fails closed on a missing `manifest_provenance` option, so an older plan degrades to `not_applicable` rather than authorising.

One framing point for the docs: the code proves *"this path is a blob at HEAD"*, while the prose in `verify.yaml`, `checks.md`, and `LOCAL_REVIEW_PROVISIONAL_NOTE` says *"committed, human-reviewed repository trust root"*. A local `git commit` satisfies the guard and not the sentence. Worth saying what is actually proven, per the rung-prose-is-a-claim rule from the #410 work.

---

### Summary of asks

1. Revert the unconditional preview → `--local-review` routing (B1) and the `ai-generated-agent-prs.md` recipe swap; make local-review reachable from an explicit signal.
2. Make local-review usable with `--base/--head`, or document its worktree-only limit honestly (B2).
3. Compute provenance on the `scan --changed-files` path; flip both defaults to fail closed (B3).
4. Derive `git_visibility` from the actual outcome; drop/negate the `local_review` block on refusal (B4).
5. Route local-review failures through `_emit_agent_mode_error_routing` (B5).
6. Publish a route to `init --write` in the control envelope (B6).
7. Update `CLAUDE.md`/`AGENTS.md`/the three SKILL.md bundles + render hashes, or don't change the routing (B7).
8. Split the `ruff format` churn into its own PR.

pengfei-threemoonslab:
Addressed the review in `1852e569` and rebased the branch onto current `main`.

Blocking items:

- **B1/B7:** restored durable `init --write` as the ordinary preview route and restored the flagship adoption recipe. Local review is now explicit by previewing with `--config .agents-shipgate-local-review.yaml`; once that manifest exists, the same explicit preview routes to verify rather than looping. Because the default workflow is unchanged, the protected agent-instruction trust roots did not need to be edited.
- **B2:** base/head verification now captures the identity-bound worktree manifest once and overlays those bytes only into the archived head tree. The documented external-review flow now exercises `--base origin/main --head HEAD`.
- **B3:** `scan --changed-files` derives actual manifest provenance, and both `VerificationContext` and the provenance helper fail closed (`unknown`/required inputs rather than repository authority).
- **B4:** refusal output no longer publishes a local-review side-effect block; Git visibility is derived from whether an exclusion outcome actually exists.
- **B5:** local-review validation and filesystem failures use the shared structured agent-mode error/control route. Broad exception handling was narrowed to `ConfigError`/`OSError` where appropriate.
- **B6:** provisional verifier control and `fix_task` now publish one human-owned, exact durable-adoption command (`init --write`) instead of an identical verify loop.

Should-fix and cleanup items:

- Refuse linked worktrees because their common `info/exclude` cannot provide checkout-local isolation.
- Added explicit `absent` provenance and suppress provenance notes on early missing-input paths.
- Marker parsing is version-independent, migrates older blocks, and refuses blocks from a newer CLI.
- Manifest provenance is derived once per verify run and threaded through scan, verifier, plan, and authorization evaluation.
- Refuse local review when `shipgate.yaml` already exists.
- Private-exclude updates are same-directory atomic replacements with `O_NOFOLLOW`, regular-file/link-count checks, identity rechecks, mode preservation, and exact-byte rollback.
- Added `init --local-review --undo --json`; it removes only the reserved manifest and this managed block, preserves reports, and is emitted in JSON and human output with the block id.
- Removed the unreachable escaping branches, made rollback state public-but-nonserialized, and removed the unrelated formatter churn from the five files called out in the review.
- Tightened wording so Git blob presence is not described as proof of human review.

Validation:

- 444 focused verifier/setup/provenance regressions passed.
- CI-equivalent suite: 6,994 passed, 7 skipped; 89.96% coverage. The isolated wheel fixture initially could not download `hatchling` under restricted networking, then all 9 packaging tests passed with network access.
- Static-only adapter trust canary passed (including updated line pins for the unchanged audited Git subprocess sites).
- `ruff check .`, compileall, schema generation/check, docs generation, and `git diff --check` passed.
- Agents Shipgate verifier: `control_state=complete`, `decision=passed`, `merge_verdict=mergeable`.

PR #487, opened 2026-09-01
Publish the capability delta as a standalone in-toto attestation (#470)

### LINKED ISSUE #470: Publish the capability delta as a standalone attestation: an in-toto predicate any consumer can verify

pengfei-threemoonslab:
Closes #470.

`verify` already computed the capability delta and bound it to a receipt; its only consumers were our own renderers. This publishes it as a **standalone, versioned attestation** so an external tool can read *"what can the agent do after this change"* without running Agents Shipgate.

## What ships

| | |
|---|---|
| Predicate type | `https://threemoonslab.com/agents-shipgate/capability-delta/v1` |
| Predicate body | `shipgate.capability_delta_attestation/v1` |
| Payload | [`shipgate.capability_payload/v1`](docs/capability-payload.md), the `delta` view, **verbatim** |
| Artifact | `agents-shipgate-reports/capability-delta-attestation.json` |
| Spec | [`docs/capability-delta-attestation.md`](docs/capability-delta-attestation.md) |
| Reference consumer | [`tools/verify-capability-delta.py`](tools/verify-capability-delta.py) |

Runtime contract **27 → 28** registers the predicate type, both schema versions, both schema paths and the artifact. That is the half #469 deliberately declined: freezing the payload while nothing emitted it was right, and advertising a schema no artifact carries would have been a contract bump with nothing behind it. Both halves land here.

`minimum_control_contract_version` stays at `21`; `AgentControl` and `shipgate.agent_control/v1` are byte-identical to v27.

## The three design decisions worth reviewing

**The subject is the reviewed tree, and the payload has to agree.** The statement names exactly one subject whose `digest.gitTree` must equal `predicate.delta.head.ref`; `base.ref` is the base tree. Without that join, four edited characters of the subject would move a valid delta onto a commit it never reviewed. The payload spec calls `ref` an opaque caller label — this surface narrows it to a git object id and enforces it.

Consequence: **only a committed-tree subject is attestable.** A run that evaluated a worktree snapshot scanned bytes that are in no tree object, so it withholds the file and appends a note to `verifier.base_notes[]`. The GitHub Action always passes an explicit `--head`, so CI is unaffected. This is the same condition the human-authorization path already requires.

**Two identities in the receipt chain, not five.** `predicate.verification` carries `input_set_id` and `subject_id` and its `status` is `bound` or `unbound` — only `bound` may name identities, the fail-closed shape `analysis_coverage` uses. `request_id`, `engine_requirement_id` and `decision_id` are deliberately *not* published: they mix in the engine build, the Python version and the platform, so an interchange format carrying them would emit different bytes for an identical review and would leak the builder's machine. They stay in `verification-receipt.json`, which `input_set_id` is the join key into.

**The digest set follows in-toto, not this schema.** Everywhere else here absence is spelled as a value; `subject[].digest` is in-toto's `DigestSet`, typed `map<string, string>`, so an uncomputed `gitCommit` is **absent**, never `null`. The published JSON Schema says so too — Pydantic's nullable form would have told an external validator to accept a document this format never produces.

## One projection, proven

`project_capability_delta_attestation` calls `project_capability_delta`; the predicate embeds its output unchanged. `diff_capability_locks` — what the PR comment renders — is the same `diff_capability_fact_sets` engine, so the capability-change count on both surfaces is one number by construction. Two tests hold it: the attested delta is re-projected byte-for-byte from the two capability locks the same run publishes, and its counts are compared against the PR comment's own printed line.

## Nothing is silently absent (#437)

`analysis_coverage` is populated **per side** from the run's own conservation law — the observed catalog minus the analysed subjects — so a tool that arrives and is never bound reaches the attestation as a *named* `newly_outside_analysis` row. A side the run could not establish (a `--diff-from` report, a committed base lock, an unreadable base catalog) publishes `unavailable`, which is not a claim that nothing was left out. A catalog row with no canonical id makes the whole side unavailable rather than quietly shortening it.

Deriving those rows needed the catalog spelling of `provider` to match the capability layer's, so `_normalize_token` moved to one shared `surface_exclusions.provider_token` — the same two-spellings-of-one-subject hazard `catalog_label_index` already documents.

## The reference verifier

Stdlib-only, one file, importing nothing of ours. It applies 20 published rules (the envelope's `E1`–`E9` plus the payload's own stage-two `P1`–`P11`), and independently re-implements the payload's semantics — the rank tables, the dimension comparison, the direction rollup. Verified against tampering:

| Tamper | Caught by |
|---|---|
| subject relabelled onto another tree | `E7` |
| summary inflated | `P4` |
| a subject row dropped | `P4` |
| subject key forged | `P1` |
| coverage named without looking | `P7` + `P8` |
| receipt binding forged | `E9` |
| **effect silently escalated to `destructive`** | `P6` |
| consumer's expected commit ≠ attested | `--expect-tree` |

A perturbation test moves each published dimension in turn and compares the script's derivation against the package's, so the two implementations cannot drift. The docs table of rule ids is pinned to the script's own `CHECKS`.

The statement is emitted **unsigned** in v1 and the docs say so plainly: a passing run establishes self-consistency and that the subject is the state the delta describes, not authorship.

## Acceptance

- [x] Attestation emitted for the shipped sample, committed as a generated golden under the `--check` drift gate (the #425 pattern, and the same `samples/ai_generated_refund_pr` fixture the #469 payload examples use). Its subject tree ids are real `git write-tree` output, and they match the ones a live `verify` on that fixture produces.
- [x] Reference verifier validates a good attestation and rejects a tampered payload and a mismatched subject.
- [x] Value-equality: delta in the attestation ≡ delta in the PR comment for the same run.
- [x] A tool added but unbound appears per the subject-based rules, named rather than counted.
- [x] Docs page with the predicate URI, schema version, and a copy-paste consumer example.
- [x] Emitted on a tagged release install — `tests/test_packaging.py` runs the projection against the extracted wheel alone, cwd outside the repository.

## Two integration points this uncovered

`_AUTHORIZATION_RECEIPT_ARTIFACTS` is deny-by-default, so a new verify artifact made every authorized push exit 3 until it was allowed there. It is now keyed off one constant and pinned by a test that asserts every artifact a verify run binds is inside the closure — the omission was found by an unrelated failing test, because nothing compared the two lists.

The attestation joined `VERIFIER_ROUTE_ARTIFACT_NAMES`. It names an `input_set_id`, so a stale one left beside a later run's receipt would offer a chain into a verification that is no longer current.

## Surface discipline

Headline metric: **independent external consumers of the format**. It extends the one decision engine — the attestation is a projection of the existing delta and introduces no second verdict, no severity and no release impact. `release_decision.decision` remains the only release gate. Non-goals respected: no runtime enforcement, no service, no new judgment path.

## Verification

`ruff check .`, `generate_schemas.py --check`, the full suite (`-n auto -m "not perf"`), `test_adapter_static_only.py`, and the perf suite all pass.


pengfei-threemoonslab (commented):
**Blocking review findings** (posted as a comment review because GitHub does not allow a PR author to submit `REQUEST_CHANGES` on the same PR).

I found two contract-level defects in the new standalone consumer, plus a robustness gap. The documented gate-input path can currently accept invalid safety data or claim a receipt binding it has not checked.

1. **[P1] Validate the closed safety vocabulary on the default stdlib path** (`tools/verify-capability-delta.py:752-777`). The membership-change branch returns before validating the carried record. Reproduction:

   `jq '.predicate.delta.subjects[0].changes[0].after.effect.effect = "harmless"' docs/examples/capability-delta-attestation.v1.example.json | python tools/verify-capability-delta.py - --json`

   This returns exit 0 with `"valid": true` and `"stage_one": "not_requested"`, even though `harmless` is outside the closed v1 effect enum. The documented 30-second command and gate example omit `--schema`, so an external policy consumer can accept an out-of-contract safety value. The same type gap lets `summary.added_subjects = true` satisfy P4 because Python equates `True` and `1`. Please enforce the structural types/enums needed by the safety rules in the stdlib verifier, or make equivalent schema validation unavoidable before `valid` can be true.

2. **[P1] Make `--require-receipt-binding` verify an actual receipt** (`tools/verify-capability-delta.py:1152-1159`). This option only checks the self-declared `verification.status` string. It never reads `verification-receipt.json`, compares `input_set_id`/`subject_id`, or verifies that the receipt manifest binds the attestation digest. Any statement can set `status: "bound"` plus two syntactically valid content IDs and pass. Because the consumer example presents this flag as the gate check for the receipt chain, please add a receipt input and validate the identity and artifact-manifest join, or rename/document the option as shape-only and do not use it as proof of binding.

3. **[P2] Keep every parseable JSON document on the rule-result path** (`tools/verify-capability-delta.py:1131-1155`). The new malformed-input tests cover six top-level shapes, but nested malformed values still escape as tracebacks. For example `predicate = 5` with `--require-receipt-binding` raises `AttributeError`; `analysis_coverage.base.status = []` raises `TypeError`; and `subjects[0].changes = true` raises `TypeError`. A mutation pass over the shipped example found 30 such crashing cases. This contradicts the documented exit/JSON contract and the promise that malformed documents fail with rule rows. Please validate before set membership/iteration and make the optional consumer checks total as well.

Verification notes: focused attestation/schema/CLI tests and `ruff` plus `generate_schemas.py --check` pass locally. PR-mode Agents Shipgate verification completes with `control_state=review_publishable`, `decision=review_required`, and no blockers; the human-review route is caused by the `AGENTS.md` trust-root change. GitHub CI is still not green: all 6,904 tests passed in 857.15s, but the `test` job was cancelled at the 15-minute limit while uploading a 2.232 MB step summary. The latest test-cost commit therefore has not yet restored a successful required check.

PR #488, opened 2026-09-01
Add action declaration review evidence

### LINKED ISSUE #428: PR comment: render the declaration change as a review surface (#410 §D follow-up)
## What is missing

§D of #410 asks the PR comment to render the declaration change as the review surface:

> the PR comment renders the declaration diff as the review surface — ✓ rows machine-verified as evidence-consistent, ⚠ rows for overrides. The reviewer reads exceptions, not 117 rows; that is what makes attestation real rather than rubber-stamped.

Half of that already ships. `acknowledged_overrides[]` (#409) puts every ⚠ row in the comment naming the action, both readings, and the reviewer's evidence and reason. #427 adds the loop that *produces* declaration changes — an agent can now write the evidence-consistent ones into `shipgate.yaml` and push them — so the volume the ✓ half exists to absorb is about to arrive.

What is not there is the ✓ half: a reviewer opening a PR that adds twelve `action_surface.actions` rows has no statement of which ones Shipgate itself agrees with, so they either read all twelve or none.

## Why it was not folded into #427

It needs two things that PR does not have, and each is a decision rather than a line of code.

**A base-vs-head manifest comparison.** "Which declarations did this PR add or change?" is a diff question. The report carries `capability_change` and `action_surface_diff`, but neither is keyed on *declaration* rows — and a PR that changes an effect from `read` to `financial_write` and one that adds the row fresh are different review objects.

**A third bucket, because two are not honest.** A declaration is not ✓ or ⚠; it is one of:

- **evidence-consistent** — the scan observed readings for this action and the declared effect covers all of them. Machine-verified; a reviewer can skip it.
- **unverified** — the scan observed nothing about this action, so the declaration is a human claim that nothing corroborates *or* contradicts. Rendering this as ✓ would be the strongest possible misreading: it is the row that most needs a person, and it looks identical to a verified one from the outside.
- **acknowledged override** — the declaration sits below inferred evidence and carries `override: {evidence, reason}`. Already rendered.

Collapsing the first two is what would make the surface rubber-stampable, which is the thing §D says it exists to prevent.

## Sketch

- A projection over `action_surface_facts` × `evidence_gaps` classifying every action carrying a reviewed effect declaration into the three buckets above (the inputs are all present; `declaration_below_inferred_evidence` and the effect readings already distinguish the cases).
- Keyed to the rows this PR *changed*, so an adopted repository with 117 declared actions does not restate all of them on every PR.
- Rendered in the `capability-review` PR comment beside the existing override rows, and in packet §1, from one rendering function — a value-join across two surfaces rendered twice is how they start disagreeing (#403 review).

## Acceptance

- A PR adding declarations shows counts per bucket and names the rows in the two that need a person, never the evidence-consistent ones.
- An action with no observed readings is never rendered as machine-verified.
- The comment says nothing new when the PR changes no declaration.

Follow-up to #427; part of #410 §D.

pengfei-threemoonslab:
## Summary

- close #428 by diffing base/head action declarations into evidence-consistent, unverified, and acknowledged-override rows
- include added, semantically modified, and removed declarations; removed rows always require review
- fail closed when base comparison, action identity, selector resolution, or semantic evidence is incomplete
- render one shared, privacy-safe declaration-review projection across report JSON/Markdown, packet JSON/Markdown/HTML, PR comments, annotations, and the GitHub summary
- publish report v0.43, packet v0.18, verifier v0.16, and runtime contract v29; preserve frozen-reader compatibility and the outgoing v28 local-contract render hash

## Scope and lineage

#431 and #326 are already implemented by merged PRs #483 and #486. This branch is rebased on the mainline implementations and now carries no diff for their MCP reader, local-review flow, or the unrelated preflight-version repair. The remaining verifier context addition is part of #428: it distinguishes a genuinely introduced gate when constructing declaration-comparison evidence.

## Validation

- full non-performance pytest suite passed under xdist, excluding only the separately run packaging/static/performance modules and the two protected public-surface files
- `tests/test_packaging.py`: 10 passed with an isolated wheel build
- static-only adapter invariant and P0 safety/binding/policy/identity/agent-boundary canaries: passed
- latency budget: 4 passed
- focused declaration-review, continuation, renderer, contract, mainline local-review, and mainline MCP regression suites: passed
- Ruff, compileall, dependency-lock verification, schema regeneration check, generated `llms-full.txt`, and `git diff --check`: passed
- committed-tree Agents Shipgate verification at `e9bd6eaf`: `control.state=complete`, `decision=passed`, `merge_verdict=mergeable`, committed capability-delta attestation emitted

## Protected-surface review required

Repository policy forbids coding agents from editing `AGENTS.md` and any `SKILL.md`. Nine intentionally surfaced contract-test failures remain until a human reviewer updates:

1. `AGENTS.md`: report v0.42 → v0.43, packet v0.17 → v0.18, anchor v017 → v018.
2. `skills/agents-shipgate/SKILL.md`: the same three updates.
3. Its byte-identical mirrors:
   - `adoption-kits/claude-code-skill/SKILL.md`
   - `plugins/claude-code/skills/agents-shipgate/SKILL.md`
4. The expected Claude Code skill render SHA-256 pin.

Then rerun:

```bash
pytest -q tests/test_public_surface_contract.py tests/test_docs_links.py
```

Closes #428


pengfei-threemoonslab:
CI completed with `6961 passed, 7 skipped, 9 failed`. All nine failures are the protected public-surface synchronization already called out in the PR description; no implementation or packaging failures remain.

Human-owned edits required by repository policy:

- `AGENTS.md`: update current report schema v0.42 → v0.43, packet schema v0.17 → v0.18, and packet anchor v017 → v018
- `skills/agents-shipgate/SKILL.md`: make the same report/packet/anchor updates
- synchronize the protected rendered skill copies through the repository's human review workflow

After those reviewed edits, rerun:

```bash
pytest -q tests/test_public_surface_contract.py tests/test_docs_links.py
```

The coding-agent-owned portion is otherwise green: Shipgate `verify` and `verify-self`, clean-checkout and Windows launchers, schema generation, lint/compile, the broad suite excluding only this protected set, and all 9 isolated packaging tests.

pengfei-threemoonslab:
## Expert review — `#488`

Reviewed at the PR head (`e4d4202e`), base current with `main` (`6f4252d9`). I ran the CLI and the test suite locally rather than reading only the diff, and reproduced each finding below before reporting it.

**Recommendation: request changes.** The engineering quality here is high and a lot of this is genuinely well built — I've listed what I verified as correct under each section, and it is substantial. But there are three P0s reachable from untrusted input, and a set of release-integrity problems that sit outside the diff.

### Blockers at a glance

| # | Area | Defect |
|---|---|---|
| **P0** | #326 local review | An ephemeral, git-invisible manifest grants `edit`/`commit`/`push`/`update_pr` that the committed trust root denies |
| **P0** | #431 MCP reader | A 15 KB `.ts` file crashes `scan` with `RecursionError` (exit 4, no report) |
| **P0** | #431 MCP reader | Quadratic recognizer — byte caps bound I/O, not runtime; ~344 s for one 1 MiB file |
| **P1** | #431 MCP reader | The `operationType` half of the feature is rejected by the evidence engine — inert |
| **P1** | #428 declarations | A *removed* declaration is invisible on every surface |
| **P1** | #428 declarations | "Could not compare" renders identically to "nothing changed" |
| **P1** | release | `CONTRACT_VERSION = "28"` collides with open PR #487 |
| — | process | #483 and #486 are open and solve two of the three issues this PR closes |

---

### 1. Two of the three issues already have open PRs, and the contract version collides with a third

| PR | Closes | Size | State |
|---|---|---|---|
| **#488** (this) | #326, #428, #431 | +32,711 / −1,056, 145 files | open |
| **#483** | **#431** | +4,273 / −24, 23 files | open |
| **#486** | **#326** | +1,186 / −341, 22 files | open, `DIRTY` |
| **#487** | #470 | — | open |

#483 and #486 are open right now, by the same author, created the same day, and they solve two of the three issues this PR closes. Merging #488 silently orphans both. That decision should be made deliberately, not by merge order.

Worth noting the designs actually diverge, so this is not a pure duplicate: #483 adds a new `tool_sources[].type: mcp_server_source` (70 references), whereas #488 extends the existing `mcp` type with an optional `idiom` enum. **#488's choice is the better one** under [CONTRIBUTING.md § Surface discipline](CONTRIBUTING.md#surface-discipline) — extend the one engine rather than add a parallel surface. That is an argument for keeping #488's approach and closing #483, but it is an argument someone should make explicitly.

**The hard blocker is #487.** Both PRs mint `CONTRACT_VERSION = "28"` from `"27"`:

```
#487: -CONTRACT_VERSION: Literal["27"] = "27"   →  +CONTRACT_VERSION: Literal["28"] = "28"   (in-toto capability delta)
#488: -CONTRACT_VERSION: Literal["27"] = "27"   →  +CONTRACT_VERSION: Literal["28"] = "28"   (typed manifest provenance)
```

Whichever lands second ships a **second, divergent definition of contract v28**, and `.well-known/agents-shipgate.json`, `docs/agent-contract-current.md`, and every downstream `.shipgate/agent-contract.json` will describe a v28 that half of consumers do not have. This is the same shape as the earlier two-divergent-v0.27-schemas incident. Whichever merges second must move to 29 and re-append its `local_contract.PRIOR_RENDER_SHA256`.

---

### 2. The PR is designed to land red, and the human handoff is understated

I reproduced the intentional failures locally — the count and attribution are exactly as the description claims:

```
9 failed — all in AGENTS.md and skills/agents-shipgate/SKILL.md
(7 × test_public_surface_contract.py, 2 × test_docs_links.py)
```

No other test fails locally. That is a genuinely clean result for a 145-file change, and the honesty about it is appreciated.

Two corrections to the handoff, though:

**a) It is more than two files.** `skills/agents-shipgate/SKILL.md` has two byte-identical mirrors:

```
skills/agents-shipgate/SKILL.md
adoption-kits/claude-code-skill/SKILL.md
plugins/claude-code/skills/agents-shipgate/SKILL.md      # all three: md5 d85ae301cc5185055b7e81fa6794e9a3
```

`tests/test_claude_code_plugin_package.py::test_plugin_skill_is_byte_identical_to_canonical` enforces that. A human who edits only the canonical file trades 9 red tests for a different red test, plus the `EXPECTED_CLAUDE_CODE_SKILL_RENDER_SHA256` pin. The handoff note should name all three files and the render-hash pin.

**b) `main` goes red on merge.** CI runs on `push: branches: [main]`, and these failures are in the main `test` step — not an advisory job. Between merge and the human follow-up, `main` is red.

---

### 3. CI did not actually finish, so several gates never ran for this PR

The `test` check is red, but it is *also* a timeout: the job hit the 15-minute cap at the moment pytest printed its summary.

```
X test in 15m16s   —  "The job has exceeded the maximum execution time of 15m0s"
Test step: 20:41:02 → 20:55:29  (14m27s)
```

Everything after `Test` was skipped:

```
- Latency budget          - Build package        - A released install reports itself as one
- Dependency audit        - Generate SBOM        - Upload SBOM
```

So the **perf/latency budget never ran** on a PR that adds ~1,300 lines of regex-driven TypeScript/Go source scanning, and packaging/SBOM/dependency-audit are unverified. `main` already runs this job at 9–14m, so the margin was thin before this PR — but this PR is what tips it over. Please split the suite or raise `timeout-minutes` before relying on a green run here.

Related: `$GITHUB_STEP_SUMMARY upload aborted, supports content up to 1024k, got 2259k` — the summary is now more than double the limit and is being silently dropped.
---

### 4. Cross-cutting findings I verified by running the tool

#### P1 — `scan` has no reserved-manifest guard, so it emits an unmarked "Release Evidence Packet" from an ephemeral manifest

`verify` gets this exactly right and fails closed:

```
$ shipgate verify --workspace . --config .agents-shipgate-local-review.yaml --format json
Config error: The reserved local-review manifest cannot be used through ordinary --config.
Re-run with --local-review so its worktree binding and non-authoritative provenance are validated.
```

`scan`, pointed at the same file, does not:

```
$ shipgate scan --config .agents-shipgate-local-review.yaml
   scan exit=0
$ head -1 agents-shipgate-reports/packet.md
# Release Evidence Packet
$ grep -c "ephemeral\|release_authoritative\|non-authoritative" agents-shipgate-reports/packet.json
0
```

`is_reserved_local_review_manifest` (`src/agents_shipgate/cli/local_review.py:476`) is called from only two places — `cli/verify/command.py:315` and `cli/verify/orchestrator.py:301`. `scan` is not one of them.

The result is a shareable artifact literally titled *Release Evidence Packet*, produced from a manifest whose whole point is that it is non-authoritative, carrying **zero** ephemeral marking. The only trace is the manifest filename appearing inside a path string. That is the exact failure mode #326 names: *"This feature must not make local assertions look like reviewed release evidence."*

Structurally this is also why `manifest_provenance` is threaded through the verifier chain (`verifier`, `verify-run`, `plan`, `receipt`, `current-control`, `agent-handoff`) but reaches neither `report.json` nor the packet — I checked, no `manifest_provenance` reference exists in `schemas/report.py`, `schemas/packet.py`, or `packet/`.

Mitigating, and why this is P1 rather than P0: no published `next_action` or doc routes a user to `scan --config <the reserved manifest>`; you have to type it. And the verifier chain itself is solid (see below).

Suggested fix: call the same guard from the `scan` entry point. That single change also resolves the next finding.

#### P1 — the ephemeral manifest declares the *durable* output directory, so `--cleanup` cannot clean up after itself

The generated manifest and the exclude block disagree about where reports go:

```
$ grep -A1 '^output:' .agents-shipgate-local-review.yaml
output:
  directory: agents-shipgate-reports          # ← the durable adoption default

$ grep agents-shipgate .git/info/exclude | grep -v '^#'
/.agents-shipgate-local-review-reports/       # ← what is actually hidden
/.agents-shipgate-local-review.yaml
```

`local_review.py` defines `LOCAL_REVIEW_REPORTS_NAME = ".agents-shipgate-local-review-reports"` (line 29) and threads it into the binding metadata and the exclude patterns, but never writes it into the generated manifest's `output.directory`. So the exclude block hides a directory nothing writes to, and anything that honours the manifest's own output setting writes to an unexcluded one:

```
$ shipgate scan --config .agents-shipgate-local-review.yaml && git status --porcelain
?? agents-shipgate-reports/

$ shipgate init --workspace . --local-review --cleanup --json   # exit 0
$ git status --porcelain
?? agents-shipgate-reports/                   # still there — 6 files, incl. packet.{md,json,html}
```

This breaks acceptance criterion 2 ("Generated reports do not appear as untracked files") and defeats the recovery path the JSON advertises. For a feature whose entire purpose is assessing *someone else's* PR without contaminating their worktree, leaving `packet.json` behind after an advertised cleanup is the failure that matters.

To be clear about scope: **the supported path is fine.** `verify --local-review` writes into the excluded directory and leaves porcelain clean. The bug is in the manifest the generator writes, and it only bites via the unguarded `scan`.

#### P2 — `STABILITY.md` now states a version the runtime does not emit

```
$ git diff origin/main...pr-488 -- STABILITY.md | grep preflight_schema_version
-- `preflight_schema_version` — currently `"0.3"`.
+- `preflight_schema_version` — currently `"0.4"`.

$ python -c "...PreflightResultV3.model_fields['preflight_schema_version'].default"
0.3
```

`docs/preflight-schema.v0.4.json` is a pre-existing misnamed orphan whose own const is `"0.3"`; the generator emits `preflight-schema.v0.3.json` and never validates the `.v0.4` file, so `generate_schemas.py --check` cannot catch this. The PR aligned the document that opens *"This document is the contract"* to the wrong artifact. Revert to `"0.3"` and handle the orphan separately.

#### P2 — stale version claims left in files that are *not* protected

The AGENTS.md / SKILL.md carve-out is legitimate, but these are neither, and they went stale in this PR:

- `prompts/fix-top-finding.md:17` — says `contract v26 / report v0.42`, and its **3 identical copies** under `skills/`, `adoption-kits/claude-code-skill/`, `plugins/claude-code/skills/agents-shipgate/`. (`v26` was already stale on `main`; `report v0.42` is new drift from this PR.)
- `docs/errors.json:13,51` — `(contract v27)`, correct on `main`, false now.
- `benchmark/safety-qualification/README.md:36` — documents the `0.42` pin while `schemas/safety_qualification.py` and `scripts/_release_support.py` both moved to `0.43`.

None of these is bound by a test, which is why the build is green on them and why they will drift again next cycle. A guard asserting any `contract v<N>` / `report v0.<N>` literal in these files equals the runtime constant would close the class.

Note also that `AGENTS.md:121` lists protected surfaces as "… and `SKILL.md`" — unqualified. Under a literal reading that covers the adoption-kit and plugin mirrors too, which is worth settling explicitly since it changes who is allowed to fix them.
---

### 5. Local review (#326) — the authority denial has a hole

The design is right and most of it is genuinely well built (see the verified list below). But the single denial rule does not cover the permissions that matter for an agent loop.

#### P0 — an ephemeral, git-invisible manifest grants `edit`/`commit`/`push`/`update_pr` that the repository gate denies

`schemas/verifier.py:1181-1189` checks `control.state == "complete"`, `completion_allowed`, `permissions.merge`, `permissions.report_complete` — and nothing else. It does not cover `permissions.edit/commit/push/update_pr`, `must_stop`, or `control.state == "review_publishable"`.

Those four are exactly the fields `PublishOnlyPermissions` (`schemas/agent_control.py:239`) pins to `Literal[True]` while leaving `merge`/`report_complete` `Literal[False]` — so a `review_publishable` local-review run sails through the validator with full progress authority. The model's own docstring names the stakes: *"`edit`/`commit`/`push`/`update_pr` are progress actions: they publish the change so a human can review it."*

A/B on the same repo, same commit, `git status` clean throughout:

```
LOCAL-REVIEW  decision=review_required  state=review_publishable   must_stop=False
              perms={'edit': True,  'commit': True,  'push': True,  'update_pr': True,
                     'merge': False, 'report_complete': False}

REPOSITORY    decision=blocked         state=human_review_required must_stop=True
              perms={'edit': False, 'commit': False, 'push': False, 'update_pr': False,
                     'merge': False, 'report_complete': False}
--- porcelain after the local-review run: (clean) ---
```

The aggravating factors are what make this more than theoretical:

- `init --local-review` and `verify --local-review` **do not refuse or even mention** an existing committed `shipgate.yaml`. Exit 0, no warning; `config` just says `.agents-shipgate-local-review.yaml`.
- The manifest is git-ignored by design, so the repo's own reward-hacking defences (`SHIP-VERIFY-TRUST-ROOT-TOUCHED`, manifest-committed checks, porcelain invariance) can never see it being tuned. `orchestrator.py:946,955` additionally force `configured_gate_introduced = manifest_introduced = False` for local review, so the run never reports that its gate is uncommitted.
- The authorized next command really is `apply-patches … --apply`, giving a complete self-contained edit → re-verify loop that git never observes.
- **No provenance anywhere in the control surface.** `control` carries no provenance field, `control.reason` says only "1 finding(s) require human review", and `current-control.json` — the artifact `read_current_control` consumes — contains zero occurrences of `ephemeral`/`local_review`/`provisional` while carrying `lifecycle_state: terminal`, `state: review_publishable`, and `commit/edit/push/update_pr: true`. (`packet.json` and `capabilities.lock.json` carry no `manifest_provenance` either.)

An agent that can write a file the repository cannot see can therefore grant itself the publish authority the committed trust root refused. For a product whose thesis is that agents must not approve their own permissions, this is the load-bearing case.

Fix: add the four progress permissions and `must_stop` to the denial; put the rule in **one** shared function instead of the four divergent copies (`verifier.py:1169` 9 conditions, `verify_run.py:320` 7, `agent_handoff.py:430` 8, `verification_identity.py:503` 3 — a mutation sweep shows only the `verifier.py` copy is covered by any test); surface `manifest_provenance` in `AgentControl`/`current-control.json`; and refuse, or loudly route, `--local-review` when a committed manifest already covers the workspace.

#### P1 — the invariant is a `raise`, not a downgrade, so it surfaces as exit 4 "Internal error"

Nothing in the orchestrator clamps the decision for non-authoritative provenance; the only thing keeping local review off that path is the *behavioural* `SHIP-VERIFY-EPHEMERAL-MANIFEST` finding downgrading the decision. Drop that one check and a would-be-passing run gives the user:

```
EXIT: 4
Internal error: 1 validation error for VerifierArtifact
  Value error, non-authoritative manifest provenance cannot carry a passed release decision
next_action: "Re-run with --verbose for a stack trace, then file an issue…"
```

And a second construction site already reaches it: `orchestrator.py:813-835` hard-codes `head_status="skipped"`, ignoring the `head_status="not_run" if local_review else "skipped"` guard added 70 lines earlier at line 757 — `merge_verdict_for(decision=None, execution="skipped")` returns `mergeable`, and the validator raises "cannot carry merge authority". It appears unreachable through today's CLI only because `TRIGGER-EXISTING-MANIFEST-PRESENT` force-runs — which also makes the line-757 special case dead code and the divergence invisible. Same "guard applied to one of two sites" class as the earlier exit-4 crash.

Clamp in the orchestrator and keep the model validator as the assertion of last resort.

#### P1 — a hostile `.gitignore` turns reports into untracked files and deadlocks `--cleanup` permanently

Ignore-effectiveness is proved once, inside `provision_local_review`; `verify` never re-checks, and `cleanup_local_review` asserts porcelain *invariance*, which is the wrong direction once the paths are visible. Reviewer inits on `main`, then checks out a PR branch whose `.gitignore` re-includes the reserved paths (`.gitignore` outranks `.git/info/exclude`):

```
porcelain BEFORE verify: ?? .agents-shipgate-local-review.yaml  (+ reports owner)
verify exit=2
porcelain AFTER verify:  11 untracked entries — verifier.json, verify-run.json,
                         agent-handoff.json, current-control.json, pr-comment.md, …

1st cleanup: exit=2  "Local-review cleanup changed Git porcelain status"
2nd cleanup: exit=2  identical
rebind:      exit=2  "the bound reports directory is missing"
```

Manifest, binding and exclude block are stuck in the worktree with no working CLI recovery — on a branch supplied by the PR under review. (The guard *is* right at provision time: a hostile `.gitignore` present before init fails closed with a complete rollback.) Fix: re-prove ignore-effectiveness at the start of `verify --local-review`, and make cleanup assert "porcelain must not gain entries" rather than invariance.

#### P1 — rebind is impossible once the generated reports directory is gone, and the error names the command that just failed

No hostile input needed:

```
$ shipgate init --workspace $R --local-review --json   # exit 0
$ rm -rf $R/.agents-shipgate-local-review-reports      # generated output
$ shipgate init --workspace $R --local-review --json   # exit 2
Config error: Local-review metadata recovery required: the bound reports directory is missing.
  …Then run `… init --local-review --json` to create and authenticate a fresh binding.
```

`provision_local_review` (`local_review.py:157-161`) passes `require_reports_owner=True`, routing to `_metadata_recovery_error` whose only remedy is the command that just failed. This is also the state a failed cleanup leaves behind — the code's own comment says reports "may already be gone after a failed cleanup".

#### P2 — a crash mid-write leaves an unexcluded temp file

`local_review.py:1002` uses `tempfile.mkstemp(prefix=f".{path.name}.", dir=path.parent)`, but the exclude patterns cover only the two final names. Killing the process before `os.replace` leaves `?? ..agents-shipgate-local-review.yaml.bkwnuze8` untracked — never reported as a side effect, never cleaned. (Ctrl-C is safe; the `finally` unlink runs.)

#### P2 — `verify --preview` silently stopped publishing artifacts

`orchestrator.py:5494-5497` moves preview into a transient `TemporaryDirectory` and `:5903` strips `artifacts` to `{}`. On `main`, preview always resolved a durable `out_dir`. Any caller reading `verifier.artifacts["verifier_json"]` or the preview's `current-control.json` breaks with no deprecation and no CHANGELOG entry — `STABILITY.md` treats stable JSON fields as contract.

#### P2 — the reserved-name guard is wired into `verify` only

Beyond the `scan` gap noted above, `preflight`, `doctor`, `baseline save -c` and `apply-patches` all accept the reserved manifest with no provenance. The `orchestrator.py:301` copy is itself vacuous — the CLI copy catches everything first, so the defence-in-depth layer is untested. (The name check is not bypassable through the obvious routes: case-variants and symlinks are both refused by pre-existing `--config` identity guards.)

**Verified correct here — and this part is strong:** binding authentication fails closed under every tamper tried (rewriting `workspace` with a recomputed `binding_id` → *"binding does not belong to this worktree"*; `release_authoritative: true` → rejected by `extra="forbid"`; editing the manifest after binding → *"manifest changed after its binding was written"*, with a rebind→cleanup recovery that restores the worktree byte-exact). Forging a binding gains nothing — `ManifestProvenance` is derived from `kind`, never read from the file. Attacker-controlled worktree content cannot claim authority: committed reserved manifest, committed file in the reserved reports dir, symlinked reports dir, hardlinked manifest, case-variant `--config` are all refused, and `checks.ignore: SHIP-VERIFY-EPHEMERAL-MANIFEST` inside the local-review manifest is ignored. Provenance is computed once and carried, with re-authentication at the orchestration boundary. Nested workspaces and linked worktrees both work, with `.git/info/exclude` restored byte-exact in either cleanup order. `init --write` is genuinely unchanged. `tests/test_local_review.py`: 37 passed.

A mutation sweep did find **28 guards that no test in the suite asserts**, including the manifest-sha256 rebind check, the tracked-path collision check, the `binding_id` content-address check, and the `--local-review` + `--head` / `--authorization` refusals. I verified by hand that most behave correctly today — the risk is that nothing stops a refactor from removing them.
---

### 6. MCP source-code discovery (#431) — two P0s reachable from the PR under review

The scanned source *is* the pull request being reviewed, so anything here is attacker-controlled input. I reproduced both of these myself.

#### P0 — a 15 KB file crashes the scan; `RecursionError` escapes the per-file guard

`inputs/mcp_code.py:1227` `_backtick_literal_end` ↔ `:1251` `_typescript_template_expression_end` recurse on nested `${` template literals with no depth bound. The same pair exists at `tools/shipgate-detect.py:1875/1893`.

```
$ python -c "open('src/evil.ts','w').write('const x = ' + '`${'*3000 + '`y`' + '}`'*3000 + ';')"
   input bytes: 15014
$ python -c "list(mcp_code._tokens(text, language='ts'))"
   *** RecursionError RAISED in _tokens ***
```

The per-file guard at `mcp_code.py:232` catches `(OSError, UnicodeDecodeError, ValueError)` — `RecursionError` is a subclass of none of them, and it is raised while tokenizing, not while reading, so it escapes either way. End to end this surfaces as `scan` exit 4, `Internal error: maximum recursion depth exceeded`, no report produced. A PR can deterministically deny the verifier a verdict — a fail-open by crash on the gate itself.

Fix: make the pair iterative with an explicit nesting cap, and add `RecursionError` to the per-file catch.

#### P0 — quadratic recognizer: the byte caps bound I/O, not runtime

`_class_bodies` (`:901`), `_property_literal_initializer` (`:934`) and `_matching` (`:945`) each start a fresh linear scan per matched token. Measured:

```
class x 2000  (  12,000 B):  _class_bodies   0.05 s
class x 8000  (  48,000 B):  _class_bodies   0.72 s
class x 32000 ( 192,000 B):  _class_bodies  11.55 s      <- 4x input, 16x time
```

Clean O(n²). The caps are `MAX_MCP_CODE_FILE_BYTES = 1 MiB` and `MAX_MCP_CODE_TOTAL_BYTES = 32 MiB`, so a single conforming file extrapolates to ~344 s (the reviewer measured 202 s on a real 1.04 MB file), and 32 MiB of such content is hours. Two TS idioms run over every TS file, once in the adapter and again in discovery.

Fix: one pre-pass recording brace/paren match positions per token stream, or bound each inner scan by a token window.

#### P1 — the `operationType` half of #431 is inert at the engine

This is the finding I'd most want checked, because it is the feature's headline. #431 option B asks for *name and operation class*. The names land; the operation classes are rejected.

`mcp_code.py:383-385` emits `source="mcp_code_operation_type"`, `basis="typed_provider_fact"`. `core/semantic_assessment.py:2369-2382` accepts `typed_provider_fact` from exactly two sources:

```python
if hint.basis == "typed_provider_fact":
    if hint.source == "anthropic_client_tool_type" and ...:
        return "typed_provider_fact"
    if hint.source == "n8n_static" and ...:
        return "typed_provider_fact"
```

`mcp_code_operation_type` is not among them, so every hint is downgraded to `basis="unknown"`, `policy_eligible=False`, **and** raises `invalid_evidence_provenance` on every tool. On a mongodb-shaped fixture:

```
drop-database | effect destructive | status protocol_default | conf low
  issues: invalid_evidence_provenance "risk hint 'destructive' has no typed evidence basis"
  claim: destructive / medium / unknown / eligible False
```

`tests/test_mcp_code.py` asserts the tuple `("destructive", "typed_provider_fact", "medium")` and passes — the emitted string is tested, the engine's acceptance of it is not. That is the "test passing for the wrong reason" shape.

#### P1 — `tool_set_proven: true` on files only partly read

`mcp_code.py:309,327` derive `surface`/`tool_set_proven` from "did a shape I matched fail", never from "is there something here I did not match":

```ts
function reg(s, name, cb) { s.tool(name, cb); }
reg(server, "drop-database", h);
server.tool("find", h);
```
```
tools = ['find']   omissions = []
extraction = {"surface": "enumerated", "tool_set_proven": true, "confidence": "medium"}
```

`drop-database` is dropped with **no ledger entry**, while the adapter asserts the tool set is proven. That is precisely what #403 exists to prevent: excluded silently meaning safe. Same for `go_addtool_v1`/`go_musttool_v1`. Relatedly, `extraction["surface"]="partial"` is read by nothing for `source_type="mcp"` — `semantic_assessment.py:2336` returns `True` unconditionally because `"mcp"` is not in `_AST_ONLY_SOURCE_TYPES`, so the adapter's surface-honesty channel is dead at the completeness ceiling.

#### P1 — discovery aborts the whole survey on the first oversized file

`cli/discovery/signals.py:3192-3212` uses `return` where the adapter's equivalent (`mcp_code.py:201-213`) uses `continue` — two implementations of the same bound that disagree. With a 1.44 MB bundle beside a real tool file:

```
suggested_sources []
excluded [{'reason_code': 'walk_capped', ...did not read aaa_bundle.js...}]
next_action "Re-run detect with --workspace pointed at the MCP server project
             rather than the enclosing monorepo."
```

The workspace *is* the MCP server project; following the instruction reproduces the result. A bundled `.js` over 1 MB is routine in exactly the repos this feature targets, so discovery will often return nothing.

#### P1 — `dist/`, `build/`, `__tests__/` are scanned, and the printed remedy cannot be followed

`_SKIP_PARTS` (`mcp_code.py:48`) omits `dist`, `build`, `out`, `lib`, `__tests__`, `coverage`, `examples`. Build output enters the catalog as real tools with the wrong effect (`delete-many` from `build/out.js` read as `op='read'`). End to end the duplicate is caught fail-closed, but the remedy says to "bring both files under a single `shipgate.yaml` tool_sources entry" — they already are one entry, and `ToolSourceConfig` has no `exclude`/`ignore` field, so no manifest edit changes the answer. That breaks the #399 rule that every published next step must be able to change the answer.

#### P1 — the two detectors disagree on `static` at line start

`mcp_code.py:919` requires the modifier on the same line; `tools/shipgate-detect.py:2094` has no such constraint. On identical input the zero-install detector reports the source and the installed adapter extracts nothing, recording an omission that misnames the shape. Also flagged: nested classes (zero-install counts 3, installed 2) and `.d.ts` handling. The same-line rule is additionally vacuous — the `;`/`{`/`}` stop already covers the donor case its comment cites — so deleting the clause fixes the divergence and a dead guard together.

#### P2 — unknown `operationType` fails open to no effect and no ledger entry

`_operation_risk_hints` returns `[]` for any value outside the closed set, so `operationType: "nuke"` is indistinguishable from a tool declaring nothing: no hint, no warning, no omission. An observed-but-not-understood claim should be recorded, not discarded.

#### P2 — the determinism-boundary matrix now publishes a false negative

`inputs/mcp.py:517` was not touched, so `docs/determinism-boundary.md:70` still ships `literal_registration … not_applicable` with the prose *"An export is a contract file, not source."* That is now false — `load_mcp_tools` dispatches to `load_mcp_code_tools`, which reads literal registrations out of source at `medium`. `mcp_tool_snapshot_v1` is unlisted too. This is the artifact #473 shipped specifically to state what shipgate can prove.

**Verified correct here:** comments, strings, template and regex literals never produce tools; the untrusted-annotation whitelist holds (a source declaring `agent_bindings: true` / `shipgate_trusted: true` gets nothing through `_annotation_bools`) — no repeat of the #268 critical; duplicate identity is fail-closed with both paths named; a self-declared effect downgrade (`delete` → `read` on `drop-database`) surfaces as `RISK_TAG_ADDED` high with before/after; the exclusion ledger names unbound tools; Go idioms behave as designed (`dynamic_tool_name`, `unresolved_tool_registration`, definition-without-registration correctly not emitted); binary files become `unreadable_source_file` omissions; `walk_input_tree` is snapshot-bound and does not follow directory symlinks; and the idiom registry is genuinely closed — `ToolSourceConfig.idiom` is a `Literal` restricted to `type: mcp`, so a PR cannot redefine the evidence used to review it.

A mutation sweep also found 10 guards no test kills, including both halves of `_test_or_vendor_path` — the directory/filename skip whose absence P1 above shows is already live.
---

### 7. Declaration-review surface (#428) — four confirmed correctness gaps

The three-bucket design is right, and the most important invariant genuinely holds: **an action with no observed readings can never be rendered as machine-verified.** It is enforced twice independently — `declaration_surface.py:421` returns `unverified`, and `DeclarationReviewRow.bucket_matches_evidence` (`schemas/report.py:930-957`) rejects such a row at the model. Mutating either half kills tests. That is the acceptance criterion #428 cares most about, and it is solid.

These four sit around it.

#### P1 — a *removed* declaration is invisible on every surface

`core/lenses/declaration_surface.py:281` iterates `sorted(head_by_id)` only. A `row_id` in base but absent in head is never added to `changed`, so deleting a reviewed declaration yields `changed_count: 0`, and all four renderers stay silent (`report/declaration_review.py:40` gates on `changed_count`).

Confirmed with a base manifest declaring `send_email: effect: read` plus an `override:{evidence,reason}`, and a head that deletes the row:

```
base had override rows: 1
head review: enabled=True  changed=0
   {'evidence_consistent': 0, 'unverified': 0, 'acknowledged_override': 0}
PR comment mentions declaration: False
```

The same report's `action_surface_diff` shows `actions_modified: 0`, so nothing else attributes it either. Where the underlying evidence is strong enough that no gap opens, deleting a human's reviewed answer — including a previously acknowledged override — is completely silent. Removing a declaration is exactly the review object #428 exists to surface.

Fix: iterate `set(head_by_id) | set(base_by_id)` and emit a `change_type: "removed"` row classified from the *base* declaration, always in an attention bucket (a removal can never be machine-verified).

#### P1 — "could not compare" renders identically to "nothing changed"

`declaration_surface.py:263-269` writes an explanatory note when no trustworthy base snapshot exists; `report/declaration_review.py:40` never reads `notes`. So `enabled=false` produces byte-identical output to a PR that changed no declaration.

Confirmed with a base report from a build predating `action_declaration_facts` (any released 0.4x is accepted by `_reference_from_report_payload`), where the head *adds* `send_email: effect: read` on an action the scan reads as `external_communication`:

```
REVIEW: {"enabled": false, "notes": ["No trustworthy base declaration snapshot was available; ..."]}
packet.md    'Declaration changes' present=False ; note present=False
packet.html  'Declaration changes' present=False ; note present=False
pr-comment   'Declaration changes' present=False ; note present=False
```

The run is otherwise healthy and green. Reachable today via `--diff-from <stored CI base report>` — the cache-epoch bump at `cli/verify/orchestrator.py:231` covers only the auto-base cache, not `--diff-from`.

This is the repo's own monotonicity rule: absence of a base is not evidence of no change. The fix pattern already exists a few files away — `cli/scan/sanitization.py:188-208` gives `BindingSurfaceDiff` a `base_comparison_requested` flag *and* rendered notes precisely so "requested but unavailable" cannot read as "not requested". `DeclarationReviewDecision` has neither.

#### P1 — identity-dimension gaps do not block the ✓

`declaration_surface.py:41-49` — `_EFFECT_GAP_KINDS` omits `unattested_surface`, `conflicting_tool_identity`, `ambiguous_tool_selector`, and `incomplete_tool_identity`. Those gaps carry `subject_kind="action"` and `subject_id=tool.id` but live on the *identity* dimension, so `semantic_assessment.effect.issues` is empty for them too — both halves of the guard at `declaration_surface.py:409` miss. With one gap on the changed row's exact subject and otherwise-covered readings:

```
unattested_surface           -> evidence_consistent
conflicting_tool_identity    -> evidence_consistent
ambiguous_tool_selector      -> evidence_consistent
incomplete_tool_identity     -> evidence_consistent
```

`evidence_consistent` asserts the scan observed readings and the declaration covers all of them — but `conflicting_tool_identity` says the scan cannot prove *which tool* those readings belong to. The row is then counted as machine-verified and, by design, never named. `unattested_surface` is the sharpest case: `ci/release_decision.py:1723` and `:2316` both treat it as one class with `incomplete_surface`, and only `incomplete_surface` made it into the set.

Fix: add the four kinds, or gate on `gap.subject_id == row.subject_id` for any gap on the identity-or-effect dimension rather than an allowlist of effect kinds.

#### P2 — the override line drops the manifest location, and the two spellings disagree

`report/declaration_review.py:174` returns only `details` — unlike `_unverified_line` it never appends `Review {manifest_path}`, so an override row gives the reviewer nothing to open. Worse, the row has the right path and the override it renders has a different, wrong one:

```
review row manifest_path : services/billing/shipgate.yaml#action_surface.actions[0]
override   manifest_path : shipgate.yaml#action_surface.actions[tool='send_email'].override
```

The second is hardcoded at `ci/release_decision.py:1535`. Pre-existing, but #428 is the first surface that joins the two spellings of the same file into one row — and both are in `report.json`.

#### P2 — a second, hand-rolled rendering of the same join

`report/pr_projection.py:85,175-200` still builds annotations / Checks-API items for `acknowledged_overrides` with its own message text, unaware of `declaration_review`. It knows nothing about the `unverified` bucket — the one #428 says *must* be named — so on the annotation surface an uncorroborated declaration produces no row at all, while overrides get a second differently-worded rendering. `report/markdown.py` and `ci/github_summary.py` carry no declaration-review line at all. #428 asked for one rendering function; there are now two.

#### P2 — guards no test covers

A 24-mutation sweep against `tests/test_declaration_review.py` (harness validated — `if base is None or True` kills 3 tests) left these alive:

| Mutation | Location | Consequence |
|---|---|---|
| drop the `effect.issues` guard | `declaration_surface.py:409` | a `declaration_drift` row becomes `evidence_consistent` |
| drop `row_char_limit` skip | `declaration_review.py:110` | a 500-char row renders in full; masked today because the one test uses an 8000-char row |
| drop `override_is_represented` dedupe | `packet/markdown.py:289`, `packet/html.py:300` | the same override renders twice in packet §1 |
| drop public-collision `ConfigError` | `declaration_surface.py:141-151`, `:232-239` | the whole collision machinery is untested |

Also dead: `declaration_review.py:169` (`if not row.acknowledged_overrides` is unreachable — the model rejects that row first), plus unused `unavailable_note` (`:251`) and `side` (`:496`) parameters.

**Verified correct here:** privacy redaction reaches every egress (a secret in `override.evidence` is absent from pr-comment, packet.md/html, report.json and packet.json, and the row degrades to `unverified` with an explicit indeterminate-after-redaction reason); no display cap reaches the machine route (`report.json`/`packet.json` carry every row, and the omission-count string is reserved before each acceptance so the disclosure can never be erased); all repository-derived values pass through `display_literal` then the surface escaper; and the `action_semantics`/`semantic_assessment` deletions are a **pure move** — all four tables are AST-identical to `main` and `declaration_covers` is behaviorally identical over all 144 effect pairs.
---

### 8. Summary of what I verified as correct

Verified by running it, not by reading it:

- **Enforcing the ephemeral-authority denial on the model is the right layer** — `VerifierArtifact._ephemeral_manifest_cannot_authorize_release` (`schemas/verifier.py:1169`) sits where an artifact cannot be constructed around it. (The defect in §5 is its *field set* and the fact that it was copied four times, not the choice of layer.)
- **`init --local-review` keeps porcelain clean** and reports every side effect with a working recovery command and byte-precise markers for the managed exclude block.
- **The reserved manifest cannot be laundered through `verify --config`** — fail-closed with an error that names the correct route.
- **The schema bump is purely additive across all seven families.** Zero removed `$defs`, zero removed properties, zero dropped `required`, zero narrowed enums. New frozen readers (`VerificationPlanV1`, `VerificationReceiptV1`, `VerifyRunArtifactV5`, `VerificationInputSetV1`) use `extra="forbid"` and dispatch on an immutable `schema_version` discriminator that *raises* on unknown — no try-v2-then-fall-back fail-open. `STABILITY.md`'s no-hard-removal rule is respected.
- **The sample goldens do not repeat the earlier `<REPO>`-before-hashing bug.** Independently recomputed: `samples/conductor_agent/expected/report.json` hashes to the pointer's `sha256` at the pointer's `size_bytes`, post-substitution, and `current_control_id` re-derives exactly.
- **`generate_schemas.py --check` is not self-certifying** — injecting a probe field into `ReadinessReport` changes the generated bytes, so it really does regenerate from the models.
- **`prior_render_sha256` was appended correctly**, and `local_contract` is the only renderer whose output changed.
- **Identity digests are deterministic** across CWD, `LC_ALL`, `PYTHONHASHSEED`, `TZ`, and CRLF conversion; no absolute path enters any identity.
- **`schemas/action_effects.py` is a move, not a fork** — `core/semantic_assessment.py` deletes its local tables and re-imports, leaving one definition each of `EFFECT_RISK_RANK`, `RISK_TAG_EFFECTS`, `BUILTIN_EFFECT_OBLIGATIONS`. Given how often this repo has been bitten by a second implementation, this is the right instinct.
- **The MCP reader records omissions rather than dropping tools silently** — named reason codes (`dynamic_tool_name`, `dynamic_tool_registration`, `ambiguous_tool_registration`, `definition_only_runtime_binding`) with ~40 omission/ledger references, which is the correct posture for #403.
- CHANGELOG entries sit under `## Unreleased`, docs are thorough (20+ files), `git diff --check` is clean, and no `TODO`/`FIXME` was introduced.

---

### 9. Process note

Three issues, 145 files, one commit. The three features are genuinely separable — the report v0.43 additions are entirely #428's (`declaration_review`, `action_declaration_facts`), while #326 drives the verifier/plan/receipt changes and #431 touches neither. Splitting now probably costs more than it saves, so I am not asking for it. But it is why the contract-28 collision with #487 and the orphaning of #483/#486 were easy to miss: a change this wide makes its own coordination problems invisible.

### Suggested order

1. **Decide the lineage.** Keep #488 and close #483/#486, or land the smaller PRs. I'd keep #488's `idiom`-on-`mcp` design over #483's new source type — it extends the one engine rather than adding a parallel surface.
2. **Resolve `CONTRACT_VERSION` with #487** — whichever lands second goes to 29 and re-appends its `PRIOR_RENDER_SHA256`.
3. **The three P0s**, in this order: the local-review permission escalation (add the four progress permissions and `must_stop` to the denial, and collapse the four copies into one function); the `RecursionError` crash; the quadratic recognizer.
4. **P1-3 on the MCP reader** — register `mcp_code_operation_type` in the evidence-basis allowlist, or the `operationType` half of #431 ships inert. Fix the test to assert the engine's acceptance, not the emitted string.
5. **The two declaration-review honesty gaps** — removed rows, and `base_comparison_requested` so "could not compare" stops reading as "nothing changed". The fix pattern already exists in `cli/scan/sanitization.py:188-208`.
6. **The `scan` reserved-manifest guard** and the generated manifest's `output.directory` — one guard call fixes both symptoms.
7. **Unprotected stale-version sites** and the `STABILITY.md` preflight line.
8. **Get the `test` job under the cap** so latency budget, packaging, dependency audit and SBOM actually run.
9. **Then** the human edits `AGENTS.md` + all three `SKILL.md` copies + the render-hash pin.

---

*Reviewed with Claude Code. Every finding above was reproduced before reporting; the P0s and the cross-cutting findings I re-ran myself.*

pengfei-threemoonslab:
## Expert review response

Addressed at rebased head `ac6a7787`. I reproduced the reported failures, rebased onto current `main`, removed superseded duplicate implementations, fixed the remaining canonical defects, and reran the relevant controls.

### Lineage and release coordination

- #483 and #486 are now in `main`; this branch uses those canonical #431/#326 implementations and removes its competing source/schema surfaces.
- #487 is also in `main`. The branch preserves #487's contract v28 and three-shard CI design, so there is no longer a divergent contract-v28 definition or CI timeout fix to duplicate here.
- The rebased branch is three commits over current `main`: `33095c880`, `5c0c38127`, and `ac6a7787`.

### #326 local-review authority

- Fixed the P0 in the canonical verifier control projection. `subject_evaluated` and declaration publication now require repository-authoritative manifest provenance.
- A provisional local-review manifest can no longer grant `edit`/`commit`/`push`/`update_pr`; the first-adoption continuation case is explicitly tested as `human_review_required` with all publication permissions false.
- The recovery/scan findings against this branch's earlier local-review implementation no longer apply because that duplicate implementation was removed in favor of #486's canonical code.

### #431 MCP source recognition

- The crash reproduction targeted the removed parser. The canonical parser is iterative; a regression test now exercises 5,000 nested template substitutions without recursion.
- Found and fixed a real quadratic path in the canonical `_enclosing_block` search. It is now a linear sweep (`_enclosing_blocks`), with a 4,000-class regression test.
- The canonical `operationType` route remains a low-confidence `basis="inferred_keyword"` challenge signal and is accepted by the evidence engine; the inert alternate implementation was removed.
- The remaining path-skip/detector-divergence findings described code that is no longer in this PR after lineage cleanup.

### #428 declaration review

- Removed declarations are now visible: comparison uses the union of base/head row ids and emits `change_type="removed"`, always `unverified`.
- Comparison failure is distinguishable from no change through `base_comparison_requested`; Markdown, packet, PR-comment, and GitHub-summary surfaces render an explicit “comparison unavailable” line only when a comparison was actually requested.
- Identity and semantic gaps (`unattested_surface`, incomplete/conflicting identity, unresolved/ambiguous selectors, legacy ambiguity, and semantic issues) prevent an evidence-consistent checkmark.
- Override rows use the exact declaration-row `manifest_path`; the release projection also resolves the precise declaration location.
- PR annotations now consume the shared declaration-review model/renderer for both unverified rows and overrides. Changed overrides are deduplicated, and report Markdown/GitHub summary carry the same reviewer surface.
- Added mutation-oriented coverage for effect issues, public collisions, 500-character row budgets, override deduplication, removal, unavailable comparison, exact paths, and renderer parity.
- A follow-up bug found by the broad suite was also fixed: ordinary scans no longer receive the “requested comparison unavailable” note when no comparison was requested.

### Published contracts and docs

- Current versions are report `0.43`, packet `0.18`, verifier `0.16`, preflight `0.4`, contract `28`; handoff remains v8 and verify-run remains v5.
- The preflight mismatch is fixed with an actual `PreflightResultV4` that emits `0.4`; v0.3 readers remain frozen.
- Generated schemas, `.well-known`, README/STABILITY/index/architecture/agent-contract pages, LLM documents, sample reports/packet, and the hash-bound Conductor pointer were regenerated and validated.
- Unprotected stale-version claims were updated. Protected trust-root files were intentionally not edited.

### Exact human-owned handoff

The only remaining failures are the same nine protected synchronization checks. A human must update:

1. `AGENTS.md`: report `0.42 → 0.43`, packet `0.17 → 0.18`, anchor `v017 → v018`.
2. `skills/agents-shipgate/SKILL.md`: the same three updates.
3. Its byte-identical mirrors:
   - `adoption-kits/claude-code-skill/SKILL.md`
   - `plugins/claude-code/skills/agents-shipgate/SKILL.md`
4. The expected Claude Code skill render SHA-256 pin.

Then rerun:

```bash
pytest -q tests/test_public_surface_contract.py tests/test_docs_links.py
```

### Verification performed

- Broad non-perf suite: all three deterministic CI shards pass when excluding only the protected sync files, plus the separately run static-only and packaging suites.
- Static-only adapter invariant: pass.
- P0 safety, binding, policy-evidence, verification-identity, and agent-boundary canaries: pass.
- Latency budget: 4/4 pass.
- Packaging/wheel suite: 10/10 pass.
- Ruff, compileall, dependency-lock verification, schema regeneration check, and `git diff --check`: pass.
- Committed-tree Shipgate verification at `ac6a7787`: `control.state=complete`, `decision=passed`, complete diff, committed-tree capability-delta attestation emitted.

The new GitHub Actions run is currently in progress. I will continue the review/check loop and address any coding-agent-owned regression it reports; the protected synchronization above remains a human gate and should be completed before merge.

pengfei-threemoonslab:
CI follow-up for `ac6a7787`: the refreshed run confirms the local attribution exactly.

- suite shard 1: **2,364 passed, 4 skipped, 7 failed** — all seven are the protected `AGENTS.md` / canonical `skills/agents-shipgate/SKILL.md` report/packet/anchor synchronization checks.
- suite shard 2: **2,371 passed, 1 skipped, 2 failed** — both are protected `AGENTS.md` current-report tracker checks.
- suite shard 3: passed.
- `test`, `verify`, `verify-self`, clean-checkout launcher, and Windows launcher: passed.

No coding-agent-owned CI regression remains. Coverage is skipped because the two protected shards are red; it will run after the human-owned synchronization listed in the preceding review response.

pengfei-threemoonslab:
## Update — #483, #486 and #487 have merged; this changes the recommendation

`main` is now `0cf68970`, three commits ahead of this PR's base:

```
0cf68970  Publish the capability delta as a standalone in-toto attestation (#470) (#487)
d7ab8f84  Add side-effect-contained local review flow (#486)
b8f8a271  Read an MCP server's tool surface out of its own source (#431) (#483)
6f4252d9  ← this PR's base
```

So two of the three issues this PR closes are **already solved on `main`, by different implementations**. I re-ran the key comparisons against the merged code. The short version: **on both overlapping features the merged implementation is the stronger one**, and the two P0s I reported are specific to this PR.

### The MCP P0s are not on `main`

#483 landed `inputs/mcp_idioms.py` + `inputs/mcp_server_source.py`, which use a masking pass with `_brace_pairs` computed once — architecturally the fix I suggested for this PR. Both adversarial inputs, run against the merged code:

```
nested-template fixture (15,014 B):  OK in 0.00s (no crash)     # this PR: RecursionError, scan exit 4
class x8000   ( 48,000 B):           0.00s                       # this PR: 0.72s
class x32000  (192,000 B):           0.00s                       # this PR: 11.55s
```

Neither the `RecursionError` crash nor the quadratic blowup reproduces. They are properties of this PR's `inputs/mcp_code.py`, not of the merged reader.

### `main`'s local review closes the shadowing vector this PR leaves open

This is the one I'd weigh most heavily. #486 refuses to create an ephemeral manifest when a committed trust root exists:

```
$ shipgate init --workspace . --local-review --json        # merged main, repo has shipgate.yaml
Config error: Refusing --local-review because shipgate.yaml already exists;
              use the repository manifest instead.
exit=2
```

This PR, same repo, same committed manifest:

```
$ shipgate init --workspace . --local-review --json        # this PR
{ "path": ".agents-shipgate-local-review.yaml", "created": true, ... }
exit=0
```

`main` prevents shadowing *at the door*. This PR permits the shadow and then tries to strip authority afterwards through `_ephemeral_manifest_cannot_authorize_release` — which is the rule with the permission-tier hole from my earlier comment (it guards `merge`/`report_complete` but not `edit`/`commit`/`push`/`update_pr`). Prevention at the door is the better design, and it is the one already shipped.

### `CONTRACT_VERSION` is now genuinely taken

`main` is at `CONTRACT_VERSION: Literal["28"] = "28"` (from #487). This PR must move to **29** and re-append its `local_contract.PRIOR_RENDER_SHA256`. The report-schema bump is still clean — `main` is at v0.42, so v0.43 remains free.

### What is still uniquely valuable here

**#428 — the declaration-review surface.** `core/lenses/declaration_surface.py` and `report/declaration_review.py` are not on `main`, and nothing merged supersedes them. That work is genuinely new, and the invariant that matters most in it holds (an action with no observed readings can never be rendered as machine-verified — enforced twice, independently).

### Revised recommendation

Reduce this PR to **#428 only**, rebased onto `0cf68970`:

1. Drop `inputs/mcp_code.py` and the #431 work — #483 already ships it, without the two P0s. If any of this PR's idiom coverage (the Go `MustTool`/`AddTool` or GitHub snapshot paths) is genuinely broader than `mcp_idioms.py`, port *that* onto the merged reader as a follow-up rather than replacing it.
2. Drop the `--local-review` work — #486 already ships it, and refuses the shadowing case this version allows. `verify --local-review` is the one addition here with no merged equivalent; if it is worth keeping, it should be a small PR on top of #486's flow.
3. Keep #428, move `CONTRACT_VERSION` to 29, and fix the four declaration-review gaps from my earlier comment — the removed-row blind spot and the "could not compare reads as nothing changed" fail-open are the two that matter.
4. The AGENTS.md / SKILL.md handoff still applies, and still needs all three byte-identical `SKILL.md` copies plus the render-hash pin.

That should also collapse this from 145 files to something reviewable, and drops most of the conflict surface against the three merged PRs (40 files currently overlap).

Worth saying plainly: the #428 work is good, and the local-review and MCP work here was not wasted — several things it does are better thought through than what merged (the typed `manifest_provenance` envelope, the exclusion-ledger reason codes). The problem is purely that it lost a race it was running against itself.

---

*Re-verified against merged `main` (`0cf68970`) with Claude Code; every comparison above was run, not inferred.*

pengfei-threemoonslab:
## Follow-up review response

Addressed at `e9bd6eaf` against `main` `0cf68970`.

- **Reduced to #428 only.** The final diff has no changes to `inputs/mcp_idioms.py`, its tests, the merged local-review flow/tests, or the unrelated preflight-version repair. The PR title, description, and closing references now name only #428.
- **Kept one verifier context field intentionally.** Passing `configured_gate_introduced` into the scan is required by #428 to distinguish first-adoption declaration evidence from a failed/missing comparison; it is not a replacement local-review implementation.
- **Advanced runtime contract 28 → 29.** Contract v28 remains the capability-delta contract from #487. I computed the outgoing local-contract render from exact `origin/main` bytes and appended its SHA-256, `bbad818d6b5b1ee34d412999720bb48d946a9dbb3cdc868b52221941c7b19477`, before publishing v29.
- **Declaration-review gaps remain fixed.** Removed rows are present and always unverified; requested-but-unavailable comparisons are explicit; incomplete/conflicting identity, ambiguous selectors, and semantic gaps cannot earn evidence-consistent; exact manifest paths, override deduplication, bounded rendering, and cross-surface parity are regression-tested.
- **Protected files remain untouched.** The nine expected failures are still exactly the human-owned `AGENTS.md` / canonical `SKILL.md` report/packet/anchor synchronization checks. The two skill mirrors and render-hash pin remain part of that handoff.

Validation after the scope reduction:

- clean aggregate non-performance suite: pass (packaging and protected files separated)
- isolated packaging: 10/10 pass
- static-only and P0 canaries: pass
- latency: 4/4 pass
- schema regeneration, Ruff, compileall, dependency locks, `llms-full`, and diff checks: pass
- committed-tree Shipgate verify: `control.state=complete`, `decision=passed`, `merge_verdict=mergeable`, committed attestation emitted

CI is running on the reduced head. I will continue the loop for any coding-agent-owned failure; the protected synchronization remains the only expected human gate.

pengfei-threemoonslab:
CI follow-up for reduced head `e9bd6eaf`:

- suite shard 1: **2,359 passed, 4 skipped, 7 failed** — all seven are the protected `AGENTS.md` / canonical `skills/agents-shipgate/SKILL.md` report v0.43, packet v0.18, and anchor v018 synchronization checks
- suite shard 2: **2,372 passed, 2 failed** — both are the protected `AGENTS.md` current-report tracker checks
- suite shard 3: passed
- `test`, Shipgate `verify`, `verify-self`, clean-checkout launcher, and Windows launcher: passed
- coverage skipped because the two protected shards are red

This confirms the #428-only scope reduction and contract-v29 change introduced no coding-agent-owned CI regression. The human-owned synchronization already listed in the PR description remains the only gate.

Issue #508, opened 2026-09-03
Cut C: produce and freeze the 56 × 2 blind labels (#456)


pengfei-threemoonslab:
Parent: #456. **Blocks the `v0.16.0` tag.**

Cut C produces the corpus's ground truth: the two blind primary labels per case, the owner's adjudication of every disagreement, and the freeze that makes them evidence.

## Why this is next

Sourcing is finished. #507 closed out Cut B: all **60 slots pinned**, 33 qualifying origins against a floor of 23, and every one of the 28 cells still holding a holdout-eligible slot. Nothing in [`strata-inventory.csv`](https://github.com/ThreeMoonsLab/agents-shipgate/blob/main/benchmark/safety-qualification/strata-inventory.csv) needs mining before this runs.

Everything downstream — the receipts, the signed artifact, the tag — waits on the labels existing and being frozen.

## The protocol is fixed, and it is why an agent cannot self-serve this

[Amendment 1](https://github.com/ThreeMoonsLab/agents-shipgate/blob/main/docs/release-evidence-policy-decision.md#amendment-1--the-pre-10-labeling-protocol-and-the-participant-validation-gate): two independent agent sessions on **different model families** produce the two primary labels; the owner adjudicates every disagreement and is **never** a rater. Six admissibility conditions apply, including mechanical blindness, a content-addressed archived transcript per label, walked-case disclosure in adjudication, a 5-case calibration round first, and a disclosure block in the published artifact.

Two consequences worth stating plainly, because they are what makes this owner-gated rather than merely unstarted:

- A session that has read the strata inventory produces **inadmissible** labels — the inventory names a target decision for every slot.
- One model family cannot supply both primary labels, so no single assistant can complete this alone.

## Clear these four first

From #489, all still open:

1. **Confirm [`LABELING.md`](https://github.com/ThreeMoonsLab/agents-shipgate/blob/main/benchmark/miner/LABELING.md).** It was restructured (rater rubric on top, the `stripe/ai#232` worked anchor removed because it stated a corpus case's label and the verifier's verdicts, which condition 2 forbids a rater to see). It is the file Amendment 1 names as the rater input, so it needs the owner's sign-off before any label is produced against it.
2. **The Claude harness is unverified** — its smoke run failed closed on an expired OAuth token (`claude login`).
3. **The OpenAI-family harness is unverified** — written against `codex exec` flags from memory, because the local codex npm install has an empty vendor binary directory.
[...]

PR #514, opened 2026-09-03
Clear what Cut C's preconditions can clear, and stop the change under judgement subtracting from its own packet (#508)

pengfei-threemoonslab:
## Summary

Cut C's four preconditions, cleared as far as this machine can clear them, and one class of defect found while clearing the fourth. Nothing under `src/` changes; no label is produced here.

The class turned out to be wider than the first commit found. Six review rounds followed it from the tree's `.gitattributes` into the operator's `~/.gitconfig`, and the last one landed back on the original failure — content the rater cannot see and cannot tell is missing.

### Precondition 4 — the packet contents: **head tree plus diff, no base tree**

The reasoning the builder already recorded stands. What it quietly assumed is that `diff.patch` is a complete textual description of base → head and that `repo/` is the commit's tree. **Neither was true**, and the base tree would not have fixed either, because the same mechanism subtracts from a base tree too.

**The tree under judgement could subtract from its own packet:**

| What | Was | Now |
|---|---|---|
| `export-ignore` | `git archive` dropped the path; `repo/` was missing files and `MANIFEST.json` hashed only what survived | read with `ls-tree` + `cat-file`, which consult no attribute |
| `-diff` / `binary` | `git diff` reduced an ordinary **text** change to `Binary files … differ` | read through a bare git dir whose `info/attributes` says `* !diff`, restated by `--text --no-textconv` |
| `diff=<driver>` | the tree's own funcname pattern chose the text after every `@@`, so the diff depended on which commit the clone was parked on | same |
| genuinely binary change | shipped as "differ", silently | refuses, naming the paths |
| a submodule the change moves | shipped as a gitlink hash | refuses, naming the paths |

The `-diff` row is the one this corpus cannot survive. Most of what the rubric calls `blocked` is a **removal** — an allowlist that no longer bounds, an approval step deleted — and a removal is visible only in the diff. A repository could hide exactly that, in a plain text file, and the packet would still build and still verify against its own manifest.

**And so could the operator's `~/.gitconfig`** (rounds 1, 4, 5, 6). Eleven `diff.*` / `core.*` keys are now pinned. Ten of them moved bytes — `diff.context` turns a 13-line patch into a 21-line one, `diff.interHunkContext` merges hunks and shifts every line number after them, `core.quotePath` decides whether `café.py` reaches a header as itself. The eleventh hid content: with **`diff.ignoreSubmodules=all`**, `git diff --raw` reports *nothing* for a changed submodule, so the refusal above found nothing to refuse, the patch carried no `Subproject commit` line, and the rater was handed a packet in which one of the change's edits simply was not present. That key is one people set to quiet noisy submodule diffs.

`git archive` is gone, and with it `tarfile`'s `data` filter, so the `../` refusal it used to provide is restored explicitly.

**The one cost:** a case whose change touches genuinely binary content or a submodule is now refused rather than shipped with a hole. Deliberate — an incomplete packet produces a label that *looks* admissible — but it means such a case needs the owner's decision. The refusal names the paths.

### Precondition 3 — the OpenAI harness: still owner-gated, for a different reason than recorded

[#489](https://github.com/ThreeMoonsLab/agents-shipgate/pull/489) recorded "the local codex npm install has an empty vendor binary directory". That is a symptom. The cached `@openai/codex@0.85.0` `darwin/arm64` binary extracts cleanly and passes `codesign -v`, but:

```
$ spctl -a -vv -t execute .../vendor/aarch64-apple-darwin/codex/codex
.../codex: CSSMERR_TP_CERT_REVOKED
```

**The signing certificate is revoked**, so macOS kills it at exec with `SIGKILL` and no output — which is why the flags could only be written from memory. Remedy is the owner's (`npm install -g @openai/codex@latest && codex login`; 0.85.0 installed, 0.153.0 current), and **the flags must then be re-verified against the installed CLI's own `--help`** rather than assumed across that gap.

`run_rater.py` now probes the CLI before handing over a packet, so the three ways this fails each name their own remedy:

```bash
python benchmark/safety-qualification/rater/run_rater.py \
  --family openai --role security_governance --check-cli
```

### Precondition 2 — the Claude harness: verified short of one live session

Against `claude 2.1.126`, with the OAuth token still expired and `ANTHROPIC_API_KEY` unset:

- every flag the harness passes exists, with the spelling it uses;
- the restrictions take effect, from a live `init` event: `tools` is exactly `["Glob","Grep","Read"]`, `mcp_servers` empty, `permissionMode` `dontAsk`, no user/project/plugin skills, no plugins;
- `_encoded_project_dir` computes the directory name the CLI actually uses — checked against the `memory_paths.auto` it reported for a known cwd, which is the only thing that makes the shared-mode memory refusal a check at all;
- the 401 arrives as a `result` event with `subtype: "success"` **and** `is_error: true` — a shape a `subtype`-only check would pass. `claude_final` gates on both, so it refuses.

### Precondition 1 — untouched, deliberately

The `LABELING.md` sign-off is the owner's. The `review_required` / `insufficient_evidence` rule is also untouched: #508 orders the calibration round to test the draft, and a correction written without the round would be the guess 56 labels are then produced against.

### Also found and closed

**Amendment 1 condition 1 — two model families — was enforced by nothing.** `SafetyCorpusCaseV1` asks only that the two `reviewer_id` values differ, and two sessions of one family differ anyway: the session uuid is part of the id. The runner now refuses the second role when the sibling label records the same family, **before** that session starts. It can only compare against a sibling it can find, so both roles of a case must go into one `--out`; the label records `family_independence`, reading `"unchecked"` when there was nothing to compare with, and refusing outright on a sibling that names no recognised family — a claimed check is worse than the silence the field replaced.

**Attribution recorded the probe, not the session.** `cli_version` now comes from the Claude `init` event's `claude_code_version`, with the probe as the fallback for the family whose stream names neither model nor version.

**Shared mode ignored instruction files above the packet.** Both CLIs discover project instructions by walking *up* from the working directory; `--bare` turns that off, so it is a shared-mode concern. `calibration.md` names a packet directory and not where it lives, so the obvious place to put one is inside this checkout — which carries `CLAUDE.md` and `AGENTS.md` at its root. Refused now. (Whether the CLI would actually have loaded them is **not established** — a local capture probe was inconclusive in both directions — so this is the packet-root rule applied consistently to the path leading there, not a reproduced leak. It is the only change here not backed by a reproduction.)

**A timed-out session archived nothing.** Condition 3's audit trail had a hole exactly where a session misbehaved; the partial transcript is now archived before the timeout propagates.

## Type

- [ ] Input adapter change
- [x] Benchmark rater harness and documentation only — nothing under `src/` changes

## Verification

CI is authoritative for `python -m ruff check .`, `python -m compileall -q src tests`, and `python -m pytest`. All checks pass on this branch.

Additional local checks run:

- `pytest -n auto -m "not perf"` over the whole suite after every round: exit 0. `ruff check .`: clean.
- **Every new guard was run against the build it was written to fail**, and each does: the five packet-completeness tests against the pre-change builder, the config tests against the branch's first commit, the `core.quotePath` and `diff.ignoreSubmodules` tests against the commit before each fix. The executable-bit test passes on both and is a regression guard, not a defect guard.
- **A systematic sweep closes the two channels by measurement rather than by argument.** 36 `diff.*` / `core.*` / `color.*` config keys set to hostile values through `GIT_CONFIG_GLOBAL`, over a fixture with separated hunks, blank lines, a non-ASCII path, a symlink and an executable: **zero drift** in the built packet. 19 `.gitattributes` directives placed in the clone's worktree: **zero drift**. And with each of `export-ignore`, `-diff`, `binary` and `diff=markdown` *committed* to the tree, the affected file still reaches `repo/` and its change still reaches `diff.patch`.
- Both original defects reproduced end to end before the fix: a `secrets/ export-ignore` path absent from `repo/`, and `notes.txt -diff` arriving as `Binary files a/notes.txt and b/notes.txt differ` while its head content sat in `repo/`.
- `--check-cli` exercised for both families: `claude: 2.1.126 (Claude Code)` (exit 0); `run_rater: openai cannot run: codex --version exited 1: Error: spawn …/vendor/aarch64-apple-darwin/codex/codex ENOENT` (exit 2).

- **The refusals are new failure modes for cases already sourced.** Cut B pinned 60 slots; if any of them touches binary content or a submodule, its packet now refuses to build. That is the intended behaviour, but it is a state Cut C can hit before the first label and it has not been swept for.
- **`_ISOLATED_CODEX_CONFIG` and the `codex exec` flags are still unverified**, and the comment saying so now names the real reason. `--check-cli` proves the CLI runs; it does not prove those flags are right.
- **`materialize_tree` holds the whole tree in memory** (one `cat-file --batch`). That is the same profile the `git archive` tar had, so it is not a regression, but a very large monorepo is the case that would find it.
- **`_DIFF_CONFIG`'s list no longer claims to be complete.** The comment says to treat it as incomplete until the hostile-`~/.gitconfig` test says otherwise; that test, not the prose, is what fails when git grows another knob. Five rounds of this list being "complete" is why.
- **`attribute_free_gitdir` writes nothing into the operator's clone** — it is a throwaway bare git dir with `objects/info/alternates` pointing at the real object store. It resolves that store through `--git-common-dir`, so a clone that is itself a worktree, or a bare clone, both work.

## Release-readiness notes

Refs #508, #456.

### DIFF
--- a/benchmark/safety-qualification/rater/build_packet.py
+++ b/benchmark/safety-qualification/rater/build_packet.py
@@ -12,14 +12,22 @@ those three things and the role's task sheet, and nothing else::
       TASK.md        the role's instructions and output contract
       MANIFEST.json  case id, pins, and a sha256 for every file above
 
-**Why ``repo/`` is the head tree.** "The pinned repository state" is the state
-the decision is about: the repository as it would stand if the change shipped.
-A rater given the head tree plus the diff that produced it sees the result and
-can reconstruct the base for any line the diff touches; a rater given the base
-tree would have to apply the diff in their head to see what the agent can do
-after the change. The diff is what makes the pair complete — a removed
-least-privilege bound is visible only there — so the packet carries both, and
-``evidence_references`` may cite either.
+**Why ``repo/`` is the head tree, and why the base tree does not ship.** "The
+pinned repository state" is the state the decision is about: the repository as
+it would stand if the change shipped. A rater given the head tree plus the diff
+that produced it sees the result and can reconstruct the base for any line the
+diff touches; a rater given the base tree would have to apply the diff in their
+head to see what the agent can do after the change. The diff is what makes the
+pair complete — a removed least-privilege bound is visible only there — so the
+packet carries both, and ``evidence_references`` may cite either.
+
+That completeness is a property of *how* the two are read, not something git
+gives for free: both of git's readers obey ``.gitattributes`` from the tree
+under judgement, so a repository can subtract from its own packet. Shipping the
+base tree as well would not fix that — ``export-ignore`` would subtract from
+the base tree too — so the fix is in the readers, and it is what makes "head
+plus diff" a complete answer rather than a hopeful one. See
+:func:`materialize_tree` and :func:`diff_pinned_states`.
 
 **What is excluded, and why.** ``agents-shipgate-reports/`` and
 ``.agents-shipgate/`` are verifier output and baselines; ``CASE.md`` is the
@@ -63,13 +71,13 @@ import argparse
 import hashlib
 import json
 import os
+import re
 import shutil
 import subprocess
 import sys
-import tarfile
 import tempfile
 from collections.abc import Iterator
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 
 REPO_ROOT = Path(__file__).resolve().parents[3]
 GUIDE_PATH = REPO_ROOT / "benchmark" / "miner" / "LABELING.md"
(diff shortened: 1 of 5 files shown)


benchmark/safety-qualification/rater/run_rater.py
```diff
@@ -374,6 +568,12 @@ def openai_invocation(
     else:
         codex_home = Path(os.environ.get("CODEX_HOME") or _real_home() / ".codex")
         check_shared_codex_home_is_instruction_free(codex_home)
+        # "Do not load $CODEX_HOME/config.toml; auth still uses CODEX_HOME" --
+        # exactly the split shared mode needs, since OAuth lives in that
+        # profile and everything else in it may not reach a blind session. An
+        # older codex without the flag exits on it, which is the right way to
+        # find out.
+        argv.insert(2, "--ignore-user-config")
```
pengfei-threemoonslab, line 610: [P1] Disable web search explicitly in shared Codex runs. `--ignore-user-config` removes the profile's `config.toml`, but this branch then supplies no replacement for `_ISOLATED_CODEX_CONFIG`'s `web_search = false`. The [official Codex config reference](https://developers.openai.com/codex/config-reference) documents the unset default as `web_search = "cached"`, so a shared-mode rater still has a search tool backed by information outside the packet. That violates the mechanical-blindness/“entire world” contract even if the prompt tells the model not to use it. Pass a command-line override such as `-c 'web_search="disabled"'` (preferably in both modes) and pin it in the invocation test.

benchmark/safety-qualification/rater/run_rater.py
```diff
@@ -628,29 +856,34 @@ def run_rater(
             home=Path(home),
             home_mode=home_mode,
         )
-        completed = runner(invocation, timeout=timeout)
+        family_independence = check_family_independence(out, manifest["case_id"], role, family)
```
pengfei-threemoonslab: [P1] Make the family claim atomic across concurrent roles. This check is a TOCTOU gate around a long-running session: if both roles start together with the same family, each observes that the sibling label does not exist, both return `unchecked`, and both sessions write valid same-family labels. I reproduced exactly that with a barrier-backed runner: the records contained families `['claude', 'claude']` and independence values `['unchecked', 'unchecked']`. Since operators can reasonably parallelize 112 sessions, reserve/lock the case in `--out` before launch (or atomically write and validate a family claim) rather than depending on sequential execution.

benchmark/safety-qualification/rater/build_packet.py
```diff
@@ -279,26 +288,381 @@ def _full_sha(repo: Path, ref: str) -> str:
...
+            "--no-color",
+            "--no-ext-diff",
+            "--no-textconv",
+            "--text",
+            "--full-index",
+            base,
+            head,
+        )
+    try:
+        diff = raw.decode("utf-8")
```
pengfei-threemoonslab: [P1] UTF-8 decodability is not a text/binary test. A Git file changing from `b"before\x00tail\n"` to `b"after\x00tail\n"` is binary by Git's NUL heuristic, but `--text` emits those NUL bytes and `raw.decode("utf-8")` succeeds. I reproduced a successful `diff_pinned_states` result whose `diff.patch` contains `\x00`, so the promised “genuinely binary change refuses” boundary is still open and Read/Grep may treat or truncate the patch as binary. Detect binary blobs/sections independently (at minimum NUL content, ideally using the changed blob IDs) and add this valid-UTF-8 binary case beside the existing invalid-UTF-8 test.

benchmark/safety-qualification/rater/build_packet.py
```diff
@@ -279,26 +288,381 @@ def _full_sha(repo: Path, ref: str) -> str:
...
+    with attribute_free_gitdir(repo) as gitdir:
+        submodules = changed_submodules(gitdir, base, head)
+        if submodules:
+            raise PacketError(
+                "the change touches submodules, whose content is in neither pinned tree: "
+                + ", ".join(submodules)
+            )
+        raw = _git_bytes(
+            gitdir,
+            *_DIFF_CONFIG,
```
pengfei-threemoonslab, line 660: [P2] Pin the `GIT_DIFF_OPTS` environment channel too. These `-c` values close config files, but `subprocess.run` inherits the caller environment; with `GIT_DIFF_OPTS=-u20`, the exact invocation here produced a different diff SHA than the normal environment despite identical base/head pins. This reopens the machine-dependent manifest and citation-line problem the PR is fixing. Remove `GIT_DIFF_OPTS` from the child environment or pass an explicit `--unified=3` after environment processing, and extend the determinism test beyond `GIT_CONFIG_GLOBAL`.

pengfei-threemoonslab (commented):
I found four correctness gaps in the blind-rating boundary. Three allow an ostensibly admissible run or packet to retain evidence the protocol says must be excluded; the fourth leaves packet identity dependent on the caller environment. All four are reproducible against the current head and should be addressed before the calibration run.

Validation performed: the 95 focused rater-harness tests pass, Ruff passes, GitHub CI is green, and Agents Shipgate reports `mergeable` / `passed` for the static capability surface. The full local suite reached completion except for the packaging fixture, whose isolated wheel build could not download `hatchling` in this network-restricted environment; that is not a branch failure.

GitHub does not permit this account to request changes on its own PR, so this is submitted as a comment review.

benchmark/safety-qualification/rater/run_rater.py
pengfei-threemoonslab: Confirmed and fixed in 75bcf9da.

You are right that the branch left shared mode with no replacement config at all — `_ISOLATED_CODEX_CONFIG` is only written in isolated mode, so `web_search = false` never applied to the mode that OAuth actually forces us into.

`-c 'web_search="disabled"'` is now passed on the command line in **both** modes, so it does not depend on any config file being loaded. I pinned the spelling by measurement rather than from the reference, since `--strict-config` makes that cheap — on 0.153.0 it accepts `web_search="disabled"` and `tools.web_search=false`, and rejects `web_search=false` (`invalid type: unit variant, expected string`) and `tools.web_search="disabled"`. The invocation test asserts the flag and its `-c` for both `HOME_MODES`, and fails on the previous commit.

benchmark/safety-qualification/rater/run_rater.py
pengfei-threemoonslab: Reproduced and fixed in 75bcf9da.

This was the sharpest of the four: the old check was a read of the sibling *label*, which does not exist until a session that runs for minutes has finished, so the window was the whole session.

Replaced with `claim_family`: an `O_EXCL` claim under `<out>/claims/<case>.<role>.json` written **before** the session starts, and the sibling read **after** it. Write-then-read is what closes it — whichever way two runs interleave, each one's write precedes its own read, so at least one of them sees the other's claim, and if they share a family that one refuses. Different families: both see a mismatch and both proceed, which is correct. Re-running a role is still allowed while the family is unchanged; swapping families under an existing claim is refused, since that is the thing the condition exists to stop.

The regression test is barrier-backed the way you described — both threads block inside the runner until the other has entered — and on the previous commit it fails with two same-family labels written.

benchmark/safety-qualification/rater/build_packet.py
pengfei-threemoonslab: Reproduced and fixed in 75bcf9da.

You are right that I conflated "decodes" with "is text". Confirmed your exact case: for `b"before\x00tail\n"` → `b"after\x00tail\n"`, `git diff --numstat` reports `-  -` (git calls it binary), `--text` emits the NULs, and `raw.decode("utf-8")` succeeds — so nothing refused.

`undecodable_paths` is now `non_text_paths`, and a section is non-text when it fails to decode **or** contains a NUL byte; the refusal names which of the two it was (`f.bin (contains NUL)`). I kept it as a per-section scan of the produced patch rather than a blob-ID check, because the patch is what the rater actually reads and it catches the case wherever it comes from — happy to switch to the changed blob IDs if you would rather the test be on the input than the output. The new test sits beside the invalid-UTF-8 one and fails on the previous commit.

benchmark/safety-qualification/rater/build_packet.py
pengfei-threemoonslab: Reproduced and fixed in 75bcf9da.

Measured it on the branch's own fixture: `GIT_DIFF_OPTS=-u20` turned a 22-line patch into a 48-line one, through exactly the invocation here. The `-c` pins outrank every config *file* and that is not one.

Git subprocesses now run with `GIT_DIFF_OPTS` and `GIT_EXTERNAL_DIFF` dropped, plus the repo-redirection variables (`GIT_DIR`, `GIT_WORK_TREE`, `GIT_COMMON_DIR`, `GIT_INDEX_FILE`, `GIT_OBJECT_DIRECTORY`, `GIT_ALTERNATE_OBJECT_DIRECTORIES`, `GIT_NAMESPACE`, `GIT_CEILING_DIRECTORIES`), which could point `-C` at another repository.

Two deliberate non-changes, both measured rather than assumed: `GIT_CONFIG_COUNT`/`KEY`/`VALUE` does **not** beat `-c` (identical output), and `GIT_EXTERNAL_DIFF` is already neutralised by `--no-ext-diff` — it is dropped anyway, but belt-and-braces. And `GIT_CONFIG_*` is kept in the child environment on purpose: it is config, `-c` does beat it, and the determinism test sets `GIT_CONFIG_GLOBAL` precisely to prove that, so stripping it would make that test vacuous.

The determinism test now covers the environment channel as well as `GIT_CONFIG_GLOBAL`, and fails on the previous commit.

Issue #520, opened 2026-09-04
insufficient_evidence is a statement about the reader, not the change: drop it from ground truth, score it as a coverage miss


pengfei-threemoonslab:
## Beta sourcing audit — remaining guide contract, 2026-09-10

At main `404e41325e62a9d71ca3c5bbd8e23711197117a3`, #512's source-inventory work found that the guide's binding reasoning and its permitted output vocabulary still disagree. The approved policy has three expected decisions, but [`benchmark/miner/LABELING.md`](https://github.com/ThreeMoonsLab/agents-shipgate/blob/404e41325e62a9d71ca3c5bbd8e23711197117a3/benchmark/miner/LABELING.md#L300) still maps four corpus decisions, and [the JSON output contract](https://github.com/ThreeMoonsLab/agents-shipgate/blob/404e41325e62a9d71ca3c5bbd8e23711197117a3/benchmark/miner/LABELING.md#L342) explicitly accepts `insufficient_evidence`. The prose says a complete packet should not produce that label; the executable instruction still permits it.

This is the remaining part of this issue's existing guide acceptance, not a new feature request. That acceptance is reopened; #519's delivered policy and binding reasoning remain delivered. Before #512 starts human primary labeling, align the rater vocabulary, example, evidence guidance and relevant packet/output validation with the three approved expected outcomes. Preserve a separate refusal/recovery for incomplete or unnameable packets rather than coercing them into a release label. Test the rendered rater input and accepted output together; do not rewrite historical labels or remove the verifier's actual-IE diagnostic vocabulary.

Deferred from #512's bounded source-inventory PR so collection can proceed without relabeling, modifying the rubric or claiming human readiness. The dated status below is historical: #559 and #561 have since merged; beta corpus/receipts and signing remain #512/#509, and the fixed-history obligations remain independently open. Sourcing candidates does not satisfy any of those bars.

Owner ruling from the first Cut C corpus round (#508), 2026-09-03: **`insufficient_evidence` should be rare, avoided where possible, and the reason worked out from first principles.** This is that reasoning, the evidence behind it, and what it changes. It is a child of #518 and it supersedes #517.

## Current implementation status — 2026-09-08

#519 has delivered the corrected labeling guide and approved 38-case pre-1.0 / 80-case beta policy. The initial 28-cell/56-case alternatives and calibration counts below remain historical rationale, superseded by that policy.

Delivered engineering:
[...]

Issue #506, opened 2026-09-03
init writes an adopter workflow pinned to a tag that does not exist (56 days and counting)


pengfei-threemoonslab:
> Product priority (2026-09-04): **P0 adoption blocker.** The generated integration fails before Agents Shipgate can evaluate the adopter's change. Fix the publication/contract rule here; #497 then protects it across distribution surfaces. Release qualification remains #456's responsibility — publishing an unqualified stable tag is not the repair.

## Symptom

`init --write --ci` writes a workflow into the adopter's repository that cannot run:

```yaml
# .github/workflows/agents-shipgate.yml, emitted today
uses: ThreeMoonsLab/agents-shipgate@v0.16.0
```

`v0.16.0` does not exist. The newest tag is `v0.15.0`. GitHub fails such a job at action-resolution time, before any step executes — so a first-time adopter's very first Shipgate run is a red check with an error about our repository, not theirs.

The bundled onboarding prompt has the same defect one layer up:

```
prompts/add-shipgate-to-repo.md:14
   SG="uvx agents-shipgate@0.16.0"    # uv: ephemeral, pinned to this exact build
```

PyPI's newest `agents-shipgate` is **0.15.0**; `0.16.0` is not on the index. So the prompt that teaches an agent to adopt Shipgate opens by pinning a build it cannot fetch.

## This is not new, and it is not the version bump

The tree has been ahead of the newest tag since **2026-07-09** (`0.16.0b1`, three days after `v0.15.0` was tagged on 2026-07-07). Every render in that window pinned a nonexistent ref — `@v0.16.0b1` … `@v0.16.0b7`, now `@v0.16.0`. **56 days and counting.** #491's release-cadence work makes the drought visible; this issue is one of its concrete adopter-facing costs.

## The repository already decided the right rule — for the other surfaces

Two conventions coexist, and only one is correct:

| Surface | Pins to | Enforced by |
|---|---|---|
| `llms.txt`, `docs/agent-contract-current.md`, `docs/quickstart.md`, `docs/integrations.md`, GitHub Actions examples | `LATEST_PUBLISHED_VERSION` | `test_public_surface_contract.py` |
| **`init --write --ci`'s emitted workflow** | `__version__` | `test_init_ci.py:53` |

`tests/test_public_surface_contract.py:75-77` states the rule outright:

> ```python
> # The source tree is on the 0.16 beta line while install snippets and Action
> # examples must continue to name the latest tag that actually exists.
[...]

Issue #504, opened 2026-09-02
Epic: the human route has no postcondition — specify the human decision step, then build surfaces from the spec


pengfei-threemoonslab:
## First advisory release scope — 2026-09-11

Owner decision: **adoption > completeness; make v1.0 ready with repository-independent work, then expand adoption.** The full human decision protocol remains open. #662/#684 own the bounded truthful presentation and routing tests required by the first advisory product.

Removed from the selected v1.0 milestone; kept open and deferred, not resolved. #643/#644/#645/#646 define current product acceptance; #572 preserves the historical qualified no-go and #648 owns the actual publication contract. The prior evidence and acceptance below remain a record of the deferred work, not newly satisfied conditions.

---

## Thesis

The authenticated human decision route does not tell the reviewer how the next run will recognize acceptance, rejection or disagreement. Human-owned declaration edits already have a completion path (#325); this epic owns the missing decision/acknowledgement path.

```python
# src/agents_shipgate/schemas/agent_control.py:75-91
class CodingAgentFetchBaseAction(BaseModel):
    """Shipgate never fetches refs itself.  ``expects`` therefore names the exact
    ref or artifact a caller must make available before rerunning verification."""
    expects: NonEmptyText          # required, non-empty

# src/agents_shipgate/schemas/agent_control.py:94-106
class HumanControlAction(BaseModel):
    """A human-owned route.  Human actions never expose executable commands."""
    command: None = None
    expects: None = None           # forbidden at the type level
```

`expects` is the postcondition — the thing that makes a route *checkable*. The docstring on the agent route states the principle exactly: when Shipgate cannot perform an action itself, it names the condition the next run will verify. **For this typed human decision action, carrying a postcondition is not merely absent; it is a type error.**

The missing postcondition prevents a caller from checking when this decision request is satisfied. `HumanActionKind = Literal["review", "stop"]` describes the requested action; it need not also encode the eventual acceptance, rejection or dispute. #536 separates those concepts and establishes the compatible publication path.

This epic is not "add an approve button." It is: **specify the human step, then build surfaces from the spec.** [#337](https://github.com/ThreeMoonsLab/agents-shipgate/issues/337) and [#293](https://github.com/ThreeMoonsLab/agents-shipgate/issues/293) are both mechanisms for a step whose semantics have never been written down, which is why both have stalled.

## Evidence: a real blocked run

[...]

Issue #498, opened 2026-09-02
The human quickstart must reach a useful review before a documentation-site rebuild


pengfei-threemoonslab:
## Problem

The human entry path gives a first-contact evaluator too many decisions before answering one practical question: **what did this capability change, and what should I do next?**

The current README is 933 lines with 28 second-level sections; `docs/` contains 225 files, including 151 versioned schemas. A marketing site and quickstart already exist. The gap is a short, coherent path through the **current, installable workflow**, not the absence of any website.

The design-partner runbook also carries older v0.15/contract-14 instructions and a blanket `CHANGE_ME` replacement step. #497 owns version/contract conformance; this issue owns a person being able to choose and finish the appropriate path.

## First increment

**1. Turn the README into an entry point.** Lead with one concrete before/after capability-change example, who needs it, what the static evidence cannot prove, the existing demo, the canonical install, and links for human adopters, coding agents and contributors. Aim for the existing proposal's approximately 150-line landing page, but do not delete useful evidence to meet a line-count metric.

**2. Rework the existing human quickstart around one review.** A reader should be able to identify the changed capability/permission, open its evidence, understand severity and coverage limits, and follow the next action without learning the internal identity or policy model first.

- Use the supported zero-manifest host-boundary path when appropriate; use the existing `detect`/`init` route for a tool-building repository. Do not make policy authoring a universal prerequisite.
- Show one resolvable published installation route. If the demonstrated workflow needs a newer contract, explicitly identify the available preview and its qualification status; #506 fixes the underlying emitted pins.
- Show a successful execution that still needs review, so exit zero is not taught as merge permission.
- Show a safe mechanical fix and a human-owned decision separately. Never tell a coding agent to infer missing authority, binding or purpose declarations from prose.
- Explain the second use: run on the next capability-changing PR, start CI advisory, and link the existing baseline/noise and strict-mode guidance. Authenticated review continuation depends on #504/#337; the guide must state current behavior until that lands.

**3. Preserve deep links and the agent path.** Move displaced README material to existing docs with navigation. Keep public anchors/links working or provide redirects, and keep generated `llms.txt`/`llms-full.txt` consistent.

## Acceptance

- [ ] README has one primary demo/install route and explicit audience links; a reader can tell whether the workflow fits their repository before installing.
- [ ] A cold human reader follows the quickstart on a committed sample using a build that actually provides its commands and contract.
[...]

PR #523, opened 2026-09-05
Measure first value and the next eligible change, and publish the zero (#521)

### LINKED ISSUE #521: The design-partner pilot stops at one PR: measure reviewer value and the next eligible change
## Problem

The existing [design-partner verifier pilot](https://github.com/ThreeMoonsLab/agents-shipgate/blob/main/docs/design-partner-verifier-pilot.md) counts three partner repositories after **one PR, one first-run feedback note, and one follow-up each**. That is a useful activation experiment, but it cannot answer the next product question:

> After the first verdict, did a reviewer make a better decision — and did the team use Agents Shipgate on the next eligible change?

A fixture completing, an advisory check returning zero, and a person saying the report looks useful are different from a reviewer acting on its evidence. We currently have no committed observation of that transition or of repeat use.

## Evidence and the gap this owns

- `docs/design-partner-verifier-pilot.md` defines a one-PR goal and exits on first-run feedback; it has no second-change or retained-integration criterion.
- #475 owns a consenting **public adopter registry**. A badge or our own `verify-self` does not establish external repeat use.
- #511 asks the authors/reviewers of **historical, frozen benchmark cases** whether a label is right and whether the gate would have helped. It does not observe a live installation or a second PR.
- #506/#485 can stop entry before a useful run; #504/#337 can stop a real review after a useful finding. The pilot must record those failures rather than excluding them from the denominator.

Extend the existing runbook and partner tracker. Do not create a second pilot program or a telemetry subsystem.

## Scope

Run a small, consent-based cohort through an actual workflow: a developer or platform engineer introduces a capability/permission change; another person reviews it; the next eligible change tests whether the integration stayed useful.

Start with the repository-declared coding-agent host boundary already supported by `check`, and include tool-building teams where a real partner has that workflow. Record which surface each participant actually uses. Do not require every partner to maintain a manifest when the supported host-boundary route does not need one.

For each consenting repository, record:

1. Entry point, published/preview build and contract, environment, attempted installation, setup steps, maintainer assistance, and time to a first valid result.
2. Time to **first value**: the reviewer can identify the changed capability, evidence, coverage limit and next action, and records a concrete decision or fix. A process exit of zero is not this event.
3. The existing alternative (manual review, CODEOWNERS, CI script, another tool, or doing nothing), and the specific work Agents Shipgate saved or added.
4. On a second eligible change within a four-week observation window: whether the team runs it again, whether CI remains enabled, whether a second reviewer can act without maintainer translation, and why it was disabled or bypassed if so.
5. Review-required outcomes separately: named request delivered, accepted/rejected/disputed, recorded continuation, or unresolved. Before #337 lands, report the missing continuation honestly; after it lands, repeat an eligible case.

A repository with no second eligible change is **not observed**, not retained or churned. A one-time audit can still have value; it does not prove a recurring CI wedge.

## Acceptance

- [ ] Update the existing runbook/tracker with the observation fields above, a consent/redaction rule, and distinct denominators for invited, attempted, first valid run, first value, and second-change eligible repositories.
- [ ] Record at least three external repository attempts with author and reviewer roles distinguished, or a dated report of the enrollment/opportunity shortfall. Include incomplete attempts and maintainer assistance. Dogfooding is reported separately and never fills this external denominator. A favorable adoption result is not required to close the experiment.
- [ ] Publish aggregate first-value timing and second-change outcomes with counts, observation windows and limitations. Set **10 minutes from starting the documented route to first reviewer-understood value** as an experiment target, not a claim already achieved.
- [ ] Preserve an actual before/after feedback bundle or redacted decision note for each reported useful outcome, using existing `feedback capture` where suitable. Public naming, source/PR links and raw bundles each require consent.
- [ ] End with an explicit continue/narrow/stop decision: which persona and workflow earned repeated use, which blockers recur, and whether any partner wants blocking CI. No response or an unobserved second change is not positive evidence.
- [ ] Turn reproduced blockers into updates to existing issues where possible. Do not add a new platform feature merely because a participant says it sounds useful.

## Sequencing

P1 product evidence alongside the adoption fixes, not after a larger platform is built. Cohort preparation and observation of the supported advisory path can start now; #506/#485 repair the entry routes under test. #498/#497 keep the runbook runnable. #504 → #337 supplies the authenticated review continuation for the later eligible-case retest.

This is independent of #509's qualification receipts and #511's historical-case outreach. It neither replaces release qualification nor changes an approved safety threshold. Contact is consent-based and must follow the repository's existing outreach rules; this issue is not permission for automated or unsolicited messages.

## Surface discipline

Headline metrics: activation, time-to-first-value and observed repeat use. Reuse `docs/design-partner-verifier-pilot.md`, its tracker, and existing feedback artifacts; no new CLI, schema, verdict, hosted service or default telemetry.

## Non-goals

- A public-adopter count (#475) or retrospective labeling exercise (#511).
- Default blocking CI before a team understands the coverage, noise and human decision path.
- Proving runtime safety or willingness to pay from three pilot attempts.
- A mandatory numeric retention target that can be met by selecting only successful users.

pengfei-threemoonslab:
Closes #521.

## What this does

The pilot counted three partners through one PR each and exited on a first-run
feedback note. That is an activation experiment; it cannot answer whether a
reviewer made a better decision or whether anyone ran it again.

**[`docs/design-partner-verifier-pilot.md`](https://github.com/ThreeMoonsLab/agents-shipgate/blob/claude/issue-521-review-255a38/docs/design-partner-verifier-pilot.md)** becomes the measurement protocol:

- **Two routes under test.** Route H (coding-agent host boundary, no manifest)
  and Route A (agent tool surface, manifest). Record which one each participant
  actually used; do not require a Route H partner to keep a manifest.
- **Six denominators** — `invited`, `attempted`, `first_valid_result`,
  `first_value`, `second_change_eligible`, `second_change_observed` — with three
  rules: failures stay in, no second eligible change means *unobserved* rather
  than churned, and dogfooding never enters any of them.
- **First value defined** as four things a reviewer who did *not* write the
  change can name — capability, evidence, coverage limit, next action — plus a
  recorded decision or fix. A process exit of zero is not the event.
- **Ten minutes** stated as an experiment target everywhere it appears.
- **Four-week window** for the second eligible change, with the disable/bypass
  reason recorded in the team's words.
- **Three separately-granted consents** (public naming, source/PR links, raw
  bundles); anything short of all three is reported aggregate-only.
- **A review-required ledger** that reports the missing continuation honestly
  until #337 lands, rather than scoring or dropping the row.
- **A pre-registered continue/narrow/stop rule**, written before the data.
- The tracker template now carries every observation field the issue lists.

## What the dry run found

> **Superseded 2026-09-05 by the merge of #506, #485 and #497 — see
> [Re-measurement](#re-measurement-after-merging-main) below. The table in this
> section surveyed two channels and there are three; the conclusion it reaches
> is corrected there and in the ledger.**

Acceptance allows a dated shortfall report in place of three attempts. Enrolling
a cohort is consent-based outreach and not something to do from this session, so
this PR does the part that determines whether an invitation is worth sending:
run the runbook's own commands on a synthetic host-boundary change (three
permission rules widened to `Bash(*)` / `Read(**)` / `WebFetch(*)`, one remote
MCP server added), against every channel a partner can install.

| Step | Released `v0.15.0` | Preview `0.16.0+preview.20260903` | Source tree |
|---|---|---|---|
| `contract --json` | **10** | 29 | 29 |
| `check --agent claude-code` | `warn` / `none`, **0 violations** | `block` / `critical`, **4** | `block` / `critical`, **4** |
| Coverage limit visible | no | yes | yes |
| `init --write --ci` pin | `@v0.15.0` — exists | `@v0.16.0+preview…` — **no such tag** | `@v0.15.0` — exists |
| `init` then `verify` on Route H | exit 3 | exit 2 | exit 2 |
| `audit --host --save-baseline` → `--drift` | **works** | works | works |
| Qualification | qualified release | **none** | not distributed |

Four findings, each a fact about a build:

1. **A build that shows the change is installable today — the preview.** All
   four first-value recognitions, and no qualification of any kind.
2. **The released build cannot show this change class.** `warn` / `none` on a
   diff granting `Bash(*)`, and no coverage surface at inventory schema 0.1.
3. **#506's repair has reached no downloadable build.** The tree writes
   `@v0.15.0`; the preview a partner downloads today predates the fix.
4. **`init` then `verify` dead-ends on Route H**, on every channel — #498's,
   and the blocker table says so.

The runbook had also demanded "runtime contract 14" in the paragraph that
installed it with `pipx install`, a precondition no published build has ever
satisfied. #497's channel table and this change both retire it.

**[`docs/design-partner-pilot-results.md`](https://github.com/ThreeMoonsLab/agents-shipgate/blob/claude/issue-521-review-255a38/docs/design-partner-pilot-results.md)** publishes six external denominators at zero, the dated
shortfall naming that cause rather than a recruiting gap, the fixture in full so
the comparison can be disagreed with, the reproduced blockers routed to #506,
#497, #520 and #504 → #337 **without opening a new issue**, and a dated standing
decision of **narrow** with what would overturn it.

## Acceptance status

| Item | Status |
|---|---|
| Runbook/tracker: observation fields, consent rule, distinct denominators | done |
| Three external attempts **or** a dated shortfall report | dated shortfall report |
| Publish first-value timing and second-change outcomes with limitations | done — the counts are zero and say so |
| Preserve a before/after bundle per useful outcome | protocol specified per route; no useful outcome exists yet to preserve |
| Explicit continue/narrow/stop decision | dated standing decision (**narrow**); the terminal decision needs observations |
| Reproduced blockers → existing issues | done; no new issue opened |

The two items that stay open depend on external observations, not on more
runbook.

## Surface discipline

No CLI, schema, verdict, hosted service or telemetry. One new doc — the
results ledger — because a runbook that also carries its own results is how
"how to observe" and "what was observed" drift into each other.

## Tests

`tests/test_design_partner_pilot.py` (26 cases) pins what is easy to soften by
accident: a dropped denominator, a blank count where zero is meant, dogfooding
folded into the external denominator, the target restated as an achievement, a
tracker field quietly removed, the unsatisfiable contract floor reintroduced, a
blocker row with no issue behind it. One guard fails the day a new tag ships, so
the build-dated findings get re-measured rather than carried forward.

A **19-case perturbation sweep** confirms each guard bites — every mutation
above was applied and caught, 19/19.

`ruff` clean; `test_public_surface_contract.py`, `test_docs_links.py` and
`test_release_pipeline.py` green.

---

## Review rounds

Two self-review passes after the first commit; six findings, all fixed in this
PR. Four were in the guards themselves — a test that passes for the wrong
reason is worse than no test.

| # | Finding | Fix |
|---|---|---|
| 1 | The build-dated guard matched the version literal *anywhere* in a section that also names the in-tree build — so tagging v0.16.0 would have satisfied it by coincidence, leaving the stale comparison standing. It went quiet at exactly the release it exists to catch. | The ledger states `**Published build measured: `0.15.0`.**` on its own line; the guard compares that value with the newest tag that exists |
| 2 | The contract-floor guard was scoped to `contract 1[0-9]`; a future "requires runtime contract 21" would have passed. | Widened to any digits |
| 3 | `_section` terminated on a `# ` shell comment inside a fenced block. Two such lines already exist under § Evidence to preserve, so any later assertion there would have been silently truncated — vacuous, not failing. | The helper tracks fences |
| 4 | "No published build has ever carried contract 14" was inferred, not measured. | Restated on what was measured: 0.15.0 carries 10, this tree carries 29, so 14 was first reached on the unpublished 0.16 line — and 0.8.0 has no `contract` command at all |
| 5 | The Continue threshold was a bare count with no denominator — the numeric retention target #521 lists as a non-goal, met by selecting successful users (2/30 fires the same as 2/3). | The rule reads its counts against the reported denominators and says it is a bar this project holds itself to, not one asked of a partner |
| 6 | The standing decision named the persona to *invite*, not the persona that *earned repeated use* — the question acceptance actually asks. | It now leads with `none`: nothing has been observed once, let alone twice |

Perturbation sweep is now 23 cases — 19 mutations that must fail, plus a
simulated v0.16.0 release, a contract floor outside the old window, a dropped
measured-build line, and a fenced shell comment in an asserted section that
must *not* fail. All 23 behave as specified.

---

## Re-measurement after merging main

`main` landed **#506, #485 and #497** — three of the four issues this ledger
cites — so its build-dated findings were re-measured rather than merged past.

**The re-measurement found an error of mine.** The 2026-09-04 shortfall said no
installable build could run the route to first value. That was wrong: it
surveyed two channels and there are three. The unqualified preview wheel
published 2026-09-03 carries contract 29 and returns `block` / `critical` with
all four violations named. The route was invitable the day the claim was
written; what was missing was a decision about putting an explicitly
unqualified build in front of a partner. The ledger now records the mistake
rather than quietly restating the conclusion.

**Conflict resolution.** #497 rewrote the same runbook, adding a channel table
and a placeholder-ownership rule. Both kept whole — and their rule *supersedes*
my Route A prompt step, which still told a coding agent to fill every
`CHANGE_ME` from the system prompt and README, which is a declaration nobody
made.

**Standing decision** stays **narrow**, with a different shape: invite Route H
on the preview channel with its unqualified status stated in the invitation,
withhold the released channel for per-change review. The tracker now records
the channel and what the partner was told about it.

**Two guard problems, both fixed.** My contract-floor guard fired on main's
*factual* "(contract 10) does not emit" — it forbade any contract number when
the defect was a *floating* one; rebound to "every contract number names the
build that carries it". And the published-tag guard never fired, because main
moved and no tag did; the offline half of that gap is closed by tying the
ledger's source-tree row to this tree's `__version__` and `CONTRACT_VERSION`.
The preview half needs the network and stays a documented limitation.

Sweeps now 45 cases across five scripts, with no-op detection — added after
three stale mutation strings silently matched nothing and read as passes. One
of them also exposed a trap worth naming: an mtime-based `.pyc` records whole
seconds, so mutating and restoring a source file inside one second leaves the
interpreter serving the mutated bytecode.


docs/design-partner-verifier-pilot.md
```diff
@@ -59,48 +233,87 @@ Prioritize teams that can share actionability feedback within one week.
...
+
+```bash
+agents-shipgate audit --host
+agents-shipgate audit --host --save-baseline   # then commit .agents-shipgate/
+```
+
+Then, on each change under review:
+
+```bash
+agents-shipgate audit --host --drift --json --out shipgate-drift.json
```
pengfei-threemoonslab, line 353: [P2] Bring the reviewed baseline into an existing PR branch

A partner bringing an already-open capability PR will have created its branch before the baseline commit this runbook adds to the default branch. Switching back to that PR removes `.agents-shipgate/host-grants.json`, so the documented `audit --host --drift --json --out shipgate-drift.json` exits 2 with `No readable host-grants baseline` instead of producing the first result. I reproduced this with the two-file fixture from the results ledger; merging the baseline commit into the change branch made the same command return all four expected expansion signals. Add an explicit step to merge/rebase the reviewed baseline into the PR branch, or carry that base snapshot through `--baseline-file`, in both this block and the partner prompt. Creating a fresh baseline from the changed checkout would acknowledge the very expansion being reviewed.

docs/design-partner-verifier-pilot.md
```diff
@@ -165,72 +401,173 @@ use `--agent-instructions=agents-md,claude-command,claude-code-skill`.
...
+
+Pre-registered so the terminal decision is not selected after the fact. Apply
+it once the cohort closes, and record the outcome in
+[`design-partner-pilot-results.md`](design-partner-pilot-results.md).
+
+| Decision | Condition |
+| --- | --- |
+| **Continue** | ≥ 2 repositories reach first value unaided **and** ≥ 1 reaches `second_change_observed`, on the same route |
+| **Narrow** | first value is reached, but only on one route, only with maintainer translation, or with no observed second change |
+| **Stop** | no repository reaches first value, or every failure traces to the product rather than to enrollment |
```
pengfei-threemoonslab: [P2] Make the terminal decision rules select a single outcome

Consider three Route H attempts: two repositories reach first value unaided, one of those runs again, and the third fails because of a product defect. `Continue` is true (two first values and one repeat on the same route), `Narrow` is true (first value occurs only on one route), and `Stop` is true (every failure is a product failure). No precedence is specified, so these pre-registered rules still let the same observations justify any terminal decision; reporting the denominators below does not resolve the overlap. Define mutually exclusive conditions or an explicit ordering, and cover mixed-success and single-route cohorts with example evaluations.

pengfei-threemoonslab (commented):
Codex review of all 8 changed files at `cd95f09f6dbf30b6a79b6bc332f62a909fe12d75`.

Two P2 issues should be addressed before merge: the Route H setup omits transferring the reviewed baseline into an existing PR branch, and the pre-registered decision rule can assign all three outcomes to the same cohort. The inline comments include a reproduced failure and a concrete cohort counterexample.

Validation: all 786 cases in `test_design_partner_pilot.py`, `test_public_surface_contract.py`, `test_docs_links.py`, and `test_release_pipeline.py` passed; Ruff passed for the new test file. I reproduced the documented baseline sequence in an isolated repository: drift exited 2 on the existing change branch; merging the reviewed baseline commit made the same command succeed and report the added MCP server plus all three widened grants. Agents Shipgate verification of the exact base/head diff returned `control_state: complete`, `decision: passed`.

docs/design-partner-verifier-pilot.md
pengfei-threemoonslab: Reproduced and fixed in b750b30e.

The trap is worse than a missing step: `--drift` exits 2 with `No host-grants baseline at …`, and the CLI's own recovery line then says `Record one first: shipgate audit --host --save-baseline`. Run from the changed checkout, that acknowledges the very expansion under review — drift afterwards reports nothing and the run looks clean. So the runbook now names the failure, gives both remedies, and prohibits the re-baseline with its reason.

Both remedies verified on the ledger's two-file fixture, each returning all four expansion signals:

- merge/rebase the reviewed baseline into the change branch;
- or leave the PR untouched — `audit --host --save-baseline --baseline-file /tmp/base-grants.json` on the default branch, then pass the same `--baseline-file` to the drift command.

Both appear in the Route H command block and in the partner prompt, since a partner follows whichever one they were handed.

`test_route_h_makes_the_baseline_reachable_from_the_change` pins it. The first version of that guard was too weak — it accepted a snapshot written with `--baseline-file` and never read back, because shell continuations put the flag and its command on separate source lines; it now joins continuations and requires both a write and a read.

docs/design-partner-verifier-pilot.md
pengfei-threemoonslab: Agreed and fixed in b750b30e. Your cohort is the proof: three independently-worded conditions are not a pre-registered rule, and reporting denominators does not resolve an overlap.

It is now an ordered ladder — first match wins, total by construction:

| # | Decision | Matches when |
|---:|---|---|
| 1 | Continue | `first_value` ≥ 2 unaided **and** `second_change_observed` ≥ 1, same route |
| 2 | Stop | `first_value` = 0 **and** `first_valid_result` ≥ 1 — a working result no reviewer could act on |
| 3 | Narrow | anything else |

Your counterexample now lands on rung 1 (Continue): the product defect goes to an existing issue, it does not cancel two first values and a repeat.

Two things the redesign changed beyond adding precedence:

- **"only on one route" is gone as a trigger.** It is vacuous when only one route was tested, which is a fact about the cohort rather than about the product. It stays as reporting guidance.
- **Rung 2 excludes the never-reached-a-valid-result cohort.** Stopping because our own entry defects blocked entry would read a fact about this repository as a fact about the market. That cohort falls to rung 3 — which is exactly where the 2026-09-04 checkpoint sits, and the ledger now shows the ladder being applied to reach it.

Six worked evaluations cover mixed success, single- vs multi-route, unaided vs maintainer-translated, all-first-value-no-repeat, and all-failed-at-entry. `test_decision_rule_selects_exactly_one_outcome` pins the precedence statement, ascending rung numbers and distinct outcomes; `test_decision_rule_works_the_ambiguous_cohorts` pins the examples.

PR #526, opened 2026-09-05
Read MCP registration sites in the zero-install detector (#485)

### LINKED ISSUE #485: The zero-install detector does not read MCP registration sites, so it still reports vendor MCP servers as non-agent projects

pengfei-threemoonslab:
Closes #485.

## The gap

`tools/shipgate-detect.py` is the documented first command an agent runs against a repository that has **not** adopted Shipgate — which is 100% of the population #431 is about. #431 taught the installed CLI to read a tool's name out of a TypeScript or Go registration site; the script did not gain it, so the two disagreed on the one question the script exists to answer.

Measured on live checkouts, before and after:

| workspace | `detect` | script (before) | script (after) |
|---|---|---|---|
| `mongodb-js/mongodb-mcp-server` | `true`, 61 tools, route `packages` | `false` — "Stop" | `true`, 61 tools, route `packages` |
| `grafana/mcp-grafana` | `true`, 114 tools, route `tools` | `false` — "Stop" | `true`, 114 tools, route `tools` |
| `github/github-mcp-server` | `true`, 114 tools, route `pkg/github` | `false` — "Stop" | `true`, 114 tools, route `pkg/github` |

On all three, `suggested_sources`, `excluded_sources` and the framework's `evidence` lines are now identical between the two detectors, character for character.

## What is ported

Into the stdlib-only single file: the masking lexer (both language grammars, both escape decoders, the regex/division heuristic), the five idioms, `is_scannable_path`, the containment rule, the declared-dependency gate, the common-ancestor route, export precedence, and the scoring. `FRAMEWORKS` gains `mcp_server_source`; `CONVENTIONAL_DIR_FRAMEWORKS` is mirrored so it stays out of the conventional-directory bonus. `SCRIPT_VERSION` → `0.5.0`.

Of the 40 reader functions shared with `agents_shipgate.inputs.mcp_idioms`, all 40 have identical bodies; of the 9 shared discovery functions, 7 do. The two that differ do so because they must, and the script says so where a reviewer diffing the files will read it: the export probe uses `json` instead of the pydantic-backed `load_mcp_tools` (the same accept rule `_probe_mcp` already mirrors), and the source read is factored into one function so the strict-decode contract is visible in one place.

## Why the duplication is affordable

A second implementation of the load-bearing matcher is the recurring bug class here, so the two are not allowed to become *different* implementations.

`tests/mcp_idiom_corpus.py` now holds every case either reader has ever been asked about — each idiom's positive sample, the whole adversarial sweep, the path predicate's cases, both escape grammars, the masking failures, and the review regressions. `tests/test_mcp_idioms.py` checks the package reader against the expected names and omissions; `tests/test_zero_install_detector.py` checks the two readers against **each other** on the same inputs, comparing every field of every site *including its byte span* — the span is what decides whether a wrapper call reports a second omission for a tool its own argument already named. The whole corpus runs a second time with CRLF line endings, because a Windows checkout is a supported one and it moves every offset.

Around that:

- `samples/mcp_source_only_server` — a TypeScript MCP server with no export, so the existing `samples/` parity sweep covers the route. Two tools named, one registration deliberately unenumerated.
- Thirteen constructed workspaces pin the branches a sample cannot reach: TypeScript, Go, both languages, a declared language that registers nothing, no declared dependency, no resolved registration, test-only registrations, covering export, partial export, wildcard export, an unresolved site outside the route, two registration directories, and the workspace root. A companion test pins the CLI's own answer for each, because a fixture set that never fires the route would agree perfectly and prove nothing.
- Parity on the strict read (undecodable bytes in a comment, placed so a lenient decode would widen the route directory), on the source-file cap, and on the export size bound.
- `mcp_server_source` is the one framework whose score, confidence, evidence lines and exclusion reason are pinned, not just its presence — the conjunction scoring, the "N registrations name themselves at runtime" line, and the sentence a reader gets when the route is withheld are all invisible to every other assertion.
- `test_framework_vocabulary_names_every_cli_omission` passes with an empty `known_omissions`; the idiom ids and omission reasons are pinned from both sides, against the CLI registry and against what the script's reader actually emits.

Six review rounds ran a perturbation sweep over the port, the discovery layer and the masker. Everything reachable is now caught by the test that claims to guard it. Rounds five and six audit the *original* reader as much as the port: a Go raw string decoded as though it were escape-processed invents `raw_name`, a `/` inside a regex character class ends the pattern early and swallows the line after it, three word-boundary lookbehinds admit constructs that merely end in `MustTool` / `NewTool` / `Tool`, and `operationType` is safe only because its pattern requires `static` — which the module says and nothing checked, while its twin `description` had both the fix and a regression. All six are corpus cases now, so they pin both readers.

Two perturbations stay invisible, provably, and the code says why at the line: the wildcard branch in the ported export reader cannot be distinguished from its fall-through for any input the probe accepts, and the `)` early exit in `_call_sites` is redundant with the second-argument check beneath it.

## Maintainer review, addressed

All four P2 findings reproduced exactly as described, in **both** readers — they are defects in the reader #431 shipped, which this port had copied rather than introduced. Fixed in `inputs/mcp_idioms.py` and mirrored into the script; the AST comparison shows all 44 shared reader functions byte-identical again.

| finding | effect before the fix |
|---|---|
| Brace inside a `${…}` counted as structural | ``const msg = `brace: ${"{"}`;`` consumed the rest of the file as one unterminated template — every later registration gone, and an MCP-declaring workspace reported as *not an agent project* over a brace in a string |
| Line break always ended the initializer | `static toolName = "safe"` + `"_delete"` on the next line published `safe` at `medium` confidence for a tool registered as `safe_delete` |
| Regex keyword resolved from raw text | A comment between `if` and its condition hid the keyword, so the pattern was scanned as code and reported a `ghost` tool — invented out of a regex body, which is what masking exists to prevent |
| CRLF continuation stepped over two characters | The identical file resolved its registration on a Unix checkout and lost it on a Git-for-Windows one |

The substitution scan is now iterative with one stack entry per open template, rather than recursive: a nested template is reached through a substitution, and recursion on attacker-shaped input is a crash rather than a wrong answer.

Nine expected-result cases in `tests/mcp_idiom_corpus.py`, so both readers are pinned to the corrected behaviour rather than to each other's agreement, and the CRLF sweep finally has a continuation case that exercises it. Reverting each fix fails its own case. **Four of the nine did not discriminate on the first draft** — a *closing* brace inside a comment or a regex ends the substitution early and the reader lands on the same backtick anyway, so the branch could be deleted and the case still passed. They carry an opening brace now.

The line-break fix is the one that can turn a resolved name into an omission, so it was checked by set rather than by count: running the pre-fix reader and the fixed one over every scannable file of the three vendor servers gives **identical name sets** — nothing lost, nothing gained, and the same unresolved counts (3 / 1 / 3). Routes, evidence and the 0.14–0.41s discovery cost are unchanged too.

And the lesson the findings carry, applied: six rounds of *code* perturbation missed all four, because mutating a branch only re-asks the questions the corpus already holds. So the corpus gains a **grammar sweep** — eighteen lexical constructs neither reader had ever been shown (a regex holding a backtick or a template opener, a comment holding a backtick or an apostrophe, a template holding a comment opener, an escaped backtick, a `$` that opens nothing, a regex beginning a statement after a block, division after an index, a private static field; on the Go side a rune holding a backslash or an escaped quote, a raw string holding a quote or a comment opener, an interpreted string holding a backtick). All eighteen passed on the first run, so they add no fix — they are recorded so the next change to the masker has to keep them passing.

## One defect fixed in both

With no MCP export in the workspace at all, `_covering_export` returned every resolved name as "uncovered", and the caller renders a shortfall as *"An MCP tool export is also present and does not name N of these registrations"*. A server whose surface exists only as source is exactly the population this input was built for, so that claim about a file that does not exist was published into the adoption evidence for every one of them — MongoDB, Grafana and GitHub included. Guarded in `cli/discovery/mcp_source.py` and in the port, with a regression test.

Also refreshed one stale prose count in `cli/scope_routing.py` (`samples/` was 23 of 25 adopted, not 21 of 22, before this PR added a sample).

## Left out, deliberately

`_probe_suggested` in the script reads every glob-matched JSON candidate with no size bound, while `probe_suggested_source` refuses inputs over 10 MB — a pre-existing divergence in `suggested_sources` and an unbounded read in a script that is curled onto unknown repositories. The new export reader here is bounded; the probe is filed separately rather than widened into this PR.

## Verification

- `ruff check .` clean; full suite green (`pytest -n auto -m "not perf"`, `tests/test_adapter_static_only.py`, and the perf run, each separately).
- The script runs standalone on a Python with no `agents_shipgate` importable, confirming it stayed stdlib-only.
- Acceptance criteria in #485: all three checked, on the live repositories.


pengfei-threemoonslab (commented):
Recommend fixing the four P2 correctness issues below before merging. I reviewed the diff at `4f5dd827c962033319f5a83139505949ed06f6d6`, including the source-discovery port, export precedence, scoring, sample, and corpus refactor.

All four findings reproduce in the newly added standalone reader and affect tool names or the public adoption verdict. They also reproduce in the existing package reader: this PR copies those defects rather than regressing that copy. Fix both readers and add expected-result corpus cases so parity is preserved while the incorrect behavior is removed.

Validation:

- `python -m pytest tests/test_zero_install_detector.py tests/test_mcp_idioms.py tests/test_mcp_server_source.py -q`: 360 passed, 1 skipped.
- Ruff passed for the changed Python files.
- The standalone script ran successfully with `python -I -S` on `samples/mcp_source_only_server`, without site packages.
- Four additional isolated regression assertions failed as described in the inline comments. I also evaluated the minimal JavaScript snippets with a local stub registration function to confirm the actual names and LF/CRLF equivalence; no MCP server was started.
- Agents Shipgate verification against the exact base/head returned `control_state: complete`, release decision `passed`; its current control identity was validated. This static gate result does not cover the parser defects above.

No repository source changes were made during review.

tools/shipgate-detect.py
```diff
@@ -677,6 +701,1434 @@ def _probe_openapi(data: Any) -> str | None:
...
+        if depth == 0 and char == "$" and index + 1 < length and text[index + 1] == "{":
+            substituted = True
+            depth = 1
+            index += 2
+            continue
+        if depth > 0:
+            if char == "{":
+                depth += 1
+            elif char == "}":
+                depth -= 1
```
pengfei-threemoonslab: [P2] Skip lexical contents inside template substitutions

The substitution counter treats a brace inside a quoted string or regex as a structural brace. For example, this is valid JavaScript and registers `real`:

```js
const msg = `Literal brace: ${"{"}`;
server.registerTool("real", {}, h);
```

The quoted `{` leaves `depth` above zero after the substitution closes, so the reader consumes the rest of the file as an unterminated template. In a workspace declaring `@modelcontextprotocol/sdk`, `detect()` then returns `is_agent_project: false`, no suggestions, and “No action.” Please skip strings, comments, regexes, and nested templates while tracking substitution braces, and add this case to the shared corpus with the expected `real` registration.

tools/shipgate-detect.py
```diff
@@ -677,6 +701,1434 @@ def _probe_openapi(data: Any) -> str | None:
...
+    can see, and reading it would publish ``backup`` for a tool the server
+    registers under some other name.
+    """
+    index = end
+    length = len(source.masked)
+    while index < length and source.masked[index] in " \t\r":
+        index += 1
+    if index >= length:
+        return True
+    return source.masked[index] in terminators or source.masked[index] == "\n"
```
pengfei-threemoonslab: [P2] Check for expression continuation across a newline

A line break does not necessarily terminate a JavaScript initializer. With:

```js
class T {
  static toolName = "safe"
    + "_delete";
}
server.registerTool(T.toolName, {}, h);
```

this branch accepts `safe` as the complete value, and the new detector publishes it as a resolved tool name at medium confidence. The registration actually receives `safe_delete`; `safe` is never registered. The same concatenation on one line is correctly reported as `name_not_literal`. Please check whether the expression continues after the newline before accepting the literal, while preserving semicolon-free fields, and pin the multiline case in the shared corpus.

tools/shipgate-detect.py
```diff
@@ -677,6 +701,1434 @@ def _probe_openapi(data: Any) -> str | None:
...
+        # validly begins the body: reading the `/` in
+        # `if (ok) /\.registerTool("fake", h)/.test(v)` as division scanned the
+        # pattern as code and reported a `fake` tool. Which of the two it is,
+        # is decided by the keyword in front of the matching `(`.
+        opener = _matching_open(out, previous_index)
+        if opener is None:
+            return False
+        return _preceding_word(text, opener - 1) in _REGEX_PRECEDING_STATEMENTS
+    if previous.isalnum() or previous in "_$":
+        return _preceding_word(text, index - 1) in _REGEX_PRECEDING_WORDS
```
pengfei-threemoonslab: [P2] Read regex context from the masked source

These keyword lookups switch back to the original text after locating the preceding token in the masked text. Consequently, a comment between `if` and its condition hides the keyword:

```js
if /*comment*/ (ok) /\.registerTool("ghost", {}, h)/.test(x);
```

The regex is then treated as code and yields a resolved `ghost` registration. With an MCP dependency and only this source file, the standalone detector reports `is_agent_project: true` and suggests source adoption, although no tool is registered. `typeof /*comment*/ /.../` has the same problem in the second branch. Please resolve the preceding word from the comment-masked representation and add expected-empty corpus cases.

tools/shipgate-detect.py
```diff
@@ -677,6 +701,1434 @@ def _probe_openapi(data: Any) -> str | None:
...
+
+def _consume_quoted(masker: _Masker, start: int, quote: str, *, allow_newline: bool) -> int:
+    text = masker.text
+    length = len(text)
+    index = start + 1
+    while index < length:
+        char = text[index]
+        if char == "\\":
+            index += 2
+            continue
```
pengfei-threemoonslab, line 1264: [P2] Consume CRLF line continuations as one escape

For a valid JavaScript quoted string with a backslash followed by CRLF, advancing two characters skips the backslash and CR but leaves LF to trigger `unterminated_string`. A registration spelled as `"my\` followed by a physical newline and `_tool"` resolves to `my_tool` with LF, but converting the same file to CRLF makes the detector lose the registration and return `is_agent_project: false`/“No action.” JavaScript evaluates both files identically. Please consume the whole CRLF continuation and update the TypeScript decoder to remove it as well; add an LF continuation case to `SOURCE_CASES` so the new CRLF sweep actually checks this behavior.

Issue #511, opened 2026-09-03
Gate 2: participant validation of the frozen corpus (#456)


pengfei-threemoonslab:
Parent: #456. **Non-gating for the tag**, under Amendment 1 and [`participant-validation.md`](https://github.com/ThreeMoonsLab/agents-shipgate/blob/main/benchmark/safety-qualification/participant-validation.md). The prerequisite is an actual frozen corpus and its terminal receipts so the case cards can show both the label and verifier outcome; closing #509 or completing independent signing is not an additional prerequisite. This historical pre-1.0 participant-validation track stays separate from #512's beta human primary labels and #571's adoption pilot.

**Current instruction reconciliation — 2026-09-08 (#566):** use the actual frozen corpus's cases and identities, not the superseded 56-case count. Keep the five pre-registered rules below. No outreach, participant response or frozen handoff is asserted by this update.

## Why it is worth doing anyway

The corpus's real-history cases were written and reviewed by real people. On the question the corpus is actually about — *should this change have been stopped?* — **those people are better authorities than any rater we can supply**, because they had the context at the time. This gate asks them, and records what comes back.

It also does double duty: the same conversation validates a case and asks the product-signal question, which makes it the first evidence-first outreach wave rather than a separate ask.

## The five pre-registered rules

Fixed in [Amendment 1](https://github.com/ThreeMoonsLab/agents-shipgate/blob/main/docs/release-evidence-policy-decision.md#amendment-1--the-pre-10-labeling-protocol-and-the-participant-validation-gate) so this cannot drift into after-the-fact relabeling:

1. **Frozen labels do not move.** A material disagreement triggers a case review whose corrections land in the `beta` corpus — except a material error found before the tag ships, which may force a re-freeze.
2. **Reviewers outrank authors.** Both are asked; recorded separately.
3. **Two questions, recorded apart.** "Is the label right?" and "would this have been useful?" have different consumers.
4. **Agreement-among-responders reporting.** Silence is not assent.
5. **Naming needs consent**, asked inside the outreach message. Absent a yes, the response is aggregate-only.

## Contact protocol

Per `real_history` / `rejected_or_reverted` case: the reviewer whose decision resolved the PR first, then the author. Design-partner cases go through the partner contact. No bulk mail, no bot comments, at most one follow-up. **A non-response is a data point, not an invitation to persist.**

## Done when

- [ ] A one-page case card is generated for every qualifying-origin case, in the committed format.
[...]

PR #531, opened 2026-09-06
Publish an opt-in adopters registry, and the rule that every adoption number comes from it (#475)

### LINKED ISSUE #475: Adoption evidence without telemetry: an opt-in adopters registry we can publish and verify
## Summary

Shipgate ships no telemetry, deliberately — local-first, static, nothing phones home. The cost of that stance is that we lack a consistent public mechanism for attributable adoption claims: download counts measure CI caches, stars measure sentiment, and the projects that quietly run the gate every day are invisible. Adoption evidence has been this project's #1 named risk in every strategic review; the counting mechanism is therefore a product deliverable, not an afterthought.

Build the lightweight, opt-in, public mechanism:

- **`ADOPTERS.md`** (or a pinned discussion): one line per adopter — name/org, repo (or "private"), what they gate — added *by the adopter themselves* via PR or comment. Being counted is an explicit act by the user; nothing is ever collected.
- **An optional README badge** adopters can add, linking back to the registry.
- **A documented claims policy**: any adoption number we publish traces to named, consenting entries in this registry. No entry, no claim.

The precedent is standard OSS practice (adopters files in CNCF projects), and it fits the product's own philosophy: adoption evidence should meet the same bar as capability evidence — **explicit, attributable, and verifiable, or absent**.

## Acceptance

- [ ] `ADOPTERS.md` exists with contribution instructions (what a valid entry contains; private-repo entries allowed with org name only).
- [ ] Badge snippet documented.
- [ ] Privacy stance stated in the file and README: no automatic collection; public entries are user-initiated and consenting. Private pilot observations are separate, require their own consent, and do not become public adopter entries automatically.
- [ ] Linked from the README and the site.
- [ ] The registry can include our own `verify-self` entry, explicitly marked maintainer dogfooding. External adopters are counted separately; a self-entry cannot satisfy an external adoption or repeat-use claim.

## What this proves

An entry proves a consenting party made the stated adoption claim. It does not prove continued use, review quality or willingness to pay. Include the entry's date and stated use (local evaluation, advisory CI or blocking CI); do not infer a stronger tier from a badge.

#521 separately measures first value and the next eligible change through the existing design-partner pilot. Names and private observations remain unpublished without explicit consent. #511 is historical-case validation, not either adoption denominator.

## Surface discipline

Repository documentation; no code surface, no data collection. Headline metric: attributable public adoption claims, with dogfooding and external entries separated. #521 supplies the workflow observations this registry cannot.

## Related

The no-telemetry positioning it protects · the incident-fixture and coverage-report pages (the funnel this registry sits at the end of).

pengfei-threemoonslab:
## Summary

Closes #475.

Shipgate collects nothing, deliberately. The price is that it cannot count its
own users: downloads measure CI caches, stars measure sentiment, and the repos
that quietly run the gate every day are invisible. This adds the counting
mechanism that stance requires — **[`ADOPTERS.md`](https://github.com/ThreeMoonsLab/agents-shipgate/blob/claude/issue-475-review-360ece/ADOPTERS.md)**,
opt-in and self-served, plus the rule that every adoption number this project
publishes comes from it.

- **One row per adopter**, added by the adopter — through the new
  [Adopter entry issue form](.github/ISSUE_TEMPLATE/adopter_entry.yml) or a
  pull request. Six fields: who, which repository (or the single word
  `private`), what it gates, the tier, the date, and a link to the public act
  where they asked. That link is the consent record, and it is what the count
  rests on.
- **Three tiers, defined tightly**: `local evaluation`, `advisory CI`,
  `blocking CI` — the last one meaning the job fails on a verdict *and* the
  check is required on the protected branch.
- **An eight-rule claims policy.** No entry, no claim; every number carries its
  as-of date; external and dogfooding are never summed; an entry is a dated
  statement rather than a measurement; a tier is self-reported and never
  upgraded; a badge is not an entry; the design-partner ledger (#521) and the
  historical corpus (#511) stay separate; removal is unconditional.
- **One optional badge**, deliberately tier-neutral, linking back to the
  registry. There is no blocking-CI variant and the file says why.
- **Today's numbers, published rather than omitted**: 0 external entries, 1
  maintainer dogfooding entry.

### The self-entry says `advisory CI`

Both self-gate workflows do fail the run on `blocked` and `unknown` — but
`main` carries no required status check, so a red run stops no merge. Under the
file's own definitions that is advisory, and a guard fails the build if the row
is ever raised without rewriting the paragraph that explains it. It is the
smallest available demonstration of the rule the registry runs on: the tier is
what is true, not what is flattering.

## Type

- [x] Documentation only

**Surface discipline.** Not a new surface under
[CONTRIBUTING §Surface discipline](CONTRIBUTING.md#surface-discipline) — no CLI
command, schema, report block, discovery surface or adapter, and no data
collection of any kind. `ADOPTERS.md` is registered in
`NOT_A_DISTRIBUTION_SURFACE` with its reason (it publishes self-reported
adopter claims, not an engine answer), which is what the top-level classifier
requires of any new root entry. Headline metric: attributable public adoption
claims, with dogfooding and external entries separated.

## Verification

Local, all green: `ruff check .`; the full `pytest -n auto -m "not perf"` gate;
`tests/test_adapter_static_only.py` and `tests/test_latency_budget.py -m perf`
in their own runs, as CI splits them.

`tests/test_adopters_registry.py` (46 tests) is the "and verify" half of the
issue title. It fails on a row missing a field, a row whose cell count is wrong
(reported by name rather than as a `zip()` error during collection), an
undefined tier, a non-ISO or future date, an entry link that does not point
into this repository, this repository listed as an *external* adopter, counts
that disagree with the rows, an as-of date older than the newest entry, an
empty table that stops saying it is empty, a claims rule quietly dropped, a
second badge variant, a README that drops the privacy stance or drifts from the
registry's counts, an issue form that offers an undefined tier or stops
requiring consent — and any adopter number stated anywhere in the repository's
prose that these rows cannot source.

**A 50-case perturbation sweep** drove every one of those mutations against the
suite and confirmed each fails, that the legitimate edits still pass (adding a
real adopter correctly, adding a second maintainer row *with* its explanation,
writing the grammatically singular "1 external adopter entry"), and — via a
no-op detector — that every mutation string actually matched something. It
found two real defects while I was writing it, both fixed here:

1. The badge guard folded case before looking for a tier token, so a
   `blocking%20CI` badge variant slipped through. It now also requires the
   section to carry exactly one shields.io URL.
2. The claim scanner's first draft misfired on real prose in this repository —
   "5 sentences of user instructions", "27 national company registries" —
   because it accepted a preposition as filler. A number now counts as a claim
   only when it reads as one: a use verb in front of it, a use verb behind it,
   or a noun with no other meaning here. Both strings are pinned as
   must-not-fire cases.

A third came out of the sweep's own failure: the negative controls hard-coded
the number `1`, which silently stopped being a control the moment a test
adopter row made `1` traceable. They now derive an unsourceable number from the
published counts.

**Two review rounds followed, both with findings** (commits 2 and 3):

- Round 1, on the guards: the README count assertion demanded the plural
  "entries", so the first real adopter had to write "1 external adopter
  entries" or take a red build over grammar; the self-entry agreement guard
  hard-coded a sentence opener that cannot be written naturally once there are
  two maintainer rows; dates were compared against the runner's local `today`,
  failing a contributor east of UTC; and the claim scanner re-parsed
  `ADOPTERS.md` once per scanned file (294 today).
- Round 2, on the prose: **the self-entry's `advisory CI` claim had been
  checked against the wrong mechanism.** Classic branch protection reports
  `main` as unprotected, but an active *ruleset* governs it. It requires a pull
  request and linear history and carries no `required_status_checks` rule — so
  the tier stands, and the file now names the command that shows it. ROADMAP.md
  was also restating the external count in prose, which is rule 2's staleness
  trap one file over; it points at the registry now instead.
- Round 3, on review feedback (three P2s, all reproduced before and after, in
  `5789ef07`): a **neighbouring** sentence about dogfooding could authorize an
  explicitly external number — the exemption is now scoped to the claim's own
  sentence, and a claim that calls itself external never takes it at all;
  **duplicate rows inflated the counts**, since counts were compared to
  `len()` — rows now carry an adopter+repository identity checked across both
  tables; and **only GitHub repositories were accepted**, which would have told
  a GitLab or self-hosted adopter to write `private` about a public repository
  — repository host and consent record are now validated independently.

## Not done here

**The site half of the acceptance is not in this repository.** "Linked from the
README and the site" — the README is done; threemoonslab.com is built from the
private `ThreeMoonsLab/web` repo, so the site link needs a change there. Happy
to file that as a tracking issue in `web` on request; it is the one acceptance
box this PR cannot tick.

## Release-readiness notes

- [x] New or changed check IDs are documented in `docs/checks.md` — n/a, no check changes
- [x] Report/schema changes are additive or documented in `STABILITY.md` — n/a, no schema changes

### DIFF
--- a/tests/test_adopters_registry.py
+++ b/tests/test_adopters_registry.py
@@ -1,0 +1,767 @@
+"""Guards for the opt-in adopters registry.
+
+This project collects nothing, so `ADOPTERS.md` is the only source for any
+adoption number it publishes. That makes the file's failure mode specific: not
+a typo, but a number that drifts upward one edit at a time — a dogfooding row
+counted as external, a count left behind when a row was removed, a tier raised
+from "we ran it once" to "it gates our merges", a badge read as an adoption
+claim, or a marketing number that no row can source.
+
+Each of those is pinned here, mechanically rather than by review attention. The
+tier vocabulary, the entry shape and the claims-policy rules live in this file
+and in the registry at once, so neither can be weakened alone.
+
+Issue #475.
+"""
+
+from __future__ import annotations
+
+import datetime as dt
+import re
+import subprocess
+from pathlib import Path
+
+import pytest
+import yaml
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+REGISTRY = REPO_ROOT / "ADOPTERS.md"
+README = REPO_ROOT / "README.md"
+TEMPLATE = REPO_ROOT / ".github" / "ISSUE_TEMPLATE" / "adopter_entry.yml"
+PILOT_RESULTS = REPO_ROOT / "docs" / "design-partner-pilot-results.md"
+PILOT_RUNBOOK = REPO_ROOT / "docs" / "design-partner-verifier-pilot.md"
+
+REPO_URL = "https://github.com/ThreeMoonsLab/agents-shipgate"
+BADGE_TARGET = f"{REPO_URL}/blob/main/ADOPTERS.md"
+
+#: The closed vocabulary for the ``Use`` column. Three tiers, ordered by what
+#: the gate can do — not by how much anyone likes it. Adding a fourth means
+#: changing the registry's definitions, this tuple and the issue form together,
+#: which is the point: a tier nobody defined is a tier nobody can check.
+USE_TIERS = ("local evaluation", "advisory CI", "blocking CI")
+
+#: Every entry has all six. ``Since`` and ``Entry`` are deliberately absent
+#: from the issue form: the date is when the row lands, not when the requester
+#: wrote it, and the entry link is the request itself.
+COLUMNS = ("Adopter", "Repository", "What it gates", "Use", "Since", "Entry")
+REQUESTER_COLUMNS = ("Adopter", "Repository", "What it gates", "Use")
+
+#: The claims policy, keyed by the bold lead-in of each rule. Wording inside a
+#: rule may be improved; a rule may not quietly disappear.
+CLAIMS_RULES = (
+    "No entry, no claim.",
+    "Every number carries its as-of date.",
+    "External and dogfooding are never summed.",
+    "An entry is a dated statement, not a measurement.",
+    "The tier is self-reported and never upgraded.",
+    "A badge is not an entry.",
+    "The pilot ledger stays separate.",
+    "Removal is unconditional.",
+)
+
+EXTERNAL_HEADING = "## External adopters"
+DOGFOOD_HEADING = "## Maintainer dogfooding"
+EMPTY_MARKER = "No external entries yet."
+
+LINK_RE = re.compile(r"\[[^\]]*\]\(([^)\s]+)\)")
+DATE_RE = re.compile(r"^\d{4}-\d{2}-\d{2}$")
+
+
+# --------------------------------------------------------------------------
+# Reading the registry
+# --------------------------------------------------------------------------
+
+
+def _read(path: Path) -> str:
+    return path.read_text(encoding="utf-8")
+
+
+def _flat(text: str) -> str:
+    """Markdown here is hard-wrapped, so any phrase asserted on below is
+    usually split across lines. Compare against the unwrapped form."""
+    return " ".join(text.split())
+
+
+def _section(text: str, heading: str) -> str:
+    """The body under ``heading``, up to the next same-or-higher heading.
+
+    Fences are tracked: a ``#`` in column one inside a fenced block is a shell
+    comment, not a heading, and truncating there would make every assertion
+    past it vacuous (the same defect found in the pilot guards, #521).
+    """
+    level = len(heading) - len(heading.lstrip("#"))
+    assert text.count(heading + "\n") == 1, f"{heading!r} must appear exactly once"
+    start = text.index(heading + "\n")
+    body = text[start + len(heading) :]
+    pattern = re.compile(rf"#{{1,{level}}} ")
+    fenced = False
+    offset = 0
+    for line in body.splitlines(keepends=True):
+        if line.lstrip().startswith("```"):
+            fenced = not fenced
+        elif not fenced and pattern.match(line):
+            return body[:offset]
+        offset += len(line)
+    return body
+
+
+def _table(section: str) -> tuple[tuple[str, ...], list[list[str]]]:
+    """The pipe table in ``section`` as (header, raw rows).
+
+    Deliberately total: a row with the wrong number of cells comes back as-is
+    rather than raising, so it fails :func:`test_every_row_has_the_columns_it
+    _claims` by name instead of breaking collection with a ``zip()`` error that
+    names neither the file nor the row.
+    """
+    lines = [line.strip() for line in section.splitlines() if line.strip().startswith("|")]
+    if not lines:
+        return (), []
+    cells = [[cell.strip() for cell in line.strip("|").split("|")] for line in lines]
+    header = tuple(cells[0])
+    assert len(cells) > 1 and all(set(cell) <= set("-: ") for cell in cells[1]), (
+        f"expected a separator row under {header}"
+    )
+    return header, cells[2:]
+
+
+def _raw_rows(heading: str) -> list[list[str]]:
+    return _table(_section(_read(REGISTRY), heading))[1]
+
+
+def _entries(heading: str) -> list[dict[str, str]]:
+    """Well-formed rows only; malformed ones are reported by their own test."""
+    header, rows = _table(_section(_read(REGISTRY), heading))
+    return [dict(zip(header, row, strict=False)) for row in rows if len(row) == len(header)]
+
+
+def _all_entries() -> list[tuple[str, dict[str, str]]]:
+    return [
+        (heading, row)
+        for heading in (EXTERNAL_HEADING, DOGFOOD_HEADING)
+        for row in _entries(heading)
+    ]
+
+
+def _entry_ids() -> list[str]:
+    return [
+        f"{heading.strip('# ').replace(' ', '-')}::{row.get('Adopter', '?')}"
+        for heading, row in _all_entries()
+    ]
+
+
+def _counts() -> dict[str, str]:
+    """The labelled numbers in ``## Counts``, by label."""
+    section = _section(_read(REGISTRY), "## Counts")
+    return {
+        m.group(1): m.group(2)
+        for m in re.finditer(r"\*\*([A-Z][A-Za-z ]+): ([^*]+?)\.\*\*", section)
+    }
+
+
+def _slug(heading: str) -> str:
+    text = heading.lstrip("# ").strip().lower()
+    text = re.sub(r"[^\w\s-]", "", text)
+    return re.sub(r"\s+", "-", text)
+
+
+def _headings(text: str) -> set[str]:
+    return {_slug(line) for line in text.splitlines() if line.startswith("#")}
+
+
+def _template() -> dict:
+    return yaml.safe_load(_read(TEMPLATE))
+
+
+def _template_fields() -> dict[str, dict]:
+    return {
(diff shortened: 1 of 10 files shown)


tests/test_adopters_registry.py
```diff
@@ -0,0 +1,783 @@
...
+    external = counts["External adopter entries"]
+    dogfood = counts["Maintainer dogfooding entries"]
+    flat = _flat(text)
+    untraceable = []
+    for match in _claims(flat):
+        number = match.group(1).replace(",", "")
+        if number == external:
+            continue
+        context = flat[max(0, match.start() - _CONTEXT) : match.end() + _CONTEXT].lower()
+        if number == dogfood and ("dogfood" in context or "maintainer" in context):
```
pengfei-threemoonslab: [P2] Keep dogfooding context local to the matched claim

The 120-character window lets an unrelated dogfooding sentence authorize an explicitly external number. With the current counts (0 external, 1 maintainer), appending `Agents Shipgate is used by 1 external team. Maintainer dogfooding is counted separately.` to `ROADMAP.md` leaves all 46 registry tests passing: this branch accepts the external `1` using the maintainer total. This defeats the separation the scanner is supposed to enforce, especially in prose that discusses both counts together. Require the matched claim itself to be qualified as dogfooding, and never use that exemption for a claim explicitly qualified as external; add this adjacent-sentences case as a regression.

tests/test_adopters_registry.py
```diff
@@ -0,0 +1,783 @@
...
+# --------------------------------------------------------------------------
+
+
+def test_the_counts_are_the_row_counts():
+    """The whole mechanism in one assertion: a published number equals rows
+    somebody consented to. Removing a row without touching the count, or
+    raising a count without a row, fails here."""
+    counts = _counts()
+    assert counts["External adopter entries"] == str(len(_entries(EXTERNAL_HEADING)))
+    assert counts["Maintainer dogfooding entries"] == str(len(_entries(DOGFOOD_HEADING)))
```
pengfei-threemoonslab, line 433: [P2] Reject duplicate adopter rows before counting them

The counts are compared only to list lengths, so the same adoption and consent record can be counted repeatedly. I duplicated the existing Three Moons Lab row byte for byte and changed the registry/README dogfooding count to 2; all 51 resulting registry tests passed, although there is still only one adopter and one public act. The same counting logic governs external entries, so accidental duplicate submissions can inflate the headline metric while every guard stays green. Add a uniqueness check for the adopter/entry identity (at minimum rejecting identical rows) before treating row counts as attributable adoption counts.

tests/test_adopters_registry.py
```diff
@@ -0,0 +1,783 @@
...
+def test_every_entry_names_a_repository_or_says_private(heading: str, row: dict[str, str]):
+    """Acceptance §1: private-repo entries are allowed at organization
+    granularity. `private` is the whole vocabulary for that — a half-named
+    private repository ("internal monorepo") discloses without being checkable."""
+    repository = row["Repository"]
+    if repository == "private":
+        return
+    hrefs = LINK_RE.findall(repository)
+    assert len(hrefs) == 1 and re.fullmatch(
+        r"https://github\.com/[\w.-]+/[\w.-]+", hrefs[0]
```
pengfei-threemoonslab: [P2] Accept public repositories hosted outside github.com

This check restricts the adopter's repository to GitHub even though the registry and issue form require a public repository link or `private`, and the CLI can be used with repositories on other hosts. A correctly filled external row containing `[acme/agent](https://gitlab.com/acme/agent)`, with its consent link still pointing into this project and both published counts updated, fails this assertion (49 passed, 1 skipped, 1 failed). Legitimate GitLab, Codeberg, or self-hosted adopters therefore cannot follow the documented contribution path without mislabeling their public repository as private. Validate the public repository URL independently of the GitHub-only consent-record requirement.

pengfei-threemoonslab (commented):
Reviewed all 10 changed files at eb71c0bdeaf5204d07b5ec2387b7408f39d23079 against db63e42b59d637dd5127cf3a2a12cc7d74028926. I found three validation issues worth fixing before merge; each inline comment includes a reproduced case.

Validation: all 830 tests in `test_adopters_registry.py`, `test_distribution_surface_parity.py`, and `test_public_surface_contract.py` passed; Ruff on the changed Python files and `git diff --check` passed. I also ran separate mutations against the new registry suite: a duplicated maintainer row and an unsupported external claim both passed, while a valid GitLab entry failed. The three GitHub Actions workflows report success for this head.

The static verifier reports `control_state: complete`, `gate.merge_verdict: mergeable`, and `release_decision.decision: passed`, with no blockers or review items. That result covers the configured capability surface; the findings below concern the new registry validation.

tests/test_adopters_registry.py
pengfei-threemoonslab: Fixed in 5789ef07, and your reproduction is now a regression test.

Reproduced first: appending that sentence pair to `ROADMAP.md` left all 46 tests green — the external `1` was authorized by the *next* sentence, which is the sum rule 3 exists to forbid.

Two changes, because the window was wrong in two independent ways:

- **The qualifier is scoped to the claim's own sentence** (`_sentence_around`), not a character window. So `used by 1 team. Maintainer dogfooding is separate.` is flagged too, not just the explicitly-external form.
- **A claim that qualifies *itself* as external never takes the exemption**, whatever its sentence says. That is checked on the matched claim text rather than the sentence, deliberately: `1 maintainer dogfooding adopter and 0 external adopters` is a legitimate sentence that names both counts, and vetoing on the whole sentence would have made honest combined prose unwritable — which is the failure mode that gets a guard deleted rather than obeyed.

Regressions added: `test_a_neighbouring_dogfooding_sentence_authorizes_nothing` (your exact case, plus the unqualified variant) and `test_an_externally_qualified_claim_never_borrows_the_dogfooding_count`. Both are also perturbation-sweep cases, so the fix is checked from outside the suite as well.

tests/test_adopters_registry.py
pengfei-threemoonslab: Fixed in 5789ef07. Confirmed your reproduction — a byte-for-byte duplicate of the maintainer row with the count moved to 2 passed everything, because the counts were only ever compared to `len()`.

`test_no_adopter_is_counted_twice` now rejects it. The identity is **adopter + repository link**, not the whole row, so a near-duplicate (same repository, reworded description, different date) is caught too — an identical-row check alone would have missed the more likely accident. Identity is compared **across both tables**, since the same party listed as external *and* as dogfooding inflates the headline number from precisely the place rule 3 forbids.

Two consequences worth stating, and the file now states them: a `private` row is identified by its adopter alone — that is all the registry knows about it — so one organization holds at most one; and a second repository is a second row, which is also how an adopter with two different tiers records that honestly.

tests/test_adopters_registry.py
pengfei-threemoonslab: Agreed, and fixed in 5789ef07 — this was the worst of the three, because the failure mode was the registry telling a truthful adopter to mislabel their public repository as `private` in order to fit the validator.

The two checks are now independent, which is the distinction the old regex collapsed:

- **`Repository`** accepts any host — `https://gitlab.com/…`, `https://codeberg.org/…`, `https://git.acme.example:8443/…` — or the single word `private`. Shipgate runs on a checkout, not on a forge API, so there was never a reason for GitHub to be privileged here.
- **`Entry`** stays restricted to an issue/PR/discussion in this repository. That one is not about hosting: it is the public act that makes the row attributable, and a consent record nobody outside the maintainers can open is not one.

Both rules are extracted into `_repository_problem` / `_entry_record_problem` and unit-tested with synthetic accept and reject cases, so the boundary is testable without adding rows to the registry. `ADOPTERS.md` and the issue form now say "on whichever host it lives", and a GitLab-hosted adopter is a must-pass perturbation case.

PR #532, opened 2026-09-06
Scope the unresolvable-root rejection to its own project (#398)

### LINKED ISSUE #398: One unresolvable application root marks every agent name candidate in the repository `selectable: false`
> Product priority (2026-09-04): **P1 first-value blocker** for observed multi-project ADK adoption. A valid project must remain selectable when an unrelated root cannot be resolved. Preserve refusal for the unresolved project itself and for truncated discovery; do not solve this by choosing the first candidate or auto-asserting deployment bindings. #521 records whether the scoped route lets the adopter reach a useful result.

## Summary

A single application root whose name is not statically resolvable marks **every** agent name candidate in the whole repository `selectable: false`, including agents in unrelated projects. On adk-samples that is roughly 55 candidates across 25–53 projects, all rejected because of one file.

## Evidence

Two runs on the same repository blamed two different single files, and in both cases rejected the entire list.

`detect` on `main`, every candidate carried:

```
"rejected: an application root is declared here but its name is not statically resolvable
 (.agents/skills/scaffold-python-recipe/resources/templates/app/agent.py: `root_agent` is last
 assigned a value that is not a statically readable agent construction); any other name would
 declare a worker as the reviewed identity"
```

`init` on `pr-1745`, every candidate carried the same sentence with a different culprit:

```
(python/agents/RAG/eval/test_eval_arize.py: `root_agent` is last assigned a value that is not
 a statically readable agent construction)
```

Affected entries include `financial_coordinator`, `cyber_guardian_orchestrator`, `data_science_root_agent` and every other real product agent — none of which has any relationship to the file named in the rejection.

That the same directory scoped down behaves correctly confirms the diagnosis: `init --workspace python/agents/smart_closer` returned `SmartCloserAgent`, `SalesforceAgent`, `SapAgent` all with `selectable: true`.

## Why this matters

The rejection rule is sound within one project — if the root cannot be identified, promoting a worker as the reviewed identity is wrong. But it is being applied with **repository scope**, so one unresolvable root anywhere disables name selection everywhere, and the message points the user at a file that has nothing to do with the project they are adopting.

Two of the culprits observed are also poor candidates for "application root" in the first place: a scaffolding template under `.agents/skills/**/resources/templates/`, and a test file under `eval/`. The ranker already discounts test code for *selection* (`declared in test code, which names fixtures rather than the product`) but apparently not for *poisoning*.

## Suggested direction

1. Scope the unresolvable-root rejection to the project the candidate belongs to — `agent_project_candidates` already computes that grouping.
2. Exclude the directories the ranker already treats as non-product (test code, `**/resources/templates/**`) from counting as an application root at all.

---

pengfei-threemoonslab:
Closes #398.

## The bug

A declared application root whose identity cannot be established statically makes **nothing** selectable — the #324 rule, and a sound one: every name still standing is by construction *not* that root, so writing one would declare a worker as the reviewed identity.

It was applied with **repository** scope. On adk-samples that meant one file rejected roughly 55 candidates across 25 projects — `financial_coordinator`, `cyber_guardian_orchestrator`, `data_science_root_agent` — and told the reader so by naming a file in a project they were not adopting.

Reproduced both halves before touching anything:

```
--- whole repository ---
  'financial_coordinator'   selectable=False
      rejected: an application root is declared here but its name is not statically resolvable
      (.agents/skills/scaffold-python-recipe/resources/templates/app/agent.py: ...)
--- scoped: python/agents/RAG ---
  'rag_root_agent'          selectable=False
      rejected: ... (eval/test_eval_arize.py: ...)
```

The second line matters: scoping alone would **not** have fixed the reproduction, because that culprit sits *inside* the project being adopted.

## The change

**1. The rejection is scoped to the project the root sits in.** That is the grouping `agent_project_candidates` already publishes, now computed once behind a `_ProjectAttribution` object that both readers share — a second, weaker "which project" here would have missed weak markers (two sibling agents whose only boundary is `requirements.txt` beside `agent.py` are two projects, #363). The rationale sentence now names the project. A name several projects declare is rejected when **any** of them is blocked; that is the fail-closed direction.

**2. Only a root the *product* declares blocks a project.** A fixture that builds an `App` is a fixture; a file under `resources/templates/` is what a generator copies. The ranker already said so for *selection* (`ORIGIN_TEST_PENALTY`) and said the opposite for *rejection* — one predicate (`_non_product_origin`) now answers both, so a scaffolding template is demoted exactly as a fixture is instead of standing in for the product. Without that symmetry the fix would have moved the failure rather than closed it: `init` would write the example's name.

The template rule is the directory **pair** `resources/templates/`, not a bare `templates/` — Django and Flask both ship a real one, and every name added to the exclusion widens what fails *open*.

## What is deliberately preserved

- A project whose own product root is unreadable still refuses, and the manifest keeps `CHANGE_ME`.
- An unreadable root under a bare `templates/` still blocks.
- Truncated discovery is untouched (`agent_scope_truncated` / `python_parse_truncated`).
- No candidate is auto-asserted and no deployment binding is inferred.

## Zero-install parity

`agent_name_candidates` is the one field `tools/shipgate-detect.py` pins **byte for byte** (`docs/zero-install.md`), so the script carries the same change (`script_version` `0.6.0`), covered by eight parametrised parity cases that also assert the *verdict* — a shared bug rejecting everything would satisfy equality alone.

Restructuring both sides surfaced a parity break that **predates this issue**: where no marker was found above the evidence, the script named the workspace's `requirements.txt` as the project marker while the CLI reported `null`. A weak marker only draws a boundary in a directory that already holds agent evidence, so one that unlocked nothing is not the boundary the project rests on. Fixed to match the CLI, with a test — every sample carries a `pyproject.toml`, which is why the per-sample parity check never reached the fallback.

## Verification

Full CI gate green (`ruff`, `pytest -n auto -m "not perf"`, `test_adapter_static_only.py`, `test_latency_budget.py -m perf`).

Every new guard was replayed against `origin/main`: 13 of them fail there and pass here. The four that pass on both are the deliberate preserved-behaviour cases (`bare_templates`, `own_root`, `shared_name`, `readable_template`), where the parity assertion is doing the work.

No new surface: `detect`'s schema is unchanged, and `zero_install_detector`'s row in `docs/distribution-surfaces.md` still claims exactly `agent_project_verdict`.

### DIFF
--- a/src/agents_shipgate/cli/discovery/signals.py
+++ b/src/agents_shipgate/cli/discovery/signals.py
@@ -211,19 +211,28 @@ ROOT_AGENT_SYMBOL = "root_agent"
 CHILD_AGENT_KEYWORDS = ("sub_agents", "handoffs")
 # Score adjustments. Hierarchy and corroboration move a candidate within
 # its origin; origin itself is meant to dominate, because the documented
-# contract is that product code outranks test code — full stop, not "unless
-# the test one happens to be a root". ORIGIN_TEST_PENALTY is therefore
-# strictly greater than the whole spread of the other signals
+# contract is that product code outranks code that is not the product —
+# full stop, not "unless the fixture happens to be a root".
+# ORIGIN_NON_PRODUCT_PENALTY is therefore strictly greater than the whole
+# spread of the other signals
 # (ROOT_BONUS + CORROBORATION_BONUS − SUB_AGENT_PENALTY = 5.5), which keeps
 # the published rank_score the single ordering key instead of needing a
 # separate tier the score cannot explain.
 # Conventional test module filenames that carry no ``test_`` prefix, so the
 # prefix rule alone reads them as product code.
 TEST_MODULE_NAMES = frozenset({"conftest.py", "test.py", "tests.py"})
+#: Consecutive directory names that mark code shipped as *material* rather
+#: than as the running application. A ``templates/`` directory nested inside
+#: a ``resources/`` one is the scaffolding convention #398 found: the file it
+#: holds is what a generator copies into a new project, not what this
+#: repository runs. The pair is required rather than a bare ``templates/``,
+#: which web frameworks use for a directory of real modules — every name
+#: added here widens what fails *open*, so it is kept to the shape observed.
+NON_PRODUCT_DIR_SEQUENCES: tuple[tuple[str, ...], ...] = (("resources", "templates"),)
 ROOT_AGENT_BONUS = 3.0
 SUB_AGENT_PENALTY = 1.5
 CORROBORATION_BONUS = 1.0
-ORIGIN_TEST_PENALTY = 6.0
+ORIGIN_NON_PRODUCT_PENALTY = 6.0
 QUALITY_FLOOR_PENALTY = 3.0
 
 # Quality floor for a value that may be written as ``agent.name``. A
@@ -377,11 +386,8 @@ def detect_workspace(
     detections.sort(key=lambda d: (-d.score, d.type))
 
     project_name_candidates = _project_name_candidates(workspace)
-    agent_name_candidates = _rank_agent_name_candidates(
-        py_facts, workspace, project_name_candidates
-    )
     codex_plugin_candidates = _codex_plugin_candidates(workspace, inventory)
-    agent_project_candidates = _agent_project_candidates(
+    evidence_paths = _agent_evidence_paths(
         py_facts,
         detections,
         # An artifact-only project — an OpenAPI spec, an MCP export, a Codex
@@ -394,6 +400,16 @@ def detect_workspace(
         + [candidate.path for candidate in codex_plugin_candidates]
         + _nested_manifest_paths(inventory, workspace),
         workspace,
+    )
+    attribution = _ProjectAttribution(workspace, evidence_paths)
+    agent_project_candidates = _agent_project_candidates(
+        py_facts, evidence_paths, attribution
+    )
+    # Ranking runs after the grouping because it reads the same attribution:
+    # an application root it cannot resolve disqualifies the names in *that
+    # project* and no others (#398).
+    agent_name_candidates = _rank_agent_name_candidates(
+        py_facts, workspace, project_name_candidates, attribution
     )
     project_root_count = _project_root_count(inventory, workspace)
     python_file_total = sum(1 for path in inventory if path.suffix == ".py")
@@ -1934,6 +1950,78 @@ def _safe_resolve(path: Path) -> Path:
         return path
 
 
+def _non_product_origin(rel_path: str) -> str | None:
+    """Why ``rel_path`` is not the product's own code, or ``None``.
+
+    One predicate, two uses, because #398 is what an asymmetry between them
+    costs: the ranker discounted a fixture for *selection* while still
+    letting the same file disable selection for every other name in the
+    repository. Both readings go through this function —
+    :func:`_score_agent_name` for the penalty and
+    :func:`_can_declare_application_root` for the block — so a file cannot
+    be product code for one and material for the other.
+
+    Two kinds of file declare agents that are not the product's:
+
+    * **Test code.** A fixture that builds an ``App`` is a fixture. #398
+      found one ``eval/test_*.py`` rejecting every real product agent in the
+      repository.
+    * **Scaffolding templates.** :data:`NON_PRODUCT_DIR_SEQUENCES` names the
+      directory shape that holds what a generator copies. Only the
+      *directory* part is examined, because the sequence is about where a
+      file sits rather than what it is called.
+
+    The return value is the rationale line the penalty publishes, so the
+    reason a name was demoted and the reason its file cannot speak for the
+    project are the same sentence.
+    """
+
+    if _is_test_path(rel_path):
+        return "declared in test code, which names fixtures rather than the product"
+    parts = Path(rel_path).parts[:-1]
+    if any(
+        parts[index : index + len(sequence)] == sequence
+        for sequence in NON_PRODUCT_DIR_SEQUENCES
+        for index in range(len(parts))
+    ):
+        return (
+            "declared under a scaffolding template directory, which names an "
+            "example rather than the product"
+        )
+    return None
+
+
+def _can_declare_application_root(rel_path: str) -> bool:
+    """Whether a root declared in ``rel_path`` speaks for the product.
+
+    The unresolvable-root rule is a statement about the application a
+    project ships: if that root cannot be identified, no other name in the
+    project may stand in for it (#324). A root declared by code that is not
+    the product (:func:`_non_product_origin`) is not that root, and reading
+    it as one disables name selection over files neither ``init`` nor the
+    runtime would ever reach.
+    """
+
+    return _non_product_origin(rel_path) is None
+
+
+def _unresolvable_root_rejection(scope: str, detail: str) -> str:
+    """Why no name a project declares may be written as its identity.
+
+    ``scope`` is the project's workspace-relative path, so the sentence
+    names the project the unreadable root belongs to. #398's report was that
+    it named only a file — one in a project the reader was not adopting.
+    """
+
+    where = "this workspace" if scope == "." else f"project `{scope}`"
+    return (
+        f"{where} declares an application root whose name is not statically "
+        f"resolvable ({detail}); every other name it declares is by "
+        "construction not that root, so none of them can be the reviewed "
+        "identity"
+    )
+
+
 def _is_test_path(rel_path: str) -> bool:
     """Whether ``rel_path`` is test code rather than product code.
 
@@ -2059,7 +2147,10 @@ class _RankedName:
 
 
 def _rank_agent_name_candidates(
-    facts: list[_PyFacts], workspace: Path, project_names: Sequence[NameCandidate]
+    facts: list[_PyFacts],
+    workspace: Path,
+    project_names: Sequence[NameCandidate],
+    attribution: _ProjectAttribution,
 ) -> list[AgentNameCandidate]:
     """Rank ``Agent(name=…)`` evidence best-first.
 
@@ -2078,17 +2169,40 @@ def _rank_agent_name_candidates(
       is ranked last and made unselectable, so ``init`` writes CHANGE_ME and
       asks for review rather than asserting something unreliable.
 
-    One rule overrides all four: if a workspace declares an application root
-    whose identity cannot be established statically, *nothing* is
-    selectable. Any name still standing is by construction not the root, so
-    writing it would declare a worker as the reviewed identity — the exact
-    failure #324 asks to fail closed on.
+    One rule overrides all four: a *project* that declares an application
+    root whose identity cannot be established statically has nothing
+    selectable. Any name still standing in it is by construction not the
+    root, so writing one would declare a worker as the reviewed identity —
+    the exact failure #324 asks to fail closed on.
+
+    That rule is scoped to the project, not to the repository. Applied with
+    repository scope it read one unresolvable root in a scaffolding template
+    as grounds for rejecting all 55 real product agents in a monorepo, and
+    told the reader so by naming a file with no relationship to the project
+    they were adopting (#398). ``attribution`` supplies the same grouping
+    :func:`_agent_project_candidates` publishes, so "which project" has one
+    answer here and there. A name several projects declare is rejected when
+    *any* of them is blocked: it is a worker in that project whatever it is
+    elsewhere, so this direction is the fail-closed one.
+
+    Only a root the product itself declares blocks a project — see
+    :func:`_can_declare_application_root`.
 
     Scores are published so a reordering regression is visible in
     ``detect --json`` instead of silently changing what the manifest claims.
     """
     by_path = {_safe_resolve(fact.path): fact for fact in facts}
-    unresolved_roots = [fact.unresolved_root for fact in facts if fact.unresolved_root]
+    # Project directory → the first unreadable application root found in it.
+    blocked_projects: dict[Path, str] = {}
+
+    def _block(fact: _PyFacts, detail: str) -> None:
+        if not _can_declare_application_root(fact.rel_path):
+            return
+        blocked_projects.setdefault(attribution.of(fact.path)[0], detail)
+
+    for fact in facts:
+        if fact.unresolved_root:
+            _block(fact, fact.unresolved_root)
     # A root whose *name* is a symbol that fails the cross-module resolution
     # below is just as unresolved as one whose name is an f-string. The
     # failure surfaces here rather than at parse time because resolution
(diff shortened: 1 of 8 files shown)


pengfei-threemoonslab:
Reviewed the branch at high effort (10 finder angles + a gap sweep); 8 findings, all addressed in 58742d04.

**Fixed (5)**

1. **Cross-project rejection omitted the shared declaration.** A value rejected because *another* project that also declares it is blocked published `path: "clean/agent.py"`, `role: "root_agent"` and a rejection naming `project \`blocked\`` — every field pointing one way and the reason pointing another. That is #398's own complaint ("the message points the user at a file that has nothing to do with the project") reproduced one project narrower. The candidate's own project is now preferred when it is blocked, and when it is not, the sentence opens `this name is also declared in project \`blocked\`, which declares an application root…`.
2. **The template match was case-sensitive.** `Resources/Templates` is the same directory as `resources/templates` on a case-insensitive checkout, so the #398 symptom survived a spelling difference the filesystem does not make. Both sides case-fold now.
3. **`attribution.of(fact.path)` ran per name-evidence site** although it is invariant across a file, paying a `Path.is_dir()` stat before reaching its cache — a stat per candidate on the discovery hot path of exactly the monorepo shape this issue is about. Hoisted, in both implementations.
4. **`test_a_bare_templates_directory_still_blocks_selection` asserted only `select_agent_name(...) is None`**, which any other reason satisfies; it now asserts the rejection names `templates/agent.py`.
5. **The precedence between the two `_block` passes was untested.** `setdefault` makes the parse-time root win over a cross-module symbol failure in the same project — the old `unresolved_roots[0]` behaviour — and `rationale[]` is pinned byte for byte, so it is a contract between two implementations. The new guard names the resolution-time file *first* in walk order, so it pins precedence rather than ordering.

Two new parity cases (`case_folded_template`, `two_unreadable_roots`) carry 2, 3 and 5 into the zero-install script.

**Documented rather than changed (2)**

Where a workspace's *only* agent evidence is non-product code, an unreadable root there no longer forces `CHANGE_ME` — the fixture or template name is written. That guard was accidental: a *readable* root in the same file already had its name written before this PR, so the change makes the two cases agree rather than opening a new hole. The general rule it points at — a name only non-product code declares should never be asserted as the reviewed identity — is #533; it would close the readable case too and is a wider product decision than a P1 first-value fix should make. The same widening reaches `init --write --allow-unresolved-scope`, whose contract already reads "adopt the first agent name it parsed" on a workspace with several unrelated projects. Both are now stated in the CHANGELOG.

**No change needed (1)**

CLAUDE.md § Surface discipline points at `docs/distribution-surfaces.md` for any surface edit. Checked: the `zero_install_detector` row still states exactly what is true — same claim (`agent_project_verdict`), same proof (`test_detector_verdict_matches_cli`), same "Narrower than the CLI in" column. The `script_version` bump is not a column, and the table carries no version field.

**Verification**

Full CI gate green after the fixes (`ruff`, `pytest -n auto -m "not perf"`, `test_adapter_static_only.py`, `test_latency_budget.py -m perf`). Each review-fix guard was replayed against 90254d10: 4 of the 6 fail there. The two that pass — the strengthened bare-templates assertion and the precedence guard — are regression guards for properties that already held, which is what they are for.

pengfei-threemoonslab:
Second review pass over 58742d04, looking only for defects the review fixes themselves introduced. One, fixed in 6a9c019b:

**The efficiency fix went too far.** Hoisting `attribution.of(fact.path)` out of the inner evidence loop moved it from once per *name* to once per *parsed file* — including the overwhelming majority that declare no agent at all, where nothing below reads the answer. On a 1000-file repository with five agent files that trades 995 stats and project walks for the handful the fix set out to save. The loop now skips a file with no name evidence outright, which is both cheaper than the original and clearer than either version.

Also recorded on both copies of `NON_PRODUCT_DIR_SEQUENCES` that its entries must be lowercase — the path is case-folded before the comparison, so a capitalised entry added later would silently match nothing and the exclusion would quietly stop working.

Checked and clean on this pass: the `sorted(..., key=lambda project: project != best_project[value])` key is evaluated eagerly inside the loop iteration, so there is no late-binding capture; `best_project[value]` is written in the same statement that writes `best[value]`, so the rejection loop cannot `KeyError`; the stable sort preserves declaration order among the remaining blocked projects; and `declared_elsewhere` with a workspace-level `scope` reads correctly ("this name is also declared in this workspace, which declares…").

Local gate green (`ruff`, `pytest -n auto -m "not perf"`, `test_adapter_static_only.py`, `test_latency_budget.py`), and the eight-case CLI↔script parity probe still agrees on every shape.

pengfei-threemoonslab (commented):
Reviewed `6a9c019b` against base `db63e42b`, including the previous review notes and the deferred scope of #533. I recommend fixing the two inline findings before merge:

1. **P1:** A candidate outside the discovered agent projects can bypass the unreadable product-root refusal, and ordinary `init --write` writes the unrelated name.
2. **P2:** The CLI and zero-install script use different evidence sets for weak project markers; feeding those attributions into ranking introduces a new `agent_name_candidates` parity failure.

Both focused regression tests pass on the base commit and fail on this head. The first exercises actual `init --write`, without `--allow-unresolved-scope`; neither reproduction uses test or template code.

Validation: discovery and zero-install tests completed with 389 passed / 1 skipped; static-only adapter tests: 458 passed; changed Python files pass Ruff and the diff passes whitespace checks. The repository verifier reports `control_state: complete`, `decision: passed`, and self-check reports ready. Those checks do not cover the two reproduced ranking regressions. GitHub's three reported PR workflows are also green.

Submitted as a comment review because the connected GitHub account is the PR author.

src/agents_shipgate/cli/discovery/signals.py
```diff
@@ -2127,19 +2274,37 @@ def _rank_agent_name_candidates(
...
+        blocking = sorted(
+            (
+                project
+                for project in declared_in[value]
+                if project in blocked_projects
+            ),
+            key=lambda project: project != best_project[value],
+        )
+        if not blocking:
+            continue
```
pengfei-threemoonslab, line 2298: **[P1] Keep unclaimed name candidates from bypassing a blocked product root**

A strong marker can put a name in a project that `agent_project_candidates` never discovered. Reproduction: root `agent.py` imports ADK, defines `build()` returning `LlmAgent(name="ActualRoot")`, and assigns `root_agent = build()`; `util/pyproject.toml` names a utilities package, and `util/agent.py` does `from domain import Agent; helper = Agent(name="CrmHelper")` using a local, unrelated `Agent` class. `_agent_evidence_paths` correctly excludes that library from the agent projects, so detection reports `agent_scope: single` and only project `.`. However, this loop exempts `CrmHelper` because its attributed directory is `util`, not the blocked `.`. Actual `init --workspace <root> --write --json` therefore writes `agent.name: CrmHelper`; the base commit retains `CHANGE_ME`.

This bypass needs neither `--allow-unresolved-scope` nor the test/template-only case deferred to #533. Keep the refusal for the selected agent project when the alternative is only an unclaimed name suggestion; sharing the marker lookup is insufficient unless the eligibility decision also respects which agent projects were established. Mirror the correction in the standalone ranker.

tools/shipgate-detect.py
```diff
@@ -3691,8 +3827,15 @@ def detect(workspace: Path) -> dict[str, Any]:
...
+    attribution = _ProjectAttribution(workspace, evidence_paths)
     agent_project_candidates = _agent_project_candidates(
-        workspace, evidence_paths, literals_by_path
+        workspace, evidence_paths, literals_by_path, attribution
+    )
+    # Ranking runs after the grouping because it reads the same attribution:
+    # an application root it cannot resolve disqualifies the names in *that
+    # project* and no others (#398).
+    name_candidates = _rank_agent_names(
+        py_facts, workspace, project_names, attribution
```
pengfei-threemoonslab, line 3902: **[P2] Match the CLI's evidence set before using attribution for ranking**

The new ranking input is built from `literals_by_path`, which includes names from files with no supported framework signal. The CLI's `_agent_evidence_paths` only adds these literal sites when `fact.framework` is true. That difference now changes eligibility: use a product `agent.py` with a factory-assigned unreadable ADK root, plus `util/requirements.txt` containing `requests` and `util/agent.py` importing a local domain `Agent` and constructing `Agent(name="CrmHelper")`. The script treats the literal as evidence that activates `util/requirements.txt`, attributes the helper to a separate project, and emits `CrmHelper` with `selectable: true`. The CLI does not activate that weak boundary, attributes it to the blocked workspace, and emits `selectable: false`.

The full `agent_name_candidates` equality test passes on `db63e42b` and fails on this head, so the new attribution dependency breaks the documented byte-for-byte parity even though the underlying project-evidence mismatch predates this PR. Use the same framework-qualified evidence on both sides and add this non-framework/weak-marker parity case.

src/agents_shipgate/cli/discovery/signals.py
pengfei-threemoonslab: Confirmed and fixed in 6bbe22f8. Reproduced exactly as described — `agent_scope: single`, only project `.` in `agent_project_candidates`, and `init --workspace <root> --write --json` writing `agent.name: CrmHelper` where `db63e42b` keeps `CHANGE_ME`.

You named the root cause precisely: sharing the marker lookup is not enough, because a *marker* and a *manifest scope* are different things. `find_project_root` is the right boundary for grouping evidence — it decides whether two agents are one manifest's business — and the wrong one for deciding eligibility. A utilities package with its own `pyproject.toml` and no agent evidence is not a scope anything could be written for, so a name found there belongs to the scope that encloses it.

`_ProjectAttribution` now carries the projects `_agent_project_candidates` actually established (its grouping keys, registered as it returns — that function is the only thing entitled to say what an agent project is here), and ranking resolves through `scope_of()`: the nearest *established* project at or above the file, falling back to the workspace, which is the fail-closed reading for a name under no project of its own. Both the block and the candidate attribution go through it, so a poison file in a non-scope directory now blocks the project enclosing it rather than a directory nothing can be written for. `scope_of()` raises if it is read before the grouping has run — answering early would silently attribute every name to the workspace, which is the same class of quiet wrongness this finding is about. Mirrored in the standalone ranker.

Two guards: `test_a_name_under_a_non_agent_marker_cannot_escape_the_blocked_scope` asserts the premise (`agent_scope == "single"`, `agent_project_candidates == ["."]`) before asserting the refusal, and `test_init_write_refuses_a_name_under_a_non_agent_marker` drives the actual command — no `--allow-unresolved-scope`, no test or template code — and asserts `CHANGE_ME` is kept and `CrmHelper` appears nowhere in the manifest. Both fail on `6a9c019b`.

tools/shipgate-detect.py
pengfei-threemoonslab: Confirmed and fixed in 6bbe22f8, at the root rather than at the ranking layer.

Reproduced: with `util/requirements.txt` and a non-framework `Agent(name="CrmHelper")`, the script emits `selectable: true` and the CLI `selectable: false` — full `agent_name_candidates` equality passes on `db63e42b` and fails on this head, as you said.

Chasing the underlying mismatch was worth it, because it is larger than a ranking difference. The script counted **every** literal as project evidence where the CLI counts only framework-attributed ones (`if literals and fact.framework`), so on both reproductions it already emitted a phantom `agent_project_candidates` entry — `[('.', ['ActualRoot']), ('util', ['CrmHelper'])]` against the CLI's `[('.', ['ActualRoot'])]` — and through it `agent_scope: "ambiguous"` where the CLI says `"single"`. That is a *verdict* disagreement on the surface whose entire registered claim is `agent_project_verdict`, sitting in `agent_project_candidates`, which `test_script_verdict_matches_cli` pins. It never fired because every sample has a `pyproject.toml` at its root and no stray non-framework literal.

So rather than special-casing the ranking input, the script now qualifies `literals_by_path` itself, which is the same dict feeding both the evidence set and each project's `agent_names`. Framework attribution is snapshotted immediately after the Python scoring pass — before package tokens, glob hits and the MCP source add candidates of their own — because that is exactly the window `_score_python_signals` measures on the CLI side; taking it later would have counted a `.py` file claimed by `mcp_server_source` as framework-attributed where the CLI does not.

Added as two parity cases, `non_agent_strong_marker` and `non_agent_weak_marker`, both asserting the CLI's `agent_scope`/`agent_project_candidates` premise alongside the byte-for-byte `agent_name_candidates` equality. Both fail on `6a9c019b`.

pengfei-threemoonslab:
Both review findings are fixed in 6bbe22f8. I reproduced each on the branch and confirmed each is a regression from this PR — `db63e42b` behaves correctly on both.

**P1 — a name under a non-agent marker escaped a blocked scope.** Scoping resolved "which project" with the nearest project *marker*. That is the right boundary for grouping evidence and the wrong one for eligibility: a utilities package with its own `pyproject.toml` and no agent evidence is not a manifest scope. `_ProjectAttribution` now carries the projects `_agent_project_candidates` actually established, and ranking resolves through `scope_of()` — the nearest *established* project at or above the file, workspace as the fail-closed fallback. `scope_of()` raises if read before the grouping runs, since answering early would quietly attribute everything to the workspace.

**P2 — the two implementations used different evidence sets.** Fixed at the root: the script now framework-qualifies `literals_by_path` the way the CLI does. Following the mismatch down was worth it, because it was already a **verdict** break, not only a ranking one — the script emitted a phantom `agent_project_candidates` entry and `agent_scope: "ambiguous"` where the CLI says `"single"`, in the field that carries this surface's entire registered claim. It never fired because every sample has a root `pyproject.toml` and no stray non-framework literal.

**Four new guards**, all failing on `6a9c019b` and passing here: two CLI tests (one asserting the `agent_scope`/`agent_project_candidates` premise before the refusal, one driving actual `init --write` and asserting `CHANGE_ME` is kept and `CrmHelper` appears nowhere in the manifest), and two parity cases covering the strong- and weak-marker shapes.

**Re-verified**: #398's own reproduction still resolves — repository-wide, `financial_coordinator`, `rag_root_agent` and `market_watcher` stay selectable, and both scoped runs return their real agent. The eight-case CLI↔script probe still agrees on every shape. Full local gate green (`ruff`, `pytest -n auto -m "not perf"`, `test_adapter_static_only.py`, `test_latency_budget.py -m perf`).

Thanks — the second finding in particular found a pre-existing hole in the surface contract that nothing was covering.

pengfei-threemoonslab:
Rebased onto `faeb91f0` (#531 landed while the fixes were in flight) and force-pushed as `9ba57199`. The only overlap was both branches adding a bullet at the top of `## Unreleased`; both are kept, #475's first.

Worth flagging because it wasted ~20 minutes: **CI reported nothing at all on `6bbe22f8` — no queued run, no pending check, `total_count: 0` on the commit's check-runs.** The cause was `mergeable: CONFLICTING`, not a queue delay; GitHub silently declines to schedule `pull_request` workflows for a conflicting head, so "no checks reported" reads identically to "CI has not started yet". Checking `gh pr view --json mergeable` is the way to tell the two apart.

All checks green on the rebased head, and the local gate was re-run after the rebase (`ruff`, `pytest -n auto -m "not perf"`, `test_adapter_static_only.py`) — green.

Issue #492, opened 2026-09-02
Two thirds of docs/ is superseded schema versions: 151 files across 34 families, 43 of them one family


pengfei-threemoonslab:
## Summary

`docs/` holds **151 versioned schema files across 34 families**, produced in four months. The largest families:

| family | committed versions |
|---|---|
| `report-schema` | **43** |
| `packet-schema` | 18 |
| `verifier-schema` | 16 |
| `capability-lock-schema` | 8 |
| `agent-handoff-schema` | 8 |
| `capability-lock-diff-schema` | 7 |
| (28 further families) | 1–5 each |

`docs/` contains 225 files in total, so **two thirds of the documentation directory is superseded machine artifacts**. `scripts/generate_schemas.py` writes them from the Pydantic models with a `--check` drift mode, so none of this is hand-maintained — the generator is good, and it is doing exactly what it was told. Nothing tells it that a version nobody reads any more should stop being shipped in the docs directory a human browses.

Two costs, both already visible:

1. **A human cannot find the current contract.** 43 files named `report-schema.v0.*.json` in one flat directory offer no signal about which is live.
2. **A version number that increments this often stops meaning "compatibility boundary" and starts meaning "build counter."** #469 was filed precisely because two upcoming surfaces were about to define the same payload twice; that instinct should become policy before 1.0.

## Proposal

**1. Retention.** Keep the current version of each family in `docs/`; move superseded versions to `docs/archive/` (still committed, still linkable from `STABILITY.md`, out of the browsing path). The generator gains the archival step so retention is enforced, not curated.

**2. One index.** A generated `docs/schemas.md`: family, current version, what it describes, which command emits it, where the migration notes live.

**3. A freeze window.** Between two tags, a family's version does not move more than once. A second bump inside one release cycle is a signal that the first was premature — that is the #469 lesson generalized.

**4. Consolidation, measured first.** Some of the 34 families are projections of one another. Before merging any of them, produce the list of which families share a payload — then consolidate only where a shared frozen payload provably serves both, the way #469 did for the delta and the lock.

## Acceptance

- [ ] `docs/` holds exactly one version per family; superseded versions live under `docs/archive/` and every `STABILITY.md` link still resolves.
- [ ] `scripts/generate_schemas.py --check` still detects drift, and additionally fails when a superseded version is left in the browsing path.
- [ ] A generated schema index exists and is regenerated by the same script.
[...]

PR #534, opened 2026-09-06
Reach one useful review from the human entry path (#498)

### LINKED ISSUE #498: The human quickstart must reach a useful review before a documentation-site rebuild

pengfei-threemoonslab:
Closes #498.

## The problem, as found

The README was 933 lines with 28 second-level sections; the quickstart opened with three commands before the reader had seen a result. That was the reported problem. Underneath it were three defects nobody was checking, because `README.md` and `docs/quickstart.md` were recorded in [`docs/distribution-surfaces.md`](docs/distribution-surfaces.md) as "repository documentation" and registered as a surface **nowhere**:

1. **The quickstart's first command fails on the build the same page tells you to install.** `shipgate check --format agent-boundary-json` — `v0.15.0` accepts only `codex-boundary-json`. Read out of the tag, not asserted: `git show v0.15.0:src/agents_shipgate/cli/check.py`.
2. **It told a coding agent to fill a human-owned declaration from prose.** "Replace `agent.name` and `agent.declared_purpose` from the prompt, main agent file, or repo README" — `agent.declared_purpose` is `placeholder_owner() == "human"`. The same wording was in `AGENTS.md`, `docs/agent-recipes.md` and `docs/troubleshooting.md`.
3. **The README's flagship example was fabricated.** "What your PR sees" showed a `### Agents Shipgate result: block` heading, an `Impact | Change | Subject | Why` table and a numbered *Required before merge* list, and said the fixture wrote it "verbatim" to `reports/pr-comment.md`. No code path renders any of it, the path is wrong, and `block` is not a value `merge_verdict` or `release_decision.decision` ever takes.

Each is one of #497's two claim families — verdict parity and executability — on the surface a stranger reaches first.

## What changed

**README → landing page** (937 → 384 lines). One before/after capability change (the real MCP-export diff from `samples/ai_generated_refund_pr`), one demo, one install with the channel caveat, what static evidence cannot prove, the accuracy numbers that are still zero, and a five-row audience-routing table. Every quoted line now comes from the artifact.

It is not the ~150 the issue sketched, and the issue's own "do not delete useful evidence to meet a line-count metric" is why: the accuracy disclosure keeps every caveat including `benign_escalation_rate` and the AI-adjudicated labels, the supported-inputs table and the `report_schema_version` line are pinned by existing guards, `#for-coding-agents` is deep-linked from `docs/agent-action-guide.md`, the Adopters section arrived from #531, and 19 lines are the retired-anchor map. Everything that could leave, left.

**`docs/quickstart.md` → one review, end to end** (380 → 643 lines; it absorbed the README's adoption material). It opens with *which build you get* — the published release, the unqualified preview and its absent qualification, the source checkout — then walks the committed `ai_generated_refund_pr` sample in seven steps that map onto the issue's four questions: what changed, why the top result matters, what the run did not establish, **exit zero is not merge permission** (advisory `would_fail_ci=false` on a `blocked` verdict), and who owns the next action. Only then does it reach the two adoption routes — Route H needs no manifest, so policy authoring is not a universal prerequisite. Safe mechanical fixes and human-owned declarations are separate sections. "The second use" covers the next PR, advisory CI, and baseline/strict, and states plainly that there is no in-repo approval mechanism (#337 not shipped).

**Displaced material moved, with a mapping table.** All 21 retired README anchors are listed in a *Moved out of this README* table pointing at their new homes: the authorization protocol to `agent-contract-current.md` (which already documented it in full), the CI recipes to `integrations.md`, the coding-agent kits to `docs/agents/README.md`, the paste-in prompt to `target-repo-agent-snippets.md`, "Why Not Just…" to `category.md`, the findings gallery to `samples/README.md`. `llms-full.txt` regenerated; `llms.txt` now names the in-repo quickstart.

## The guards

Documentation rots unless something reads it, which is how all three defects survived. So:

- **Two surfaces registered.** `human_entry_path` (`README.md`, `docs/quickstart.md`) claims `executable_pin`, `contract_floor`, `placeholder_ownership` and both vocabularies; `agent_instructions` (`AGENTS.md`, `docs/agent-recipes.md`, `docs/agents/`, `docs/target-repo-agent-snippets.md`) claims `executable_pin` and `placeholder_ownership`. Both were in `NOT_A_DISTRIBUTION_SURFACE`; they are not any more.
- **`test_the_human_entry_path_states_what_the_published_build_provides`** reads the `--format` values `check` accepts *at the published tag* — the way #506 reads that tag's `CONTRACT_VERSION` — and fails if the entry path teaches one it does not without saying so. `test_the_published_format_reader_sees_the_difference_it_exists_for` is its non-vacuity control: when a release finally carries the current option set, that test fails and the caveat comes out.
- **`test_the_entry_path_quotes_lines_the_pr_comment_actually_renders`** runs the fixture and compares every quoted comment line against the artifact, per rendered line, with Markdown escaping and hard wrapping normalised.
- **The vocabulary reader learned Markdown tables.** The braced-literal reader could not see either of the quickstart's verdict tables, which is the only shape the entry path uses. Deleting the `unknown` row now fails.

Three prose guards were made whitespace-insensitive (`_resolve_freeform`, the contract-floor and release-literal checks): a multi-word alias resolved or not depending on where a paragraph happened to wrap, which is a claim about the shape of the prose, not its content.

## Verification

Each new guard was replayed against the pre-fix content and fails there:

| Guard | On pre-fix `README.md` / `docs/quickstart.md` |
| --- | --- |
| `test_surface_routes_human_owned_placeholders_to_a_human[docs/quickstart.md]` | FAILS — `agent.declared_purpose` unrouted, blanket `Replace any CHANGE_ME` |
| `test_the_human_entry_path_states_what_the_published_build_provides` | FAILS — no channel table |
| `test_the_entry_path_quotes_lines_the_pr_comment_actually_renders` | FAILS — `'### Agents Shipgate result: block Decision: \`block\` · Risk: \`critical\`…'` |
| `test_surface_enumerations_match_the_engine_vocabulary` | FAILS on a quickstart with the `unknown` row deleted |
| the `--format` sweep | FAILS with the "accepts only" caveat removed |

I walked the quickstart as a cold reader, which found two of my own defects: `fixture run` writes to a temporary directory it names on stdout (not `agents-shipgate-reports/`, and a *relative* `--out` resolves inside the fixture copy), and the `fix_task` excerpt was abridged without saying so. Both fixed. Route H's `audit --host` / `--save-baseline` / `--drift`, Route A's `verify --preview` / `check`, and every flag the two pages print (`init`, `apply-patches`, `feedback export`, `baseline save`, `scan`, `verify`, `check`) were run or read off `--help` against this build.

Full suite green (`pytest tests/ -n auto`, exit 0); `ruff check .` clean; all CI checks pass.

## Review rounds

Three passes over the branch; every finding is fixed in this PR.

**Round 1 — three findings, all in the new guards or the new prose.**

- `_first_column_sets` required a trailing newline on every table row, so a verdict table that *ends a file* lost its last row: the complete five-row table returned four members and failed a document that was correct. Reproduced directly; the newline is optional now.
- `_published_check_formats` understood only `format_ == "literal"`. The next release will likely spell acceptance the way this tree does (`format_ not in CHECK_FORMATS`), and the old reader would have returned the option's default alone — a *partial* set, which is worse than an empty one because it passes the emptiness assertion and silently stops detecting a stale caveat. A name is now resolved when its module-level binding is a set of string literals; any other shape raises.
- The retired-anchor table — this change's own answer to keeping public anchors reachable — omitted five of the 21 anchors and claimed the framework notes had moved when half of them stayed. Checked now by diffing the old README's heading slugs against the table; zero unlisted.

**Round 2 — two findings in material I moved rather than wrote.**

- `docs/agents/README.md` carried "require contract v15 or newer" for the Codex plugin. That number arrived in #271, is stated nowhere else, and contradicts the kit it describes: `adoption-kits/codex-skill/SKILL.md` renders `{{ minimum_control_contract_version }}`, which is 21. Replaced with the check rather than a literal, so there is no second copy of the number to go stale.
- `docs/integrations.md`'s strict-mode block said "save a baseline first" and then showed a bare strict scan as step one. The three commands are numbered now and say what each is for; step one is expected to exit 20, and that is the point of running it.

**Round 3 — no new findings.** Anchor and file-link sweeps over every changed Markdown file report zero broken links introduced.

**Round 4 — four P2 findings from the PR review, all reproduced and fixed.**

- **Ownership was routed through a field the CLI does not emit.** `placeholders[]` entries carry `path`, `current`, `line` — `collect_placeholders()` emits no `owner`, and `placeholder_owner()` is internal, so an agent following "read each entry's owner" could not perform the lookup. The routing the engine *does* publish is `control.next_action`: `actor: "human"`, `permissions.edit: false`, and a `why` naming the human-owned fields and their line numbers. `AGENTS.md`, the quickstart, the recipes and troubleshooting now say that, with the message quoted so a reader recognises it; `llms-full.txt` regenerated.
- **The walkthrough promised the published release and quoted the source tree.** Running the fixture from an extracted `v0.15.0` shows the gap: `Capability delta: +2, 3 modified, -0` with a flat change list and no `support.search_kb` row, `Evidence coverage: static (human review recommended)` with no counts, and the static-verdict boundary printed *nowhere* — not in the comment, the artifacts, or stdout. Steps 3, 4, 5 and 7 now name `v0.15.0` and show what it prints; the scoping sentence claims only what holds on both channels (same commands, same verdict, same exit `0`). The README carries the same two notes.
- **A scaffold was read as a route.** `tool_surface_origin: "scaffold"` says discovery could not *read* a surface, not that there is none. Rebuilding the `_fastmcp_server` reproduction standalone: `init` scaffolds, and `audit --host` returns three generic `instruction_trust_root` grants — `grep -c budgets` on that output is `0`. The quickstart keeps those repositories on Route A and asks for an export or inventory; the same inference is corrected at its source in the design-partner runbook.
- **The pointer does not validate itself.** After an empty commit, `agents-shipgate agent control` exits `4` with `workspace_changed` while reading `current-control.json` still returns the old `review_publishable`. This was a regression I introduced compressing the README's correct sentence; `integrations.md`, `docs/agents/README.md` and quickstart step 7 now put the command first and say what the file alone cannot tell you.

**Round 5 — two more P2 findings from the follow-up review, both in round 4's fixes.**

- **`control.next_action.why` is bounded, so absence from it grants nothing.** Reproduced with `doctor --json` on eight unresolved placeholders: the sentence names three paths then `and 3 more in placeholders[]`, and the three it dropped (`agent.prohibited_actions[1]`, `[2]`, `permissions.scopes[0]`) are human-owned too. "The rest of `placeholders[]` is repository reading" therefore handed real declarations to the agent — the fail-open round 4 removed, reintroduced by the sentence that removed it. Every surface now routes on `control.next_action.actor`, which routes the **turn**: verified in both directions — any unresolved human-owned value gives `actor: "human"` / `permissions.edit: false`, and only once they are all supplied does a re-run give `actor: "coding_agent"` with the field to replace.
- **Step 7's command did not run where step 2 left the reader.** `fixture run` writes into a temporary copy and reports under `<copy>/reports`, so `--workspace .` looked for `./agents-shipgate-reports` in the caller's directory and exited `3`. The step now prints the `cd` and `--reports-dir reports`, sourced from the run's own `Fixture copy left at` and `Reports:` lines, and names the envelope's real spelling — `agent control` returns top-level `control_state`, the pointer file is nested `control.state`. `v0.15.0` has no `agent control` subcommand at all (`agent` is `handoff` only), which the released-build note now says.

Also fixes the second copy of the scaffold-implies-Route-H inference, in the runbook's partner agent prompt — one file over from the copy corrected in round 4.

Two more guards, both replayed against the text they replace:

- `test_surfaces_say_the_routing_prose_elides` — the elision marker is asserted to be the engine's own string via `_placeholder_review_why`, and the guard is scoped to surfaces that route on the bounded prose rather than every file naming both tokens (the setup prompts and the runbook enumerate the human-owned fields by name and derive nothing from `why`).
- `test_the_walkthrough_control_command_runs_in_the_walkthrough` — reads step 7's flags out of the page and runs them where step 2 leaves the reader, taking the printed `cd` literally. Its first draft passed `--workspace <abs>` without changing directory and disagreed with a hand-run terminal, because `--reports-dir` is relative to cwd.

Two guards changed in round 4 as a result, both strictly stronger:

- `test_the_entry_path_quotes_lines_the_pr_comment_actually_renders` now renders **both** engines — the reviewer was right that one could not catch the channel mismatch. A quoted line must come from some channel, and a line only one channel renders must sit in a step naming the release. Tag extraction skips cleanly if the release's dependencies do not resolve on the running interpreter. Its section scanner had to learn to mask fenced blocks: the PR-comment excerpt quotes the comment's own `## Agents Shipgate` heading, so the first draft placed the block inside itself.
- `test_runbook_routes_host_boundary_partners_away_from_a_manifest` pinned the disproven sentence as a literal and failed. It keeps its property — the runbook must still say what to do on a scaffold — pinned now on "Never invent a manifest." and on naming the input that would let `verify` read the surface, plus a new assertion that the old inference is gone.

## Also fixed, same class

- Four `uvx agents-shipgate@0.18.0` pins in `docs/incidents/` and `samples/README.md` named a version the index does not carry. `agent_weakens_gate` *is* in `v0.15.0` and now says so; the other two name the checkout and say what to pin once a release carries them.
- `docs/integrations.md` and `docs/agents/README.md` told a reader to read `agent-handoff.json` first; the contract's `agent_read_order[0]` has been `current-control.json` since v20.
- `docs/agent-recipes.md` described a workflow-collision check as `ThreeMoonsLab/agents-shipgate@*`, which reads as a pin.

## Not done, deliberately

- **No timing was recorded in #521.** The 10 minutes is stated as a target, never as a result, on both pages. Recording an observation needs a real cold reader who is not the author; manufacturing one would be the thing #521 exists to prevent.
- **No MkDocs deployment and no `docs/` schema archival** — #498 lists both as non-dependencies, and #492 stays independently useful.
- **`docs/incidents/` and `samples/README.md` stay unregistered.** Their pins are fixed but still unguarded, which is why they were wrong. Registering them means reconciling `samples` in `NOT_A_DISTRIBUTION_SURFACE`, which is a third registry change in one PR; filed as a follow-up instead.
- **Eight pre-existing broken in-page anchors** (`AGENTS.md` and `docs/agent-contract-current.md` → deleted `STABILITY.md` sections) are untouched: their targets no longer exist, and choosing new ones is not this change's context. Happy to file.

## Note on this PR's own verdict

`./shipgate verify` on this branch returns `review_required` / `human_review_required`, `control.state: review_publishable` — it touches `AGENTS.md`, a trust root, and `SHIP-VERIFY-AGENT-INSTRUCTIONS-WEAKENED` cannot rule out weakening statically. That is the correct routing: the `AGENTS.md` change *narrows* what a coding agent may fill in, and a person should confirm that.


AGENTS.md
```diff
@@ -565,12 +565,17 @@ Otherwise proceed to `init`. MCP/OpenAPI tool-surface repos and Codex plugin pac
...
 pipx install agents-shipgate
-agents-shipgate init --workspace . --write
-# edit shipgate.yaml to replace any CHANGE_ME values
+agents-shipgate init --workspace . --write --json
+# resolve the placeholders init reports, by owner (below), then:
 agents-shipgate scan -c shipgate.yaml
 ```
 
-`init` writes a manifest with `CHANGE_ME` placeholders for `agent.name` and `agent.declared_purpose`. Replace them by reading the agent's prompt or main file.
+`init --json` returns `placeholders[]`, and **each entry names its owner**. Read that field; do not treat the array as one list of edits. The rule behind it:
```
pengfei-threemoonslab: [P2] Route ownership through a field the CLI actually publishes

`placeholders[]` does not contain an `owner` field, even on this PR's current build. Replaying `init --write --json` yields entries such as `{"path":"agent.declared_purpose[0]","current":"CHANGE_ME","line":13}`; `collect_placeholders()` emits only `path`, `current`, and `line`, and `placeholder_owner()` is used internally. The new instruction makes this nonexistent per-entry value authoritative over the prose, so an agent following it cannot perform the required ownership lookup. Document the emitted `control.next_action.actor` routing, or add and test a public per-placeholder ownership field before teaching it. The same claim needs correcting in the quickstart, recipes, troubleshooting, and regenerated `llms-full.txt`.

docs/quickstart.md
```diff
@@ -1,310 +1,526 @@
...
-  | python3 - --workspace . --json
+Open `pr-comment.md` — the same text the GitHub Action posts on a pull
+request. It leads with the capability delta, by subject:
+
+```text
+- Capability delta (analysed surface): 2 subjects across 6 changes (+1 added, 2 modified, -0 removed)
+- Top capability changes by subject:
+  - `stripe.create_refund`: added action — blocks release; added tool — blocks
+    release; broadened action destructive — high-risk effect destructive added;
+    blocks release; …
```
pengfei-threemoonslab, line 158: [P2] Match the walkthrough excerpts to the promised release channel

The page explicitly says this walkthrough runs on published `v0.15.0`, but these are current-tree results. I ran `fixture run ai_generated_refund_pr` from the `v0.15.0` tag: its comment says `Capability delta: +2, 3 modified, -0`, has no per-subject grouping or `support.search_kb` binding-change row, and its report says only `Evidence coverage: static (human review recommended)` rather than the reachability/pass-eligibility counts quoted in step 5. A reader using the recommended install cannot find the evidence that steps 3–5 ask them to interpret. Quote the released artifacts for the release walkthrough, or explicitly mark these steps as requiring a newer build and supply the released alternative. The new fixture quotation test runs only this tree's engine, so it currently cannot catch this channel mismatch.

docs/quickstart.md
```diff
@@ -1,310 +1,526 @@
...
+**When discovery reads no tool surface at all.** `init --json` reports
+`tool_surface_origin`: `"detected"` when every source in the manifest was read
+out of the workspace, `"scaffold"` when none was, and `null` when this run's
+render reached neither disk nor the payload. On `"scaffold"` the `tool_sources`
+block is a placeholder — `id`, `type` and `path` are all `CHANGE_ME`, all three
+are in `placeholders[]`, and `manifest_message` says nothing was inferred. That
+is the common outcome for an MCP server whose tools are registered in code
+rather than exported to a file. **A scaffolded `tool_sources` block means this
+repository is on Route H**: record the dead end and switch, rather than
+inventing a manifest to fill.
```
pengfei-threemoonslab: [P2] Keep undiscovered tool surfaces on the tool-review route

A scaffold means discovery could not supply a source; it does not establish that the repository only has a coding-host boundary. The existing `_fastmcp_server` reproduction in `tests/test_init_scaffold_disclosure.py` demonstrates this directly: it defines a Python MCP tool, `init --json` returns `tool_surface_origin: "scaffold"`, and the suggested `audit --host --json` returns `grants: []` with no observed host sources. Following this instruction therefore stops reviewing the actual MCP capabilities and substitutes an audit that never examines them. Keep these repositories on Route A and request an explicit MCP export/local inventory or report unsupported extraction; switch to Route H only when the user's intended subject is genuinely the coding-host configuration.

docs/integrations.md
```diff
@@ -100,8 +100,10 @@ Action outputs:
 The action runs `agents-shipgate verify`, which writes Markdown, JSON, SARIF,
 packet JSON, verifier JSON, verify-run JSON, and PR-comment Markdown
 artifacts. It intentionally emits `packet.json` only for the packet;
-`pr-comment.md` is the human PR surface. Read `agent-handoff.json` first for
-the compact agent handoff, `verifier.json` for detailed control context,
+`pr-comment.md` is the human PR surface. Read `current-control.json` first —
+it names which run is current and refuses the read when HEAD or the working
+tree has moved since the decision — then `agent-handoff.json` for the compact
```
pengfei-threemoonslab: [P2] Run the control validator before relying on the pointer

Reading `current-control.json` cannot refuse stale evidence: it is an ordinary persisted file. In an isolated fixture, I made an empty commit after verification; loading the pointer still returned the old control state, while `agents-shipgate agent control --workspace ... --reports-dir ...` exited 4 with `workspace_changed`. The new wording attributes that validation to the raw read and omits the command that performs it, allowing a consumer to treat an intact old artifact set as current after HEAD changes. Make `agents-shipgate agent control` the first step, require a successful read, and then inspect the bound artifacts. The same wording was added to `docs/agents/README.md` and step 7 of the quickstart.

pengfei-threemoonslab (commented):
Reviewed all 28 changed files at `76dd9215c9dba986728164717fe9262732612a7a`. I recommend addressing the four P2 findings below before merge: the new ownership instructions refer to a field the CLI does not emit; the published-release walkthrough quotes source-checkout output; a discovery failure is incorrectly treated as a host-only repository; and the artifact-reading guidance attributes freshness validation to a plain JSON read.

Validation: all **847 tests passed** in the four affected test modules (`test_distribution_surface_parity.py`, `test_docs_links.py`, `test_fixture.py`, and `test_public_surface_contract.py`, using `pytest -n 4`); `ruff check .` and `git diff --check` passed; `self-check --json` returned `ready: true`. I also replayed the refund fixture from the `v0.15.0` tag, exercised `init` and host audit against the existing FastMCP reproduction, and reproduced stale-pointer behavior after an empty commit. These checks found gaps that the current documentation guards do not cover.

Agents Shipgate verification against base `faeb91f0eb433868ae571bb640a11dd0401b062e`: **`control_state: review_publishable`**, **`merge_verdict: human_review_required`**, **`release_decision: review_required`**. Publishing this review is allowed; merge and completion remain denied pending human review of the trust-root change. This review does not approve that change.

Submitted as a comment review because the connected account is the PR author.

AGENTS.md
pengfei-threemoonslab: Confirmed and fixed. Replaying `init --write --json` on a FastMCP scratch repo returns exactly what you describe — `{"path": "agent.declared_purpose[0]", "current": "CHANGE_ME", "line": 13}`, no `owner` — so the instruction pointed at a lookup an agent cannot perform.

Documented the emitted routing instead of adding a field. The same run publishes:

```json
"control": {"next_action": {"actor": "human", "kind": "review",
  "why": "In .../shipgate.yaml: line 13 (agent.declared_purpose[0]) must be supplied by a human. ..."}}
```

`AGENTS.md`, `docs/quickstart.md`, `docs/agent-recipes.md` and `docs/troubleshooting.md` now say the split plainly: `placeholders[]` is a *location* list (`path`, `current`, `line`, no owner); `control.next_action` is the routing, and its `why` names the exact fields and lines a person must supply. The quickstart quotes that message so a reader recognises it. `llms-full.txt` regenerated in the same commit.

docs/quickstart.md
pengfei-threemoonslab: Confirmed and fixed — this was the sharpest of the four, because it is the exact defect the PR exists to close, committed by the fix itself.

I extracted `v0.15.0` (`git archive`) and ran the fixture on it. Your reading is right, and there is a fourth difference: it prints the static-verdict boundary sentence *nowhere* — not in the comment, not in the artifacts, not on stdout. It carries the `## Disclaimer` at the foot of `report.md` instead.

The scoping sentence now claims only what is true of both channels — same commands, same verdict, same exit `0` — and steps 3, 4, 5 and 7 each name `v0.15.0` and show what it prints:

- step 3: `Capability delta: +2, 3 modified, -0` and the flat change list, no `support.search_kb` row;
- step 4: same four check IDs, two messages worded differently ("adds destructive capability without rollback controls") — match on the ID, not the sentence;
- step 5: `Evidence coverage: static (human review recommended)`, no counts, plus where the boundary actually lives;
- step 7: `fix_task` is identical in `actor` / `safe_to_attempt`.

The README carries the same two notes for its own excerpts.

On your last sentence — the guard could not catch this, so it now renders **both** engines. A quoted line must come from one of them, and a line only one renders must sit in a step naming the release. Replayed against the pre-fix text and against a fabricated "v0.15.0" line; both fail. Tag extraction skips cleanly if the release's dependencies do not resolve on the running interpreter. Its own section scanner had to learn to mask fenced blocks first — the PR-comment excerpt quotes the comment's `## Agents Shipgate` heading, so the first draft placed the block inside itself.

docs/quickstart.md
pengfei-threemoonslab: Confirmed and fixed. I rebuilt your `_fastmcp_server` case standalone: `init --json` returns `tool_surface_origin: "scaffold"`, and `audit --host --json` on the same workspace returns three `instruction_trust_root` grants for claude-code/cursor/codex — `grep -c budgets` on that output is `0`. The tool the reader came to review never appears.

The quickstart now says a scaffold means discovery could not *read* a surface, not that there is none, keeps those repositories on Route A, and asks for an exported MCP tool list, an OpenAPI spec or a local tool inventory — switching to Route H only when the coding-host configuration is genuinely the subject.

I also fixed it at its source: I took the sentence from `docs/design-partner-verifier-pilot.md`, which carried the same inference ("stop: this repository is on Route H"). That paragraph is corrected too.

`test_runbook_routes_host_boundary_partners_away_from_a_manifest` pinned the disproven sentence as a literal, so it failed. It keeps its property — the runbook must still say what to do on a scaffold rather than leave a partner to invent a manifest — pinned now on "Never invent a manifest." and on naming the input that would let `verify` read the surface, plus a new assertion that the old inference is *gone*. Strictly stronger than before, and its docstring records the FastMCP evidence.

docs/integrations.md
pengfei-threemoonslab: Confirmed and fixed, and it was a regression I introduced: the pre-#498 README said this correctly — "it names which run is current, **and `agents-shipgate agent control --workspace .` validates it** … refusing the read when HEAD or the working tree has moved". I compressed that into three copies that dropped the command and left the refusal attributed to the file.

Reproduced your scenario in an isolated fixture — verify, then `git commit --allow-empty`:

```
agent control  -> exit 4: workspace_changed ("head_commit_sha is 1da845b4…,
                  but the pointer was published against 27692be3…")
plain JSON read -> control.state = review_publishable   (stale)
```

`docs/integrations.md`, `docs/agents/README.md` and quickstart step 7 now put `agents-shipgate agent control --workspace .` first, require it to succeed, name the exit code and the `workspace_changed` reason, and say explicitly that reading the pointer file does not check freshness. Only then do they describe reading the bound artifacts.

(The README was already correct and is unchanged.)

AGENTS.md
```diff
@@ -565,12 +565,17 @@ Otherwise proceed to `init`. MCP/OpenAPI tool-surface repos and Codex plugin pac
...
 pipx install agents-shipgate
-agents-shipgate init --workspace . --write
-# edit shipgate.yaml to replace any CHANGE_ME values
+agents-shipgate init --workspace . --write --json
+# resolve the placeholders init reports, by owner (below), then:
 agents-shipgate scan -c shipgate.yaml
 ```
 
-`init` writes a manifest with `CHANGE_ME` placeholders for `agent.name` and `agent.declared_purpose`. Replace them by reading the agent's prompt or main file.
+`init --json` reports the placeholders two ways, and only one of them routes. **`placeholders[]` is a location list** — each entry is `path`, `current` and `line`, and carries no owner. **`control.next_action` is the routing**: while any human-owned value is unresolved it returns `actor: "human"`, `permissions.edit: false`, and a `why` naming those exact fields and their line numbers. Read `control.next_action` and treat the fields it names as a person's; everything else in `placeholders[]` is yours. The rule behind that split:
```
pengfei-threemoonslab: [P2] Do not infer ownership from fields absent from the bounded summary

`control.next_action.why` is deliberately incomplete when there are several unresolved declarations: `_placeholder_review_why()` fits it to the envelope's prose budget and emits `and N more in placeholders[]`. I reproduced this with both `init --json` and `doctor --json`: seven human-owned placeholders produced three named paths plus `and 4 more`; the omitted paths included `agent.prohibited_actions[1..3]` and `permissions.scopes[0]`. The new “everything else in placeholders[] is yours” rule therefore hands those declarations to the coding agent, despite the human route. Absence from this prose must not grant authorship: preserve the human obligation when the summary is elided, or publish complete ownership metadata before teaching a per-entry partition. The same inference appears in Route A and the troubleshooting table in `docs/quickstart.md`, and in the regenerated `llms-full.txt`.

docs/quickstart.md
```diff
@@ -1,310 +1,598 @@
...
-The older static scan fixture is still useful when you want a 5-minute path to
-confirm the CLI works and inspect a real Tool-Use Readiness Report before
-touching your own repo:
+On a build that implements contract `21` or newer, the same answer arrives as
+an explicit state machine. **Validate the pointer before you read anything it
+binds** — `current-control.json` is an ordinary file and goes stale silently:
 
 ```bash
-uvx agents-shipgate fixture run support_refund_agent
+agents-shipgate agent control --workspace .
```
pengfei-threemoonslab: [P2] Point this command at the fixture workspace and its reports directory

Step 2 runs the fixture in a separate temporary repository and writes its artifacts under `<fixture-copy>/reports`, but this command looks for `./agents-shipgate-reports` in the caller's directory. After following the walkthrough on this head, it exits 3 with `Current control is unavailable (missing)`. Merely changing into the fixture root still fails because its report directory is named `reports`; if the caller already has a pointer, the command instead validates that unrelated run. Show the workspace from `Fixture copy left at` and pass the directory printed by `Reports:` explicitly. I verified that running from the fixture root with `agent control --workspace . --reports-dir reports` exits 0 and returns the expected `human_review_required` state.

pengfei-threemoonslab (commented):
Follow-up review of all 30 changed files at `3417910c91e7c4e2b97d5a9bceb1d262f8345625`, against base `ef5d94f6667d83f0f54de0489e67f6b70f98ce6f`.

The release-specific excerpts and the corrected scaffold routing address the earlier findings. Two reproducible issues remain in the revised ownership and control-command guidance; I recommend fixing the inline P2 findings before merge.

Validation: **877 tests passed** across `test_distribution_surface_parity.py`, `test_fixture.py`, `test_docs_links.py`, `test_public_surface_contract.py`, and `test_design_partner_pilot.py` (`pytest -n 4`), including the source/released-engine quotation guard. `ruff check .` and `git diff --check` passed; `self-check --json` returned `ready: true`. I additionally exercised `init --json` and `doctor --json` with multiple unresolved declarations, and replayed the refund fixture followed by the documented control command from both the caller directory and the fixture root. These reproductions expose cases the passing prose guards do not cover. The three current PR workflow runs also report success.

Agents Shipgate verification and the refreshed control validator report **`control_state: review_publishable`**, **`merge_verdict: human_review_required`**, **`release_decision: review_required`**. Review publication is permitted; the `AGENTS.md` trust-root edit still requires human review, with merge and completion permissions false.

Submitted as a comment review because the connected account is the PR author.

AGENTS.md
pengfei-threemoonslab: Confirmed and fixed — and you are right that it was the *fix* that introduced it. Reproduced with `doctor --json` on eight unresolved placeholders:

```
why: In .../shipgate.yaml: line 4 (project.owner), line 8 (agent.declared_purpose[0]),
     line 10 (agent.prohibited_actions[0]), and 3 more in placeholders[] must be
     supplied by a human. …
placeholders[]: project.name, project.owner, agent.name, agent.declared_purpose[0],
     agent.prohibited_actions[0..2], permissions.scopes[0]
```

The three it dropped — `agent.prohibited_actions[1]`, `[2]`, `permissions.scopes[0]` — are all human-owned, so "the rest of `placeholders[]` is repository reading" handed them to the agent.

I took the second option you offered and stopped teaching a per-entry partition. `control.next_action.actor` routes the **turn**, not fields, and I verified both directions: any unresolved human-owned value gives `actor: "human"` / `permissions.edit: false` — surface everything and stop; only once they are all supplied does a re-run give `actor: "coding_agent"` / `kind: "edit"` naming the field to replace. Every surface now says the `why` is where to *start*, quotes the `and N more in placeholders[]` shape, and states that absence from it grants nothing. Corrected in `AGENTS.md`, Route A and the troubleshooting row in `docs/quickstart.md`, `docs/agent-recipes.md` and `docs/troubleshooting.md`; `llms-full.txt` regenerated.

Guarded by `test_surfaces_say_the_routing_prose_elides`, with the marker asserted to be the engine's own string via `_placeholder_review_why`. I scoped it to surfaces that route on the bounded prose itself rather than every file naming both tokens — `prompts/add-shipgate-to-repo.md` and the runbook enumerate the human-owned fields by name and are not deriving anything from `why`, so requiring the caution there would have been noise.

docs/quickstart.md
pengfei-threemoonslab: Confirmed and fixed, with your invocation. Reproduced exactly: after step 2, `agent control --workspace .` from the caller's directory exits 3 with `Current control is unavailable (missing): No current-control.json is present in agents-shipgate-reports`. So the step said "validate the pointer first" while printing a command that could not.

Step 7 now prints both lines, sourced from the run's own output:

```bash
cd /tmp/shipgate-fixture-ai_generated_refund_pr-<random>/ai_generated_refund_pr
agents-shipgate agent control --workspace . --reports-dir reports
```

— the first path from `Fixture copy left at`, `reports` from the last segment of `Reports:` — and says why both flags matter, including your point that a caller with an existing pointer otherwise validates an unrelated run.

Two more things that came out of checking it. The successful output is the compact `shipgate.agent_control/v1` envelope, whose fields are **top-level** (`control_state`, `next_actor`, `decision`, `permissions`), while the step went on to describe the pointer file's nested `control.state`; both spellings are named now, each against its own artifact. And `v0.15.0` has no `agent control` subcommand at all — its `agent` app is `handoff` only — so the released-build note says that rather than just "those fields do not exist".

`test_the_walkthrough_control_command_runs_in_the_walkthrough` now reads the flags out of the page and runs them where step 2 leaves the reader, taking the printed `cd` literally via `monkeypatch.chdir` — my first draft passed `--workspace <abs>` without changing directory and disagreed with a hand-run terminal, because `--reports-dir` is relative to cwd. Replayed against the previous text: exit 3.

AGENTS.md
```diff
@@ -565,12 +565,22 @@ Otherwise proceed to `init`. MCP/OpenAPI tool-surface repos and Codex plugin pac
...
+agents-shipgate init --workspace . --write --json
+# resolve the placeholders init reports, by owner (below), then:
 agents-shipgate scan -c shipgate.yaml
 ```
 
-`init` writes a manifest with `CHANGE_ME` placeholders for `agent.name` and `agent.declared_purpose`. Replace them by reading the agent's prompt or main file.
+`init --json` reports the placeholders two ways, and only one of them routes. **`placeholders[]` is a location list** — each entry is `path`, `current` and `line`, and carries no owner. **`control.next_action.actor` routes the turn**, not individual fields:
+
+- `actor: "human"` — *any* human-owned value is still unresolved. `permissions.edit` is `false`: surface the whole thing and stop. Do not edit the manifest, and do not split the array.
+- `actor: "coding_agent"` — every human-owned value has been supplied and only fields you own are left. `why` names the one to replace.
```
pengfei-threemoonslab, line 576: [P2] Keep setup authority scoped to the selected action

`actor: coding_agent` does not establish that every human-owned value has been supplied. Replaying the existing `test_a_refused_instruction_target_still_reports_what_failed` scenario returns an agent-owned `kind: edit` for the refused Cursor instruction file while `placeholders[]` still includes `agent.declared_purpose[0]`. `setup_control_envelope()` deliberately gives blocking repairs precedence over that deferred human obligation. Following this sentence would misclassify the unresolved declaration as agent-owned. The unprotected guide copies are corrected in this commit; preflight prohibits me from applying the corresponding protected-file edit.

Minimal replacement:

```suggestion
- `actor: "coding_agent"` authorizes only the exact `control.next_action`: read its `kind`, `path` or `command`, perform that step, and rerun the stated check. A blocking setup repair can take precedence while human-owned declarations remain unresolved. This actor never grants ownership of other placeholders.
```

Please also qualify the nearby unconditional “While one is unresolved, init returns … human” sentence to preserve the blocking-repair precedence, make clear that `placeholders[]` is authoritative for locations only, and regenerate `llms-full.txt`. The strengthened setup-precedence test in this commit demonstrates the required behavior.

pengfei-threemoonslab (commented):
Reviewed the latest PR and implemented the independently permitted fixes in `10f9d689d22a8bb8c0f1b5e49207924f352c5020`, against base `20968551eb1d2b081756bea3ca44d82558c18c0c`.

Both findings from my previous review are addressed: the guidance no longer assigns omitted `why` entries to the agent, and the fixture control command now changes into the copied workspace and selects `reports`. The preceding release-excerpt and scaffold-routing fixes also remain in place.

Additional corrections pushed in this review:

- Quickstart/recipes now authorize only the exact setup action. A blocking repair can return `actor: coding_agent` while human declarations remain unresolved; the actor does not assign ownership to the rest of the manifest.
- The quickstart and both runbook examples now reflect Python MCP extraction merged in #535: direct FastMCP decorators are detected, while the factory-bound `app = create_server()` reproduction still scaffolds. The source-selection table names the supported source adapter.
- The walkthrough guard now checks that the page actually prints its `cd`, and uses `shlex.split` for the flags. Removing the documented `cd` now fails the guard. The existing setup-precedence regression explicitly asserts that an agent edit route coexists with unresolved human placeholders. `llms-full.txt` was regenerated.

**One actionable P2 remains**, attached to `AGENTS.md:576`. I did not edit this protected file: preflight of the concrete correction returned `control.state: human_review_required`, `next_action.kind: stop`, and every permission false because it matches `**/AGENTS.md`. The inline finding includes a minimal proposed replacement and the reproduction; its generated copy in `llms-full.txt` must follow once a human addresses the source. The unprotected plan separately routed to verification, and refreshed controls authorized the edits, commit, push and review publication.

Validation: 1,054 targeted tests passed. The full suite produced **8,635 passes and 5 skips**; ten packaging checks initially errored during their shared build setup because sandbox DNS blocked the build-backend download. All **10 packaging tests passed** on the network-enabled rerun, for **8,645 passed / 5 skipped** across that validation. `ruff check .`, `git diff --check`, and `self-check --json` (`ready: true`) passed. I replayed both FastMCP discovery shapes and the setup-precedence case. A transient boundary snapshot error during concurrent tests cleared on the quiet rerun, which requested verification.

Verification on the final committed SHA and the refreshed artifact validator report **`control_state: review_publishable`**, **`merge_verdict: human_review_required`**, **`release_decision: review_required`**. Merge and completion remain denied. Submitted as a comment review because the connected account is the PR author; no merge or trust-root approval was performed.

PR #535, opened 2026-09-06
Read the Python `@mcp.tool` decorator, in both detectors (#484)

### LINKED ISSUE #484: Python FastMCP servers have no route either: `@mcp.tool` is the largest measured registration idiom
# Python FastMCP servers have no route either: `@mcp.tool` is the largest measured registration idiom

## Summary

#431 shipped a built-in registry of MCP tool-registration idioms covering TypeScript
and Go — the two languages its three vendor walks measured. The 30-server survey it
required before choosing that set is published at
[`docs/mcp-registration-idioms.md`](../blob/main/docs/mcp-registration-idioms.md), and it
found one shape above the coverage line that #431 deliberately did not ship:

| idiom | repositories | largest single repo |
|---|---|---|
| **Python `@mcp.tool` (FastMCP decorator)** | **5** | `awslabs/mcp` (488 call sites) |
| Go `Tool{Name: "…"}` | 3 | `github/github-mcp-server` (135) |
| TS `.tool(` / `.registerTool("…"` | 3 | `cloudflare/mcp-server-cloudflare` (85) |

Python FastMCP is the largest single shape in the survey by both measures. `redis/mcp-redis`,
`neo4j-contrib/mcp-neo4j`, `chroma-core/chroma-mcp` and `awslabs/mcp` are all in the state the
MongoDB and Grafana servers were in before #431: `detect` returns `is_agent_project: false`.

## Why it was deferred rather than skipped

It is a **different extraction mechanism**, and #393's lesson is that the mechanism is
reusable but each one needs its own adversarial probe list before it can claim anything.
Concretely, none of these hold for the lexical reader #431 shipped:

- The name defaults to the decorated **function's** name and is only sometimes a literal
  (`@mcp.tool(name="…")`).
- The input schema comes from the **signature**, so unlike TS/Go this route could carry
  real parameters — which changes what the `SHIP-SCHEMA-*` family sees.
- Provenance is an **import binding** (`from mcp.server.fastmcp import FastMCP`,
  `from fastmcp import FastMCP`), not a declared dependency, and the binding is what stops
  `@anything.tool` from reading as a registration.
- Python has a real AST in-tree, so a masking lexer would be the wrong tool: the existing
  `inputs/_python_framework.py` machinery is the right base.

The lexical reader had **eight fail-open constructs in its first draft**, all found by
writing the probe list. A Python idiom bolted onto #431's increment without its own list
would have shipped without that check.

## Scope

- One idiom, `py_fastmcp_decorator`, in the existing built-in registry — same registry,
  same `medium` ceiling, same per-file completeness rule, same exclusion-ledger accounting
  for a name this reader cannot resolve.
- AST-based, following `google_adk`'s binding-table approach rather than a text scan.
- Its own adversarial probe list, at minimum: a decorator on an alias
  (`t = mcp.tool`), a name built at runtime, a decorator applied by a factory,
  `functools.wraps`, a conditional registration, a re-bound module symbol, and
  `@app.tool` where `app` is not a FastMCP instance.

## Acceptance

- [ ] `detect` on `redis/mcp-redis` and `neo4j-contrib/mcp-neo4j` returns
      `is_agent_project: true` with a named source.
- [ ] `scan` enumerates their tools, with parameters read from the signature.
- [ ] A decorator whose name is built at runtime is reported unenumerated and reaches the
      exclusion ledger, as in #431.
- [ ] The published survey table in `docs/mcp-registration-idioms.md` moves the Python row
      to "shipped".
- [ ] `IDIOM_REGISTRY_VERSION` bumps, and the trigger catalog's token list follows.

Filed from #431's survey; the measurement is the evidence the
[roadmap non-goal](../blob/main/ROADMAP.md#explicit-non-goals) clause asks for.

pengfei-threemoonslab:
Closes #484.

## What was missing

[#431](https://github.com/ThreeMoonsLab/agents-shipgate/issues/431) shipped a built-in registry of MCP tool-registration idioms covering TypeScript and Go, and [its own 30-server survey](https://github.com/ThreeMoonsLab/agents-shipgate/blob/main/docs/mcp-registration-idioms.md) named what it left out: Python's `@mcp.tool` decorator, the largest single shape by both repositories and call sites. `redis/mcp-redis`, `chroma-core/chroma-mcp` and `neo4j-contrib/mcp-neo4j` were all in the state MongoDB and Grafana were in before #431 — `detect` returned `is_agent_project: false` for a first-party server publishing dozens of tools.

## Measured, on live checkouts

Identical verdicts, routes and evidence from the installed CLI and the zero-install detector:

| repository | route | tools | unenumerated |
|---|---|---:|---:|
| `redis/mcp-redis` | `src` | 53 | 0 |
| `chroma-core/chroma-mcp` | `src/chroma_mcp` | 13 | 0 |
| `neo4j-contrib/mcp-neo4j` | `servers` | 0 | 40 |
| `awslabs/mcp` | `src` | 334 | 107 |

Before this change all four reported `mcp_server_source` nowhere; `mcp-neo4j` and `chroma-mcp` were `is_agent_project: false` outright.

## Why it is a mechanism, not a sixth pattern

Three facts, each measured before it was designed for:

- **The name is usually not written down.** `@mcp.tool()` on `def dbsize()` registers `dbsize` — all 53 of `redis/mcp-redis`'s registrations are the bare form, so a reader looking for a literal beside the call finds none of them.
- **The schema is the signature.** This idiom publishes an input schema where the lexical idioms genuinely have nothing to publish. The request `Context` parameter is dropped, because the server drops it too; so are `*args`/`**kwargs`.
- **Whether it registers anything is a binding fact.** `@app.tool` is a tool registration when `app` is a server and somebody else's decorator otherwise. The reader follows the decorated name back to a server construction — **across modules**, because `redis/mcp-redis` constructs in `src/common/server.py` and decorates in `src/tools/*.py` — and refuses out loud when it cannot.

## Two findings that changed the design after measurement

**The official Python SDK v2 renamed `FastMCP` to `MCPServer`.** `mcp.server.fastmcp` is now a module that exists only to raise `ModuleNotFoundError`, and 41 of `awslabs/mcp`'s servers had already moved. A reader that knew one class name read **105 tools where there are 334**. The reader knows `(module, class)` pairs, and the index prefilter is derived from that table so the two cannot drift.

**A server can register 40 tools and name none of them.** Every `neo4j-contrib/mcp-neo4j` tool is `name=namespace_prefix + "…"`. The lexical idioms' rule — offer no route without a resolved name — would have reported that repository as "not an agent project" *because* its names are dynamic, which is the silent miss this input exists to end. A Python site carries its own provenance (it was only emitted after the decorator was followed back to a server construction, and a client does not construct a server), so the route is offered and the evidence says `40 registration(s), none of which this reader can name`. A committed export cannot displace such a route either: containment is the test, and an export contains an empty set vacuously.

## What it still refuses

The ceiling stays `medium`. A runtime-built name stays an omission the exclusion ledger accounts for. A decorator on an object the reader cannot follow — `@self.mcp.tool` on an injected server, `app = create_server()` on a factory's result, both measured in `awslabs/mcp` — is recorded rather than guessed at or dropped.

## Both detectors move together

`tools/shipgate-detect.py` goes 0.5.0 → 0.6.0 in the same commit. #485 made it a second implementation of this reader, and `tests/mcp_idiom_corpus.py` is what keeps it from becoming a different one: the whole Python probe list and the cross-module trees live there once, and both readers are driven through all of it — every site field, span included, plus `parameters`, `returns` and `proves_server`, which decide the schema and whether a route exists at all.

## Verification

- **32 adversarial probes** covering the list #484 required (alias, runtime name, factory, `functools.wraps`, conditional registration, re-bound symbol, `@app.tool` on a non-server) plus star imports, scope shadowing, the three constructor spellings, PEP 695 generics, and docstring/class/positional shapes.
- **A 30-case perturbation sweep: 27 caught.** The three that are not are inert by construction and say so at the line — the index prefilter is a cost bound; a `type_params` subtree cannot bind a name in any module Python will compile (`ast.parse` accepts a walrus in a TypeVar bound, `compile` refuses it); and reading an empty module through `or` returns the same text.
- All CI checks green. `IDIOM_REGISTRY_VERSION` → `2`; `TRIGGER-MCP-TOOL-REGISTRATION-SOURCE` routes `@mcp.tool`.

## Review rounds

Two self-review passes ran on this branch and their fixes are in it. Four changed behaviour, and the first is the one worth reading:

- **The Poetry half of the Python language gate was dead.** The dependency walker returned a table's *constraint* where the distribution is the key, so `mcp = "^1.6.0"` produced `^1.6.0` and never `mcp` — three of the five tables named in the constant could never match, and a Poetry-managed server got no route at all. Fixed, and the reason it shipped is closed: every fixture in the file wrote PEP 621, so a parametrised test now drives all five tables plus `requirements*.txt`.
- **Class scopes were searched from a nested function**, where Python does not close over them — a real registration was reported unenumerable.
- **The route repeated its proving module** once per importing file. The fixture now has two tool modules, because with one importer the defect is invisible.
- **The test-file *prefix* rule was applied to every language**, silently dropping `test_helpers.ts` and `test_util.go` from the surface. Scoped to Python, where the reason (pytest collects `test_*.py`) lives.

#441's scaffold tests moved to the factory-bound shape, since their original reproduction is a workspace `detect` now reads — that is the fix, not a regression.

### DIFF
--- a/src/agents_shipgate/cli/discovery/mcp_source.py
+++ b/src/agents_shipgate/cli/discovery/mcp_source.py
@@ -43,7 +43,7 @@ from __future__ import annotations
 
 import json
 import tomllib
-from collections.abc import Iterable, Sequence
+from collections.abc import Iterable, Iterator, Sequence
 from dataclasses import dataclass, field
 from pathlib import Path, PurePosixPath
 
@@ -61,7 +61,10 @@ from agents_shipgate.inputs.mcp_idioms import (
     language_for_path,
     scan_source,
 )
-from agents_shipgate.inputs.mcp_server_source import SOURCE_TYPE
+from agents_shipgate.inputs.mcp_server_source import (
+    MAX_CACHED_SOURCE_BYTES,
+    SOURCE_TYPE,
+)
 
 #: How many source files discovery reads before it stops. Discovery runs on an
 #: unknown workspace with no manifest and has to stay bounded; the scan-time
@@ -182,7 +185,7 @@ def discover_mcp_server_source(
     scannable.sort(key=lambda pair: pair[1])
     truncated = len(scannable) > max_source_files
     read = scannable[:max_source_files]
-    server_index = _python_server_index(read)
+    server_index, python_texts = _python_server_index(read)
     names: set[str] = set()
     languages: set[SourceLanguage] = set()
     unresolved_by_file: dict[str, int] = {}
@@ -191,7 +194,7 @@ def discover_mcp_server_source(
         language = language_for_path(path)
         if language is None:  # pragma: no cover - filtered above
             continue
-        text = _source_text(path)
+        text = python_texts.pop(path, None) or _source_text(path)
         if text is None:
             continue
         result = scan_source(
@@ -296,15 +299,18 @@ def discover_mcp_server_source(
         unresolved_count=unresolved,
         evidence=evidence,
         framework_evidence=tuple(sorted(framework.reasons)),
-        candidate_files=tuple(sorted(candidate_files)),
+        # De-duplicated: one module can prove the binding for every file
+        # that imports it, so a plain sort repeated `src/common/server.py`
+        # once per decorating module.
+        candidate_files=tuple(sorted(set(candidate_files))),
         truncated=truncated,
     )
 
 
 def _python_server_index(
     scannable: Sequence[tuple[Path, str]]
-) -> PythonServerIndex:
-    """Index every Python module in the walk that constructs a FastMCP server.
+) -> tuple[PythonServerIndex, dict[Path, str]]:
+    """Index every Python module in the walk that constructs an MCP server.
 
     Built ahead of the scan for the same reason the adapter builds one:
     ``redis/mcp-redis`` constructs the server in ``src/common/server.py`` and
@@ -314,17 +320,30 @@ def _python_server_index(
     The keys are workspace-relative here and root-relative in the adapter, and
     that is not a divergence: both are the path *inside the tree being read*,
     which is what an import's segments are matched against.
+
+    Returns the texts it read alongside the index, up to
+    :data:`MAX_CACHED_SOURCE_BYTES`, so the scan pass reads each Python file
+    once. Past the bound a file is read twice rather than held, which changes
+    no answer.
     """
 
-    return PythonServerIndex.build(
-        (relative, text)
-        for relative, text in (
-            (relative, _source_text(path))
-            for path, relative in scannable
-            if language_for_path(path) == "python"
-        )
-        if text is not None
-    )
+    texts: dict[Path, str] = {}
+    cached = 0
+
+    def _modules() -> Iterator[tuple[str, str]]:
+        nonlocal cached
+        for path, relative in scannable:
+            if language_for_path(path) != "python":
+                continue
+            text = _source_text(path)
+            if text is None:
+                continue
+            if cached + len(text) <= MAX_CACHED_SOURCE_BYTES:
+                texts[path] = text
+                cached += len(text)
+            yield relative, text
+
+    return PythonServerIndex.build(_modules()), texts
 
 
 def _source_text(path: Path) -> str | None:
@@ -461,12 +480,21 @@ def _pyproject_requirements(text: str) -> list[str]:
 
 
 def _toml_requirements(node: object, depth: int = 0) -> list[str]:
-    """Requirement strings inside a dependency list, table, or table of lists.
+    """Every string in a dependency table that could name a distribution.
 
-    One walker for three shapes, because they are three spellings of one fact:
-    PEP 621 writes a list of requirement strings, PEP 735 and Poetry's groups
-    write a table of them, and Poetry's own table writes the *name* as the key
-    with a constraint as the value.
+    Four spellings of one fact, and the walker admits **both** halves of a
+    table rather than choosing between them: PEP 621 writes a list of
+    requirement strings, PEP 735 and Poetry's groups write a table of such
+    lists, and Poetry's own table writes the *name as the key* with a
+    constraint as the value — sometimes an inline table (``{version = "^2"}``)
+    rather than a string.
+
+    Choosing was the bug. A first draft emitted the key only when the value
+    yielded nothing, so ``mcp = "^1.6.0"`` produced ``^1.6.0`` and never
+    ``mcp`` — every Poetry table in the list above was dead, and the gate
+    withheld the route from a Poetry-managed server. Emitting the key too
+    costs nothing: a group name like ``dev`` is not a distribution this gate
+    looks for, and :func:`normalized_distribution` drops ``^1.6.0`` anyway.
     """
 
     if depth > 3:  # pragma: no cover - a dependency table is never this deep
@@ -478,12 +506,8 @@ def _toml_requirements(node: object, depth: int = 0) -> list[str]:
     if isinstance(node, dict):
         found: list[str] = []
         for key, value in node.items():
-            nested = _toml_requirements(value, depth + 1)
-            # A Poetry table maps ``name -> constraint``, so the *key* is the
-            # requirement; a group table maps ``group -> [requirements]``, so
-            # the values are. Both are admitted, and the key only when the
-            # value was not itself a requirement list.
-            found.extend(nested if nested else [str(key)])
+            found.append(str(key))
+            found.extend(_toml_requirements(value, depth + 1))
         return found
     return []
 
--- a/src/agents_shipgate/cli/discovery/mcp_source.py
+++ b/src/agents_shipgate/cli/discovery/mcp_source.py
@@ -195,7 +195,9 @@ def discover_mcp_server_source(
         if language is None:  # pragma: no cover - filtered above
             continue
         # ``is None``, not ``or``: an empty module caches as ``""``, which is
-        # the right answer and a falsy one, so ``or`` would read it again.
+        # the right answer and a falsy one, so ``or`` would read it again. The
+        # re-read returns the same text, so this costs a read and never an
+        # answer — a perturbation sweep reports it untested and is right to.
         text = python_texts.pop(path, None)
         if text is None:
             text = _source_text(path)
--- a/src/agents_shipgate/cli/discovery/mcp_source.py
+++ b/src/agents_shipgate/cli/discovery/mcp_source.py
@@ -194,7 +194,11 @@ def discover_mcp_server_source(
         language = language_for_path(path)
         if language is None:  # pragma: no cover - filtered above
             continue
-        text = python_texts.pop(path, None) or _source_text(path)
+        # ``is None``, not ``or``: an empty module caches as ``""``, which is
+        # the right answer and a falsy one, so ``or`` would read it again.
+        text = python_texts.pop(path, None)
+        if text is None:
+            text = _source_text(path)
         if text is None:
             continue
         result = scan_source(
--- a/src/agents_shipgate/cli/discovery/mcp_source.py
+++ b/src/agents_shipgate/cli/discovery/mcp_source.py
@@ -5,7 +5,7 @@ agent project". Both publish dozens of tools — ``drop-database``,
 ``delete-many``, ``update_incident`` — and neither commits an export, so every
 route discovery had was filename-shaped and found nothing. This module supplies
 the missing route: it asks whether a workspace *is* an MCP server written in
-TypeScript or Go, and if so which directory holds the registrations.
+TypeScript, Go or Python, and if so which directory holds the registrations.
 
 Two facts have to hold together before the route is offered, and the pairing is
 the whole design.
@@ -14,26 +14,35 @@ the whole design.
 is not an MCP server just because a class of its own happens to spell a field
 ``toolName``. #393's lesson is that a proof resting on a spelling is the
 fail-open shape, so the dependency — ``@modelcontextprotocol/*`` in a
-``package.json``, an MCP module in a ``go.mod`` — is what turns a spelling into
-provenance. It is also the cheap half: a workspace with no such dependency is
-answered without reading a single source file.
+``package.json``, an MCP module in a ``go.mod``, ``mcp`` or ``fastmcp`` in a
+``pyproject.toml`` — is what turns a spelling into provenance. It is also the
+cheap half: a workspace with no such dependency is answered without reading a
+single source file.
 
-**A resolved registration.** Dependency evidence alone says the repository
-*uses* MCP, which every client does too. At least one tool name has to be
-readable at a registration site before this claims a tool surface.
+**A registration this reader can stand behind.** Dependency evidence alone says
+the repository *uses* MCP, which every client does too. For the lexical idioms
+the second fact is a **resolved name**: they match a spelling, and a spelling
+is all `.registerTool(` in a client repository ever is. The Python idiom
+already carries the second fact in the site itself — it is only emitted after
+the decorator has been followed back to a server construction, and a client
+does not construct a server — so there the route is offered for a registration
+whose *name* is built at run time. That is not a corner: all 40 of
+``neo4j-contrib/mcp-neo4j``'s tools are ``name=namespace_prefix + "…"``, and
+requiring a readable name would report the repository as "not an agent project"
+precisely because its names are dynamic.
 
 Where both hold and the workspace *also* has a parseable MCP export, the export
-wins: it is the server's own published contract, it carries the input schemas
-this route deliberately does not read, and it is ``high`` confidence against
-``medium``. The source route is then withheld and recorded in
-``excluded_sources`` — named and visible, never silently dropped, because a
-route that vanishes without a reason is indistinguishable from one nobody
-implemented.
+wins: it is the server's own published contract, published in the shape a
+client receives it, and it is ``high`` confidence against ``medium``. The
+source route is then withheld and recorded in ``excluded_sources`` — named and
+visible, never silently dropped, because a route that vanishes without a reason
+is indistinguishable from one nobody implemented.
 """
 
 from __future__ import annotations
 
 import json
+import tomllib
 from collections.abc import Iterable, Sequence
 from dataclasses import dataclass, field
 from pathlib import Path, PurePosixPath
@@ -45,7 +54,9 @@ from agents_shipgate.inputs.mcp_idioms import (
     MAX_SOURCE_FILE_BYTES,
     SKIP_DIRECTORY_NAMES,
     TYPESCRIPT_FRAMEWORK_PACKAGES,
+    PythonServerIndex,
     SourceLanguage,
+    declares_python_mcp_framework,
     is_scannable_path,
     language_for_path,
     scan_source,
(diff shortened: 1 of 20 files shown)


pengfei-threemoonslab (commented):
Reviewed `e46365894156d3547362c1f05845d75cb5b4393d` against `ef5d94f6667d83f0f54de0489e67f6b70f98ce6f`. **I recommend changes before merge.** Five reproducible correctness findings are attached inline: two ways to publish the wrong action identity, loss of the source route for a mixed dynamic/literal surface, incorrect cross-module binding evidence, and removal of valid MCP input parameters.

Validation:

- Focused suite: **920 passed, 1 skipped** across `test_mcp_idioms.py`, `test_mcp_server_source.py`, `test_zero_install_detector.py`, `test_init_scaffold_disclosure.py`, and `test_distribution_surface_parity.py`.
- Ruff on all changed Python files and `git diff --check`: passed. CLI self-check: `ready: true`.
- Independent adversarial probes reproduced the name, wrapper, import-resolution, and export-routing issues in both readers. Name/signature discrepancies were compared with schemas from local SDK `mcp==1.27.2` on Python 3.12.13, using inert toy registrations; no tool handlers were invoked. Full installed and standalone detection both suggested only the incomplete export in the mixed-name case.
- Required verification of the exact PR refs, with the current control envelope refreshed: **`control_state: review_publishable`**, **`gate.merge_verdict: human_review_required`**, **`release_decision.decision: review_required`**. The verifier permits publishing this review and withholds merge authority. Its human-review items concern `AGENTS.md` and `docs/triggers.json`.

The passing existing suite does not cover these reproductions. This review does not approve the protected-surface changes.

src/agents_shipgate/inputs/mcp_idioms.py
```diff
@@ -1402,7 +1561,857 @@ def _brace_depth(masked: str, open_brace: int, index: int) -> int:
...
+    there would publish a name that server does not serve.
+    """
+
+    if isinstance(decorator, ast.Call):
+        given = _python_keyword(decorator, "name")
+        if given is None and decorator.args:
+            given = decorator.args[0]
+        if given is not None:
+            return _resolve_name(_python_string(given), True)
+    return _resolve_name(node.name, True)
```
pengfei-threemoonslab, line 2337: [P1] Do not default the tool name when decorator options are unpacked

`_python_keyword()` ignores keywords whose `arg` is `None`, so `@mcp.tool(**options)` is treated as if no name was supplied. For example:

```python
options = {"name": "delete_all"}
@mcp.tool(**options)
def harmless() -> str:
    return "unused"
```

With a proven SDK `FastMCP` instance, both readers emit `harmless`; the adapter marks it `surface: "enumerated"` with no omissions. The SDK actually publishes `delete_all` (reproduced with `mcp==1.27.2`). This substitutes a nonexistent action identity for the real tool instead of preserving the unresolved-name boundary. Treat an unpacked mapping that could supply `name` as unresolved unless its value is statically established, and cover the case in the shared corpus so both readers retain the omission.

src/agents_shipgate/inputs/mcp_idioms.py
```diff
@@ -1402,7 +1561,857 @@ def _brace_depth(masked: str, open_brace: int, index: int) -> int:
...
+                server_modules.add(proving_module)
+            # Built on the first site, not per file: most modules that pass the
+            # `"tool"` prefilter register nothing — a variable of that name, a
+            # word in a docstring — and the line table would be built and
+            # thrown away for every one of them.
+            if offsets is None:
+                offsets = _PythonOffsets(text)
+            sites.append(
+                _python_site(node, decorator, offsets, proven=proven)
+            )
```
pengfei-threemoonslab, line 2154: [P1] Account for inner decorators before using the original function metadata

The object received by the MCP decorator can already have been replaced by a decorator below it, but `_python_site()` always reads the original `FunctionDef` name and signature. A valid reproducer is:

```python
def replace(fn):
    def delete_all(target: str) -> str:
        return "unused"
    return delete_all

@mcp.tool()
@replace
def harmless(query: int) -> str:
    return "unused"
```

Both readers report `harmless`, and the adapter reports `query` as its input with an enumerated surface and no omissions. The SDK instead registers `delete_all(target: str)`. The `functools.wraps` probe exercises only a metadata-preserving wrapper; it does not justify assuming every inner decorator preserves the callable. When an inner decorator's effect on the callable is unknown, retain an omission rather than asserting the original name/schema. Apply this boundary to both readers.

src/agents_shipgate/cli/discovery/mcp_source.py
```diff
@@ -201,8 +251,15 @@ def discover_mcp_server_source(
     evidence = _evidence_lines(
         framework, languages, names, root, truncated, unresolved
     )
-    covering_export, uncovered = _covering_export(
-        workspace, exported_source_paths, names
+    covering_export, uncovered = (
+        _covering_export(workspace, exported_source_paths, names)
+        if names
```
pengfei-threemoonslab: [P1] Preserve the source route when some registrations are still unnamed

The new guard protects an entirely unnamed surface, but a mixed surface is still discarded. Put `@mcp.tool()` on `visible` and `@mcp.tool(name=PREFIX + "delete_all")` on another function, then commit an MCP export containing only `visible`. Both discovery implementations return `path: null`, `unresolved_count: 1`, and an exclusion claiming the export covers the registrations. Full `detect` suggests only the export. Consequently, following that route never scans the dynamic registration and cannot put it in the source omission ledger.

Adding one readable tool should not remove the protection added here for dynamic names. Do not let an export displace this Python source route while any registration in its scope remains unresolved; extend the all-dynamic regression test with this mixed case in both detectors.

src/agents_shipgate/inputs/mcp_idioms.py
```diff
@@ -1402,7 +1561,857 @@ def _brace_depth(masked: str, open_brace: int, index: int) -> int:
...
+
+    @classmethod
+    def build(cls, modules: Iterable[tuple[str, str]]) -> PythonServerIndex:
+        """Index ``(relative posix path, source text)`` pairs."""
+
+        indexed: dict[str, frozenset[str]] = {}
+        for path, text in modules:
+            exports = python_server_exports(text)
+            if exports:
+                indexed[path] = exports
```
pengfei-threemoonslab: [P2] Include non-server modules when checking import ambiguity

The index discards every module without a proven server export before `resolve()` checks whether the import has exactly one candidate. Consider `src/common/server.py` exporting `mcp = Other()` where `Other.tool()` is an ordinary decorator, plus `archive/src/common/server.py` constructing a real FastMCP instance. A tool module containing `from src.common.server import mcp` and `@mcp.tool()` is resolved against the archive module: both readers emit a named tool with `proves_server=True`, even though the exact imported module is the non-server one.

Keep the paths of modules with empty/unproven exports in the resolution universe, then check whether the resolved module exports the requested server symbol. Otherwise a known conflicting module disappears before the uniqueness check and an unrelated server supplies false binding evidence. The existing ambiguity fixture needs a case where only one candidate exports a server.

src/agents_shipgate/inputs/mcp_server_source.py
```diff
@@ -304,12 +413,103 @@ def _tool_from_site(
...
+#: Matched on the annotation as well as the name because the framework matches
+#: on the *type*: ``SKIPPED_TOOL_PARAMETERS`` catches the conventional ``ctx``
+#: and ``context`` spellings, and this catches the parameter that is annotated
+#: ``Context`` and called something else.
+_CONTEXT_ANNOTATIONS: frozenset[str] = frozenset({"Context"})
+
+
+def _is_context_parameter(parameter: SignatureParameter) -> bool:
+    if parameter.name in SKIPPED_TOOL_PARAMETERS:
+        return True
```
pengfei-threemoonslab: [P2] Preserve ordinary MCP inputs that share conventional context names

FastMCP identifies injected context by its annotation, not by the parameter-name blacklist shared with other framework adapters. For a valid tool `def configure(config: dict, context: str, runtime: str) -> str`, this branch drops all three parameters and publishes an empty input schema and `configure()` with no omissions. The SDK publishes all three as required inputs; I checked its generated schema using `mcp==1.27.2`.

Names such as `config`, `context`, `runtime`, and `ctx` can be normal user-supplied MCP inputs. Removing them hides actual schema changes and gives schema/policy consumers an incorrect parameter inventory. Apply MCP-specific, type-backed context detection instead of treating membership in `SKIPPED_TOOL_PARAMETERS` as proof of injection, and add a negative test alongside the annotated-Context test.

src/agents_shipgate/inputs/mcp_idioms.py
pengfei-threemoonslab: Fixed in c37e56c2 — reproduced exactly as described.

`_python_tool_name` now withholds the name whenever any keyword in the decorator call has `arg is None`, so `@mcp.tool(**options)` is `name_not_literal` rather than the function name. I did not try to statically establish a literal mapping: no surveyed server writes one, and a resolver for it is speculative surface on the exact path where being wrong publishes an id nobody serves.

Corpus case `an unpacked mapping can carry the tool name`, driven through both readers; sweep perturbation `an unpacked mapping cannot carry the tool name` is caught.

src/agents_shipgate/inputs/mcp_idioms.py
pengfei-threemoonslab: Fixed in c37e56c2. You are right that `functools.wraps` proves nothing at the use site, and the probe was doing exactly the work you say it was.

A decorator later in `decorator_list` is applied *first*, so it is the object `.tool` receives. The site now keeps its provenance and loses both its name and its signature — the schema comes from the same object, so publishing parameters beside an unreadable name would describe a function that may never have been registered.

I measured the cost before choosing fail-closed over proving the wrapper. 106 sites in `awslabs/mcp` have an inner decorator; 72 of those were previously named, so it goes 334 → 262 with the 72 in the exclusion ledger. The other six measured servers (redis, chroma, neo4j, mongodb, grafana, github) have none. The four decorators behind most of them — `handle_exceptions`, `write_operation`, `readonly_safe`, `log_tool_call_with_response` — are `functools.wraps` wrappers or a plain `return func`, so proving them is feasible; it needs one-hop cross-module *body* analysis, which is a mechanism rather than a rule, and #393 says a mechanism needs its own probe list. Deferred with the 72 recorded as the evidence for whether it earns one.

`docs/mcp-registration-idioms.md` gains a "Three ways the registered identity is not the one on the page" section. The corpus keeps the positive control the boundary must not swallow: a decorator *above* the registration is applied after it and still resolves.

src/agents_shipgate/cli/discovery/mcp_source.py
pengfei-threemoonslab: Fixed in c37e56c2, and you identified the more general rule — my guard was the special case.

There is now one rule with two readings: an export displaces the source route only when the route has nothing the export left out, **unreadable registrations included**. With no readable name the comparison was vacuous (`names <= covered` is true for the empty set); with some readable, an export naming exactly those looks like containment and is not. Both are answered by the scoped `unresolved` count instead of a special case, and the evidence names the shortfall.

Applied to all three languages rather than to Python, since a TypeScript `.registerTool(NAME, …)` beside a covering export loses its ledger entry the same way.

Regression `test_an_export_cannot_displace_a_route_it_only_partly_accounts_for` asserts the route survives *and* that following it puts the dynamic registration in the omission ledger; `python_names_are_all_dynamic` keeps the all-dynamic case, and both detectors are driven through the workspace table.

src/agents_shipgate/inputs/mcp_idioms.py
pengfei-threemoonslab: Fixed in c37e56c2. "Ambiguity has to be measured against what is there" is the right framing and I had it backwards.

`PythonServerIndex.build` now records every scanned module, mapping to an empty frozenset where it constructs nothing. Resolution finds the unique path match across all of them, and only then asks whether that module exports the requested symbol — so your `src/common/server.py` binding `mcp = Other()` stays in the universe, makes the import ambiguous against `archive/src/common/server.py`, and the site is `server_binding_not_proven` with `proves_server=False`.

New tree case `a conflicting module that exports no server is still a conflict` reproduces your fixture; sweep perturbation `non-server modules leave the resolution universe` is caught.

src/agents_shipgate/inputs/mcp_server_source.py
pengfei-threemoonslab: Fixed in c37e56c2 — verified against the SDK the same way, and you are right that the name list is the wrong instrument.

`_is_context_parameter` is annotation-only now: `Context`, `Context | None`, `Context[ServerSession, None]` and a quoted forward reference are dropped; `list[Context]` is not. `SKIPPED_TOOL_PARAMETERS` is no longer imported here, and the module comment says why it cannot be shared — `config`, `context` and `runtime` are ordinary MCP inputs.

I also stopped dropping `self`/`cls` for the same reason: a decorated method registers the plain function and the server puts `self` in the schema, so this reader now says what the server says rather than what looks tidy.

`test_a_conventional_parameter_name_is_still_a_tool_input` carries both arms — your `configure(config, context, runtime)` keeping all three with all three required, and the annotated cases still dropped — and the sweep perturbation `the context is detected by name again` is caught.

pengfei-threemoonslab:
All five findings reproduced and are fixed in c37e56c2. Four are the same class — the reader asserted a name or a schema that something between the `def` and the server can change — and the fifth was the wrong instrument for detecting an injected context. Replies are on each thread; the summary:

| finding | resolution |
|---|---|
| P1 `@mcp.tool(**options)` | any `**` in the decorator call withholds the name |
| P1 inner decorator replaces the callable | a decorator *below* the registration withholds name **and** signature |
| P1 mixed surface loses its route | an export displaces the route only when nothing it found was left out, unreadable registrations included — applied to all three languages |
| P2 ambiguity universe | every scanned module is in the index, exporting nothing where it constructs nothing |
| P2 conventional parameter names | context detected by annotation only; `SKIPPED_TOOL_PARAMETERS` no longer used here |

**Measured cost, both detectors agreeing on all seven repositories.** The wrapper boundary is the only one that costs recall, and only on one repository:

| repository | tools | unenumerated |
|---|---:|---|
| `redis/mcp-redis` | 53 | — |
| `chroma-core/chroma-mcp` | 13 | — |
| `neo4j-contrib/mcp-neo4j` | 0 | 40 `name_not_literal` |
| `awslabs/mcp` | 262 *(was 334)* | 107 `server_binding_not_proven`, 72 `wrapped_before_registration` |
| `mongodb-js/mongodb-mcp-server` | 61 | 3 |
| `grafana/mcp-grafana` | 115 | — |
| `github/github-mcp-server` | 114 | 3 |

On the wrapper boundary specifically: I measured before choosing fail-closed. 106 sites in `awslabs/mcp` carry an inner decorator, 72 of which were previously named; the six other servers have none. The decorators behind most of them are `functools.wraps` wrappers or a plain `return func`, so proving them is feasible — but it needs one-hop cross-module *body* analysis, which is a mechanism rather than a rule, and #393 says a mechanism gets its own probe list. Deferred, with those 72 recorded as the evidence for whether it earns one.

Each boundary keeps the positive control it must not swallow — a decorator *above* the registration is applied after it and still resolves; `list[Context]` is a list, not a context. Every fix is in both readers and in `tests/mcp_idiom_corpus.py`, so the port cannot diverge on any of them.

The perturbation sweep is now 34/37. The three uncaught are inert by construction and say so at the line: the index prefilter is a cost bound, a PEP 695 `type_params` subtree cannot bind a name in any module `compile` accepts, and reading an empty module through `or` returns the same text.

Re-checked: full suite across all three CI shards, `ruff`, and the regenerated `docs/determinism-boundary.*` and `llms-full.txt`. `docs/mcp-registration-idioms.md` gains a section for the three identity boundaries and one for what the request context is and is not.

PR #540, opened 2026-09-07
Name the ADK MCP endpoint and credential-reference change (#538)

### LINKED ISSUE #538: ADK MCP endpoint and credential-reference changes disappear before the capability diff
Parent: #520. A bounded engine slice of “name the binding”; related #515, #440 and #474.

## Problem

A reviewer should see that a PR changed the remote MCP endpoint or credential reference even when its downstream tool inventory is unavailable. Currently the ADK reader discards that connection evidence before the capability diff can use it.

Reproduced on `20968551` with a `McpToolset` bound to an ADK agent:

```python
McpToolset(
    connection_params=StreamableHTTPConnectionParams(
        url="https://readonly.example/mcp",
        headers={"Authorization": os.environ["READ_KEY"]},
    ),
    tool_filter=["search"],
)
```

Change only the literal endpoint to `https://admin.example/mcp` and the reference to `ADMIN_KEY`, preserving source lines and filter:

- Both readers produce identical `GoogleAdkToolset` records and empty tool catalogs.
- Full scans produce `capability_facts=[]`.
- Head `capability_change.enabled=true`, but added/removed/broadened/narrowed are empty; `binding_surface_diff` is also enabled with no changes.
- Both verdicts are `insufficient_evidence`; the related finding retains only MCP kind, source reference and agent name.

**This is lost evidence, not a demonstrated unsafe pass.** The variable name `ADMIN_KEY` proves no privilege level. What is established is a changed endpoint and credential reference, which the reviewer never sees.

[Extraction](https://github.com/ThreeMoonsLab/agents-shipgate/blob/20968551eb1d2b081756bea3ca44d82558c18c0c/src/agents_shipgate/inputs/google_adk.py#L1453) reads filter/inventory but not connection identity; [the model](https://github.com/ThreeMoonsLab/agents-shipgate/blob/20968551eb1d2b081756bea3ca44d82558c18c0c/src/agents_shipgate/core/artifact_models.py#L186) has no connection/credential-reference facts; [the finding](https://github.com/ThreeMoonsLab/agents-shipgate/blob/20968551eb1d2b081756bea3ca44d82558c18c0c/src/agents_shipgate/checks/adk.py#L196) cannot recover discarded evidence.

## Proposed direction

1. **Read and preserve evidence.** Start with the explicit ADK remote connection constructor above. Capture literal endpoint, direct `os.environ[...]` / `os.getenv(...)` references, literal filter and source spans. Dynamic expressions remain explicitly unresolved. Never import target code, connect to MCP or read environment values.
2. **Project the binding into existing review/diff surfaces.** Compare agent → remote binding before/after, independently of whether leaves were enumerated. Report endpoint/reference changes and established filter expansion/contraction. Reuse existing capability/review machinery; review persisted-schema compatibility if representation must change.
3. **Name the missing input.** Identify which binding lacks which inventory/configuration evidence. Supplying an inventory supplements that binding; it cannot conceal an independent connection change.

## Acceptance

- [ ] Endpoint-only, credential-reference-only and filter-only changes are named with base/head evidence in existing reviewer output.
- [ ] Line movement, formatting and comments alone produce no capability delta.
- [ ] Two agents using the same endpoint retain distinct attribution.
- [ ] Shadowed/rebound constructors, dynamic URLs, missing base and parse failures retain explicit limitations.
- [ ] Only credential-reference names are published; literal secrets and credentials in URL userinfo/query are protected by the existing redaction rules.
- [ ] A fixture containing `raise RuntimeError("must never execute")` is analyzed statically.
- [ ] Adding or updating leaf inventory cannot clear the independent endpoint/reference/filter delta.
- [ ] No hostname, variable or tool name is promoted to business-risk evidence; no remote leaves or reviewed authority/effect/binding declarations are invented.
- [ ] Existing release enum and qualification thresholds remain unchanged.

## Sequencing and non-goals

The reader/evidence slice can start now, alongside #516 and #536. Coordinate its comparison representation with #515; do not wait for a general dependency graph. It does not depend on #474's proposed committed state or #337's GitHub UI to deliver local value.

No adapter expansion campaign, endpoint execution, whole-Python evaluator, automatic reviewed declaration or blanket conversion of incomplete coverage to `review_required`.

## Surface discipline

Headline metric: the paired connection changes above produce correctly attributed, actionable review evidence, with zero deltas on the unchanged controls. Reuse existing reports and identity machinery; missing leaf inventory must not erase known binding facts.

pengfei-threemoonslab:
Closes #538. Parent #520; comparison shape published up front for #515 at [this comment](https://github.com/ThreeMoonsLab/agents-shipgate/issues/538#issuecomment-5574779300).

## The defect, reproduced

Two workspaces differing **only** in the endpoint literal and the `os.environ[...]` key — every source line preserved, filter unchanged — produced **byte-identical** `report.json` payloads apart from `manifest_dir`. Empty `capability_facts`, empty `tool_surface_facts`, one identical source warning, and the related finding carrying exactly `{"kind": "mcp", "source_ref": "agent.py:11", "agent_name": "search_agent"}`.

This is lost evidence, not a demonstrated unsafe pass. `ADMIN_KEY` proves no privilege level, and nothing this PR adds claims it does.

**After**, in existing reviewer output:

```
broadened scope 'search_agent -> MCP binding #1 endpoint'
  ('https://readonly.example/mcp' -> 'https://admin.example/mcp')
  — MCP endpoint changed; direction of change is not established: a host name
    or an environment variable name proves no privilege level; review required
```

## Three layers, all on machinery that already exists

**Read** — `_read_mcp_connection` parses `connection_params=` with `ast`: literal endpoint, the `os.environ` / `os.getenv` **names** credentials are read from, the transport, and the literal `tool_filter`. Every axis carries a status, so *"the reader did not look"*, *"the argument is not there"* and *"it is there and is not readable"* stay three different answers, and a per-binding limitation code names each doubt.

**Carry** — each binding rides base-to-head as **four independent per-axis** `ToolSurfacePolicyFact` rows, the carriage `core/toolkit_scope.py` already uses for toolkit bounds. Four rows rather than one is the point: the axes diff independently through the ordinary `_diff_policies`, which is what makes an endpoint-only, credential-reference-only or filter-only change land as its own named delta. `inventory_path` is deliberately not one of the four.

**Project** — into the existing `capability_change` block, which never gates. `tool` stays empty: no remote leaf is synthesized. A changed endpoint, credential reference or transport takes the block's documented opaque-direction bucket (`broadened`, `medium`) with a rationale that says so in words. The tool filter is the one axis where a direction is claimed.

## Acceptance

- [x] Endpoint-only, credential-reference-only and filter-only changes are named with base/head evidence in existing reviewer output.
- [x] Line movement, formatting and comments alone produce no capability delta — identity is `<agent>:<source_id>:<slot>` and no line number enters any key or hash.
- [x] Two agents using the same endpoint retain distinct attribution; so does one toolset shared by two agents, which now records **every** binding agent rather than the first.
- [x] Shadowed and rebound constructors, dynamic URLs, missing base and parse failures retain explicit limitations.
- [x] Only credential-reference names are published; URL userinfo and sensitive query values are redacted at the reader, and a literal credential is reported as present-and-withheld.
- [x] Every fixture carries `raise RuntimeError("must never execute")`.
- [x] Adding or updating leaf inventory cannot clear the connection delta — a test supplies the inventory *and* changes the endpoint, and both the enumerated leaves and the endpoint delta appear.
- [x] No hostname, variable or tool name is promoted to business-risk evidence; no remote leaves and no reviewed authority/effect/binding declarations are invented.
- [x] Release enum and qualification thresholds unchanged.

## Schema and identity effects

- **No `report_schema_version` bump.** `python scripts/generate_schemas.py --check` is clean. `ToolSurfacePolicyFact.kind` is a free string, so the four `remote_binding.*` kinds are additive.
- **No new check id, no new CLI surface, no new committed inventory.** Nothing in `docs/distribution-surfaces.md` changes.
- **Finding fingerprints move** for `SHIP-ADK-DYNAMIC-TOOLSET-NOT-ENUMERABLE` and `SHIP-ADK-MCP-TOOLSET-UNFILTERED`: their evidence now names the binding. The endpoint being part of the finding's identity is deliberate — a baseline that accepted the old endpoint must not silently accept a new one.
- **`input_set_id` / `subject_id` / `decision_id`** move for any workspace with an ADK MCP toolset, because `tool_surface_facts` gained content. `samples/google_adk_cold_start_agent/expected/report.json` is regenerated; no other golden changed.

## Three defects my own self-review found and fixed

1. **Cross-source key collision.** A two-component `agent:slot` key silently dropped one binding when two configured sources each declared an agent of the same name. The key now carries the source id.
2. **Member-id collision.** The capability member's id is hashed from its subject, and a subject that omitted the `#n` ordinal collapsed two *inline* bindings of one agent into one member, discarding one binding's evidence. Proven by re-running the new guard against the pre-fix rendering.
3. **An inverted filter direction, twice.** `["search"] -> (unresolved)` was classified as a **narrowing** because the sentinel decoded as an empty set — an unreadable filter may be wider, not narrower. And `tool_filter=[]` encoded identically to an absent filter, which reads the *narrowest* state as the *widest*. Both sides of a filter comparison now carry their kind, and only three transitions are treated as directional.

## Tests

`tests/test_adk_remote_bindings.py`, 40 cases: the reproduction; each axis in isolation; filter expansion, contraction, addition, removal, emptying and becoming unreadable; line movement / reformatting / comments; two agents on one endpoint; a shared toolset and one agent dropping it; a new and a removed binding; dynamic URL, rebound reference, shadowed constructor; missing base; parse failure; redaction; the leaf-inventory independence; the unchanged verdict and finding set; and the codec's round trip including values spelled like its own sentinels.

Full suite green (`pytest -n auto`), `ruff check .` clean, `scripts/generate_schemas.py --check` clean.

## Shipgate's own controls

`./shipgate verify --base origin/main --head HEAD --ci-mode advisory` on this branch: `control_state: complete`, `decision: passed`, exit 0. `./shipgate self-check --json` → `ready: true`. No gate was suppressed, lowered, waived or baselined.

## Coverage limits, published rather than implied

- Only the Python `McpToolset(connection_params=…)` constructor is read. An ADK YAML agent config records the binding with `not_read` statuses and a `connection_not_read_from_agent_config` limitation, plus a literal `connection_params.url` — it never claims the binding carries no credential.
- A change confined to **redacted** bytes (a rotated password in URL userinfo, a rotated hardcoded header) is not named: `value_hash` is computed over the published value, so a withheld secret is never hashed. Presence/absence of literal credential material still flips the axis, and the limitation codes stand while it is there. Referencing the credential through an environment variable is what makes the change comparable.
- A **renamed slot is a replacement**, not a rename: the diff is a set comparison, so a renamed toolset variable reports the old binding removed and the new one added. Both members carry their own endpoint, so a rename cannot hide a connection change — it only declines to call the two the same binding. (This corrects the pairing claim in my design comment on #538, which the implementation did not adopt.)
- One adapter. No adapter-expansion campaign; the carriage, identity rule and projection are framework-neutral and documented for reuse in [`docs/engineering/remote-binding-evidence.md`](docs/engineering/remote-binding-evidence.md).

## For #515

The only subject-bearing strings are the policy `key` and the capability member's `scope`. Everything else — pairing, per-axis hashing, the four rows — is subject-agnostic, so those two are the swap points when a canonical agent→binding subject lands.


pengfei-threemoonslab:
## Perturbation sweep: no vacuous guards

Eight ways to break the behavior this PR claims, each applied to the shipped code and run against the guard that is supposed to catch it. All eight caught:

| Perturbation | Guard |
| --- | --- |
| identity carries the source line | `line_movement…` |
| identity drops the agent | `two_agents_on_one_endpoint…` |
| connection facts suppressed once an inventory resolves | `leaf_inventory_cannot_clear…` |
| URL credentials not redacted at the reader | `literal_credentials_are_withheld…` |
| a literal header credential published instead of withheld | `literal_credentials_are_withheld…` |
| an empty filter folded into an absent one | `empty_filter_is_a_narrowing…` |
| a changed endpoint claims a direction | `claims_no_direction_and_no_risk…` |
| a new binding emits one member per axis | `new_binding_is_one_added_member…` |

The collision guard added in review was checked the same way, by replaying it against the pre-fix subject rendering.

Ready for independent review.

pengfei-threemoonslab (commented):
Reviewed commit `912419f41ffcf8eee67e7774ee9d2f4410596a29`. **Recommendation: changes required before merge.** Five actionable findings are attached inline: two P1 issues (empty-filter direction is opposite to ADK's behavior; credential query values can leak into reports) and three P2 issues (cross-source member collision; nested stdio credentials reported absent; mixed literal/reference credential additions missing from the delta).

Validation: 208 targeted tests passed across the ADK reader, remote bindings, tool-surface diff, verifier blocks, privacy, and capability-change schema/hash coverage. Ruff and schema generation checks passed. Independently reproduced the five report behaviors with static fixtures, and checked ADK's upstream source for filter and stdio semantics. No target agent was imported or executed.

Agents Shipgate evidence for the PR base/head: `control_state: complete`, `decision: passed`; self-check: `ready: true`. Those results coexist with the implementation defects above. The checked-out files remain unchanged.

Submitted as a COMMENT review because the connected account is also the PR author.

src/agents_shipgate/core/findings/verifier_blocks.py
```diff
@@ -408,6 +432,203 @@ def build_capability_change(
...
+            "",
+            scope=axis_subject,
+            before_scope=drift.before_summary,
+            after_scope=drift.after_summary,
+            rationale=rationale,
+        )
+
+    if before_kind == LIST_VALUES and after_kind == LIST_VALUES:
+        gained = sorted(set(after) - set(before))
+        lost = sorted(set(before) - set(after))
```
pengfei-threemoonslab: [P1] Treat an empty ADK filter as unbounded when calculating direction

ADK does not interpret `tool_filter=[]` as selecting no tools: [`BaseToolset._is_tool_selected`](https://github.com/google/adk-python/blob/main/src/google/adk/tools/base_toolset.py#L187-L198) returns `True` when the filter is falsy, and [`McpToolset.get_tools`](https://github.com/google/adk-python/blob/main/src/google/adk/tools/mcp_tool/mcp_toolset.py#L477-L502) uses that predicate. Comparing the existing fixture's `["search"]` against `[]` currently emits a high-confidence `narrowed` member, `release_impact="informational"`, and “search no longer reachable,” although the change exposes every advertised tool, including search. Normalize the ADK empty-list semantics before this set comparison and cover both directions; the new `test_an_empty_filter_is_a_narrowing_not_a_widening` currently asserts the opposite of the framework's behavior.

src/agents_shipgate/core/privacy.py
```diff
@@ -394,6 +394,59 @@ def looks_like_secret_value(value: str) -> bool:
...
+    head, question, query = redacted.partition("?")
+    if question:
+        parts: list[str] = []
+        for pair in query.split("&"):
+            key, equals, _value = pair.partition("=")
+            if equals and is_sensitive_key(key):
+                parts.append(f"{key}={marker}")
+                changed = True
+            else:
+                parts.append(pair)
```
pengfei-threemoonslab, line 491: [P1] Redact decoded credential query keys before publishing endpoints

This checks the raw query-key spelling, so a URL such as `https://host.example/mcp?api%5Fkey=demo-password` bypasses redaction even though its decoded parameter name is `api_key`. I reproduced `redact_url_credentials` returning the original URL with `False`, and a full scan retaining `demo-password` in `report.json`'s endpoint policy summary and finding evidence. The ordinary OAuth spelling `access_token=demo-password` also leaks because it is absent from the shared sensitive-key vocabulary. Parse/decode query keys before classification and cover credential query names such as `access_token`; add end-to-end assertions that neither form reaches the public report or its value hash.

src/agents_shipgate/core/findings/verifier_blocks.py
```diff
@@ -408,6 +432,203 @@ def build_capability_change(
...
+    axis = axis_label(drift.policy_kind)
+    # The slot is always part of the subject, including the ``#n`` ordinal
+    # form. It reads slightly heavier for an agent with one binding, and it is
+    # what keeps two *inline* bindings of one agent apart: the member id is
+    # hashed from this string, so dropping the ordinal collapsed both bindings
+    # into one member and silently discarded one binding's evidence.
+    binding_subject = " ".join(
+        part for part in (agent, "-> MCP binding", slot) if part
+    )
+    axis_subject = f"{binding_subject} {axis}"
```
pengfei-threemoonslab, line 488: [P2] Preserve the source ID in capability-member identity

The policy key distinguishes `(agent, source_id, slot)`, but this subject drops `source_id`, and `_MemberFactory` hashes the subject into the member ID. With two configured ADK files both declaring `search_agent` and inline slot `#1`, changing both endpoints produces two distinct `remote_binding.endpoint` policy-drift rows but only one capability member after deduplication. In my reproduction, the `adk_other` change survives and the `adk_search` endpoint change disappears from `capability_change`. Include the full binding identity in the member subject/ID and extend the cross-source test through report projection; its current codec-only assertion misses this collision.

src/agents_shipgate/inputs/google_adk.py
```diff
@@ -1450,9 +1701,101 @@ def _extract_openapi_toolset(self, call: ast.Call, agent_name: str) -> list[Load
...
+            transport = None
+            transport_status = "unresolved"
+        else:
+            transport_status = "literal"
+        endpoint, endpoint_env_ref, endpoint_status = _read_endpoint(
+            expr, self.aliases, limitations
+        )
+        credential_refs, credential_status = _read_credential_refs(
+            expr, self.aliases, limitations
+        )
```
pengfei-threemoonslab: [P2] Read credentials inside StdioConnectionParams.server_params

For the recognized `StdioConnectionParams` constructor, this passes the outer call to `_read_credential_refs`, which only examines top-level `headers` and `env`. ADK places the server environment in [`server_params: StdioServerParameters`](https://github.com/google/adk-python/blob/main/src/google/adk/tools/mcp_tool/mcp_session_manager.py#L216-L225). Scanning `StdioConnectionParams(server_params=StdioServerParameters(command="npx", args=["server"], env={"API_KEY": os.environ["READ_KEY"]}))` therefore reports `credential_status="absent"` with no limitation. Changing only `READ_KEY` to `ADMIN_KEY` produces no binding delta. Resolve the nested literal server parameters, or explicitly report this credential axis as unread/unresolved instead of claiming absence.

src/agents_shipgate/core/remote_bindings.py
```diff
@@ -0,0 +1,356 @@
...
+            key=key,
+            value_hash=_value_hash(binding.endpoint_status, endpoint_value),
+            summary=_endpoint_summary(binding),
+        ),
+        ToolSurfacePolicyFact(
+            kind=CREDENTIAL_REFS_KIND,
+            key=key,
+            value_hash=_value_hash(
+                binding.credential_status, sorted(binding.credential_refs)
+            ),
```
pengfei-threemoonslab, line 291: [P2] Carry literal-credential presence alongside environment references

When a connection already has `headers={"Authorization": os.environ["READ_KEY"]}`, adding `"api_key": "demo-password"` produces no credential policy drift or capability member. `_read_credential_refs` keeps status `environment_reference` whenever any reference exists; it records the new literal only in `limitations`, which this hash and summary omit. I reproduced an unchanged credential fact despite the new `literal_credential_value` limitation in finding evidence. This is a credential being added, beyond the documented limitation for rotating already-withheld bytes. Carry non-secret presence/completeness information alongside the references so adding or removing literal credential material remains comparable without hashing or publishing its value.

src/agents_shipgate/core/findings/verifier_blocks.py
pengfei-threemoonslab: Fixed in `41641bf1`, and you are right about the framework. I downloaded the `google-adk` 2.8.0 wheel and read it rather than trusting my reasoning: `BaseToolset._is_tool_selected` opens with `if not self.tool_filter: return True`, so an empty list is no filter and exposes every advertised tool.

I had introduced the empty-list sentinel *during self-review* precisely to avoid a direction inversion, and produced the opposite one by reasoning from set semantics instead of from ADK. The direction logic now asks whether a side **bounds** anything (`_filter_is_unbounded`), and an empty literal list answers no. Two unbounded sides now produce **no member at all** — the policy-drift row still records the edit, because `tool_filter=[]` and an absent argument are different source text, but nothing about what the agent may call moved.

`test_an_empty_filter_is_a_narrowing_not_a_widening` is replaced by `test_an_empty_filter_is_a_widening_because_adk_ignores_it` (citing the upstream predicate) plus `test_adding_an_empty_filter_where_there_was_none_is_no_capability_change`.

src/agents_shipgate/core/privacy.py
pengfei-threemoonslab: Fixed in `41641bf1`. Both halves reproduced before the fix.

Query keys are now classified **after** `unquote_plus`, so `api%5Fkey` lands on `api_key`. And the classification moved off the exact `SENSITIVE_VALUE_KEYS` vocabulary onto a new public `privacy.is_credential_key`: the exact set plus a suffix rule (`…token`, `…secret`, `…password`, `…apikey`, `…signature`, `…credential`, `…authorization`), which covers `access_token` and `x-api-key` without needing to enumerate compound spellings.

It deliberately excludes a bare `key`, so `sort_key` and `partition_key` stay ordinary parameters — over-redaction is the safe direction for a value, but claiming a credential where there is none is its own false statement. The same predicate now classifies connection header and `env` keys, which closes the matching under-report there (`X-Access-Token: "literal"` previously read as non-credential).

`test_encoded_and_oauth_query_credentials_are_redacted` asserts end-to-end that neither spelling reaches `report.json`, while `page=1` survives. I left `is_sensitive_key` and `SENSITIVE_VALUE_KEYS` untouched so the generic output-redaction pass keeps its existing behavior.

src/agents_shipgate/core/findings/verifier_blocks.py
pengfei-threemoonslab: Fixed in `41641bf1`. This is the same class as the `#n` ordinal collision I found in self-review, one level up, and I stopped at the policy key without following it into the member id — thank you for pushing it through the projection.

The member subject now carries the whole binding identity, spelled `search_agent [adk_search] -> MCP binding #1 endpoint`. The bracketed form is the one the release-decision reason text already uses, so it stays a `tool_sources[].id` an adopter can open rather than a bare internal identifier.

`test_two_sources_with_the_same_agent_name_keep_both_capability_members` builds exactly your reproduction — two configured ADK files, both declaring `search_agent` at inline slot `#1`, both endpoints changed — and asserts both members survive with their own before/after. It goes through the report projection, not the codec.

src/agents_shipgate/inputs/google_adk.py
pengfei-threemoonslab: Fixed in `41641bf1`. Confirmed against the wheel: `StdioConnectionParams.server_params: StdioServerParameters`, so `env` is never inline. My earlier probe used `StdioConnectionParams(command=…, env=…)`, which is not a shape ADK accepts — the fixture was wrong, so the reader looked right.

`_read_connection_credentials` now resolves the nested call, inline or through a module-level name the module binds exactly once, and merges its entries under a `server_params.` prefix. The status is the strongest claim either side supports, so `absent` survives only when both sides are.

One extra defect surfaced while writing your test: an *unrecognized* nested call (`server_params=build_params()`) still fell through to the credential reader and returned `absent` — the same false absence one level down. It now requires a recognized connection-params constructor and otherwise reports `unresolved` with an `unresolved_nested_server_params` limitation. Three tests: the nested read, a changed nested reference producing a delta, and the unresolved case.

A bare `StdioServerParameters` passed directly as `connection_params` (the deprecated path the session manager still accepts) was already covered by the top-level `env` read.

src/agents_shipgate/core/remote_bindings.py
pengfei-threemoonslab: Fixed in `41641bf1`. Agreed on the distinction — this is a credential being *added*, which is not what the documented rotation limitation covers, and the limitation-only record was invisible to exactly the two fields that decide whether anything changed.

The credential axis now lists **every** entry rather than only the references: a value read and withheld renders as `headers.api_key=<literal credential withheld>`, one that could not be read as `<not statically readable>`. The key name is not a secret, so presence and completeness are carried without a byte of the value, and the summary, the hash and any set comparison all move when a literal credential is added or removed. Angle brackets rather than parentheses so a one-entry list can never be mistaken for a whole-summary status sentinel.

`test_adding_a_hardcoded_credential_beside_a_reference_is_a_delta` runs your exact scenario and additionally asserts `demo-password` appears nowhere in the report.

pengfei-threemoonslab:
## All five findings addressed in `41641bf1`

Every one was valid. I checked both P1 claims against `google-adk` 2.8.0's own source rather than reasoning about them, and both held.

**Three were the same mistake**: reasoning about a framework's semantics from the shape of the data instead of from the framework.

| # | Was | Now |
| --- | --- | --- |
| P1 empty filter | `[]` compared as the empty set → `narrowed`, "search no longer reachable" | `BaseToolset._is_tool_selected` returns `True` for any falsy filter, so `[]` is *no* filter → `broadened`; two unbounded sides produce no member at all |
| P2 stdio | outer call only → `credential_status: absent` for a binding that had one | nested `server_params` resolved; an *unrecognized* nested call reports `unresolved`, not the same false absence one level down |
| P1 query keys | raw spelling matched against the exact vocabulary | classified after `unquote_plus`, against a new `privacy.is_credential_key` (exact set + credential-name suffixes); bare `key` deliberately excluded so `sort_key` stays ordinary |

**Two were silent losses**, both where a value that decides identity or change was left out of the thing that computes it:

- The member id is hashed from the subject, and the subject dropped `source_id` — two same-named agents from different sources merged into one member and one endpoint change disappeared. The subject now carries the whole identity as `agent [source_id]`.
- A hardcoded credential added beside an environment reference was recorded only in `limitations`, which the summary and hash never see. The credential axis now lists every entry, with `<literal credential withheld>` standing in for a value it will not publish — presence and completeness carried, not a byte of the value.

Worth naming: the empty-filter sentinel was something **I added during self-review** to avoid a direction inversion, and it created the opposite one. Reasoning from set semantics was the wrong tool; the answer was in `base_toolset.py`. Writing the P2 stdio test also surfaced a sixth defect of the same family — my fix's own unrecognized-nested-call path returned `absent` — now fixed and covered.

### Verification

- 6 new tests for these findings (47 total in `tests/test_adk_remote_bindings.py`), each running the reproduction from your comment.
- **Seven-perturbation sweep, one per fix, all caught.** Reverting any single fix fails its guard.
- Full suite green, `ruff` clean, `generate_schemas.py --check` clean. No sample golden moved.
- `./shipgate verify --base origin/main` on the new head: `control_state: complete`, `decision: passed`, exit 0.

No schema bump, no new check id, and the release enum and qualification thresholds remain untouched.

PR #541, opened 2026-09-08
FastMCP signature projection resolves the injected Context and discloses unsupported schema (#539)

### LINKED ISSUE #539: FastMCP signature projection must resolve injected Context identity and disclose unsupported schema
## Summary

#535 makes Python MCP servers discoverable and reads their signatures. The signature projection can still **erase a real required input** or **invent a required argument the framework injects** because request-context exclusion matches the annotation's spelling rather than its binding. Unsupported annotations are also emitted as concrete `string` schemas without a signature coverage gap.

This is a correctness follow-up to the delivered adapter, not another discovery expansion or a new parity framework.

## Evidence

Reproduced with the production `load_mcp_server_source()` on main `20968551e`, reading the following source files statically. None of the target files were imported or executed.

```python
from mcp.server.fastmcp import FastMCP
from pydantic import BaseModel

class Context(BaseModel):
    account_id: str

mcp = FastMCP("s")

@mcp.tool()
def update(context: Context) -> str:
    return "ok"
```

The application model's required input disappears:

```json
{"input_schema":{"type":"object","properties":{},"required":[]},"function_signature":"update() -> str"}
```

The opposite spelling fails in the opposite direction:

```python
from mcp.server.fastmcp import FastMCP, Context as RequestContext
mcp = FastMCP("s")

@mcp.tool()
def lookup(query: str, ctx: RequestContext) -> str:
    return "ok"
```

The injected framework context becomes a user-supplied required string:

```json
{"input_schema":{"type":"object","properties":{"query":{"type":"string"},"ctx":{"type":"string"}},"required":["query","ctx"]},"function_signature":"lookup(query, ctx) -> str"}
```

A third case, `def lookup(limit: int | None = None) -> str`, emits `limit: {"type":"string"}`. All three report `extraction.surface: "enumerated"`, `confidence: "medium"`, no warnings and no omissions.

Negative controls were also run: canonical framework `Context` is correctly omitted, and an ordinary parameter named `context` annotated `str` is correctly retained. Neither behavior should regress.

Reproduce by writing either source to `server.py` and reading it through:

```python
from pathlib import Path
from agents_shipgate.inputs.mcp_server_source import load_mcp_server_source
from agents_shipgate.schemas.manifest import ToolSourceConfig

loaded = load_mcp_server_source(
    ToolSourceConfig(id="s", type="mcp_server_source", path="server.py"),
    Path("/absolute/path/to/fixture"),
)
tool = loaded.tools[0]
print(tool.input_schema, tool.function_signature, tool.extraction)
```

Run against the checkout with `PYTHONPATH=src <supported-python> reproduce.py`.

## Why it happens

- `src/agents_shipgate/inputs/mcp_server_source.py::_context_annotation` strips quotes, optional/generic syntax and the module prefix, then `_is_context_parameter` compares the last token with `Context`. Its docstring explicitly says the binding is not resolved.
- `_signature_parameters` removes a matching input; `_signature_input_schema` publishes the result as an object schema, including an empty object when the only input was removed.
- `inputs/python_static.py::json_schema_type` falls back to `string` for unrecognised annotations. The FastMCP projection carries no local schema-fidelity gap for that case.
- The SDK's [context injection reader](https://github.com/modelcontextprotocol/python-sdk/blob/v1.x/src/mcp/server/fastmcp/utilities/context_injection.py) resolves type hints and checks class identity/subclassing against its actual `Context`. Its [tool builder](https://github.com/modelcontextprotocol/python-sdk/blob/v1.x/src/mcp/server/fastmcp/tools/base.py) skips the identified parameter. An application class with the same spelling is not that class.

## User impact and boundary of the observation

The parameter inventory and signature are facts a reviewer and downstream schema/diff consumers use. One case hides `context.account_id` behind a tool presented as taking no arguments; the other asks the caller for a framework-provided value. The action projection also derives its input field inventory from `tool.parameters`.

These reproductions establish wrong input facts, **not a reproduced false-pass release decision**. The adapter's medium confidence ceiling remains appropriate; it does not repair a concrete schema assertion the source does not support. Tool-name enumeration and signature fidelity are different claims.

## Implementation direction

1. Add the two context counterexamples at the production loader boundary before changing the implementation. Preserve the canonical-context and ordinary-name controls.
2. Establish bounded annotation identity using the existing Python import/alias/rebinding machinery: known framework context, known non-framework annotation, or unresolved. Exclude a parameter only when the injection identity is established. Never import or execute target code.
3. Keep a visible parameter and a precise signature gap when injection identity is unresolved; do not silently erase it or state it is definitely user-supplied. Reuse existing extraction/coverage structures rather than adding a parallel verdict.
4. In a separately reviewable slice, check schema fidelity before publishing a concrete annotation-derived type. Represent supported nullable/container forms accurately; for other forms retain the parameter and expose the local unknown. The ADK `_annotation_is_faithful` / `_function_surface_gaps` work provides an existing precedent. Do not require a complete Python type system or change every adapter's shared emitter in this fix.
5. Preserve known tool names and the discovery route when only signature evidence is partial. If registration-site facts change, extend the existing shared `tests/mcp_idiom_corpus.py` parity cases in both readers.

## Acceptance criteria

- [ ] A required application model named `Context` remains in the parameter inventory and required inputs; its unresolved schema is not silently asserted to be an empty tool interface.
- [ ] A canonical or aliased, statically bound framework context is excluded from caller inputs.
- [ ] Qualified names, import aliases, same-name application classes, rebinding and unresolved imports are covered; unresolved identity produces a named signature limitation.
- [ ] `int | None` and an unsupported model annotation are represented faithfully or explicitly partial, never silently asserted as string.
- [ ] `input_schema`, `parameters`, `function_signature` and derived action input fields describe the same evidence.
- [ ] A partial signature retains the known tool and route, does not increase extraction confidence, and cannot satisfy missing effect/authority/binding declarations.
- [ ] The new defect regressions fail on `20968551e`; preserved-behavior controls pass before and after. Both detector readers remain aligned where shared site facts change.

## Non-goals

Executing a server to discover its schema; evaluating arbitrary type expressions; complete Pydantic support; importing external framework packages in scans; promoting medium-confidence extraction; broadening the list of supported framework adapters.

## Related

#535/#484 — delivered Python registration support. #357/#440 — evidence versus projections and honest reader-facing conclusions. #497 — existing distribution conformance, reused if shared facts change.

## Source anchors and surface discipline

[Context exclusion](https://github.com/ThreeMoonsLab/agents-shipgate/blob/20968551eb1d2b081756bea3ca44d82558c18c0c/src/agents_shipgate/inputs/mcp_server_source.py#L442), [schema fallback](https://github.com/ThreeMoonsLab/agents-shipgate/blob/20968551eb1d2b081756bea3ca44d82558c18c0c/src/agents_shipgate/inputs/python_static.py#L94), [current canonical-context test](https://github.com/ThreeMoonsLab/agents-shipgate/blob/20968551eb1d2b081756bea3ca44d82558c18c0c/tests/test_mcp_server_source.py#L1241).

Headline metric: no lost real parameter or invented caller requirement in the paired context fixtures, with unsupported signature evidence explicitly localized. Reuse existing extraction/coverage and conformance structures; no new top-level verdict or report summary.

pengfei-threemoonslab:
Closes #539.

`@mcp.tool` signatures are read by #535, and two of the facts that reading
published could be wrong. Both were reproduced through the production
`load_mcp_server_source()` on `20968551e`, statically, with none of the target
files imported or executed.

## What was wrong

**A real required input disappeared.** An application model named `Context`
was matched as the framework's request context and dropped, so the catalog
published `update() -> str` — a tool presented as taking no arguments, with
`context.account_id` behind it, on `enumerated` evidence, no warning and no
omission.

**A framework-supplied value became a caller requirement.** `from
mcp.server.fastmcp import Context as RequestContext` did not match, so
`ctx: RequestContext` shipped as a required `string` the caller is asked to
supply.

One cause: the request context was identified by the **last token of the
annotation's spelling**, while the SDK's injection reader resolves the type
hint and checks class identity against its own `Context`.

**And a third, separately reviewable:** `json_schema_type` reads the *rendered*
annotation and answers `string` for everything it does not recognise, so
`int | None`, `Annotated[int, Field(ge=1)]`, `typing.List[str]` and a Pydantic
model all shipped as concrete string schemas — with no gap recorded anywhere.

## What this changes

**Injection identity (commit 1).** The reader answers from the module's own
import and class table: the machinery that already follows `@mcp.tool` back to
a `FastMCP(...)` construction, with the same prefix rule and the same refusal
on a doubly-bound name, plus the base list, because the SDK injects any
*subclass* of its context. The context module family is derived from
`PYTHON_SERVER_CONSTRUCTORS`, so it cannot go stale on the rename that hid 229
tools in #484. `SignatureParameter.injection` is a tri-state — "not
established" is a different answer from "established as a caller input", and
publishing them as one is how a parameter disappears.

**Schema fidelity (commit 2).** A parameter's type is what its annotation
*denotes*, read from the annotation tree against the same binding table.
Nullable unions retain a local coverage gap because this projection cannot express both a value and null. `Annotated[T, ...]` projects `T`, and builtins and
`typing` aliases are canonical only while the module has not rebound them.
Where the type was not read, **nothing is published**: the property is `{}`,
`ToolParameter.type` is `None`, and the tool carries `untyped_parameter` or
`unrepresentable_annotation` — the vocabulary `google_adk` established, pinned
equal by a test. `output_schema` follows the same rule; an *absent* return
annotation stays the honest omission it was.

**Measured, then corrected (commit 3).** Running the finished reader over
`redis/mcp-redis` and `chroma-core/chroma-mcp` and reading every refusal found
two defects of its own: a builtin outside the published-type table (`bytes`)
read as an *unresolved injection identity*, and requiring a container's
elements to be representable reported `Dict[str, Any]` — the most common return
spelling in `mcp-redis` — as a type nobody read. A container's kind does not
depend on what it holds; a mapping's *key* does, so `dict[int, str]` is still
refused.

Where identity or type is not established, the parameter stays visible and the
tool says which question is open. This input's failure modes are a real input
erased and an invented caller requirement, and both are worse than an admitted
gap.

## What does not change

The tool keeps its name, its file and line, the `medium` ceiling and the
exclusion ledger it already had; a per-tool signature gap is not a registration
the reader failed to read, so nothing enters the omission ledger for it, and it
closes no effect, authority or binding declaration (asserted through the
engine's own `surface_is_complete` / `extraction_is_complete`). No new idiom, no
new verdict, no new report block: `IDIOM_REGISTRY_VERSION` stays `2` and
`report_schema_version` is untouched. The zero-install detector carries the same
resolution — its published verdict, frameworks and evidence are unchanged, so
`script_version` stays `0.6.0`.

## Measurements before review corrections

| Repository | Tools named | Held `partial` | Remaining unreadable annotations |
|---|---|---|---|
| `redis/mcp-redis` | 53 (unchanged) | 14 (21 before commit 3) | unions of different JSON kinds, one opaque library alias |
| `chroma-core/chroma-mcp` | 13 (unchanged) | 0 | — |

## Review corrections

Round 1 identified two remaining #539 acceptance defects, addressed in `477f16ee` after preserving the concurrent cycle-guard test commit `5ef848cd`:

- Nullability is kept separate from whether a default permits omission. Eight nullable variants now retain the parameter and explicitly report the unrepresentable type.
- A relative module's existence does not establish the class it exports. A local framework-Context re-export now retains `unresolved_context_identity` rather than claiming caller ownership.

The measurements above describe the earlier implementation, not the corrected head. Broader SDK whole-signature/generic-context injection behavior is tracked separately in #542 and deferred from this PR.

## Verification

- **16 new tests fail on `20968551e`**; every pre-existing test in
  `tests/test_mcp_server_source.py` passes there, and so do the
  preserved-behavior controls this PR adds (canonical context, ordinary
  parameter named `context`, the qualified and forward-reference spellings).
- **14 new shared-corpus cases** in `tests/mcp_idiom_corpus.py`, driven through
  both readers; the parity comparison now includes `injection`, `json_type` and
  `returns_json_type`, so a port returning the dataclass defaults would agree on
  every name and disagree about every signature.
- **An AST comparison** (annotations and docstrings stripped) reports 14/14
  shared reader functions byte-identical across the two implementations.
- **A 17-case perturbation sweep** over every decision these commits add — the
  prefix match, the base walk, its cycle guard, the relative-import branch,
  `binds_name`'s stricter test, the `typing`-module gate, union-arm agreement,
  the mapping-key requirement, `Annotated`'s first element, the
  forward-reference parse, the exclusion itself, each gap, and holding a gapped
  tool at `partial`: **17/17 caught**. The cycle guard is the one that found a
  live defect in this PR — it carried `# pragma: no cover - a class cannot be
  its own base`, which is true of one class and false of two, and deleting it
  raises `RecursionError` on `class A(B)` beside `class B(A)`.
- `ruff check .`, `python scripts/generate_schemas.py --check`, and the three
  CI-shaped pytest runs (`-n auto -m "not perf"`, `test_adapter_static_only.py`,
  `test_latency_budget.py -m perf`) are all green.
- `shipgate check --agent claude-code` → no violations, no issues;
  `shipgate preflight --changed-files` → routed to `verify`;
  `shipgate verify --ci-mode advisory` → `control.state=complete`,
  `release_decision.decision=passed`.

## Surface discipline

No new surface. `SURFACE_GAP_UNRESOLVED_CONTEXT` is one new reason token inside
the existing `extraction["surface_gaps"]` list; the other two reuse
`google_adk`'s spellings. The `mcp_server_source` coverage cell is updated and
`docs/determinism-boundary.{md,json}` regenerated from it.

**Headline metric:** blocked-recall and noise rate. No lost real parameter and
no invented caller requirement in the paired context fixtures, with unsupported
signature evidence localized to the tool it belongs to.


src/agents_shipgate/inputs/mcp_idioms.py
```diff
@@ -2415,6 +2512,422 @@ def _python_annotation(node: ast.expr | None) -> str | None:
...
+    An optional is the type it wraps. Our property schema carries one type, and
+    the framework's own spelling for "this input may be omitted" is exactly
+    ``T | None`` — so naming ``T`` is the same *kind* the source names, while
+    ``string`` for an integer is a different one. A union of several types
+    denotes one only when every arm agrees.
+    """
+
+    if node is None or depth > _MAX_ANNOTATION_DEPTH:
+        return None
+    members = _python_annotation_members(node, module, scope, depth=depth)
```
pengfei-threemoonslab: [P2] Preserve the nullable alternative or disclose that it was not represented. `_python_annotation_members` discards None, so `def lookup(limit: int | None = None)` currently emits `{type: number}` and `surface: enumerated` with no gap. Nullability and omission are distinct: without a default, the same nullable parameter is required but still accepts null. #539 explicitly requires this shape to be faithful or explicitly partial. Keep null in the JSON-type reasoning (or emit the existing unrepresentable-annotation gap); do not silently narrow the accepted value domain.

src/agents_shipgate/inputs/mcp_idioms.py
```diff
@@ -2415,6 +2512,422 @@ def _python_annotation(node: ast.expr | None) -> str | None:
...
+        imported = module.import_of(node.id, scope)
+        if imported is None or imported.symbol is None:
+            # ``import a.b`` binds a module, and a module is not a class.
+            return "unresolved"
+        if imported.level:
+            # A relative import names a module of *this repository*, which is
+            # not the installed framework — but only once the walk has found
+            # it. An import pointing outside the scanned tree is one this
+            # reader cannot place at all.
+            return (
```
pengfei-threemoonslab: [P2] A resolved module path does not resolve the imported class identity. A local `models.py` containing `from mcp.server.fastmcp import Context`, consumed as `from .models import Context`, makes this branch claim caller_supplied even though it re-exports the exact injected class. The production loader then requires ctx and reports only an unrepresentable type, not the unresolved injection identity. Until symbol provenance is actually followed, retain `unresolved`; merely finding the file is not the missing evidence. Add the re-export negative control alongside the ordinary local-model case.

pengfei-threemoonslab (commented):
Review/address round 1 of at most 3 for #539. The original application-Context and framework-alias counterexamples now behave correctly, and the targeted loader/idiom/standalone-detector suite passes. Two remaining acceptance defects are attached inline: nullable annotations still publish a non-null concrete type without a gap, and locating a relative-import module is treated as proof that its exported class is not the framework Context. Both were reproduced through the production static loader; no target module was executed. I will address these within this PR. Separately discovered SDK whole-signature/generic injection questions will be filed for later work, as requested, rather than expanding this fix. This is a COMMENT review because the connected account is also the PR author; it is not an independent human approval.

pengfei-threemoonslab:
Round 1 addressed in `477f16ee0fec4e7f3aa82d5b66f14038ee043187`.

Both review findings were reproduced before editing. All 9 added cases failed on the reviewed head and pass after the repair: 8 nullable spellings/default combinations plus the local Context re-export. Nullable unions now publish no invented single type and carry the existing local gap; defaults alone decide requiredness. Finding a relative module no longer asserts the identity of its export. Both readers were updated together, and the old tests that encoded those incorrect claims were corrected.

Preserved the author's concurrent `5ef848cd` cycle-guard test commit with a three-way checkout; no force push. Targeted loader/idiom/standalone checks pass on the resulting head. The full non-perf run had no functional failures; its 10 wheel setup errors cleared on rerunning all packaging tests with dependency-download access. Performance tests (4), Ruff and schema generation checks also pass. Committed base/head verification returns `control_state: complete`, `decision: passed`.

The additional SDK generic/whole-signature injection uncertainty is filed as #542 and deferred, rather than broadening this issue.

pengfei-threemoonslab (commented):
Review/address round 2 of at most 3, on `477f16ee`.

Rechecked both round-1 findings against the committed fixes and their negative controls. Nullable values are no longer silently excluded by a concrete non-null schema, and a found relative-import file no longer substitutes for class provenance. The application Context, canonical/aliased framework Context and ordinary context: str controls retain their behavior; the concurrent base-cycle guard test is preserved. Both inline findings are resolved by this commit. No additional in-scope change is requested in this round.

Validation: targeted loader/idiom/standalone suite, full non-perf run plus successful network-enabled wheel retry, four latency cases, Ruff, generated schema check, and committed-head verifier complete/passed. Remaining SDK injection semantics are explicitly deferred in #542. Submitted as a COMMENT review because this connected account is also the PR author; this records the requested review loop, not independent human approval.

Issue #543, opened 2026-09-08
Preserve why agent.name is unresolved through init and doctor recovery


pengfei-threemoonslab:
Found while implementing #533; deferred as a separate routing question at the user's request.

## What was observed

#533 prevents a test/template-only candidate from being automatically written as the product's identity. That selection decision does not survive as a specific ownership condition in later setup routing.

On a synthetic test-only ADK workspace after that fix, `init --write --json` writes `agent.name: CHANGE_ME`. Its human review action names `agent.declared_purpose[0]`, not the unresolved identity. After supplying only the fixture's purpose and running `doctor --json`, the ranked actions include an ordinary agent edit: `Replace CHANGE_ME at field 'agent.name'`. A missing-tool warning outranks it in this exact fixture; it is still an offered recovery, not a named human identity decision.

`cli/discovery/placeholders.py::placeholder_owner` classifies `agent.name` as coding-agent-owned. That is useful when a name can be read from product code, but it does not distinguish that case from discovery explicitly rejecting every candidate as non-product. A generic edit instruction cannot tell a caller which situation it is in.

## Question to settle

When there is no supported product-identity evidence, should the route require a person to name the product, request a supported source that establishes it, or retain an explicit unresolved state? Do not make every missing name human-owned: many names are mechanically recoverable from legitimate product declarations.

## Proposed follow-up

- Carry or recompute the reason the identity is unresolved using existing discovery/diagnostic mechanisms, without trusting a stale discovery artifact.
- Distinguish a mechanically readable product name from a choice no scanned source establishes.
- Make init and doctor name the required evidence/decision consistently after unrelated purpose/permission placeholders have been filled.
- Keep scope ambiguity, truncated discovery and all-false setup permissions intact.

## Acceptance

- [x] A fixture with only test/template names never offers an instruction that implies any nonempty replacement is adequate identity evidence.
- [x] A product-declared name remains recoverable by the coding agent where static evidence establishes it.
- [x] The missing-identity route remains visible after purpose is supplied; tests do not use an unrelated purpose obligation as proof that identity itself was routed to a human.
- [x] Ranked actions and compact control agree; new persisted context, if necessary, follows the existing compatibility discipline.

This does not block #533's bounded auto-selection correction. It owns the later recovery semantics, not a new naming heuristic or general approval framework.

## Delivered — PR #594

[...]

PR #544, opened 2026-09-08
Keep fixture-only names out of the reviewed product identity (#533)

### LINKED ISSUE #533: A name only non-product code declares should not be asserted as the reviewed identity
Found while reviewing #532 (the fix for #398).

## Summary

`select_agent_name` takes the first `selectable` candidate. The ranker penalises a name declared only in **non-product** code — test fixtures, and (since #532) scaffolding templates under `resources/templates/` — by `ORIGIN_NON_PRODUCT_PENALTY`, but the penalty only moves it down the list. When it is the *only* candidate, it is still selected and `init` writes it as `agent.name`.

## Reproduction

A repository whose only Python agent file is `tests/test_agent.py`:

```python
from google.adk.agents import LlmAgent

worker = LlmAgent(name="AlphaFixture")
```

`detect` reports `agent_scope: single` and `AlphaFixture` with `rank_score: -2.0`, `selectable: true`. `init --write` declares a test fixture as the repository's reviewed identity.

The same holds for a scaffolding template:

```
skills/recipe/resources/templates/app/agent.py  →  TemplateWorker
```

## Why it matters

The ranker's own documented contract is that "product code outranks code that is not the product". Outranking nothing is not the same as being the product. `agent.name` is what the alignment layer compares the observed capability surface against, so an example's name makes that comparison meaningless — the same class as #320 (a one-character test literal became a repository's declared identity) and #324.

`CHANGE_ME` plus an explicit review action is the honest output here, exactly as it is for a value below the quality floor or a generic scaffolding name.

## Why it was not fixed in #532

Two reasons, both worth weighing before acting:

1. **It is pre-existing and wider than #398.** The readable-root case above already behaved this way before #532. What #532 changed is that the *unreadable*-root variant no longer accidentally forces `CHANGE_ME` — so the two cases now agree, in the direction the rest of the ranker already went.

2. **It trades against first value.** #398 is a P1 first-value blocker; turning more repositories into `CHANGE_ME` runs the other way. Whether a test-only or template-only repository should get a name at all is a product call, not a bug fix.

## Suggested direction

Track whether *every* site declaring a value is non-product (`_non_product_origin` already answers this per site, and `_rank_agent_name_candidates` already accumulates the sites a value is declared in for #398's project scoping), and set `selectable: false` with a rationale saying so. It is a quality-floor rule, so it belongs beside the two that already exist rather than inside `select_agent_name`.

Any change here must be mirrored in `tools/shipgate-detect.py`: `agent_name_candidates` is pinned byte for byte (`docs/zero-install.md`).

## Product decision and implementation — 2026-09-06

**P1: preserve the reviewed identity's meaning.** A successful setup populated with a fixture's identity is misleading first value. Keep `CHANGE_ME` and its existing human-owned review route when every declaration site is non-product; do not invent a replacement name or ask a coding agent to assert one. This does not make an otherwise recognized tool surface “not an agent project”.

Implement as a bounded quality-floor change in the existing ranker, using the complete declaration-site set already collected. A value also declared in product code must not be rejected merely because an earlier site was a test/template. Mirror the result in the standalone detector and reuse the existing byte-level parity contract, delivered through #497.

### Acceptance

- [ ] Test-only and `resources/templates/`-only agent names remain visible candidates with `selectable: false` and an exact non-product rationale.
- [ ] `init --write` does not assert those names as reviewed identity, lists the unresolved name, and preserves existing human-declaration routing. The later identity-specific recovery is tracked in #543.
- [ ] A value declared in both fixture and product code remains eligible under the existing quality rules, regardless of file order.
- [ ] Product-only valid names, scope ambiguity and truncated-discovery protections retain their current behavior.
- [ ] Installed and standalone detector candidate payloads remain byte-equivalent; use existing parity tests rather than a new framework.
- [ ] Tests cover detect → init output, not only the ranker's sorting function.

This can ship independently of #536/#537 and #515. Coordinate changes to the shared init file with #328's small error-route fix; the two do not depend on one another semantically.

## Implementation boundary discovered during validation

The selection correction and init output are covered here. A human route immediately after init currently names the purpose declaration; it is not proof that the unresolved identity itself has a human-owned postcondition. After purpose is supplied, doctor can offer a generic agent-name edit without retaining why discovery rejected every candidate. That separate recovery question is reproduced and deferred in #543, as requested. This issue does not claim to solve it by merely leaving `CHANGE_ME`.

pengfei-threemoonslab:
## Summary

When every declaration of a name lives in tests or scaffolding templates, `init` currently selects it even though discovery already identifies those sites as non-product. Reject that candidate for selection while retaining its evidence and rationale. A matching declaration in product code stays eligible regardless of file order; recognized tool surfaces remain adoptable.

The installed and standalone rankers apply the same quality floor. `init` writes an unresolved `agent.name` placeholder instead of asserting a fixture's identity. The later identity-specific recovery question is reproduced separately in #543 and deferred: the existing human route for purpose/permissions does not prove that the name itself has been confirmed.

Closes #533.

## Type

## Verification

- Eleven new origin, order, and installed/standalone parity cases pass; three intended regressions failed before the fix.
- Full non-performance pytest suite passed with eight workers; focused discovery/init/standalone suites passed.
- Ruff and generated-schema consistency check passed.
- Committed base/head verification reports `complete` / `passed`, with no new authority declarations.

## Release-readiness notes

- [x] No check IDs or persisted schema grammars changed

This intentionally changes a previously selectable test/template-only name into a visible rejected candidate; it does not discard discovery or fabricate another name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+- **Test- and template-only agent names require review.** (#533) Discovery
+  keeps these names visible with their source and rationale, but `init` no
+  longer asserts them as the product's reviewed identity. A name also declared
+  in product code remains eligible under the existing quality and scope rules.
+  Installed and standalone detectors apply the same declaration-site floor;
+  tool discovery remains available and unresolved names use the existing
+  `CHANGE_ME` placeholder. The later identity-specific recovery question is
+  tracked separately in #543; existing purpose/permission review remains.
+
 - **The FastMCP signature projection now resolves the injected `Context`, and
   says so when it cannot read a type.** (#539) #535 made Python MCP servers
   discoverable and read their signatures. Two of the facts it published about
--- a/src/agents_shipgate/cli/discovery/signals.py
+++ b/src/agents_shipgate/cli/discovery/signals.py
@@ -2181,7 +2181,8 @@ def _rank_agent_name_candidates(
     - **Origin.** Product code outranks test code.
     - **Corroboration.** A name the project name independently agrees with
       is two sources, not one.
-    - **Quality floor.** A value too short or too generic to be an identity
+    - **Quality floor.** A value too short, too generic, or declared only in
+      non-product code cannot establish the reviewed product's identity. It
       is ranked last and made unselectable, so ``init`` writes CHANGE_ME and
       asks for review rather than asserting something unreliable.
 
@@ -2247,6 +2248,7 @@ def _rank_agent_name_candidates(
     # candidate points at.
     declared_in: dict[str, list[Path]] = {}
     best_project: dict[str, Path] = {}
+    product_declared: set[str] = set()
     order = 0
     for fact in facts:
         if not fact.agent_names:
@@ -2263,6 +2265,8 @@ def _rank_agent_name_candidates(
             if resolved is None:
                 continue
             value, provenance, detail = resolved
+            if _non_product_origin(evidence.rel_path) is None:
+                product_declared.add(value)
             ranked = _score_agent_name(
                 value=value,
                 role=evidence.role,
@@ -2283,6 +2287,14 @@ def _rank_agent_name_candidates(
                 best_project[value] = project
 
     for value, ranked in best.items():
+        # Selection is a claim about the product, not the highest score left
+        # in the list. Inspect every declaration site before applying this
+        # floor so a product/test shared name remains eligible (#533).
+        if value not in product_declared:
+            ranked.selectable = False
+            ranked.rationale.append(
+                "rejected: declared only in non-product code, not a reviewed product identity"
+            )
         # The candidate's own project first when it is blocked, so the
         # sentence names the project every other field already points at;
         # declaration order decides the rest.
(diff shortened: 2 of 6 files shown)


pengfei-threemoonslab (commented):
Review round 1 — inspected all six changed files at a211db89 against the issue's declaration-site boundary.

No in-scope correction requested. The floor is applied only after collecting every resolved declaration site, so a test encountered first cannot disqualify a later product declaration. Existing generic-name and unreadable-root rules still run; discovery recognition is not replaced by identity selection. Installed and standalone results use the same predicate and rationale, with full candidate-payload parity tests. The end-to-end tests check the written manifest and placeholder list, rather than treating a ranking result as successful setup.

The updated old expectations described precisely the now-rejected fixture behavior. Scope ambiguity, capped discovery and unreadable-root regressions remain covered by the passing broader/full suites. #543 correctly retains the separate unresolved-identity recovery question: the purpose review route is not evidence of identity-specific authorization.

This is a recorded coding-agent review on the author's connected account, not an independent human approval. Complete the address disposition and require CI plus fresh verifier authority before merge.

pengfei-threemoonslab:
Address disposition for review round 1: no code changes requested or required. Confirmed the six-file scope, negative controls and deferred #543 boundary against a211db89. Full local non-performance suite, focused origin/parity tests, Ruff, schema consistency and committed-diff verification pass. This completes one review/address round; merge remains conditional on the actual PR checks and a current verifier result.

Issue #561, opened 2026-09-08
Coverage recovery: distinguish reader repair from missing input without inferring ownership from prose


pengfei-threemoonslab:
## Problem

A user receiving `insufficient_evidence` needs to know whether to supply an existing artifact or wait for an Agents Shipgate reader fix. Today `EvidenceGap` names the subject/source and `EvidenceGapAction` names a repair, but neither records the evidence for that distinction. `authorable_by` answers who may draft a declaration; it does not identify the owner of a reader defect.

## Evidence

While implementing #520 against `0be83d2f87597c5b21989315fac0e1bdeaee96f4`, `src/agents_shipgate/schemas/report.py` exposes generic `source_warning`/`incomplete_surface` gaps and recovery kinds such as `review_warning` and `provide_source`. The same kind can arise from unavailable input or unsupported static syntax. Classifying the case from `why` text or from human-only declaration authorship would produce a confident but unsupported handoff. #328's generator-error slice and #357's effect-provenance presentation do not establish this distinction for coverage gaps.

## Scope

Start with one reproduced loader path that can distinguish unavailable source from unsupported syntax using typed parser evidence. Carry that reason through the existing gap and recovery projections; name the exact required input when it exists, or the bounded product repair when it does not. Preserve an explicit unresolved classification where the reader cannot decide. Extend the existing schema under its compatibility policy rather than adding another verdict or summary surface.

Do not infer deployed authority, suggest inventing a declaration, or weaken `insufficient_evidence` to make the workflow complete. Keep `authorable_by`'s existing meaning.

## Acceptance

- [x] A fixture with missing input and one with supported input containing unsupported syntax receive distinct, evidence-backed recovery routes.
- [x] The CLI/report/PR surfaces preserve the named source and exact remedy; an unresolved case stays visibly unresolved.
- [x] Qualification diagnostics can consume the same classification without parsing prose or assigning ownership from `authorable_by`.
- [x] Existing release decisions and declaration permissions remain unchanged.

## Sequencing

Child of #328; follows the diagnostic accounting slice of #520. Discovered during the implementation pass and deliberately deferred at the user's request. #520 records actual gaps and explicit unclassified cases in the meantime. Refs #357, #547, #543.

[...]

Issue #559, opened 2026-09-08
Qualification: hash and parse the same artifact bytes before scoring release evidence


pengfei-threemoonslab:
## Problem

Found while implementing #520's coverage diagnostics. `scripts/run_safety_qualification.py::_receipt_artifact` calls `sha256_file(artifact_path)` and then opens the same path again through `_read_json_object`. `_evaluate_receipt` does the same for the terminal receipt. A file replacement between those reads can make the parsed object differ from the bytes whose hash was accepted.

## Reproduction

A deterministic local probe uses a temporary `report.json` containing `{"probe":"hashed bytes"}` and a receipt reference to that hash. Wrap `sha256_file` so it computes the real digest, replaces that same file with `{"probe":"different parsed bytes"}`, then returns the original digest. `_receipt_artifact` returns the second object without a hash error. A final independent hash check returns false against the receipt's bound digest. The probe is synthetic; it demonstrates the read/parse gap, not an observed compromised release.

The path must be canonical in the macOS probe (`Path(...).resolve()`), because confinement resolves `/tmp` to `/private/tmp` before the hash call.

## Required implementation

Read bounded bytes once and use those exact bytes for both digest validation and parsing. Apply this to the receipt and every artifact whose validated data participates in qualification. Prefer the existing private receipt-closure snapshot machinery where its archived-receipt contract fits; do not silently import live-workspace assumptions into an offline release scorer.

Keep input containment, the full receipt/artifact identity joins, wheel/policy bindings and all approved qualification bars. Failure to obtain one coherent validated snapshot must produce a named input failure, never a score from unbound bytes.

## Acceptance

- [ ] An interleaving that changes an artifact after its first read cannot cause newly read, unbound content to be scored.
- [ ] The terminal receipt follows the same byte-consistency rule.
- [ ] Missing, replaced, malformed and escaping inputs remain explicit failures; no fallback scan or loosened identity assertion is introduced.
- [ ] Tests cover the actual scorer's successful path and deterministic hash/read interleavings, with real artifact bindings.
- [ ] #509 uses the corrected scorer before signing a candidate qualification result.

Deferred under the user's instruction to file newly discovered problems separately. #520 may ship diagnostic accounting without claiming this reader-hardening work is complete. Refs #509, #520, #552; the latter concerns compact current-control closure, not this offline scorer read.

PR #562, opened 2026-09-08
Qualification: report actual coverage misses without changing safety thresholds

pengfei-threemoonslab:
The approved three-label corpus has no expected-IE cases, but the legacy expected-IE metric printed `0/0` with `passed=true` without saying it was inapplicable. Actual coverage misses also lacked per-profile case evidence. This adds diagnostic accounting to the existing qualification result while preserving the six metrics, approved thresholds, labels and release decisions.

## What changes

- Qualification v6 records actual-IE count/rate over all corpus cases per profile, every missed case, its existing typed gap rows, and explicit unscored IDs. A rate with unscored cases is a lower bound; missing gap rows remain `unclassified`.
- Metrics gain explicit applicability. Expected-IE is not applicable only with zero denominator and zero policy floor. Actual IE continues to lose the applicable exact-outcome score, including the approved safe-pass tolerance.
- Diagnostics reuse the report object already validated for scoring. V5 remains readable without fabricated diagnostics; v1/v2/v4 keep their original beta/test restriction and normalization to v5. New diagnostic fields mislabeled as an older closed grammar are rejected. The stdlib gate recognizes v6 without changing its tier policy.

## Validation

21 new diagnostic/compatibility cases; 64 focused qualification/release tests. The first review found a malformed-old-metric container raising TypeError; the final head fixes it with three negative controls, and the second review found no remaining blockers. Full non-performance suite completed with only ten packaging setup errors from the restricted build dependency download; those ten passed unchanged when rerun with network access. Ruff and generated-schema checks pass. CI must also pass, including combined coverage, before merge.

## Surface discipline and limits

This makes the coverage component of noise/usable-outcome measurement interpretable; it does not claim a measured adoption gain. The existing qualification JSON carries the diagnostics. Its closed grammar requires v6; report, verifier, receipt and control grammars do not change. There is no new verdict, runtime execution or threshold.

The rest of #520 remains open: truthful reader-repair versus missing-input ownership is deferred to #561, and corpus/signing obligations remain #509/#456. The separately discovered artifact read race #559 is not repaired here; the final signed candidate must wait for that repair. Refs #520, #328, #357, #538, #519.

### DIFF
--- a/src/agents_shipgate/schemas/safety_qualification.py
+++ b/src/agents_shipgate/schemas/safety_qualification.py
@@ -763,11 +763,15 @@ class SafetyQualificationResultV1(BaseModel):
             declared in LEGACY_QUALIFICATION_ENVELOPES
             or declared == "shipgate.safety_qualification/v5"
         ):
+            intervals = data.get("intervals")
+            # Leave malformed containers to typed validation. A before-hook
+            # must not turn invalid input into an uncaught TypeError.
+            intervals = intervals if isinstance(intervals, (list, tuple)) else ()
             if "coverage_misses" in data or any(
                 "applicability" in item
                 if isinstance(item, dict)
                 else getattr(item, "applicability", None) is not None
-                for item in data.get("intervals", [])
+                for item in intervals
             ):
                 raise ValueError("coverage diagnostics and applicability require the v6 envelope")
         if declared not in LEGACY_QUALIFICATION_ENVELOPES:
--- a/src/agents_shipgate/schemas/safety_qualification.py
+++ b/src/agents_shipgate/schemas/safety_qualification.py
@@ -8,13 +8,13 @@ from pydantic import BaseModel, ConfigDict, Field, field_validator, model_valida
 
 from agents_shipgate.schemas.common import ReleaseDecisionStatus
 from agents_shipgate.schemas.disclaimers import STATIC_VERDICT_DISCLAIMER
+from agents_shipgate.schemas.report import EvidenceGap
 
 SAFETY_CORPUS_SCHEMA_VERSION = "shipgate.safety_corpus/v4"
 SAFETY_RECEIPT_INDEX_SCHEMA_VERSION = "shipgate.safety_receipt_index/v4"
-# v5 carries the ``pre_1_0`` tier and the production_qualified/tier invariant
-# added for issue #341. The corpus and receipt-index envelopes stay at v4: their
-# grammar did not move, and versions here track grammar, not release batches.
-SAFETY_QUALIFICATION_SCHEMA_VERSION = "shipgate.safety_qualification/v5"
+# v6 adds diagnostic coverage misses and explicit metric applicability. v5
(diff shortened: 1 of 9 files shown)


src/agents_shipgate/schemas/safety_qualification.py
```diff
@@ -720,6 +758,18 @@ def _upgrade_prior_envelopes(cls, data: Any) -> Any:
...
+        # An older closed grammar must not be widened by changing its label.
+        if (
+            declared in LEGACY_QUALIFICATION_ENVELOPES
+            or declared == "shipgate.safety_qualification/v5"
+        ):
+            if "coverage_misses" in data or any(
+                "applicability" in item
+                if isinstance(item, dict)
+                else getattr(item, "applicability", None) is not None
+                for item in data.get("intervals", [])
```
pengfei-threemoonslab: [P2] Inspect diagnostic keys only after checking the intervals container shape. `intervals: null` on an old envelope currently raises TypeError here, before normal schema validation can report a malformed qualification artifact.

pengfei-threemoonslab (commented):
Review round 1 — coding-agent review, not independent human approval.

Reviewed all nine published files, the metric-preserving path and the old-envelope round trip. One P2 compatibility defect needs an address: the new before-validator iterates `data.get('intervals', [])` before Pydantic can reject its shape. A v5 payload with `intervals: null` raises raw TypeError rather than ValidationError, so the release validator's documented malformed-artifact handling is bypassed. Reproduced directly against this head. Guard the diagnostic-key inspection to actual list/tuple inputs and let normal typed validation reject other shapes; add malformed old-artifact controls.

The counting/denominator/tolerance behavior otherwise matches the issue. Keep #559 and #561 deferred and all release thresholds unchanged.

src/agents_shipgate/schemas/safety_qualification.py
pengfei-threemoonslab: Addressed in 60d44cae17d1ce75b47cff9f9b0a8dfe579a027f. The before-validator inspects diagnostic keys only in list/tuple containers and leaves null, scalar and object inputs to normal typed validation. Three new regression controls cover those malformed shapes; all 64 focused qualification/release tests pass. No policy or metric behavior changed.

pengfei-threemoonslab (commented):
Review round 2 — coding-agent review, not independent human approval.

Re-read the published address diff and the regression controls. The malformed-container path now yields ValidationError, and the old-version diagnostic-field refusal remains intact. No remaining blocking findings in this bounded slice. Counts, zero denominators, missing receipts, positive legacy floors, original safe-pass tolerance and old-envelope round trips are covered; #559/#561 remain explicit limitations. Final CI (including combined coverage) and fresh current-control authority are still required before merge.

pengfei-threemoonslab:
Address round 2: no further blocking findings. The final head is 60d44cae17d1ce75b47cff9f9b0a8dfe579a027f; the inline round-1 finding is resolved. Awaiting the final-head CI and coverage result; no additional feature work is folded into this review.

Issue #577, opened 2026-09-08
Evidence readers can block on a FIFO before rejecting non-regular inputs


pengfei-threemoonslab:
## Problem

Found during the independent design review for #559. `core/verification_identity.py::read_regular_file_beneath` opens the leaf with `O_RDONLY | O_NOFOLLOW`, then checks `S_ISREG` with `fstat`. On POSIX, opening a FIFO for reading can block before that type check. The bounded byte limit therefore does not bound time to reject a non-regular input.

## Reproduction

Create a temporary directory and `os.mkfifo(root / 'input.json')`, with no writer. In a subprocess call `read_regular_file_beneath(root, 'input.json', max_size=1024)`. A local macOS probe timed out after three seconds and killed the child; the reader never reached its regular-file refusal. No release evidence or repository files were changed by the probe.

## Required implementation

Make opening the leaf non-blocking where supported, reject non-regular descriptors before reading, and retain descriptor-relative containment, no-follow handling, size limits and stability checks. Check other evidence readers for the same open-before-type-check pattern rather than fixing only a caller.

## Acceptance

- [ ] A FIFO with no writer is rejected promptly with the existing named read failure.
- [ ] A FIFO with a writer, directories and supported non-regular inputs cannot become evidence.
- [ ] Regular files keep the same digest, size and confinement behavior on supported platforms.
- [ ] A bounded subprocess regression cannot hang the test suite.

Refs #559, #552. Deferred separately under the user's instruction to file new findings for later; this does not change qualification thresholds or add a release claim.

pengfei-threemoonslab:
## Triage — 2026-09-14: still reproduces on `main` `45d6ca7a` (post-1.0 reliability reserve, #778)

`os.mkfifo(<tmp>/input.json)` with no writer, then in a child process `read_regular_file_beneath(Path(<tmp>), "input.json", max_size=1024)` with a 3-second timeout: **the child timed out; the reader never reached its regular-file refusal.** The current code still opens the leaf (`os.open(parts[-1], file_flags, dir_fd=current)`, `src/agents_shipgate/core/verification_identity.py:808`) before the `S_ISREG` check (`:811`), and `file_flags` carries no `O_NONBLOCK`. macOS, Python 3.12.

The issue's scope and acceptance stand as written (non-blocking leaf open, reject non-regular descriptors before reading, sweep other evidence readers for the same open-before-type-check pattern, bounded subprocess regression).

[...]

PR #578, opened 2026-09-08
fix(qualification): score the bytes bound by input digests

### LINKED ISSUE #559: Qualification: hash and parse the same artifact bytes before scoring release evidence

pengfei-threemoonslab:
The qualification scorer could accept a file's digest and then score different bytes reopened from the same path. Capture bounded bytes once for the wheel, corpus, index, policy inputs, terminal receipt and consumed artifacts, and use that capture for both validation and parsing.

Preserve detached receipt files and independent artifact roots beneath the index directory, all identity joins and approved qualification thresholds. Enforce artifact byte lengths and a separate decompressed wheel METADATA limit; unsafe or oversized evidence keeps the existing named failure routes. Document the input bounds.

Validation:
- Six deterministic hash/reopen regressions failed on the original implementation and pass with this repair.
- 151 qualification/release/verification-identity tests passed, including successful detached archives, replacement of every consumed input, corrupt/missing evidence, symlinks and size limits.
- Whole-repository Ruff passed; working-tree and committed-ref verification passed with fresh complete control.
- Independent coding-agent review/address round 1 is recorded at `73e47dd9`; no actionable findings and no follow-up code changes.
- All nine applicable GitHub checks passed, including the three full-suite shards and combined coverage; the release-only tag check is intentionally skipped.

Fixes #559. The shared reader's separately reproduced FIFO open blocking is deferred in #577. This repairs scoring integrity; it does not supply human labels, qualify a release or grant signing authority.

### DIFF
--- a/benchmark/safety-qualification/README.md
+++ b/benchmark/safety-qualification/README.md
@@ -40,6 +40,19 @@ contain binding and semantic coverage, and agree with the verifier receipt. Miss
 failed, unknown, hash-mismatched, or fallback receipts fail closed; the runner
 never substitutes a cold-start scan result.
 
+The scorer hashes and parses the same captured bytes for every consumed
+input. It does not reopen a path after accepting its digest. Detached receipt
+files and separate `artifact_root` directories remain supported beneath the
+receipt index directory; archive paths cannot traverse symlinks or escape
+that directory. Artifact byte lengths must also match the receipt.
+
+Reads are bounded: 4 MiB per receipt, 64 MiB per report, verifier, verify-run,
+corpus, index or policy file, and 256 MiB per wheel. Wheel `METADATA` has a
+separate 4 MiB decompressed limit. Unreadable or oversized top-level inputs
+are configuration errors (exit 2); invalid receipt evidence produces a named
+`invalid_verifier_receipt` failure and cannot be scored. These bounds do not
+change either qualification policy.
+
 ## What to build first
 
 [`strata-inventory.csv`](strata-inventory.csv) maps the known candidate pool onto
--- a/scripts/run_safety_qualification.py
+++ b/scripts/run_safety_qualification.py
@@ -25,6 +25,7 @@ from collections.abc import Iterable
 from email.parser import BytesParser
 from email.policy import default as email_policy
 from fractions import Fraction
+from io import BytesIO
 from pathlib import Path
 from typing import Any, Literal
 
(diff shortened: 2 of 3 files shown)


pengfei-threemoonslab (commented):
Independent coding-agent review — round 1 (not human approval).

Reviewed published head 73e47dd9fb4f07924355404156d1788e85cbf730 against base 8e8d5530dc314836d8a34f4b6261868a64b57b79. All three GitHub file blobs match the independently reviewed local diff. No actionable findings.

The repair hashes and parses the same captured bytes across all consumed qualification inputs, preserves detached archives and existing identity/policy joins, and checks artifact lengths plus decompressed wheel metadata bounds. Tests cover successful scoring and deterministic replacement across every input, with explicit invalid-evidence routes. The separately reproduced shared-reader FIFO limitation is deferred in #577.

Read-only review; the reviewer did not rerun tests or supply signing, release or human merge authority. Full CI and current repository control remain separate merge requirements.

pengfei-threemoonslab:
Address round 1: the independent review reports no actionable findings at 73e47dd9. No follow-up code change is needed. The 151 targeted tests, whole-repository Ruff and committed-ref verifier pass are recorded in the PR description; final full CI is still required before merge. #577 remains a separately scoped deferred issue.

PR #599, opened 2026-09-09
Compare imported SDK guard predicates with bound input evidence

pengfei-threemoonslab:
A shared imported guard can be weakened while the SDK tool declaration and finding fingerprints remain unchanged. The existing finding buckets give the reviewer no dependency evidence for that change.

This adds reader-owned evidence for a bounded OpenAI Agents SDK Boolean guard profile. Each row preserves its canonical observation/capability, call and definition locations, source/dependency digests and Boolean input domain. Base/head comparisons distinguish unchanged, wider, narrower and otherwise changed predicates; unavailable, ambiguous, redacted and unsupported evidence stays unresolved. Markdown shows relevant comparisons and JSON retains the full typed records. The records use separate fields rather than generic policy drift, so a guard improvement is not projected as capability broadening.

The verifier binds actual dependency bytes and named present/absent import candidates in its input identity. Current-control reads recheck those inputs for both committed and worktree pointers, including ignored paths. Attempted reads that could not capture bytes remain explicitly unconfirmable and use the existing workspace-unverifiable refusal until repaired and reverified. Base cache epoch advances so earlier reports are regenerated.

Refs #557. This is its first bounded reader/evidence implementation: every row retains incomplete dependency coverage and `finding_exclusion_eligible: false`. Existing findings and the release decision are unchanged. Full capability dependency closure, #312's fixed-history safety measurement and #515's default scope change remain open.

## Validation

- Actual paired repository scans and committed base/head verification prove that a guard-only change identifies its callers and carries evidence from each side. Old main's CLI preserves equal fingerprints and emits no guard evidence for the same pair.
- Shared callers, canonical observation membership, duplicate/invalid evidence, Boolean equivalence, dynamic/configuration cases, reader limits, late Python bindings and public redaction have targeted regressions.
- Real current-control tests cover changed helper bytes, appearing/disappearing ignored package candidates, missing-to-present helper/initializer recovery, and refused symlink/oversize/directory input followed by repair.
- Review-corrected frozen-source full suite: **9,193 passed, 5 skipped**. All **124 focused guard, input-identity and current-control tests** pass, including helper and package-initializer self-referential symlinks failing against the reviewed head and recovering after repair. Ruff, diff and generated-schema checks pass. Final-head independent review and CI remain required before merge.

[...]

Issue #572, opened 2026-09-08
Epic: qualify a stable v1.0 capability-review release from the exact shipped distribution


pengfei-threemoonslab:
## Product scope decision — 2026-09-11

Owner direction: **adoption > completeness; v1.0 contains only work this repository can independently implement, verify and deliver. Make the product ready first, then actively expand adoption.**

#643 now owns the selected advisory host-review acceptance; #644/#645/#646 define its sequence. Correct covered zero-change results count as first value. New host families, a manifest/lock-state rewrite, external participation and a completed authenticated human workflow are not its prerequisites.

This issue still owns the release decision and the historical qualified-gate audit below. None of its missing evidence is marked satisfied by narrowing product scope. #648 must resolve the stable advisory version, claims and actual publisher before choosing a 1.0 tag: the current `v*` path still requires qualified evidence. #569/#570 remain selected compatibility and exact-distribution obligations. #659/#660 supply engineering measurements, not independent human qualification.

#512/#509/#510 keep their qualified-corpus, independent signing and real rehearsal requirements open. #573/#494 retain publication-control and operational responsibilities; the selected publisher must preserve applicable integrity controls and describe support truthfully, while independent people's acceptance cannot be fabricated. #563/#557/#515/#580 remain deeper tool-source work; #548 remains an unresolved qualified-rehearsal diagnosis. Removing a whole issue from the selected milestone is not resolving its defect or changing runtime behavior.

No new go decision is recorded here. The earlier no-go remains historical evidence; readiness and publication require the new selected acceptance and a final candidate-specific decision. Active outreach (#653/#571) begins afterward.

---

## Release decision — 2026-09-08

**No-go for v1.0 at main `452bdeb80`.** This issue owns the release decision and sequencing, not a second scanner verdict or an instruction to close every open epic. The audit covers all 45 open issues (the three open Dependabot PRs are separate), recent implementation, stability/distribution contracts, release workflows and the recorded user/corpus evidence.

## What v1.0 promises

A developer or platform reviewer can review a supported repository-declared agent capability change, see exact base/head evidence and coverage limits, and follow a truthful next action. Local/static operation and advisory-first CI remain the defaults. The stable CLI, reports and control contracts must mean the same thing in the downloaded build and its generated CI.

Do not claim runtime-effective permission, universal MCP/tool coverage, proven organizational adoption, or authenticated GitHub continuation without the separate evidence those claims require. Existing human-owned stops remain human-owned.

[...]

Issue #563, opened 2026-09-08
Fixed-history release regression: establish reviewed scope for refused monorepo cases


pengfei-threemoonslab:
## First advisory release scope — 2026-09-11

Owner decision: **adoption > completeness; make v1.0 ready with repository-independent work, then expand adoption.** Reviewed historical monorepo tool-source scope is deferred from the selected host workflow. Missing scope remains an explicit limitation, not a benign result.

Removed from the selected v1.0 milestone; kept open and deferred, not resolved. #643/#644/#645/#646 define current product acceptance; #572 preserves the historical qualified no-go and #648 owns the actual publication contract. The prior evidence and acceptance below remain a record of the deferred work, not newly satisfied conditions.

---

## Problem

The unchanged W27 cold-start workflow no longer reaches a verifier on the fixed 19-case history. This is a release-evidence failure, not evidence of unsafe auto-passing. Setup can correctly refuse ambiguous scope while the end-to-end adopter workflow still fails to deliver an answer.

## Reproduction and evidence

The #312 rerun on candidate `60d44cae17d1ce75b47cff9f9b0a8dfe579a027f` retains all original labels and base/head SHAs. `benchmark/miner/results/2026-W37-reeval.{csv,jsonl,comparison.json}` records:
- all 19 root init attempts report `refused_unresolved_scope`;
- 15 cases finish without a manifest; one fallback writes a placeholder manifest and scan exits 2; three fallback scans produce answers but none reaches verify;
- both real `must_block` cases, crewAI-examples#169 and stripe/ai#232, are unscored;
- legacy `must_block_caught = 0/2` and `needs_human_caught = 1/3`, against unchanged required 1.0. The sole caught review case uses the historical scan fallback, not a PR verifier result.

Reproduce with the committed #312 driver and already-fetched exact refs. No scope override, semantic declaration, label change or threshold relaxation was used. The fixed population is five repositories, not the eight in the wider mining history.

## Implementation direction

First determine, from each original PR and the existing candidate routes, what reviewed project scope the case needs. Preserve a separate unassisted cold-start measurement; do not erase the setup refusal by silently injecting a scope. If a reviewed scoped corpus is needed, record its scope provenance before reading candidate outcomes and publish it as a distinct view against the same frozen PR labels/SHAs. Provide a bounded route for multi-project changes instead of selecting the first parsed agent or adding `--allow-unresolved-scope` indiscriminately.

[...]

Issue #557, opened 2026-09-08
Diff attribution: establish dependency coverage before excluding a standing finding


pengfei-threemoonslab:
## First advisory release scope — 2026-09-11

Owner decision: **adoption > completeness; make v1.0 ready with repository-independent work, then expand adoption.** Deeper tool-source dependency attribution is outside the selected host first-value path. Existing evidence gaps and exclusion restrictions remain unchanged.

Removed from the selected v1.0 milestone; kept open and deferred, not resolved. #643/#644/#645/#646 define current product acceptance; #572 preserves the historical qualified no-go and #648 owns the actual publication contract. The prior evidence and acceptance below remain a record of the deferred work, not newly satisfied conditions.

---

## Whole-function source comparison — 2026-09-09

Delivered in #606, merged `7a5abcdd897b71205d6d53e111b98b6b3e4e96ac`. The additional `sdk_boolean_function/v1` profile inspects the entire finite Boolean function, the selected pure imported predicate and one literal source Agent binding. It distinguishes return values, true-return input sets, source membership and the bound true-return relation. A narrowed first guard can coexist with a wider whole-function true domain; unchanged returns can coexist with added membership. Unknown operations/configuration, unsupported or unreachable statements, ambiguous identities and redaction retain unresolved evidence.

The model's complete Boolean-domain table is a **source comparison**, not action effect, human approval or deployed authority. The outer capability dependency coverage remains `incomplete`, every finding-exclusion flag remains false, and the acceptance checklist below is deliberately unchanged. Newly isolated sub-issue #607 owns the relationship between a modeled operation and an existing finding's policy predicates; it is deferred from this implementation pass. The supported operation/effect/dependency boundary, #515's TypeScript `cal-1` case and #563/#312's reviewed historical evidence remain outstanding.

[...]

PR #606, opened 2026-09-09
Compare complete Boolean source functions and literal Agent membership

pengfei-threemoonslab:
An unchanged first guard can hide a changed continuation or different Agent tool membership. This extends #599's guard evidence with a bounded `sdk_boolean_function/v1` model that inspects the entire selected function, a pure imported predicate and one literal source Agent binding. Reports compare return values, true-return input sets, tool membership and the bound true-return relation separately.

Every statement and branch is inspected before evaluating the finite Boolean model. Unknown calls, configuration/free variables, mutation, alternate bindings, unsupported signatures and missing or redacted evidence remain unresolved. The helper's entire AST is validated before its finite truth table is built; each possible helper input is evaluated once, independently of repeated calls. Input bytes use the existing verifier dependency identity; real committed base/head and stale-control tests exercise the path. Markdown and the generated report schema document the distinction.

A true return is not approval, effect or authority. All finding exclusions remain false and capability dependency coverage remains incomplete. This is the next bounded source-proof slice of #557, not its closure: operation-to-finding predicate attribution is now isolated in deferred #607; modeled external operations, the TypeScript `cal-1` case, #563's reviewed historical scope and #515's scope/default migration remain open. No gate, declaration, policy, receipt authority or release qualification is changed.

Validation: **9,347 passed, 5 skipped** in the final full local suite; 113 focused reader/verification checks passed. Regressions cover narrowed-guard/widened-function, unchanged-guard/changed-membership, unreachable unsupported code, ambiguous identities, redaction and ignored-source currency. The full-size bounded helper fixture first reproduced 202,496 helper invocations, then required exactly two after the fix and compared the actual unstubbed result. Ruff, generated schemas, sample-golden checks and current-base verification passed.

Independent GitHub review [5149940982](https://github.com/ThreeMoonsLab/agents-shipgate/pull/606#pullrequestreview-5149940982) found the helper-cost multiplication; the fix has been independently rechecked. This PR has one review/address loop, bringing #557 to three across #599/#606 in this implementation pass. Final published-head confirmation and author address are recorded in the conversation.

Refs #557, #515, #563, #607, #572.

### DIFF
--- a/src/agents_shipgate/inputs/sdk_boolean_source.py
+++ b/src/agents_shipgate/inputs/sdk_boolean_source.py
@@ -1,0 +1,262 @@
+"""Finite Boolean source relations over fully inspected function and binding ASTs.
+
+This module interprets a private Boolean model, not Python/application code.
+The caller has already captured the imported predicate and package bytes. No
+opaque call, free/configuration variable or alternate binding is a model leaf.
+A true return is a value, never evidence of approval, authority or an effect.
+"""
+
+from __future__ import annotations
+
+import ast
+from collections.abc import Callable
+
+from agents_shipgate.schemas.guard_dependencies import BooleanSourceBehavior, BooleanSourceBinding
+
+Value = Callable[[dict[str, bool]], bool]
+Step = Callable[[dict[str, bool]], tuple[bool, bool | None]]
+
+
+class _Unsupported(ValueError):
+    pass
+
+
+def _body(node):
+    body = node.body
+    if (
+        body
+        and isinstance(body[0], ast.Expr)
+        and isinstance(body[0].value, ast.Constant)
+        and isinstance(body[0].value.value, str)
+    ):
+        return body[1:]
+    return body
+
+
+def _parameters(node: ast.FunctionDef) -> list[str]:
+    args = node.args
+    if (
+        args.posonlyargs
+        or args.kwonlyargs
+        or args.vararg
+        or args.kwarg
+        or args.defaults
+        or args.kw_defaults
+        or getattr(node, "type_params", ())
+    ):
+        raise _Unsupported("function_signature_not_closed")
+    if not (
+        node.returns is None or isinstance(node.returns, ast.Name) and node.returns.id == "bool"
+    ):
+        raise _Unsupported("function_annotation_not_boolean")
+    if any(
+        not isinstance(arg.annotation, ast.Name) or arg.annotation.id != "bool" for arg in args.args
+    ):
+        raise _Unsupported("function_parameters_not_boolean")
+    names = [arg.arg for arg in args.args]
+    if len(names) > 8 or len(names) != len(set(names)) or "bool" in names:
+        raise _Unsupported("function_parameter_domain_not_supported")
+    return names
+
+
+def _expression(
+    node: ast.expr,
+    names: set[str],
+    guard_name: str | None,
+    guard: tuple[list[str], Value] | None,
+    depth: int = 0,
+) -> Value:
+    if depth > 24:
+        raise _Unsupported("source_model_depth_limit")
+    if isinstance(node, ast.Constant) and type(node.value) is bool:
+        return lambda values: node.value
+    if isinstance(node, ast.Name) and node.id in names:
+        return lambda values: values[node.id]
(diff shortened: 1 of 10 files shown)


src/agents_shipgate/inputs/sdk_boolean_source.py
```diff
@@ -0,0 +1,262 @@
...
+        and isinstance(node, ast.Call)
+        and isinstance(node.func, ast.Name)
+        and node.func.id == guard_name
+        and not node.keywords
+        and len(node.args) == len(guard[0])
+    ):
+        arguments = [_expression(arg, names, None, None, depth + 1) for arg in node.args]
+        return lambda values: guard[1](
+            {name: arg(values) for name, arg in zip(guard[0], arguments, strict=True)}
+        )
```
pengfei-threemoonslab: [P2] Bound helper evaluation independently of repeated call sites

Each supported call re-evaluates the entire helper expression for every tool input. The separate 4,096-node limits therefore multiply: a helper returning an `and` of 4,000 True literals is 36,030 bytes / 4,009 AST nodes; a tool with eight Boolean parameters, its required early guard, and an `and` of 790 `allow(p0)` calls is 11,327 bytes / 4,010 nodes. Both are admissible, but this path performs 256 × 791 × 4,000 = 809,984,000 helper leaf evaluations during one source model. A small actual-reader counting probe confirmed the same multiplication (53,760 evaluations for 10 literals and 20 repeated calls).

Compile and structurally validate the helper once, then bind its finite truth table (at most 256 parameter tuples) or memoize by the complete ordered Boolean argument tuple, so call sites perform bounded lookups. Preserve validation of every operand before evaluating any input, including unsupported expressions behind short circuits. Add a count/budget regression for a repeated-call fixture rather than a timing-sensitive test.

pengfei-threemoonslab (commented):
Independent coding-agent review, round 1, at 2c539f0fdf92ff981d97cc905cbb28cdd9366436 (base 826557678271b5fdabb0727f2103b5ed1a00ce0c). This is a coding-agent COMMENT review, not human approval, release qualification, or merge authority.

I verified GitHub's current head and all 10 changed-file blob SHAs against the reviewed snapshot. One P2 finding remains: repeated helper calls multiply evaluation cost beyond the intended bounded source model.

Validation: a counting probe against the actual static reader returned observed with 256 return rows and exactly 53,760 helper leaf evaluations for eight Boolean tool parameters, a ten-literal helper, and twenty repeated helper calls, although the helper has only two possible inputs. A separate AST-only construction fits the existing byte and AST limits but would require 809,984,000 helper leaf evaluations; I did not run that expensive case. Fixture application/SDK code was never imported or executed, and no worktree files or control artifacts were changed.

The remaining reviewed source semantics, separate return/binding comparisons, redaction, and unknown-case handling are coherent within this deliberately narrow profile. Complete capability dependencies, deployed wiring, finding-predicate attribution, historical qualification, and #557/#515 closure remain unproved, as the PR states. I did not rerun the full suite or verifier during this review.

pengfei-threemoonslab:
Addressed review [5149940982](https://github.com/ThreeMoonsLab/agents-shipgate/pull/606#pullrequestreview-5149940982) in `00fd44190dc40ce4fbd092f306a16743820025a7`.

The helper's entire expression is validated first, then each possible Boolean helper input is evaluated once and calls use positional table lookups. The full admissible 4,000-leaf / 790-call fixture reproduced **202,496 helper invocations** before the fix; it now requires **exactly two**, and its unstubbed static read produces the identical model. This eliminates multiplicative helper recomputation while preserving validation of unsupported/unreachable operands.

Final full local suite: **9,347 passed, 5 skipped**; 113 focused checks passed; Ruff, generated artifacts and committed current-base verification passed. The independent reviewer rechecked the fix and confirmed GitHub's final head and all 10 published blobs. Only the two reviewed fix files changed; no remaining actionable findings. The recheck belongs to this address loop, not a new formal review or human approval.

This completes one review/address loop in #606 and three total for #557 across #599/#606 in this implementation pass. Final-head CI, aggregate coverage and a fresh current-control check remain required before merge. #607 records the newly isolated operation/predicate relationship for later; #557/#515 remain open and no finding-exclusion or release qualification claim is made.

Issue #613, opened 2026-09-09
Host audit treats a recognized configuration path that is a directory as absent


pengfei-threemoonslab:
Found while implementing #568 on main `649992803952da63c087f3143403447bf8528a46`. Deferred from this implementation pass as requested.

## Problem

A repository containing a directory named `.mcp.json` produces a repository host inventory with `artifacts: []`, `issues: []`, and Claude Code `host_coverage.status: "complete"`. A recognized configuration location exists, but its invalid file type disappears rather than becoming an input failure. The same enumeration branch applies to directories named `.claude/settings.json`.

## Reproduction

In a disposable workspace, create the empty directory `.mcp.json`, then run `agents-shipgate audit --host --workspace <workspace> --json`. The current source API `host_audit_inventory(workspace)` returns the empty/complete result above. This was reproduced on the stated main revision with the source checkout explicitly on the import path; no external host state was read.

`core/host_grants.py::_repository_paths` descends directories before boundary candidate classification. Unlike a regular file, symlink or other nonregular entry, a directory matching an exact configuration path never reaches the parser's input refusal.

## Proposed direction

Retain recognized configuration locations regardless of file type and report the existing typed nonregular/input failure through the host snapshot and audit projections. Preserve bounded enumeration, containment, no-follow reads and actual directory traversal needed to discover other supported sources. Do not make a directory count as a successfully parsed configuration.

#568 will retain these filenames during applicability discovery and explicitly route the directory defect for inspection; it will not claim the existing audit has repaired or validated them.

## Acceptance

- [ ] A directory at each supported exact configuration location remains observed and reports a typed input failure.
- [ ] Inventory completeness and host coverage cannot treat that input as absent or successfully parsed.
- [ ] Audit, check and verifier projections retain the same input identity and failure reason.
- [ ] Regular files, symlink refusals, nested supported paths, entry caps and unrelated directories preserve their contracts.

Related: #568, #561, #572. This is an input-coverage repair, not a new host adapter.

## Selected-path review — 2026-09-12

The selected manifest-free host reader independently reproduces a directory at `.mcp.json` as no artifacts and no issues while reporting complete coverage. This can turn an unreadable recognized input into a quiet comparison, so it is a selected-path correctness defect, not merely additional host support.

[...]

PR #612, opened 2026-09-09
Compare instruction structure consistently across review and edit workflows

pengfei-threemoonslab:
A prose-only instruction edit currently reaches human review through the generic trust-root gate and surviving word-based boundary checks even after #549 retired the path-only weakening check. This change compares supported instruction structure across the final verifier, local boundary check, preflight and host drift. Complete unchanged structure clears only that instruction touch; settings, permissions, MCP declarations, executable hooks, CI and unresolved inputs keep their existing protection.

The shared bounded classifier handles plain guidance, supported skill/command frontmatter and preprocessing, and Cursor rule metadata. It requires complete captured sides, exact applicable hunks and unique file identity. Directory role and configured manifest/policy precedence prevent familiar filenames from exempting structured inputs. Historical word-based rule IDs remain registered and deprecated; malformed or incomplete comparisons produce explicit structural/input evidence instead of semantic claims.

Generated Claude Code hooks preview one exact contained Edit/Write through the actual CLI and reconfirm file identity. Conditional prose previews neither seed nor consume per-path approval memory, so accepting a prose edit cannot authorize later allowed-tools or hook expansion. Unknown payloads, old CLI output, timeouts and races retain the review prompt; final verification is still required.

Contract 32 publishes verifier 0.17, handoff v9, preflight 0.5 and host inventory/baseline/drift 0.3. Frozen predecessor schemas remain unchanged. Old verifier/handoff artifacts retain their deny-lists without synthesized conditional authority; old host baselines remain readable and explicitly incomparable. Commands retain the actual workspace/config, including spaces. The documented synthetic Route H dry run was rerun against this candidate; it still blocks all four known expansions and contributes no external adoption or release-qualification evidence.

Validation: the complete local run finished with 9,416 passed and 5 skipped; its one stale word-heuristic assertion was replaced with a real prose/preprocessing pair, and the entire 28-test host-boundary module then passed. The additional 62 review regressions and 41 contract/pilot checks passed; committed mixed permission/MCP/hook/CI controls and required-field legacy refusals are covered. Ruff, schema generation, all 24 unchanged sample goldens, and committed verification against current main passed. Final published-head CI and independent GitHub review follow on this PR.

Closes #545.
Closes #516.
Refs #572. The discovered frozen preflight URL mismatch and pre-existing completed-plan permission semantics are separately filed and deferred in #609 and #610. #611 separately records header-like hunk rows lost by the existing diff parser; those comparisons remain unresolved and review-required.

[...]

PR #614, opened 2026-09-09
Route supported host-only repositories into their first review

### LINKED ISSUE #568: The first discovery decision rejects supported host-only repositories before they reach Route H
## Problem

The lead adoption route supports repository-declared host permissions and MCP configuration without a manifest, but the first applicability decision tells that user to leave.

## Evidence — 2026-09-08

On main `452bdeb80`, supported Python 3.12, a repository containing only `.claude/settings.json` with `permissions.allow: ["Bash(npm test)"]` and `.mcp.json` with one stdio `mcpServers` entry produces:

```json
{"is_agent_project": false, "suggested_sources": [], "codex_plugin_candidates": [], "python_parse_truncated": false}
```

`docs/quickstart.md:89–95` interprets this complete negative tuple as “not the right tool for this repository.” The same page's Route H at lines 404–419 explicitly supports those inputs. `tools/shipgate-detect.py` also ends this case with a no-action recommendation. The detector is correct to exclude host `mcpServers` config from MCP *tool exports*; the unsupported conclusion is about applicability of the whole product.

## Implementation direction

Separate “no manifest/tool-source candidate” from “no supported host-boundary review.” Route supported host-only users to the existing `audit --host` / `check` workflow, without generating a placeholder manifest or inventing tool inventory. Repair the quickstart and machine/instruction consumers of the same stop rule together.

Use existing host recognition and the existing discovery result where possible. If applicability metadata is needed, give it a bounded meaning and maintain CLI/zero-install distribution parity. No second detector or new command.

## Acceptance

- [ ] Supported host-only fixtures reach Route H from the first README/quickstart/discovery decision on the declared supported runtime.
- [ ] The caller can distinguish host review, agent-builder setup, unsupported input and incomplete/truncated discovery without reading prose as authority.
- [ ] Truly unrelated repositories still receive an accurate negative result; parse caps and malformed host inputs cannot become a false negative.
- [ ] Host config is never emitted as an MCP tool export, and no purpose/authority/binding declaration is fabricated.
- [ ] A cold run of the installed candidate and zero-install path demonstrates the corrected route, with distribution parity tests updated.

## Sequencing

P1 v1.0 first-value blocker for the existing lead wedge. Reuse #485/#497's delivered detector/parity machinery and #498's quickstart; those issues stay closed. Coordinate #543/#547/#561 recovery and #550's canonical temporary-path recipe. This does not expand support to Python 3.9 or add a new framework.

pengfei-threemoonslab:
Closes #568. Refs #572, #613.

A repository containing only supported host configuration (for example `.claude/settings.json` and `.mcp.json`) currently reaches a complete negative discovery result and leaves the adoption workflow, although Route H already supports it. This change carries filename-backed host candidates through the existing discovery result and routes host-only `detect`, `init` and `bootstrap` to the existing host audit without writing a placeholder manifest or setup files.

The runtime and zero-install detector use the existing host path vocabulary, bounded metadata traversal, explicit incomplete-link evidence and no config execution or authority inference. Parse caps, scope ambiguity, existing manifests, builder inputs and plugin inputs retain their routes. A host-config directory receives a human inspection route; the separate host-audit directory omission is filed as #613 and deferred. The complete-negative stop rule requires both new fields to be explicitly empty, including in trigger consumers and bundled instructions.

Contract 33 adds `host_boundary_candidates` and `host_discovery_incomplete_paths`; the report, preflight and verifier versions are unchanged. README, quickstart, diagnostics, prompts, generated bundles, current contract and the synthetic Route H pilot ledger are synchronized. The new applicability fields improve activation and time to first useful review through the existing discovery surface; they grant no operational authority and introduce no adapter, command or second verdict.

Validation:
- Full local suite: 9,461 passed, 5 skipped, 2 stale expected-shape assertions failed, and 10 packaging fixture errors because the sandbox could not fetch the isolated build backend. Both assertions were updated for the new discovery fields. The complete 670-test affected regression set then passed; all 10 packaging tests passed with the unchanged isolated-build tests and permitted dependency access.
- Ruff, schema drift check, rendered-bundle parity and all 24 sample golden checks passed.
- Cold installed-wheel replay outside the checkout: detect/init/bootstrap/audit all exited 0, both detectors reported the same two host candidates, init wrote no manifest/workflow, and host audit returned two grants. Wheel SHA-256: `f50b436cff0c1cb26458143bc5eb5c792b0a04623f49582287ed50b25e087b9c`. This is an unqualified implementation build, not release qualification evidence.
- Committed verification is `review_publishable` / `review_required`: commit, push and PR updates are allowed; merge and completion remain denied because the change updates protected adoption instructions and the trigger catalog. No findings, baselines, waivers or control permissions were weakened.

Independent GitHub PR review/address loop 1 is requested on the published commit. CI on the PR will validate the full final commit.


docs/quickstart.md
```diff
@@ -87,14 +87,28 @@ Continue when `is_agent_project: true`, or when `suggested_sources` or
...
 Python parse stopped at its cap, so that negative describes the files that were
 read rather than the repository — re-run with
 `--max-python-files <workspace_signals.python_file_total>` before concluding
 anything. [`zero-install.md`](zero-install.md) covers the `uvx` and GitHub
 Action variants that also avoid a local install.
 
+If `host_boundary_candidates` is non-empty, the repository has recognized
+host configuration paths. Follow [Route H](#route-h--no-manifest)
+with `agents-shipgate audit --host --workspace . --json`; no manifest is
+needed. Detection reads filenames, not permissions. Invalid contents remain
```
pengfei-threemoonslab: [P2] Limit the no-manifest handoff to host-only repositories

This condition also matches a builder repository containing both exported tools and `.claude/settings.json`, or a repository that already has `shipgate.yaml`. In those cases `detect` intentionally keeps the init/doctor route (covered by `test_mixed_tool_source_preserves_the_builder_route` and `test_existing_manifest_keeps_doctor_route`), but this first-adoption paragraph tells the user to take Route H and that no manifest is needed. That can leave the actual tool surface outside the intended setup workflow. Qualify the handoff as host-only after framework/tool/plugin candidates, existing-manifest handling, and incomplete-discovery precedence; direct callers to follow the emitted route instead of deriving it from this one nonempty field. Host audit can remain an additional review for mixed repositories.

pengfei-threemoonslab (commented):
Independent coding-agent review, loop 1, at de921a2d11cbf0ac555b05647f1e041715291e81 (base 649992803952da63c087f3143403447bf8528a46). This is a coding-agent COMMENT review, not human approval or a verifier override; it grants no merge or completion authority.

One P2 remains in the quickstart's applicability routing: it sends mixed builder/host repositories to the host-only no-manifest route, despite the implementation correctly preserving their existing init/doctor route.

I verified GitHub's current head and all 45 changed-file blob SHAs against the reviewed source. The three pre-PR code findings are addressed: normalized eligibility exclusions match the stdlib detector, mixed human output retains tool/plugin candidates, and the stdlib census rejects directory reparse/junction facades before enumeration. The reviewed implementation keeps ignored host names visible, treats arbitrary links as incomplete evidence rather than host presence, retains config-directory inspection, and hands host-only init/bootstrap off before setup writes. Setup permissions remain false.

Independent validation: all 217 tests in tests/test_host_discovery.py, tests/test_setup_control.py and tests/test_bootstrap.py passed against this immutable source, using separate temporary test artifacts and no pytest cache. The worktree remains clean. Reparse/junction validation here is simulated, not a native Windows run; I did not repeat the author's full suite, installed-wheel replay, verifier, or release qualification. The existing audit config-directory omission remains explicitly deferred in #613.

docs/quickstart.md
pengfei-threemoonslab: Addressed in d6a206d6: the first-run guide now limits the no-manifest handoff to host-only repositories with complete discovery and no manifest/builder/plugin candidates. It preserves the mixed-builder setup, existing-manifest doctor and incomplete/scope-resolution precedence, and directs the caller to the emitted typed next action. The public-surface and distribution-parity suites passed on this correction. No runtime behavior or operational authority was changed.

pengfei-threemoonslab: Independent coding-agent recheck of the existing loop 1 finding: confirmed GitHub head d6a206d67ab13a25bdeb3a7ad28511c02d832b5b and the published comparison from de921a2d11cbf0ac555b05647f1e041715291e81. Only docs/quickstart.md changed (blob ca4ee088aea5dbec59bce756596e20651e7840a5), matching the reviewed correction. The paragraph now limits no-manifest Route H guidance to host-only input, preserves mixed-builder and existing-manifest routes, and directs callers to the emitted typed action with incomplete-discovery/scope handling. The P2 is addressed; resolving this thread. This is a recheck within review 5159765676, not another formal review, human approval, or merge/completion authority.

pengfei-threemoonslab:
Review loop 1 is addressed and independently confirmed in the resolved thread. Final commit d6a206d67ab13a25bdeb3a7ad28511c02d832b5b has completed CI successfully: run 34405136833 includes all three full-suite shards, aggregate coverage, Windows launcher, clean-checkout launcher and supporting checks; the inapplicable main-only release-tag check was skipped. Agents Shipgate run 34405136815 and Self Dogfood run 34405136880 also succeeded. The verifier remains review_publishable, so this does not claim human approval or merge/completion authority.

pengfei-threemoonslab:
## Independent review — loop 1

Reviewed `d6a206d` (contract 33, 45 files) against `main`, reproducing each finding on a live tree rather than reading the diff alone. CI was green and the dedicated `tests/test_host_discovery.py` corpus is genuinely good — every finding below is a shape the corpus does not contain.

Two of them are **P0-class**: both turn a settled, correct answer into a stop, and both are reachable on ordinary repositories that have nothing to do with host configuration. The rest are consistency and honesty gaps. All are fixed in the commit pushed to this branch.

---

### 1. A symlink to a **file** defeats the product-wide negative (High)

`discover_host_boundary` put *every* symlink into `host_discovery_incomplete_paths`, because `_symlink_may_hide_boundary_glob` returns `True` for any name once a `**/`-prefixed glob is in the registry. A link that resolves to a regular file has no descendants, so it can conceal nothing — but it withheld the complete negative anyway.

Reproduced on a plain Python library with one ordinary `docs/README.md -> ../README.md`:

```
before: diagnostics ['SHIP-DIAG-NON-AGENT-LIBRARY'] | decision setup_not_applicable
        Next: Stop: Python project with no agent framework, prompts, or tool surface.

after adding only the symlink:
        diagnostics []                              | decision setup_incomplete
        Next: Review: Unfollowed links may conceal host configuration: docs/README.md.
```

The blast radius is every consumer of the negative at once: `SHIP-DIAG-NO-AGENT-SURFACE` / `-NON-AGENT-LIBRARY` / `-PURE-PROMPT-EXPERIMENT` are all suppressed by the same guard in `diagnose_detect`, `bootstrap` returns `host_review_required`, the exclusion ledger gains a row per symlink whose `detail` is factually false for a file link, and the `trigger` stop conditions cannot fire. Both implementations had it.

**Fix:** a link is concealment only when it can actually hide a tree — `stat()` for the target *type* only (never enumerated, never followed into), with any failure (broken, cyclic, unreadable) staying concealment. A synthesized child under an ancestor link that resolves to a non-directory is dropped rather than published as `unresolved`, because it cannot exist. Mirrored in `tools/shipgate-detect.py`; verified byte-identical on the file-link, directory-link, broken-link and symlinked-`.claude`-directory shapes.

### 2. One unreadable directory made `detect` / `init` / `bootstrap` exit 4 — permanently (High)

The census raised `DiscoveryError` for any `HostInventoryReadError`, and it ran *first* in `detect_workspace`. So a single `chmod 000` directory anywhere aborted classification of a repository `main` classifies fine:

```
main:  {"is_agent_project": true, "frameworks": [{"type": "openai_agents_sdk", ...
this:  Workspace discovery could not establish bounded coverage: Host configuration
       discovery could not read a stable, bounded repository inventory.  exit=4
```

Three things make this worse than a conservative default:

- **`audit --host` — which owns this exact walk — does not do this.** `host_grants.py:1273` catches the same three exception classes, records an `inventory_failures` row, and carries on. I ran it on the same repository: exit 0, full `host_coverage` payload. An applicability *hint* was stricter than the audit it routes to.
- **The published recovery can never succeed.** `expects` says "rerun detect" (and reuses the git-failure wording, "inspect the Git failure"), but the directory is still unreadable on the next run. No flag raises the 100 000-entry bound either, so a large monorepo — a shape this codebase supports on purpose, with `--max-python-files` and #363 scope routing — has no way back.
- **Nothing about the failure is published.** The message says "Inspect unreadable paths" and names none.

**Fix:** the census now returns `([], [subject])` naming the path that stopped it, instead of raising. That is strictly *more* fail-closed on the thing that matters — no candidate is published from a walk that did not finish, and the incomplete row withholds the negative and the trigger stop — while the framework, source and scope answers stand. It is also what the audit already does with the same failure. Both implementations name the same subject (`secret`, or `.` when the failure carries no path); the script tracks it at the two granularities the canonical reader reports.

### 3. `detect`'s human output claimed the host route on an adopted workspace (Medium)

`needs_host_route` does not consider `has_manifest`, and the new human branch sat ahead of every other one, so a repo with `shipgate.yaml` + `.cursor/mcp.json` printed a "Host configuration discovery" heading over `Next: … doctor …`. The JSON path was right; only the text disagreed with its own next step. The branch also skipped `_echo_excluded_sources`, dropping the one sentence that explains why `.cursor/mcp.json` is not a tool source. Both fixed; and the adopted-workspace negative branch now prints its `Next:` line at all, which it never did.

### 4. `manifest_status: "not_applicable_host_review"` was missing from every published enum (Medium)

`llms-full.txt`, `docs/agent-recipes.md` and all four copies of `prompts/add-shipgate-to-repo.md` still listed only `written | skipped_existing | refused_unresolved_scope | not_attempted`. An agent following the prompt this PR edits meets a value the same prompt says cannot occur. Added, with the render SHA and `llms-full.txt` regenerated. (The intermediate `a25dc9c8…` was removed from `prior_render_sha256` — it never shipped, so `b600b97d…` stays the outgoing hash.)

### 5. Truncated lists published as if complete (Low)

`", ".join(paths[:5])` in `host_discovery_action` and `[:10]` in the human branch: five names and a full stop assert there are five. Now `(and N more)`, in both implementations — the rule from #397.

### 6. The parity harness's two new rows could not fail (Low)

No `PARITY_WORKSPACES` entry has host configuration — and none can, since a `.mcp.json` under `tests/` would be a candidate of *this* repository — so `detector_divergences`' new lines only ever compared `[] == []`, and `test_detector_comparison_reports_a_seeded_divergence` did not seed them. Added a tmp-tree parametrization over the config-file, config-directory and unreadable-directory shapes, plus the seeds. `docs/distribution-surfaces.md` records which test proves it.

### 7. Three spellings of one predicate (Low)

`bootstrap` restated `needs_host_route` inline over the JSON payload, and already differed from it (`not manifest_already`). Now one `needs_host_route_payload(payload)`, with `needs_host_route(result)` as the model-side wrapper; `bootstrap` keeps its own manifest check on top.

### 8. Smaller items

- `init`'s host payload dropped `tool_surface_origin` and `control_pack` — the latter contradicting the comment beside it ("emitted for every run, including a refused one"). Restored via a shared `_control_pack_payload` rather than a second dict literal.
- The ledger `reason` was `source_rejected`; nothing rejected a source. Now `census_not_seen_through`, alongside `walk_capped` / `scope_contested` / `scope_unknown`.
- Misindented dict entries in `test_schema_boundaries.py` and `test_trigger_command.py`.
- `tests/test_adapter_static_only.py`'s `cli/bootstrap.py` allowlist is line-number pinned and shifted by the new import.
- Docs that described the field as "unfollowed links" only now describe what it actually holds.

---

### Not changed — worth your call

- **`--local-review` is no longer honored on a host-only repo**, while `--minimal` is explicitly carved out. Both are explicit user requests; `STABILITY.md` singles out only `--minimal`. Deliberate and tested (`test_direct_init_hands_off_without_writing_any_setup` parametrizes it) — flagging in case the asymmetry was not intended.
- **`detect` now walks the repository twice.** Measured on this checkout: 9.4 s → 10.6 s (+12 %). The census duplicates the inventory `detect_workspace` already builds. Fine for now; an issue if large-repo adoption matters.
- **`accounting="not_claimed"`** on the incomplete rows, though the run *did* act on them when the host route is selected. `route_blocked` would need `core.surface_exclusions` to import the predicate from `cli.*`, so I left it.
- **`host_config_adapters_for_path` hard-codes `.claude/commands/` and `.cursor/rules/`** beside a registry lookup. The conformance test catches CLI-vs-script drift, not a new adapter that lands a JSON glob under an instruction directory.

### Validation

Full suite green (`-n 4`), `ruff check .` clean, `scripts/generate_schemas.py --check` clean, rendered-bundle parity and `llms-full.txt` regenerated. The Route H pilot fixture replays end to end on the fixed tree: two candidates, `audit --host` route, `init --write --ci` writes nothing and returns `not_applicable_host_review`, `bootstrap` returns `host_review_required`, `audit --host` returns grants. CLI/zero-install parity re-verified by hand on the file-link, directory-link, broken-link, symlinked-config-directory, unreadable-directory and many-links shapes.

Note the PR description's cold installed-wheel replay (`f50b436c…`) predates these fixes and no longer describes this branch.

pengfei-threemoonslab:
## Review loop 2 — merge conflict resolved, deferred items closed

`main` moved (#616), and the four items I left as "worth your call" are now three closed and one open.

### Merge conflict — `ROADMAP.md`

Resolved as a merge commit rather than a rebase, since another worktree holds this branch and a force-push would strand it.

- **The #612 sentence** and **the deferred-issue list**: took `main`'s. It is the later authority on its own prose, and its list already carries #613 alongside the new #615/#617.
- **The trailing deferral note**: kept `main`'s three-issue sentence and re-attached the one fact only this branch states — that #568's discovery routes the #613 defect to inspection and claims no audit repair.
- **One claim went stale in the merge and is corrected**: `main` says #614 has "its first independent PR review correction published". There are two now, so the paragraph says two and names what the second repaired.

`docs/distribution-surfaces.md` and `tests/test_distribution_surface_parity.py` auto-merged — #616's changes and this branch's do not overlap.

### `--local-review` vs `--minimal` — behavior kept, reason stated

I asked whether the asymmetry was intended. Looking again, it is principled; it just was not written down. The line is not *how explicit a flag is*, it is *whether the flag uses discovery at all*:

- `--minimal` short-circuits at `_register_init.py` before `detect_workspace` is ever called, so there is no classification for a host-only route to act on.
- `--ci`, `--claude-code`, `--agent-instructions` and `--local-review` all render their manifest **from** that classification. On a host-only repository each would write a manifest declaring nothing — the dead end #568 exists to remove.

So the carve-out is a rule, not a list. `STABILITY.md` and `docs/agent-recipes.md` now state it. A host-only workspace that genuinely wants a manifest still asks for the template by name.

### `accounting` — corrected to `route_blocked`

My own reason for leaving this (`core.surface_exclusions` would have to import a predicate from `cli.*`) does not survive contact with the precedent. `walk_capped` is `route_blocked` **while `detect` publishes a runnable higher-cap retry command** — so the token means *this stage withheld its verdict*, not *routing stopped*. That makes it unconditional and predicate-free.

It is also the accurate reading: `total - gated` is documented as what was "recorded and deliberately not acted on", and these rows are the opposite — withholding the product-wide negative **is** the act. An unreadable directory now reports `total 1 gated 1` with `('secret', 'census_not_seen_through', 'route_blocked')`.

### Instruction-prefix denylist — drift made loud

`host_config_adapters_for_path` still carries `.claude/commands/` and `.cursor/rules/` beside an otherwise registry-driven lookup, and it cannot be made registry-driven without adding a field to `BoundaryAdapterSpec` — new surface for a maintenance concern. So the drift is now a failing test instead of a silent widening: every one of the 38 registry entries is pinned as host-config or not, and a new entry fails by name.

Two details keep it from passing for the wrong reason: every expansion ends in `.json`, so a negative cannot come from the suffix gate rather than the prefix rule; and `.claude/commands/hook.json` / `.cursor/rules/rule.json` are asserted directly, together with an assertion that they are still boundary surfaces at all. Verified non-vacuous by adding a hypothetical `.cursor/hooks/*.json` adapter — the guard fails and names it.

### Still open: `detect` walks the repository twice

Not cheaply fixable and deliberately not attempted here. `_repository_paths` is the audit's identity-bound walk with TOCTOU re-validation; `detect`'s own inventory goes through `git ls-files`. Sharing them is a real refactor with different safety guarantees on each side, not a review fix. Measured at +12% (9.4s → 10.6s on this checkout). Happy to file a tracking issue if you want it on the ROADMAP's deferred list.

### Validation

Full suite green on the merged tree, `ruff check .` clean, `scripts/generate_schemas.py --check` clean, `llms-full.txt` regenerated.

**The cold installed-wheel replay was re-run**, because the ROADMAP row commits to it as evidence and the previous one predated the loop-2 fixes. Fresh wheel `c2988502c6413650ac231284e5f67ee1db64b6ab7066b2e7156d5527df3a7cb5`, installed into a clean venv, driven from outside the checkout: contract 33; Route H reports two candidates and routes to `audit --host`; `init --write --ci` writes nothing and returns `not_applicable_host_review`; `bootstrap` returns `host_review_required`; `audit --host` returns grants. Both loop-2 P0 fixes reproduce through that installed wheel — the file-symlink repository keeps `SHIP-DIAG-NON-AGENT-LIBRARY` / `setup_not_applicable`, and the unreadable-directory repository classifies as an agent project, records `secret`, and routes to `init` at exit 0.

This remains an unqualified implementation build, not release qualification evidence, and this comment grants no merge or completion authority.

Issue #512, opened 2026-09-03
Deliverable 2: the 80-case human-labeled beta corpus required for 1.0 (#456)


pengfei-threemoonslab:
## First advisory release scope — 2026-09-11

Owner decision: **adoption > completeness; make v1.0 ready with repository-independent work, then expand adoption.** The blind human-labeled corpus remains required for qualified claims. #659 is a distinct engineering corpus and cannot satisfy or impersonate this evidence.

Removed from the selected v1.0 milestone; kept open and deferred, not resolved. #643/#644/#645/#646 define current product acceptance; #572 preserves the historical qualified no-go and #648 owns the actual publication contract. The prior evidence and acceptance below remain a record of the deferred work, not newly satisfied conditions.

---

## Direct-source extension — 2026-09-10

[PR #626](https://github.com/ThreeMoonsLab/agents-shipgate/pull/626), merged at `61d21527124ed35d006c1a1b0157ba2beab820c1`, adds four pinned real-history sourcing hypotheses in the existing register: MCP Git argument guards, a Terraform current-token identity read, an ADK file-search restriction, and optional parent conversation text forwarded to a delegated child. Public full merge/first-parent comparisons match the existing sweep pins. Actual handler behavior and whole-change context, rather than a title, annotation or new verifier result, support the provisional targets.

The current register has **46 pinned candidates, 38 gaps and six reserves** across the unchanged 84 slots. All four additions use the existing `diff_substance` convention and conservatively retain `maintainer_walk`/`tuning_only`; they were source inspections, not engine development or beta labeling. There are now 14 tuning-only candidates; **32 potential holdout candidates and three missing per-cell holdout slots remain unchanged**. The two sequential ADK changes and the later unplaced declaration change retain joint admissibility and one-split obligations. Distinct related PR numbers do not establish independent diversity.

Beta guards now enforce the direct-inspection exposure floor and require evidence in the actual beta register, with immutable source/comparison links. Existing miner-label consistency, source-pin contradiction checks, historical exposure floors and scoring policy remain unchanged.

[...]

PR #58, opened 2026-05-09
Add explain-finding CLI + companion prompt for per-finding prose

pengfei-threemoonslab:
## Summary

- Ships W4-3 / W4-4 of the agent-adoption strategy. Closes the gap between `explain <check-id>` (static catalog metadata) and what coding agents actually need when summarizing one finding for a human reviewer: catalog metadata + the specific finding's evidence + a deterministic 3–5 sentence prose explanation.
- New CLI: `agents-shipgate explain-finding <fingerprint> --from <report.json> [--json]`. Companion prompt: `prompts/explain-finding-to-user.md` (+ skill mirror).

## Type

- [ ] Check or risk-model change
- [ ] Input adapter change
- [x] CLI or GitHub Action behavior _(new `explain-finding` command + new `unknown_fingerprint` agent-mode error kind)_
- [ ] Report, schema, or SARIF output
- [ ] Documentation only

## Verification

CI is authoritative for `python -m ruff check .`, `python -m compileall -q src tests`, and `python -m pytest`.

Additional local checks run:

- `python -m pytest` → 903 passed (was 891 on main; +10 new + parametrization expansion), 3 skipped
- `python -m ruff check .` → clean
- Manual end-to-end: `agents-shipgate explain-finding fp_<...> --from agents-shipgate-reports/report.json --json` against the support_refund sample returns the canonical 21-key payload and a tool-named, severity-named, action-aware prose explanation.

## Release-readiness notes

- [x] No user-code import added to default scan paths _(read-only command — loads `report.json` and the catalog only)_
- [x] No network access added to default scan paths
- [x] New or changed check IDs are documented in `docs/checks.md` _(no new check IDs)_
- [x] Report/schema changes are additive or documented in `STABILITY.md` _(no schema change; new stable command added to STABILITY.md)_

## Notable design points

- **Determinism**: `explain_finding_payload` is a pure function over `(fingerprint, report_path, plugins_enabled)`. Same inputs always produce byte-identical output (pinned by `test_payload_is_deterministic`). Callers can cache without surprises.
- **Action-aware prose**: every `agent_action` enum value (`auto_apply`, `propose_patch_for_review`, `escalate_to_human`, `suppress_with_reason`, `informational`) maps to a tailored closing sentence. The prompt's "what NOT to do" section forbids the agent from saying "I'll apply this automatically" when `agent_action` isn't `auto_apply`.
[...]

PR #642, opened 2026-09-10
Freeze the report contract at 1.0 and publish what it promises (#569)

### LINKED ISSUE #569: Freeze the report 1.0 contract and prove migration before calling the product v1.0
## Problem

`STABILITY.md:7–12` promises that a 1.0 line will not begin until the report schema reaches 1.0 and holds without a breaking change. Main `452bdeb80` still emits report `0.43` and runtime contract `31`; the qualification policy deliberately requires report `0.43`. No open delivery issue owns satisfying this published release condition.

Bumping the package version alone would break the promise. Renumbering every schema would create work without establishing compatibility.

## Scope and implementation direction

After the selected v1.0 behavior and defaults converge, inventory the actual stable CLI/exit-code/Action/control/report surfaces and name any provisional integrations explicitly. Freeze report 1.0 using the existing schema generator and compatibility rules; preserve the current operational control grammar where it suffices. Update the qualification schema requirement deliberately in the same reviewed compatibility change, preserving all scoring floors.

Document migration from the actual shipped v0.15 contract and the supported preview/RC path, including baseline/receipt re-verification, changed defaults, parser direction and refusal behavior. An old signed receipt must not gain current authority through conversion or a restamped digest.

## Acceptance

- [ ] The published report 1.0 freeze condition is satisfied, with a stable/provisional inventory and explicit 1.x compatibility/deprecation rules.
- [ ] Compatibility fixtures cover shipped inputs and consumer paths: installed CLI, generated Action workflow and machine control/report readers.
- [ ] Unsupported historical artifacts fail with a usable migration/reverification route; no silent schema reinterpretation or authority upgrade.
- [ ] Published schema URLs, stable aliases and check IDs keep their promised compatibility. Time on unreleased main does not count as a shipped deprecation cycle.
- [ ] The schema holds through a recorded RC exercise (an untagged, unpublished candidate is sufficient) and migration replay; every later breaking change restarts the freeze and invalidates affected qualification evidence.
- [ ] The final report schema, reviewed policy requirement and exact candidate wheel agree before production receipts are collected.

The already approved retirement of active pre-1.0 tier issuance belongs in this final 1.x compatibility change **before** the final wheel freezes. Preserve historical artifact readability and diagnostics; do not reinterpret old evidence as beta qualification or change scoring floors.

## Sequencing

P1 hard v1.0 gate from the existing stability contract. Behavior repairs precede freeze; beta sourcing and blind human labeling can run in parallel and do not need to wait. Final #512 receipts and #509 promotion follow freeze. #493 owns accepted-decision documentation, #499 reproducible goldens, #553 discovery enumeration, and #566 stale qualification instructions. Broad schema archival (#492) is not a prerequisite. This is not a numeric-version-only change.

pengfei-threemoonslab:
Closes #569.

`STABILITY.md:33` promised that a `1.0` product line would not begin until the report schema reached `1.0` and held without a breaking change. Main emitted `0.43`, and no delivery issue owned satisfying that published condition. This is that delivery.

## The freeze is a promotion, not a break

`docs/report-schema.v1.0.json` and `docs/report-schema.v0.43.json` are **byte-identical apart from `$id`, `title` and the version constant**. A consumer written against `0.43` needs no edit.

That matters because `STABILITY.md`'s own rule reads "`report_schema_version` bumps minor on additive changes, **major on breaking**" — read literally, `0.43 → 1.0` announces a break. It doesn't, and `test_the_1_0_schema_is_a_promotion_of_the_last_pre_freeze_schema` makes that a fact somebody can check rather than a sentence in a migration note.

`docs/report-v1-consolidation-rc.md`'s regrouping proposal was **not** adopted and is marked superseded: re-parenting the whole report is a way to do work, not a way to establish compatibility.

## What actually moved: the promise

[`docs/report-1-0-contract.md`](https://github.com/ThreeMoonsLab/agents-shipgate/blob/claude/agents-shipgate-issue-569-1b8782/docs/report-1-0-contract.md) is the new document. It carries:

- **The `1.x` rules** — additive-only; anything else needs `2.0`; a deprecation cycle counts **shipped releases**, never time on unreleased `main`; every published schema URL keeps its bytes forever.
- **The stable/provisional inventory** — all 54 report top-level fields on two independent axes (`presence` = in the schema's `required` list; `stability` = name/type/meaning frozen for `1.x`), plus CLI, exit codes, Action, control and the other published artifacts.
- **Migration from the contract that actually shipped** — `v0.15.0`, which emits `report 0.28` / `contract 10` and predates `minimum_control_contract_version` entirely. Not from `0.43`, which only ever existed on `main`.
- **The recorded RC exercise** (below).

`tests/test_report_1_0_contract.py` binds that document to the runtime: every field classified exactly once, every row naming a real field, the presence column equal to the published schema's `required` list, and **nothing `STABILITY.md` already promised stable quietly demoted** — a demotion in the freeze commit is a removal with no deprecation cycle at all, dressed as an inventory entry.

## Pre-freeze reports stop being engine input

`ReadinessReport` is `extra="allow"` with defaults nearly everywhere, so `model_validate` accepts a `0.9` payload and hands back an object whose newer blocks hold *this build's* defaults — with nothing distinguishing a value that was recorded from one that was invented.

New `schemas/report_compatibility.py` gates every boundary where a user hands the engine a report path — `scan --diff-from`, `explain-finding`, `findings`, `scenario suggest`, `evidence-packet` — **before** validation, with a stable `reason_code` and a route:

| `reason_code` | Cause |
|---|---|
| `report_schema_pre_freeze` | `0.x` |
| `report_schema_newer_than_engine` | a `1.x` minor this build does not have |
| `report_schema_future_major` | `2.x`+ |
| `report_schema_malformed` / `report_schema_missing` | not a version this engine emits |

**Nothing is converted.** No artifact gains current authority through conversion or a restamped digest. Every superseded schema stays published, so archived reports remain validatable. `org` and `attest` stay ungated deliberately — they read a run's own bundle, neither is a release gate, and refusing there would break the diagnostics this freeze is required to preserve.

### The two parser directions are not symmetric

A reader that only *projects* its input accepts a later `1.x` minor — that is what additive means. An evidence *comparison* refuses it, because the additive rule is a promise to consumers, not a licence to diff against blocks this build cannot interpret. Strict is what a new caller gets by forgetting; `accept_newer_minor=True` is a claim a caller makes about itself.

## One fail-open defect found on the way

This is why the coupling is now structural rather than reworded.

The release decision routed an incomparable `--diff-from` base to `insufficient_evidence` only when its `source_warning` gap matched **three substrings of one refusal's prose** (`"predates report schema"`, `"semantic evidence"`, `"not comparable with --diff-from"`). Rewording the refusal — which this change does — silently downgraded the verdict to `review_required`: the run stopped withholding a verdict it had no evidence for, and **nothing that named the decision failed**. It surfaced as one unrelated-looking assertion in a differently-named test.

The refusal now carries its own reason code and the classifier looks it up. `tests/test_release_decision.py`'s guard was also building its fixture warning by hand-copying the prose, so it could not have caught the drift it exists to catch — it now sources the warning from the producer.

## Qualification

- Production `beta` policy's `required_report_schema_version` → `1.0`, equal to what the engine emits.
- **`pre_1_0` tier issuance is retired.** The runner produces no artifact carrying it; `--policy-tier pre-1.0` is refused *by name* rather than dropped from the choice list, so an operator running a pre-freeze runbook line gets the reason instead of "invalid choice".
- Retirement is **issuance only**. The policy, its thresholds and every reader remain, and it deliberately keeps its historical `0.43` pin: `tier_for_requirements` names a policy by byte-equality, so re-pinning it would demote every existing `pre_1_0` artifact to the unnamed `test` tier and replace an accurate diagnosis with a misleading one. **No scoring floor moved.**
- The retirement guard sits at `select_release_requirements` (the only path `main`, and therefore the release workflow, can reach) and deliberately *not* on the `requirements=` keyword — refusing there would destroy the historical re-scoring the retirement is required to preserve.

## Contract

Runtime contract `33 → 34`. `minimum_control_contract_version` stays `21`: every operational control shape is byte-identical. Packet, verifier, receipt, capability-lock, attestation, preflight and host-grants schemas keep their versions — renumbering every schema would be work without compatibility.

## Compatibility fixtures (acceptance 2)

`tests/test_report_1_0_compatibility_fixtures.py`, over the three consumer paths:

- **the CLI as a process** — driven as a subprocess with every entry-point env var cleared; emitted schema equals what `contract --json` advertises, and the report validates against the schema it names. It runs *this* working tree (`PYTHONPATH` is pinned to the worktree's `src/` on purpose — the `.venv` editable install points at the main checkout). A genuinely **installed** distribution is exercised by the RC exercise below, not by these fixtures;
- **generated Action workflow** — through `scripts/github_action_outputs.py::extract_outputs`, asserting *values* (a re-parented block would publish `''` and every `== '0'` condition would silently flip), plus that the emitted workflow hardcodes no schema version;
- **machine report readers** — the relabel-refusal at every boundary, the `--diff-from` case that withholds rather than aborts, and the acceptance case, because a gate nothing passes is indistinguishable from a broken one.

## Recorded RC exercise (acceptance 5)

Against an **untagged, unpublished candidate** — which #569 explicitly permits — built with the locked `hatchling==1.32.0` backend and installed into an isolated virtualenv, driven as `python -I -m agents_shipgate`:

Observed: `cli_version 0.16.0`, `contract_version 34`, `report_schema_version 1.0`, `min_control 21`.

| Replay | Observed |
|---|---|
| `scan` | report declares `1.0`, equal to what `contract --json` advertises |
| relabelled `0.43` → `findings` | exit `3`, `[report_schema_pre_freeze]`, names the route |
| `1.99` → `findings` | accepted (projection) |
| `--diff-from` the `0.43` base | scan completes, diff disabled, refusal recorded, verdict not `passed` |
| `--diff-from` the `1.99` base | refused `report_schema_newer_than_engine` |
| own `1.0` report | accepted |

The relabelled input is the candidate's own report with **only** its version rewritten — a single-variable experiment, and precisely the move the freeze forbids: had it been accepted, an old signed receipt would acquire current authority through a one-field edit.

Automated as `report_schema_exercise()` in `scripts/release_engine_smoke.py`, so the final candidate re-runs it on the disposable job that already proves CLI/Action parity. **Not** release provenance: the wheel is unsigned, unqualified, and built from a dirty tree, so its SHA-256 identifies nothing reproducible. It establishes the report contract and nothing about qualification, adoption or the final candidate — #512/#509/#570 own those.

## Design-partner ledger

`docs/design-partner-pilot-results.md`'s source-tree column was **re-measured**, not carried forward — a contract bump is exactly what its guard exists to catch. The Route H fixture was rebuilt and every cell reproduced unchanged (`block`/`critical`, 4 violations; `not_applicable_host_review`; forced verify exit 2; all 4 drift expansion signals) except the contract number. "Nothing changed" is a measurement here, not an assumption.

## Sequencing note

Epic #572 places #569 after behavior convergence, and #563/#548/#557 remain open. The owner's call was to proceed: there are no external consumers today, so a later breaking change costs the renumbering rather than a migration. The freeze is explicitly conditional — `docs/report-1-0-contract.md` § Restarting the freeze states that a breaking change ships as `2.0`, closes the `1.x` line, and invalidates every qualification receipt collected against `1.0`.

## Review round 1 — addressed in `dfd8cf5e`

An extra-high-effort review of the freeze surfaced 10 findings; all are fixed. The two that mattered were both a claim being stronger than the code:

- **The compatibility fixtures did not test what they said.** `_isolated_cli` documented itself as running an installed package while actually pinning `PYTHONPATH` to the worktree. The pin is correct — without it a worktree run tests the main checkout — but the docstrings and this PR read as though a built distribution were covered by CI. Renamed `_worktree_cli`; both docstrings now say what it does and point at the RC exercise for the installed-distribution claim. The bullet above is corrected too.
- **`explain-finding` asserted an invariant in a comment.** "1.x carries `agent_action` by construction" is true of the published schema but not of the Pydantic model (`AgentAction | None = None`), so a payload merely *declaring* `1.0` could still produce the `"agent_action": null` explanation #58 P2.2 exists to prevent. It now refuses structurally, the way `findings.py` already refuses a missing `provenance_kind`.

Also: a bare `assert` guarding a runtime invariant in the refusal gate (stripped by `python -O`, leaving `None[1]` → `TypeError` past callers that wrap only `ValueError`); three now-dead helpers deleted; a vacuous `>= 0.31` check; a promotion test that skipped exactly when it started mattering, now comparing the two frozen documents permanently; a doc snippet still teaching a fallback the same release made unusable; a discarded `wheel_version` parameter; and the RC exercise now states which side each fact comes from.

**`docs/distribution-surfaces.md` gains a `report_schema_pin` claim.** Five registered surfaces restate which `report-schema.v<X>.json` to validate against, and that is an engine answer (`contract --json` publishes it) — so per the registry's own rule it belongs there rather than in hand-maintained per-file lists. Both directions of the proving test were verified by perturbation: a pin left behind fails, a pin ahead of the build fails, and a frozen-reference list of older versions correctly does not.

## Review rounds 2–4

The loop ran until each round's fixes stopped producing new ones.

**Round 2 (`ad37f8e4`) — the round-1 fix introduced the defect it was about.** Replacing the bare `assert` added a refusal carrying `report_schema_engine_version_unreadable`, but `report_schema_refusal_code` built its recognizable codes by walking `ReportSchemaStatus` — a second copy of the producer's vocabulary — and that code is not an *input* status. So the new refusal was invisible to the classifier, and an incomparable base carrying it would have routed to `review_required` instead of withholding the verdict: the same fail-open class, reintroduced in the one branch that fires when the install itself is broken. The recognizer now matches the bracket marker, so producer and consumer are in sync by construction.

**Round 3 (`1f122627`) — that branch was still never executed.** It is unreachable from any input, which is why it needed forcing. `test_a_build_that_cannot_read_its_own_version_refuses_readably` patches the engine's declared version and asserts the three things rounds 1 and 2 claimed. Perturbation-verified.

**Round 4 (`c50ef7ea`) — the contract doc named a command that does not exist.** The migration section listed `agents-shipgate packet`; the command is `evidence-packet`. The one document whose job is telling a reader what to do with a refused artifact told them to run something that errors, and nothing caught it — the docs tests check links, the parity harness checks pins. `test_every_command_the_contract_names_exists` now checks every `agents-shipgate <subcommand>` in that doc against the CLI's registered commands. Perturbation-verified.

## Verification

- `pytest -n auto -m "not perf" --ignore=tests/test_adapter_static_only.py` — **exit 0**
- `tests/test_adapter_static_only.py` — exit 0
- `tests/test_latency_budget.py -m perf` — exit 0
- `ruff check .` — clean; `scripts/generate_schemas.py --check` — no drift
- self-dogfood `verify --config shipgate-self.yaml --base origin/main --head HEAD` — **`human_review_required` / `control.state: review_publishable`**, reason: *"This PR edits a release trust root … a coding agent cannot self-approve that change."* That is the correct and expected verdict for a change touching `AGENTS.md` and the `SKILL.md` copies, and it is why the `verify-self` CI job stays green: it fails only on `blocked,unknown`, leaving `human_review_required` to the PR reviewer.

  **Correction:** an earlier revision of this PR body reported this run as `mergeable` / `complete`. That was wrong. It ran while the changes were still uncommitted, so `--base origin/main --head HEAD` had `base == head == 8e81672b` and the verifier evaluated an *empty diff* — it verified nothing. The number above is from the real range.

Protected-surface edits (`AGENTS.md`, the three `SKILL.md` copies) were routed by `./shipgate preflight` as `human_review_required` and approved by the repository owner in-conversation before any were touched.


src/agents_shipgate/schemas/report_compatibility.py
```diff
@@ -0,0 +1,335 @@
...
+            f"`{regenerate_command}`. There is no conversion: a pre-freeze report "
+            "does not record what the current contract needs, and re-labelling one "
+            "would assert evidence nothing measured. Re-verify anything derived "
+            "from it (baselines with `agents-shipgate baseline save`, receipts by "
+            "re-running the verification that produced them) against the "
+            "regenerated report."
+        )
+    return f"Regenerate the artifact with `{regenerate_command}`."
+
+
```
pengfei-threemoonslab, line 256: [P1] Cover the report-consuming mutation command with this boundary

The new guard is wired into the projection readers and --diff-from, but `apply-patches --from <report.json>` still consumes arbitrary report versions without calling it. Using the existing test_apply_patches stale-suppression fixture, I generated a report, rewrote only report_schema_version to `0.43` and then (in a fresh fixture) `2.0`, and ran `apply-patches --confidence high --apply --json`: both returned exit 0/status=applied and changed shipgate.yaml. Thus a pre-freeze or unknown-major report that findings refuses can still drive a file mutation under this engine. Validate the report version before extracting any patches or writing files, and add current-version acceptance plus pre-freeze/future-major/missing-version non-mutation regressions. If patch payloads are intended to be a separately versioned exception, that needs an explicit validated contract rather than unconditional report-version acceptance.

STABILITY.md
```diff
@@ -68,6 +94,54 @@ thresholds and release permissions are unchanged. See
...
+stable/provisional inventory of report fields, CLI, exit codes, Action and
+control surfaces is in
+[`docs/report-1-0-contract.md`](docs/report-1-0-contract.md), checked against
+the runtime by `tests/test_report_1_0_contract.py`.
+
+**Pre-freeze reports are no longer engine input.** `scan --diff-from`,
+`explain-finding`, `findings`, `scenario suggest` and `evidence-packet` refuse a `0.x`
+report by name, with a stable `reason_code` and a regeneration route, instead
+of validating it against a model whose defaults would stand in for blocks it
+never recorded. A report from a *newer* `1.x` minor is refused the same way:
```
pengfei-threemoonslab: [P2] Publish the implemented forward-read exception consistently

This states that a newer 1.x minor is refused by the listed readers, but all four projection integrations now pass `accept_newer_minor=True`; the projection compatibility test (and the RC table in docs/report-1-0-contract.md) deliberately expect acceptance. The new contract document repeats the blanket refusal in its 1.x rules and input section. A consumer following the frozen contract therefore gets the opposite result from findings/explain-finding/scenario suggest/evidence-packet for a 1.99 report. Distinguish projection acceptance from --diff-from comparison refusal in STABILITY.md and every new contract statement, and test the published distinction. Keep the stricter unknown-major refusal.

pengfei-threemoonslab (commented):
Reviewed commit c50ef7ea27c1011fab1b0ad03191a7365652d09c. Recommendation: request changes for the report-reader coverage and published-contract contradictions below.

Independently ran tests/test_report_1_0_contract.py, tests/test_report_1_0_compatibility_fixtures.py, and tests/test_release_engine_smoke.py: all 55 tests passed. Inspected the version classifier, reader integrations, schema promotion comparison, qualification-policy transition, and smoke exercise. GitHub CI is green at this commit.

Additional reproduction: take the existing stale-suppression patch fixture, generate its report, change only report_schema_version to 0.43 (and separately 2.0), and run apply-patches --from <report> --confidence high --apply --json. Both exit 0, report status=applied, and modify the fixture manifest. This does not meet the advertised unsupported-report input boundary.

This is a code review, not qualification or release evidence. Posted as COMMENT because the authenticated account owns the PR and GitHub disallows requesting changes on one's own PR; the technical recommendation is to fix and re-review.

pengfei-threemoonslab (commented):
Re-reviewed the published fixes. The two earlier findings are addressed: apply-patches checks the report version before extracting or applying mutations, and the compatibility documentation now distinguishes forward-compatible projection readers from comparison/mutation readers. Regression tests cover missing, malformed, pre-freeze, newer-minor and future-major inputs without modifying the manifest. Structured errors include a typed recovery route.

Validation: full suite completed; the invocation-policy fixture was updated to reach its intended missing-manifest-field route, and the focused compatibility/invocation/hook rerun passed (159 passed, 2 skipped). The integrated public-contract/comparison/reader checks passed (977 passed, 2 skipped). Performance tests and Ruff pass. Fresh CI is running.

Human merge review is still required by Shipgate. Exact-head control reports `review_publishable`, with commit/push/update_pr allowed and merge/report_complete denied. Three suppression-immune `SHIP-VERIFY-TRUST-ROOT-TOUCHED` findings identify:

- `adoption-kits/claude-code-skill/SKILL.md`
- `plugins/claude-code/skills/agents-shipgate/SKILL.md`
- `skills/agents-shipgate/SKILL.md`

These edits publish the report 1.0 schema reference and migration promises in protected skill instructions. I have preserved them and have not weakened or suppressed the gate. The maintainer must review these edits and perform the merge; I will leave this PR open. No remaining actionable code-review findings were found in this pass.

pengfei-threemoonslab (commented):
Reviewed final head `6648515a6b951ae568c98446e9e5b9503f58019d` against current main. 869 final integration tests, dependency-lock checks and Ruff passed. All GitHub CI checks passed on this exact published head. The human-review requirements below remain outstanding.

Technical re-review found no remaining code defect in the reviewed report-contract changes. The fixes enforce supported report schemas before apply-patches can parse or mutate a manifest, preserve older read-only projection compatibility, and cover malformed, missing, future-minor and future-major schema inputs. Final integration preserves current reader vocabulary and regenerated fixture artifacts.

This is not approval to merge the protected change. The PR updates the shipped agent skill instructions in `adoption-kits/claude-code-skill/SKILL.md`, `plugins/claude-code/skills/agents-shipgate/SKILL.md`, and `skills/agents-shipgate/SKILL.md`. Those instructions and the 1.0 report-contract promise require human review. Contract version 34 and report schema 1.0 do not themselves constitute a package release.

Fresh local control:

```json
{
  "control_state": "review_publishable",
  "current_control_id": "sha256:49bdde0afcc16f0d4ae7557a1a24dbc8b04744bae0e79bbd7076b06e4dfa76e7",
  "decision": "review_required",
  "next_actor": "human",
  "permissions": {
    "commit": true,
    "edit": true,
    "merge": false,
    "push": true,
    "report_complete": false,
    "update_pr": true
  }
}
```

pengfei-threemoonslab (commented):
Re-reviewed final head `ddb0842710251fdf4a4736c9bea0f2ae95c96f40` after integration of the reviewed #679 changes from main. Local integration result: 903 passed in 21.92s. Ruff passed. All GitHub CI checks passed on this exact published head. The human-review requirements below remain outstanding.

Technical re-review found no remaining code defect in the reviewed report-contract changes. The fixes enforce supported report schemas before apply-patches can parse or mutate a manifest, preserve older read-only projection compatibility, and cover malformed, missing, future-minor and future-major schema inputs. Final integration preserves current reader vocabulary and regenerated fixture artifacts.

This is not approval to merge the protected change. The PR updates the shipped agent skill instructions in `adoption-kits/claude-code-skill/SKILL.md`, `plugins/claude-code/skills/agents-shipgate/SKILL.md`, and `skills/agents-shipgate/SKILL.md`. Those instructions and the 1.0 report-contract promise require human review. Contract version 34 and report schema 1.0 do not themselves constitute a package release.

Fresh local control:

```json
{
  "control_state": "review_publishable",
  "current_control_id": "sha256:a0828466376c7f58a4b9219bc2cc627ef2aecc53009a68c9037d4ef372335143",
  "decision": "review_required",
  "next_actor": "human",
  "permissions": {
    "commit": true,
    "edit": true,
    "merge": false,
    "push": true,
    "report_complete": false,
    "update_pr": true
  }
}
```

pengfei-threemoonslab:
Reconciled with `main` and re-verified. Head is now `b65974f8`.

**The contract collision.** After this branch froze the report at contract v34, `main` shipped v34 (#685, host drift 0.4) and v35 (#684, `agent_boundary_result/v3`). The same numbers meant different things. The freeze is now **v36**, on top of `main`'s history. Current-state pins moved to 36: `contract.py`, `.well-known`, the three contract assertions, the `agent-contract-current.md` stamp, `architecture.md`, and this branch's own `STABILITY.md`, `passed-verdict-contract.md`, `report-reading-for-agents.md` and the migration table. `main`'s v3 boundary schema and host schema 0.4 are **kept**. This branch's side still carried v2 and 0.3, and taking it would have silently regressed both. `MINIMUM_CONTROL_CONTRACT_VERSION` stays 21 on every side.

**Recorded measurements were re-measured, not retyped.**
- `test_route_readiness_source_tree_row_matches_this_tree` was the one failure in the full local suite after the merge. It exists to stop a contract bump from being carried forward. I rebuilt the documented Route H fixture and observed all seven source-tree cells on `871786ba`. Every cell reproduced except the contract number, so only that cell and the dated recheck line changed.
- The installed-candidate replay in `docs/report-1-0-contract.md` named contract 34. Release Engine Smoke [run 34726385422](https://github.com/ThreeMoonsLab/agents-shipgate/actions/runs/34726385422) rebuilt the wheel from `8a14470d`, and its `report_schema_exercise` passed. I then installed the downloaded wheel in isolation and replayed all six rows by hand; every one reproduced. Two earlier dispatches landed on the stale head `ddb08427` and were cancelled. Their output is not cited.

**Still to do before merge.** Merges are serial under the up-to-date rule. This will be updated to `main` again after #678, then merged once CI is green on that head.

PR #668, opened 2026-09-11
fix(diff): read every counted hunk row, so header-shaped content is not lost (#611)

### LINKED ISSUE #611: Unified-diff header filtering drops Markdown separator rows inside real hunks
Found while completing #545's paired instruction/structure regressions on base `42818ec25b06cf39d5c95024e2ba7884bffa4166`. Deferred from this implementation pass as requested.

## Problem

The shared unified-diff parser drops real hunk content when a removed line starts with `--` or an added line starts with `++`. Its header exclusions apply inside a hunk. A real Markdown horizontal-rule removal consequently loses a line before the complete-text comparison can validate it.

## Reproduction

Create and commit `AGENTS.md` with three lines: `Explain results.`, `---`, `Keep examples concise.`. Remove only the horizontal rule. Actual `git diff` produces:

```diff
@@ -1,3 +1,2 @@
 Explain results.
----
 Keep examples concise.
```

`core/boundary_diff.py::parse_unified_diff` reports `old_count=3` but retains only the two context rows. The removed `---` row is absent. On the #545 candidate, the local check correctly refuses that incomplete comparison as `BOUNDARY-INPUT-INCOMPLETE` / `human_review_required`, even though both captured instruction profiles are plain guidance with the same structural digest. This is reproduced input loss and unnecessary review, not evidence of an unsafe pass.

The corresponding code excludes every `raw_line.startswith("---")` / `raw_line.startswith("+++")` rather than distinguishing file headers from content using parser state. Header-looking content with spaces also needs examination because the earlier file-header branches run before hunk consumption.

## Proposed direction

Make hunk state and declared row counts determine whether a line is hunk data. Preserve existing path identity checks, rename validation and malformed-input refusals. Do not weaken #545's completeness requirement to accommodate missing rows.

## Acceptance

- [ ] Real Git diffs removing `---` and adding `+++` retain every counted hunk row.
- [ ] Header-looking lines inside hunks cannot overwrite the file's identity.
- [ ] Multi-file/multi-hunk, rename and no-newline cases retain their existing contracts.
- [ ] The same plain-guidance horizontal-rule removal reaches a complete structural comparison through local and final verification.
- [ ] Malformed counts and genuine path/header mismatches still refuse rather than becoming successful comparisons.

Related: #545, #516, #572. Keep this parser repair separate from changing the instruction semantic boundary.

pengfei-threemoonslab:
Closes #611.

## The defect

The shared unified-diff parser decided what a line *was* from how it was spelled. Two shapes Git really emits collide with the file-header prefixes:

| file content | rendered in the diff | old behaviour |
| --- | --- | --- |
| removed `---` (Markdown rule) | `----` | row dropped; `old_count` 3 vs 2 retained rows |
| removed `-- note` | `--- note` | matched the `--- ` **file header** branch: `old_path` became `\0invalid-diff-path` |

The first cost an unnecessary `human_review_required` on a plain-guidance edit, which is what #611 reports. The second is worse and was not in the report's title: a file's own content rewrote the file's identity.

Reproduced with real `git diff` before touching anything:

```
old_path='\x00invalid-diff-path' new_path='AGENTS.md'
removed []
```

## The fix

Hunk state and the header's declared row counts decide, not spelling. While a hunk still owes rows, any line opening with a unified-diff marker (space, `+`, `-`, `\`) is that hunk's data and is dispatched on its first character alone. `\ No newline at end of file` is a marker, so it consumes no row budget.

**Nothing is loosened.** A truncated or over-declared hunk ends at the first line that cannot be hunk data, keeps its short row list, and is still refused by the declared-count comparison in `core.instruction_structure`; the next file's header is still read as a header. Path identity, rename, new/deleted and malformed-header contracts are unchanged. Lines outside a hunk keep their existing branches, so callers that synthesise headerless diffs are untouched.

## Acceptance

- [x] Real Git diffs removing `---` and adding `+++` retain every counted hunk row.
- [x] Header-looking lines inside hunks cannot overwrite the file's identity.
- [x] Multi-file/multi-hunk, rename and no-newline cases retain their existing contracts.
- [x] The same plain-guidance horizontal-rule removal reaches a complete structural comparison (`instruction_pair_is_complete` and `unchanged_instruction_structure` both hold).
- [x] Malformed counts and genuine path/header mismatches still refuse.

## Evidence

16 cases in `tests/test_boundary_diff_hunks.py` — real `git diff` output where the defect is about what Git emits, handwritten text where the point is input Git would never produce.

**Negative control** (the part that matters): replayed against pre-fix `main`, the 11 defect cases fail and the 5 contract-preservation cases pass in both states. A guard that passes before the fix is not a guard.

Locally green: every test matching `boundary|instruction|diff|host|codex|hook`, plus `ruff check`. The file is not `ruff format` clean on `main` either and CI runs only `ruff check`, so it is left unreformatted rather than adding unrelated churn.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 ## Unreleased
 
+- Read every counted hunk row, so header-shaped content stops being lost.
+  Hunk state and the header's declared row counts, not a line's spelling,
+  decide what the shared unified-diff parser treats as content: a removed
+  `---` renders as `----` and a removed `-- note` renders as `--- note`,
+  and matching those against the file-header prefixes dropped the row and,
+  in the second case, replaced the file's identity with the invalid-path
+  sentinel. A plain Markdown horizontal-rule removal now reaches a complete
+  structural comparison instead of an unnecessary `human_review_required`.
+  Truncated and over-declared hunks still end at the first line that cannot
+  be hunk data, keep their short row list, and are still refused by the
+  declared-count comparison; path, rename, no-newline and malformed-header
+  contracts are unchanged (#611).
+
 - Name what a change did to the bound each finding depends on.
   `tool_surface_diff.finding_attributions[]` projects the shipped
   `openapi_delete/v1` and `sdk_boolean_guard/v1` comparisons onto active
--- a/src/agents_shipgate/core/boundary_diff.py
+++ b/src/agents_shipgate/core/boundary_diff.py
@@ -105,17 +105,38 @@ def git_diff_path_token(prefix: str, path: str) -> str:
 
 
 def parse_unified_diff(diff_text: str) -> list[DiffFile]:
+    """Parse unified diff text into per-file added/removed lines and hunks.
+
+    Hunk state, not line spelling, decides what is content.  A removed
+    ``---`` renders as ``----`` and a removed ``-- note`` renders as
+    ``--- note``; matching those against the file-header prefixes drops a
+    counted row, and in the second case rewrites the file's own identity
+    from its content.  While a hunk still owes rows, any line opening with
+    a unified-diff marker (space, ``+``, ``-``, ``\\``) is that hunk's data
+    and is dispatched on its first character alone.
+
+    The row budget comes from the hunk header, so a truncated or
+    over-declared hunk ends as soon as a line cannot be hunk data.  The
+    short hunk is kept, and the declared-count comparison in
+    ``core.instruction_structure`` still refuses it — repairing the parser
+    must not turn an incomplete comparison into a successful one.
+    """
+
     files_out: list[DiffFile] = []
     current: dict[str, Any] | None = None
     current_hunk: DiffHunk | None = None
+    old_remaining = 0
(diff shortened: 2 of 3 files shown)


pengfei-threemoonslab (commented):
Reviewed commit bbbbb7edf71126a196e7b87a1c90784d5331d795. No actionable correctness findings in this revision.

I inspected the parser state transitions and the downstream instruction-completeness check, including header-shaped content, exhausted and truncated hunks, multi-file boundaries, rename handling, and no-newline markers. Independently ran all 16 tests in tests/test_boundary_diff_hunks.py successfully, plus 67 tests across test_instruction_structure.py, test_instruction_structure_contracts.py, test_instruction_structure_hooks.py, and test_instruction_structure_workflow.py. The latter suite passed with AGENTS_SHIPGATE_PYTHON pointing to the existing virtualenv; without that override, the temporary checkout launcher selected an interpreter lacking typer. GitHub CI is green at this commit.

This is a code-review assessment, not a merge or release authorization.

PR #669, opened 2026-09-11
fix(scan): one availability contract for required tool sources (#585)

### LINKED ISSUE #585: Reconcile required missing SDK entrypoints with the documented scan input-error contract
## Problem

The documented scan error contract says a missing required `tool_sources[].path` is `InputParseError(3)`, while the OpenAI Agents SDK loader returns a warning even when `optional: false`. An integration that uses execution status to distinguish bad input from a completed advisory scan receives a different shape depending on the source reader.

## Evidence

Found while validating #561 against main `c079864e8a37209a7712d74197fd1cf1b4ebf06e`. A manifest with an explicit purpose and `tool_sources: [{id: sdk, type: openai_agents_sdk, path: agent.py, optional: false}]`, with no `agent.py`, completes `agents-shipgate scan -c shipgate.yaml` with advisory exit 0, a source warning `OpenAI Agents SDK entrypoint not found: .../agent.py`, and `release_decision.decision: insufficient_evidence`.

`inputs/openai_sdk_static.py::load_openai_sdk_static_tools` returns `LoadedToolSource(warnings=[...])` on `not path.exists()` without consulting `optional`. The required-path resolver in `cli/scan/validation.py` is called by doctor/inspect; the scan orchestration does not use it as a common precondition. #561 adds an explanation to the existing warning and preserves its release/exit behavior.

This reproduction is a documented execution-contract mismatch, not a reproduced unsafe auto-pass: the release decision still abstains. Do not interpret advisory exit 0 as merge authority.

## Scope

Decide and implement one supported contract for required source availability before adapter dispatch, preserving containment checks and optional-source warning behavior. Audit the existing reader-specific exceptions before changing shared behavior. If there is an intentional SDK exception, describe it explicitly in the error catalog and instructions instead of promising universal exit 3. Missing input must never be repaired by inventing a binding or authority declaration.

## Acceptance

- [ ] Public scan, doctor JSON/human output and verify agree on the documented required/optional availability boundary.
- [ ] A missing required SDK entrypoint has a tested structured result and the exact missing source path; optional behavior remains explicit.
- [ ] CLI/module invocation, advisory/strict mode and other existing adapter behavior have regression coverage.
- [ ] Error catalog, docs and protected agent instructions agree with the chosen implementation. Release/merge authority cannot be inferred from advisory exit 0.

## Sequencing

Deferred from #561; related #547/#543 recovery clarity. Keep #561's explanation-only change separate from this execution-contract decision. Protected instruction edits follow the repository's preflight and human-review boundary.

pengfei-threemoonslab:
Closes #585.

## The mismatch

`docs/diagnostics.md:267` and `AGENTS.md` both publish the same contract: a required `tool_sources[].path` that does not resolve is `InputParseError(3)` from `scan`. Measured on `main`, the answer depended on the reader:

| required source, path absent | `main` | this PR |
| --- | --- | --- |
| `mcp`, `openapi`, `mcp_server_source` | exit 3, `Input file not found: …` | exit 3, names the source id |
| `openai_agents_sdk` | **exit 0**, source warning, `insufficient_evidence` | exit 3 |

`tests/test_adapter_registry.py::test_dispatch_required_source_raises` already asserted the dispatcher raises for a required missing source. It passed only because its fixture used `mcp`.

## The decision

The issue allowed two outcomes: enforce the published contract, or document an SDK exception. Enforcing it is the right one — the docs already promise it, the majority of readers already do it, and exit 3 naming one fixable thing is more useful to a first-time user than exit 0 with a buried warning and an abstaining verdict.

So: **one availability precondition, before any adapter runs**, reading the same resolver `doctor` uses for `SHIP-DIAG-MISSING-SOURCE-FILE`. Scan and doctor now agree by construction rather than by each reader deciding for itself.

```
Input parsing error: Required tool source unavailable: 'sdk' at 'agent.py' does not exist.
Restore the declared file, correct tool_sources[].path, or remove the source entry.
```

The message deliberately does not suggest `optional: true`. Telling someone to downgrade their own required declaration to silence the error is the weakening this issue warns against.

## Deliberately unchanged

- **Optional sources.** They keep the warning and the `coverage_recovery` evidence; the resolver never reports them. Covered for all four path-taking source types.
- **doctor.** `--json` still exits 0 with the diagnostic, human form still exits 3.

## The one behaviour change worth reviewing

`verify` applies the precondition per tree, so a **base** commit whose manifest declares a path absent from that tree changes:

| | `main` | this PR |
| --- | --- | --- |
| `base_status` | `succeeded` | `scan_failed` |
| `capability_review` | `added: 2` | `added: 0` |
| `merge_verdict` | `insufficient_evidence` | `insufficient_evidence` |

Measured on the same fixture, not asserted. The verdict does not move. What changes is the claim: `main` reported the base as successfully scanned with zero tools and a `+2` delta, having never read the source the base manifest declared. The new `base_notes` entry is explicit and actionable, and says the head gate is untouched:

> Base scan failed without changing the head gate: Required tool source unavailable: 'sdk' at 'agent.py' does not exist. …

## Two existing tests moved with the decision

- `test_required_sdk_source_keeps_existing_advisory_execution_result` was #561's pin that *recovery metadata must not alter the execution contract*, and its own comment named #585 as the owner of the contract. It keeps that invariant — both paths equal — and now asserts exit 3.
- Four dispatch-order fixtures in `test_adapter_registry.py` declared placeholder paths (`a.py`, `b.py`) that never existed. They are marked `optional`, which is what they always were; the ordering assertions are untouched.

## Acceptance

- [x] Public scan, doctor JSON/human and verify agree on the required/optional boundary — by sharing the resolver.
- [x] A missing required SDK entrypoint has a tested structured result naming the exact path; optional behaviour is explicit.
- [x] CLI, `python -m` invocation, advisory and strict mode covered; other adapters regression-covered.
- [x] Error catalog, `docs/diagnostics.md`, `AGENTS.md` and `llms-full.txt` agree with the implementation.

15 cases in `tests/test_required_source_availability.py`, one of which fails if any adapter runs before the check. **Negative control:** replayed against pre-fix `main`, 10 fail; the 5 that pass are the optional-source and doctor cases that must not change.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -495,7 +495,8 @@ When a required `tool_sources[].path` does not resolve under the manifest direct
 
 - `agents-shipgate doctor --json` exits **0** with a `SHIP-DIAG-MISSING-SOURCE-FILE` diagnostic and an `unresolved_sources: [{id, declared_path, line, reason}]` field in the payload, so an agent can route to a fix without parsing the error message. `reason` is `"missing"` or `"outside_manifest_dir"`.
 - `agents-shipgate doctor` (no `--json`) prints the same `unresolved_sources` + diagnostic block in human-readable form and **exits 3**, preserving the pre-feature loud failure for interactive users.
-- `agents-shipgate scan` is unchanged — it still raises `InputParseError(3)` regardless of `--json`. Once you're past doctor, missing sources are real scan failures.
+- `agents-shipgate scan` is unchanged — it still raises `InputParseError(3)` regardless of `--json`. Once you're past doctor, missing sources are real scan failures. One precondition enforces this for every source type before any adapter runs, off the resolver doctor uses, so the answer does not depend on which reader the source would have gone to. `optional: true` sources are not covered: they keep their source warning and recovery evidence, and the scan completes.
+- `verify` applies the same precondition to each tree it scans. A base commit whose manifest declares a path that tree does not contain yields `base_status: "scan_failed"` with the reason in `base_notes` and no capability delta; the head gate is unchanged.
 
 ### Missing vs invalid manifests
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
   absent entrypoint finish as an advisory exit-0 scan; an integration using
   execution status to tell bad input from a completed scan got a different
   answer per reader. The error names each offending source id, its declared
-  path, and whether it is missing or escapes the manifest directory.
+  path, and whether it was not found or escapes the manifest directory.
   `optional: true` sources are unchanged, keeping their warning and
   `coverage_recovery` evidence. `verify` applies the same precondition per
   tree: a base commit declaring a path absent from that tree reports
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,25 @@
 
 ## Unreleased
 
+- Refuse a required tool source whose declared path is unavailable, once,
+  for every reader. `scan` applies one availability precondition before any
+  adapter runs, reading the same resolver `doctor` renders as
+  `SHIP-DIAG-MISSING-SOURCE-FILE`, so the two agree by construction. The
+  contract `docs/diagnostics.md` already published — a required
+  `tool_sources[].path` that does not resolve is `InputParseError(3)` — was
+  true of the shared loaders and `mcp_server_source` and false of
+  `openai_agents_sdk`, which returned a source warning and let a required,
+  absent entrypoint finish as an advisory exit-0 scan; an integration using
+  execution status to tell bad input from a completed scan got a different
+  answer per reader. The error names each offending source id, its declared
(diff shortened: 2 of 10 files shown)


pengfei-threemoonslab (commented):
Reviewed commit 2e73f0d8182b42bce01522a40ba6b12cf28b2c39. No actionable correctness findings in this revision.

I checked the shared resolver, pre-adapter enforcement, doctor filtering, optional-source behavior, and the base-scan failure boundary. Independently ran tests/test_required_source_availability.py, tests/test_adapter_registry.py, and tests/test_coverage_recovery.py: all 61 tests passed. The first run's module-invocation test lacked `python` on PATH; it passed with the existing virtualenv on PATH. GitHub CI is green at this commit.

The required-source execution-status change is explicit and regression-tested; it does not establish runtime tool behavior. This is a code-review assessment, not approval of protected declarations or merge/release authority.

PR #670, opened 2026-09-11
fix(host): keep machine-written tool caches out of the host inventory (#598)

### LINKED ISSUE #598: Local check routes an incomplete generated-cache inventory to human review during concurrent tests
## Problem

During #557 validation, `shipgate check --agent codex --format agent-boundary-json` ran while the full pytest suite was active. It returned `human_review_required`, all permissions false, with `Directory inventory could not complete at tests/__pycache__`. No release declarations or protected workflow edits were involved. The repository was then left unchanged until tests exited; a fresh read-only verifier completed and current control returned `passed / complete`.

This is an observed local workflow interruption. Concurrent cache creation is a plausible cause, but the present diagnostic does not establish whether the directory changed during capture, hit a bound, or encountered another reader failure. Do not label that cause proven from the path alone.

## Proposed direction

Use the actual identity-reader failure origin to distinguish a recoverable unstable input snapshot from a semantic/human-owned decision. Where sound, provide a bounded retry after the writer is quiescent, preserving captured-input identity and real trust-root review. Examine whether generated cache inventory is needed for recognized host coverage before changing traversal; do not blindly skip paths or turn incomplete reads into success.

## Acceptance

- [ ] A deterministic fixture reproduces the relevant generated-directory inventory failure during a local check.
- [ ] The diagnostic identifies its typed cause and an exact safe next step without relying on exception-text inference.
- [ ] A quiescent rerun obtains new evidence; no cached denial or cached completion overrides it.
- [ ] Genuine missing, unreadable or protected capability inputs remain fail-closed and human-owned decisions retain their existing route.

## Sequencing

Deferred from #557, which validates after tests are quiescent. Related #547 delivered bounded recovery metadata; #588 owns deeper typed shared-reader provenance. This issue owns the concurrent local test/check workflow and appropriate recovery routing.

pengfei-threemoonslab:
Closes #598.

## What was observed

`shipgate check --agent codex` run while the pytest suite was active returned `human_review_required`, every permission false, with *"Directory inventory could not complete at tests/__pycache__"*. A quiescent rerun passed.

## What was actually wrong

**Not the refusal.** `IdentityBoundReadSession` revalidates every directory it scanned, so a `.pyc` appearing — or the directory vanishing — between the scan and its revalidation is precisely the incoherent read the session exists to refuse. Softening that would be the wrong repair.

The defect is one level up: **a bytecode cache was an identity-bound input at all.** No host reads configuration from a `__pycache__`, so walking it buys no coverage, and one such directory failing discards the entire repository inventory — `repository_paths = []`.

That is the question #598 asked to settle: *"Examine whether generated cache inventory is needed for recognized host coverage before changing traversal."* It is not.

## The change

`__pycache__`, `.pytest_cache`, `.mypy_cache`, `.ruff_cache`, `.tox` and `.nox` join the skip set that already holds `.git`, `.hg`, `.svn`, `node_modules`, `site-packages`, `.venv` and `venv`. Same principle, same list, one line of reasoning.

## What deliberately does not move

- A **recognized** host directory that changes mid-read is still refused and yields no grants.
- An **unreadable** directory is still fail-closed.
- A rerun after the writer goes quiet performs a real read: no cached denial, no cached completion.

Each of those is a test, and each passes both before and after the change — they are the guarantees, not the fix.

## Acceptance

- [x] A deterministic fixture reproduces the generated-directory inventory failure during a local check.
- [x] The diagnostic identifies its typed cause and next step without exception-text inference — the existing `HostInputFailure(reason, phase, source)` already does, and the reported instance no longer arises.
- [x] A quiescent rerun obtains new evidence; no cached denial or cached completion overrides it.
- [x] Genuine missing, unreadable or protected capability inputs remain fail-closed.

10 cases in `tests/test_host_inventory_stability.py`. **Negative control:** against pre-fix `main`, the 7 cache-boundary cases fail and the 3 integrity guards pass in both states.

## One thing I built and threw away

A new `input_unstable` failure reason, distinguishing "changed while read" from "cannot be read", with its own timing-based recovery text. I reverted it: the existing model already names a typed cause without text inference, and the new value turned out reachable only from a narrow parent-identity path I could not demonstrate end to end. An enum value no test can reach is surface, not a diagnostic — and this is a freeze window.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,22 @@
 
 ## Unreleased
 
+- Stop inventorying machine-written tool caches, so an ordinary concurrent
+  test run no longer collapses the host inventory. `check` run while pytest
+  was active returned `human_review_required` with *"Directory inventory
+  could not complete at tests/__pycache__"*; the refusal was correct — the
+  identity reader revalidates every directory it scanned, and a `.pyc`
+  landing between the two reads really is a directory that changed while it
+  was read — but a bytecode cache should never have been an identity-bound
+  input. `__pycache__`, `.pytest_cache`, `.mypy_cache`, `.ruff_cache`,
+  `.tox` and `.nox` are excluded from the repository walk for the same
+  reason `.venv`, `node_modules` and `.git` already are: no host reads
+  configuration from one, so inventorying them buys no coverage. Nothing
+  else is softened — a recognized host directory that changes mid-read is
+  still refused with no grants, an unreadable one is still fail-closed, and
+  a quiescent rerun still performs a real read rather than serving a cached
+  denial or a cached completion (#598).
+
 - Read every counted hunk row, so header-shaped content stops being lost.
   Hunk state and the header's declared row counts, not a line's spelling,
   decide what the shared unified-diff parser treats as content: a removed
--- a/src/agents_shipgate/core/host_grants.py
+++ b/src/agents_shipgate/core/host_grants.py
@@ -996,7 +996,18 @@ def _repository_paths(
     """Enumerate repository sources exclusively from the boundary registry."""
 
     indexed: dict[tuple[str, str], tuple[Path, str, str, str]] = {}
-    skipped = {".git", ".hg", ".svn", "node_modules", "site-packages", ".venv", "venv"}
+    skipped = {
+        ".git", ".hg", ".svn", "node_modules", "site-packages", ".venv", "venv",
+        # Machine-written tool caches. No host reads configuration from one, so
+        # inventorying them buys no coverage — and it made an ordinary
+        # concurrent test run collapse the whole repository inventory, because
+        # a `.pyc` appearing between the scan and its revalidation is a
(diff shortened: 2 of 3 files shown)


pengfei-threemoonslab (commented):
Reviewed commit e5bb71fdbd92f12ee9a3af470542dc1e8dbff45d. No actionable correctness findings in this revision.

I checked how the new exclusions interact with directory enumeration, symlinks, the boundary registry, identity revalidation, and failed-inventory handling. Independently ran all 10 tests in tests/test_host_inventory_stability.py successfully, including recognized-directory churn, unreadable inputs, and a fresh quiescent rerun. GitHub CI is green at this commit.

The fix removes cache contents from traversal; it does not eliminate races involving creation/removal of entries in a parent directory that is still inventoried. That remaining fail-closed behavior is consistent with this patch's scope. This is a code-review assessment, not a merge or release authorization.

PR #671, opened 2026-09-11
feat(check): compare against the detected base by default (#649)

### LINKED ISSUE #649: `check` defaults to the merge-base against the worktree; `--base` alone is accepted
## Problem

On a branch with a committed change to `.claude/settings.json`, `.mcp.json` and `.github/workflows/ci.yml`, `shipgate check` returns `decision: allow`, `changed_files: []`, "No recognized coding-agent boundary change". `check --base main` exits 2 with "`--base` and `--head` must be provided together". Only `check --base main --head HEAD` finds the five violations. Nothing in `--help` or the output suggests the flags. Every CI tool a reviewer knows compares against the merge-base by default.

## Proposal

- `--base` defaults to the merge-base between HEAD and the repository's default branch (resolved from `origin/HEAD`, then `main`/`master`, then the manifest `ci.base_ref` if present); `--head` defaults to the worktree. The chosen refs are printed in the first line of output.
- `--base X` alone means "X against the worktree"; `--head Y` alone means "default base against Y".
- When no default branch can be resolved, `check` says so in one line and names the flag, instead of returning `allow`.
- The worktree-vs-HEAD comparison (today's implicit behaviour) stays available as `--base HEAD`.

## Acceptance

- [ ] S1 scenario: `shipgate check` on the branch returns `block`/`critical` with the four widenings named, no flags.
- [ ] `check --base main` exits 0 and compares against the worktree.
- [ ] A repository with no resolvable default branch prints the one-line explanation and exits 2, never `allow`.
- [ ] Existing hook callers (`install-hooks`) and the Action pass their explicit refs and are unaffected; goldens unchanged.
- [ ] `docs/quickstart.md` and the skills drop the explicit `--base/--head` from the first command.

Parent #644. Related #348 (effective worktree diff).

pengfei-threemoonslab:
Closes #649.

## The defect

`shipgate check` compared the working tree with `HEAD`. On a branch whose changes were already committed it answered:

```
decision: allow    changed_files: []    "No recognized coding-agent boundary change"
```

It was truthfully reporting *"nothing is uncommitted"* to someone asking *"what does this branch change"*. The real answer needed `--base main --head HEAD` — a pair nothing in `--help` suggested, and which the CLI rejected unless **both** were given.

Measured on a repository that widens `.claude/settings.json` to `Bash(*)`, adds an MCP server, and gives its workflow `contents: write` + `pull_request_target`, all committed:

| | before | after |
| --- | --- | --- |
| `shipgate check` (no flags) | `allow`, 0 changed files | `block`/`critical`, 3 changed files, **5 violations** |

## The ladder

| invocation | compares |
| --- | --- |
| *(no flags)* | detected base's merge base → working tree |
| `--base <ref>` | that ref's merge base → working tree |
| `--head <ref>` | detected base → that ref |
| `--base A --head B` | `A...B`, unchanged |
| `--base HEAD` | uncommitted only — the previous default, by name |
| on the default branch | working tree, unchanged and complete |
| no base detectable | **exit 2** naming `--base`, never `allow` |

`subject.base` now names the ref actually compared. A verdict whose base is unnamed cannot be reviewed.

## Two places the issue's sketch was wrong

**It proposed falling back to local `main`/`master`.** `detect_default_base` refuses those *on purpose* — a remote is the authority a local branch might be stale against. Rather than override that reasoning, I narrowed the fallback to the case it does not cover: a repository with **no remote at all**, where the local branch is the only base in existence. Opt-in, so `verify` is untouched. A test pins that a remote still wins over a locally-ahead `main`.

**It assumed "no base detected" means "cannot be determined".** That detector also returns nothing when HEAD *is* the default branch — the commonest situation there is — which would have turned an ordinary `check` on `main` into an error. The two are now separated: on the default branch the working-tree answer is complete; only a genuinely undetectable base stops.

## A behaviour change worth reviewing

The issue assumed hook and skill callers pass explicit refs. **They do not** — `adoption-kits/*/SKILL.md`, `docs/agents/*.md` and the plugin all invoke `shipgate check --agent … --workspace . --format agent-boundary-json` with no refs. Those calls now get branch scope instead of session scope.

I believe that is right, and it is the fix for a known complaint: the 2026-07-26 in-loop review recorded that *"`git commit` silently clears check (worktree-vs-HEAD diff)"*. An agent that commits mid-session no longer loses the finding. But it is a real change to the in-session loop's noise profile, so it should be a conscious call rather than a side effect — if you would rather the local guard keep session scope, the fix is `--base HEAD` in the rendered hook, and I will make that change.

## Also touched, for consistency

`verify_command_for` required both refs or neither — a mirror of the CLI rule this removes. It now emits whichever ref was resolved, since `verify` accepts either alone. Leaving it would have made the emitted command disagree with the check that emitted it, which is this repository's recurring second-implementation defect.

## Evidence

12 cases in `tests/test_check_default_comparison.py`. **Negative control:** against pre-fix `main`, 8 fail; the 4 that pass are the default-branch, empty-ref and option-smuggling cases that must not change.

`test_codex_check_rejects_one_sided_git_refs` pinned the removed rule. It now pins the new one and, separately, the empty-ref refusal it was also covering. Two line-pinned entries in the static-only allowlist moved with the code they point at.

Locally green: everything matching `check|boundary|agent_mode|agent_result|trigger|control|invocation|verify|hook|static_only`, plus the surface-contract and parity suites and `ruff check`.

### DIFF
--- a/src/agents_shipgate/cli/agent_result.py
+++ b/src/agents_shipgate/cli/agent_result.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import json
 import os
 import stat
+from dataclasses import replace
 from pathlib import Path, PurePosixPath
 from typing import Any
 
@@ -77,21 +78,25 @@ def build_codex_agent_result(
     """
 
     return freeze_codex_boundary_result(
-        _assessment_for_diff(
-            agent=agent,
-            workspace=workspace,
-            diff_text=diff_text,
-            config=config,
-            policy=policy,
-            input_mode=input_mode,
-            input_issues=input_issues,
-            base=base,
-            head=head,
-            verification_replayable=verification_replayable,
-            base_manifest_absent=base_manifest_absent,
-            requested_workspace=requested_workspace,
-            changed_files_override=changed_files_override,
-            manifest_text_snapshot=manifest_text_snapshot,
+        _with_compared_refs(
+            _assessment_for_diff(
+                agent=agent,
+                workspace=workspace,
+                diff_text=diff_text,
+                config=config,
+                policy=policy,
+                input_mode=input_mode,
+                input_issues=input_issues,
+                base=base,
+                head=head,
+                verification_replayable=verification_replayable,
+                base_manifest_absent=base_manifest_absent,
+                requested_workspace=requested_workspace,
+                changed_files_override=changed_files_override,
+                manifest_text_snapshot=manifest_text_snapshot,
+            ),
+            base,
+            head,
         ).legacy_result
     )
 
@@ -129,7 +134,36 @@ def build_agent_boundary_result(
         changed_files_override=changed_files_override,
         manifest_text_snapshot=manifest_text_snapshot,
     )
-    return project_agent_boundary_result(assessment)
+    return project_agent_boundary_result(_with_compared_refs(assessment, base, head))
+
+
+def _with_compared_refs(assessment, base: str | None, head: str | None):
+    """Name the refs the check compared, on the assessment it produced.
+
+    `check` resolves an omitted base to the default branch's merge base
+    (#649), so which question was answered is no longer evident from the
+    request. `AgentResultSubject` already carries `base`/`head` and nothing
+    ever filled them. Set only when a ref was actually resolved, so a plain
+    worktree check is byte-identical to before.
+
+    Both projections read `assessment.legacy_result`, so replacing the
+    subject there reaches the current and the frozen v2 contract from one
+    place rather than two.
+    """
+
+    if base is None and head is None:
+        return assessment
+    legacy = assessment.legacy_result
+    return replace(
+        assessment,
+        legacy_result=legacy.model_copy(
+            update={
+                "subject": legacy.subject.model_copy(
+                    update={"base": base, "head": head}
+                )
+            }
+        ),
+    )
 
 
 def _assessment_for_diff(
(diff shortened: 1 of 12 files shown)


src/agents_shipgate/cli/agent_result.py
```diff
@@ -527,6 +561,119 @@ def _declared_codex_marketplace_roots(
...
+    from agents_shipgate.cli.verify.git import (
+        LOCAL_BASE_CANDIDATES,
+        REMOTE_BASE_CANDIDATES,
+        commit_sha,
+    )
+
+    head_sha = commit_sha(workspace, "HEAD")
+    if head_sha is None:
+        return False
+    for candidate in (*REMOTE_BASE_CANDIDATES, *LOCAL_BASE_CANDIDATES):
```
pengfei-threemoonslab: [P1] Preserve the detector's authority when deciding that HEAD is the default

This fallback accepts any local main/master at HEAD even when a remote is configured and none of its tracking refs is available. I reproduced a repository with local main, an origin remote, and no origin/main or origin/HEAD: the detector correctly returns no base, but this helper converts that into `--base HEAD`; check exits 0 with `decision=allow`, `control.state=complete`, and every permission true. Committed changes are therefore omitted in exactly the unavailable-base case the new contract promises to refuse. It also has the reverse problem: origin/HEAD -> origin/trunk at HEAD is recognized by the detector but absent from this hard-coded candidate list, so an ordinary check there exits 2. Carry the selected default ref and the reason for no comparison out of one detector; only take the HEAD fallback when that authoritative default actually equals HEAD, preserving the no-local-fallback-with-a-remote rule. Add both fixtures.

src/agents_shipgate/cli/agent_result.py
```diff
@@ -527,6 +561,119 @@ def _declared_codex_marketplace_roots(
...
+    """
+
+    from agents_shipgate.cli.verify.git import detect_default_base, merge_base_sha
+
+    if base and head:
+        return base, head, f"{base}...{head}", "HEAD"
+
+    def default_base() -> str:
+        detected = detect_default_base(
+            workspace, "HEAD", allow_local_when_no_remote=True
```
pengfei-threemoonslab: [P2] Detect the base relative to the requested head

For `--head <ref>` this call always passes the checkout's HEAD to detect_default_base, which intentionally excludes a candidate equal to that commit. Reproduction: create main, commit a widened .claude/settings.json on feature, check out main, then run `check --head feature --format agent-boundary-json`. It exits 2 with "No base ref could be detected", even though main...feature is available and the explicit `--base main --head feature` comparison works. The added test only exercises `--head HEAD` while already on the feature branch, so it misses this. Resolve the default against the requested head and cover an off-checkout head plus the requested head equal to the default.

pengfei-threemoonslab (commented):
Reviewed commit 99eb426690e498eb2e1dfa786662d7bf3b516482. Recommendation: request changes for the two findings below.

The branch-wide default fixes the reported clean-feature-branch case, and all 12 tests in tests/test_check_default_comparison.py passed independently. Additional real-Git fixtures exposed gaps in default-base resolution: an unavailable remote base can still yield `control.state=complete`, and `--head feature` can reject an available base while checked out on main. The same default-branch fallback also rejects a correctly resolved origin/trunk at HEAD (details inline).

All three CI suites and the other completed jobs are now green; coverage is still running at the last check. This review is posted as COMMENT because the authenticated account owns the PR and GitHub disallows requesting changes on one's own PR. The technical recommendation is to fix and re-review, not to merge.

pengfei-threemoonslab (commented):
Re-reviewed 7f034819 after resolving the main conflict. Both earlier findings are addressed: the shared resolver distinguishes an authoritative base at HEAD from unavailable remote refs, and --head uses the requested comparison head. Added regression coverage for remote-without-refs, custom origin/HEAD, alternate explicit heads, and equal-head precedence. No remaining actionable findings from this review.

Validation: full non-performance suite run; seven subprocess-interpreter environment failures passed on rerun with Python 3.12 on PATH. Separate performance tests and Ruff pass. Exact committed-head verification and a fresh agent-control read report complete/passed with merge permission. Fresh GitHub CI is running; merge remains conditional on its success.

PR #672, opened 2026-09-11
fix(control): make every emitted next action lead somewhere (#650)

### LINKED ISSUE #650: Every emitted next_action chain terminates: property test over all samples; fix the verify → init → audit loop on host-only repositories
## Problem

On a host-only repository: `verify --base main --preview` prints "Agents Shipgate verify: failed", exits 0, and says "Next command: init --write --json". `init --write` replies "Host-only discovery needs no shipgate.yaml; no setup files were written. Next: audit --host --json". `audit --host` ends with "Next: verify --preview --json". Four commands, one circle. `next_action` is the product's central promise to a coding agent; a next action that points at a command which refuses is worse than none.

## Proposal

- A property test: for every sample under `samples/`, every fixture, and the three first-run scenarios, follow `control.next_action.command` from each of `check`, `verify --preview`, `init`, `audit --host`, `scan`; the chain must reach a terminal state (`complete`, `human_review_required` with a named human step, or `review_publishable`) in <= 3 steps, and no step may return a refusal or repeat an earlier command with identical arguments.
- Host-only repositories: `verify --preview` routes directly to `audit --host --save-baseline` on the base ref followed by `audit --host --drift`, or (after #651) to `shipgate diff`; `init` on a host-only repository emits the same route, never `verify --preview`.
- "failed" is never printed with exit 0; the preview says `not_evaluated` with the reason.

## Acceptance

- [ ] The termination property test exists, runs in CI, and is green.
- [ ] S1 scenario: from `verify --preview`, the chain reaches a diff/drift output in <= 2 steps.
- [ ] `verify --preview` never prints "failed" on exit 0.
- [ ] The test is documented in `docs/agent-contract-current.md` as a guarantee of the control envelope.

Parent #644. Related #322 (second-implementation class), #323, #575.

pengfei-threemoonslab:
Closes #650.

## The reported cycle

On a repository with recognized host configuration and no manifest, following the tool's own advice never ends:

```
verify --preview  ->  "Next: init --write"
init --write      ->  refuses; a host-only repo needs no manifest, "Next: audit --host"
audit --host      ->  "Next: verify --preview"            <- back to the start
```

`control.next_action` is what a coding agent follows. A next action that points at a command which refuses, or back at the command that emitted it, is worse than none.

## The second defect, found by writing the test

The host-audit footer emitted `agents-shipgate verify --preview --json` **with no `--workspace`**. Walking the chain therefore did not loop — it *left*:

```
1. audit --host --workspace <repo>   ->  verify --preview --json        (no workspace!)
2. verify --preview --json           ->  ... mainwt/shipgate.yaml       (a different repository)
3. terminal
```

That is the invocation policy's own prohibition — *"Control commands are commonly executed after a tool/cwd transition; retaining a relative spelling would silently retarget the authorized operation"* — emitted by the tool itself. Every emitted command now carries the workspace it was produced for.

## The fixes

- **Conditional footer.** The baseline comparison where there is no manifest, `verify` where there is, `--drift` once a baseline exists. It names the **base ref** deliberately: recording a baseline from the changed checkout acknowledges the change under review rather than comparing it — the trap recorded in the 2026-09-04 pilot notes.
- **Workspace carried** on every emitted command.
- **`not evaluated`, not `failed`.** "failed" was the catch-all whenever no release decision existed, so a preview with nothing to gate printed `Agents Shipgate verify: failed` directly above `Exit code: 0` while its own fields said `execution: "not_run"` and `applicability: "not_evaluated"`. Only the human line disagreed.

## The guard, and three passes to stop it being vacuous

`tests/test_next_action_chains_terminate.py` walks every entry command on a host-only and a manifest repository, failing on a repeat, a lost workspace, or a step this CLI cannot run. Getting it to actually detect the defect took three corrections, each now recorded in its comments:

1. **Anchored the `Next:` reader to end-of-line.** The footer reads ``Next: `… --json` for release gating.`` — the text continues past the backtick, so the reader saw no command and called it a terminal.
2. **Keyed a step on raw argv.** `audit --host` and `audit --host --json` became different nodes, so a loop through both looked finite. The key is now the question asked, with presentation flags stripped.
3. **Only asserted termination.** The real chain terminated legitimately — in the wrong repository. The workspace assertion is what finally reproduced it.

**Negative control:** replayed against pre-fix `main`, both `audit --host` chain cases fail, plus the two verdict-word cases. Before correction 3 the chain cases passed on `main`, which is the definition of a guard that is not one.

## Acceptance

- [x] The termination property test exists, runs in CI, and is green.
- [x] From `verify --preview`, the chain reaches the host route in 2 steps (`verify` → `init` → `audit`), pinned separately.
- [x] `verify --preview` never prints "failed" on exit 0.
- [x] Documented in `docs/agent-contract-current.md` as a guarantee of the control envelope: no cycle, no workspace loss, every step runnable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,25 @@
 
 ## Unreleased
 
+- Make every emitted next action lead somewhere, and prove it. Following
+  `control.next_action` on a repository with recognized host configuration
+  and no manifest went round three commands forever: `verify --preview`
+  named `init --write`, `init` refused because a host-only repository needs
+  no manifest and named `audit --host`, and `audit --host` named
+  `verify --preview` again. The host-audit footer is now conditional — the
+  baseline comparison where there is no manifest, `verify` where there is,
+  `--drift` once a baseline exists — and it says to record the baseline on
+  the base ref, never on the changed checkout, which would accept the change
+  instead of comparing it. That footer also emitted its command without
+  `--workspace`, so following it ran the next step against the caller's
+  current directory rather than the repository just audited; every emitted
+  command now carries the workspace it was produced for. A preview that
+  evaluated no gate reports `Agents Shipgate verify: not evaluated` instead
+  of `failed`, matching the `not_run` its machine fields already carried.
+  `tests/test_next_action_chains_terminate.py` walks every entry command on
+  two repository shapes and fails on a repeat, a lost workspace, or a step
+  this CLI cannot run (#650).
+
 - Read every counted hunk row, so header-shaped content stops being lost.
   Hunk state and the header's declared row counts, not a line's spelling,
   decide what the shared unified-diff parser treats as content: a removed
--- a/docs/agent-contract-current.md
+++ b/docs/agent-contract-current.md
@@ -94,6 +94,32 @@ well. In verifier, verify-run, and preview output, follow
 only `control.next_action` and `control.allowed_next_commands`. See the
 [embedded-trigger migration note](../STABILITY.md#migration-note-unreleased-embedded-trigger-routing).
 
+### Emitted next actions terminate
+
+Following `control.next_action.command` — or the human `Next:` line, which is
+the same routing rendered for a person — reaches a terminal. Three properties
+hold, and `tests/test_next_action_chains_terminate.py` walks every entry
+command on a host-only and a manifest repository to keep them holding:
+
+- **No cycle.** No chain returns to a question it already asked. Output
+  format is not part of a step's identity: `audit --host` and
+  `audit --host --json` ask the same thing, and treating them as different
+  steps is what once hid a three-command loop
+  (`verify --preview` → `init --write` → `audit --host` → `verify --preview`).
+- **No workspace loss.** Every emitted command names the same `--workspace`
+  it was produced for. A step that drops it does not continue the walk; it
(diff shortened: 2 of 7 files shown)


pengfei-threemoonslab (commented):
Review: changes needed before merge.

The host-only cycle fix is useful, but `_audit_next_step` drops `--scope` and a custom `--baseline-file` from its drift command. It also interpolates the workspace without shell quoting and hardcodes the entry point. An audit of a path containing spaces, or a local-static inventory with a custom baseline, therefore routes to a different or invalid request. The no-manifest advice also omits the workspace when naming the baseline command.

I am fixing these routes to preserve the audited target, scope, baseline and current CLI invocation, with regression coverage. The full test suite will be rerun before merge.

pengfei-threemoonslab (commented):
Re-reviewed the complete audit follow-up routes. Commands now preserve the workspace, custom baseline, scope, shell quoting, and current CLI entry point; baseline creation remains an explicit human decision. The earlier review findings are addressed. Full tests completed with the seven interpreter-dependent cases passing on rerun; performance tests and Ruff pass. The integrated comparison/routing/host-audit checks pass (96 tests). Exact-head verification is complete/passed. No remaining actionable findings; waiting for fresh CI before merge.

Issue #657, opened 2026-09-11
Permission-width lattice and precedence resolution for host configs; calibrate `Read(**)`; narrowing never reports `expanded`


pengfei-threemoonslab:
## Problem

`_is_wildcard_allow` recognizes only wildcard *shapes* (`*`, `Bash(*)`, `Bash(*:*)`); "expanded" is the set-difference of rule strings. Consequences measured on 2026-09-10: `Read(**)` is reported `risk: critical, access: admin`; `Bash(pytest *)` is `high`; two pre-existing MCP servers are `high` on an unchanged baseline (4 of 8 grants are noise); replacing `Bash(npm *)` with `Bash(npm test:*)` — a narrowing — fires allow-expanded. User/project/local settings precedence is refused (`unresolved_precedence`) rather than resolved.

## Proposal

- Per host, a pattern lattice: `Bash(*)` ⊃ `Bash(pytest *)` ⊃ `Bash(pytest tests/*)`; `Read(**)` ⊃ `Read(src/**)`; MCP server entries compared by identity then by command/url/env keys. Direction = comparison in the lattice, not string inequality.
- Severity by tool class and width: `Read(**)` on a code repository is `low` unless the repository marks sensitive paths; `Bash(*)` stays `critical`; `WebFetch(*)` `high`.
- Precedence resolved as user < project < local with the effective layer named in the row; a genuinely ambiguous merge still fails closed but says which two rules conflict.
- Fixture set: 20 widen/narrow/unchanged pairs per host, with the expected direction.

## Acceptance

- [ ] Zero false `expanded` rows on the narrowing fixtures; zero missed widenings on the widening fixtures.
- [ ] S1 without a baseline: `audit --host` noise 4/8 -> 0/8; `Read(**)` no longer `critical/admin` by shape alone.
- [ ] Precedence cases print the effective layer; the fail-closed case names the conflicting rules.
- [ ] `policies/host-boundary.shipgate.yaml` severities updated with a rationale line each.

Parent #645. Related #545, #516, #538.

pengfei-threemoonslab:
Splitting this issue's two halves across two PRs, so the second is not silently dropped.

**PR #678 — the lattice.** Covers three of the four acceptance items:

- [x] Zero false `expanded` rows on the narrowing fixtures; zero missed widenings on the widening fixtures.
[...]

Issue #655, opened 2026-09-11
`shipgate diff` v1 across Route A: manifest-free base synthesis; the capability delta is never `unavailable` on first adoption; verdict only as a footer


pengfei-threemoonslab:
## Selected execution — 2026-09-24

Days 1–5 design is in PR #869. Head selection may seed base discovery but cannot prove old coverage or copy head inventory into base. Generated input origin must survive all evidence identities and must never become a reviewed verifier base or release authority. Refine the absolute never-unavailable target: emit established structural changes plus precise unresolved scope; unsupported/ambiguous input can legitimately remain incomplete. No reader or runtime implementation is claimed by this design. Selected after #580 in milestone 13.

---

The earlier advisory-release deferral below is historical. This issue is now selected for the application proof milestone; its implementation is not yet complete.

## First advisory release scope — 2026-09-11

Owner decision: **adoption > completeness; make v1.0 ready with repository-independent work, then expand adoption.** Synthesizing a tool-source base is not required for the selected repository-declared host diff. Preserve provenance and absence semantics when this is revisited.

Removed from the selected v1.0 milestone; kept open and deferred, not resolved. #643/#644/#645/#646 define current product acceptance; #572 preserves the historical qualified no-go and #648 owns the actual publication contract. The prior evidence and acceptance below remain a record of the deferred work, not newly satisfied conditions.

---

## Problem

On a FastMCP server repository, after `init --write` and three hand-filled fields, `verify --base main` still reports `Capability delta (analysed surface): unavailable` / `Base status: missing_manifest` because the base commit has no `shipgate.yaml`. The pre-existing `delete_item` and the newly added `purge_account` are reported identically. The one question a PR gate exists to answer — what did this PR add — is unanswered on the first PR of every adoption.

## Proposal

- When the base tree has no manifest, synthesize the base surface by applying the head manifest's `tool_sources` (and, absent a manifest, the discovery result) to the base tree; record `base_surface_origin: synthesized_from_head_manifest` in the receipt so the provenance is explicit and reproducible.
- `shipgate diff` extends the #651 table to tool-source subjects: `added tool purge_account — destructive (keyword: purge), no approval/confirmation declared — critical`; `modified stripe.create_refund — scope broadened …`; unchanged subjects are not printed.
- Rows carry the evidence basis (`reviewed_declaration`, `inferred_keyword`, `protocol_default`) so a reviewer sees what was read versus inferred.
[...]

Issue #654, opened 2026-09-11
Adoption freeze until M2 exit: no new check IDs or schema families; PR size budget of 800 added lines


pengfei-threemoonslab:
## Problem

In the last 30 days 130 PRs merged (median +1,139 lines, +372k lines total) and 36 issues were filed in the three days after 2026-09-08 alone; 27 of 63 open issues begin "found while implementing #N; deferred". The engine has 144 check IDs, 159 schema files in `docs/`, report schema 0.43, and nine functions over 400 lines. New surface is being produced faster than any user could evaluate it, and every review round lands a defect in the same four functions (#496).

## Proposal

- Until the M2 exit (#645) is recorded: no new `SHIP-*`/`SEC-*`/`LINT-*` check ID, no new `shipgate.*` schema family, no new adapter. Bug fixes, renderers, readers for existing families, and the items on #644/#645 are in scope.
- PR budget: <= 800 added lines excluding goldens and generated schemas; larger changes split or justified in the PR body with a named reviewer decision.
- A CI check reads the diff and fails on a new check ID or schema family without a `freeze-exception` label; the budget is a warning, not a failure, for 30 days, then a failure.

## Acceptance

- [ ] CONTRIBUTING.md § Surface discipline records the freeze window and budget with their end condition.
- [ ] The CI check exists and is green on `main`.
- [ ] At M2 exit, the freeze is lifted or extended by a recorded decision in #643.

Parent #644. Related #492, #495, #496, #493.

pengfei-threemoonslab:
Not planned. [PR #675](https://github.com/ThreeMoonsLab/agents-shipgate/pull/675), the implementation of this issue, was closed unmerged by the repository owner.

The reason is the operating rule, not a defect in the implementation: **"adoption > completeness"**, and **"v1.0 contains only what this repository can achieve on its own"** (#643). A freeze is a constraint on work rather than a delivery — it advances no v1.0 or M2 exit criterion — and short-term it is more likely to slow the milestone than to protect it.

Closing the issue as well as the PR, rather than leaving it open, so this does not become phantom work someone re-implements. The analysis below is worth keeping, which is why it stays recorded here rather than being deleted.

**What the measurement said, then and now:**

| | #654 (2026-08-12ish) | 2026-09-11 |
|---|---:|---:|
| check IDs in `src/` | 144 | 162 |
| schema files / families in `docs/` | 159 / — | 166 / 43 |
[...]

Issue #652, opened 2026-09-11
Human-language output surfaces: vocabulary table, `--help` shows six commands, a text format for `check`, un-escaped PR-comment Markdown, an outsider summary atop ROADMAP


pengfei-threemoonslab:
## Problem

stdout speaks engine: `insufficient_evidence`, `control.state`, `binding graph`, `Route H`, `input_set_id`, exclusion ledger. Buyers say allowlist, permission, MCP server, workflow permission. `--help` shows 3 of 53 commands while README leads with hidden ones (`fixture`, `init`, `scan`, `doctor`). `check` has no human text format. The generated PR comment escapes Markdown (`stripe\.create\_refund`, `\(`) and repeats the same recommendation in `violations`, `suggested_fixes`, `agent_repair_instructions` and `repair.instructions`. ROADMAP.md is 483 lines of release obligations; a stranger cannot tell whether the project is alive or what to run.

## Proposal

- A vocabulary table in `docs/` mapping engine terms to user terms, enforced on stdout headlines: `insufficient_evidence` -> "Not enough was read to decide — here is what was read"; `human_review_required` -> "Needs a person: <named step>"; `agent_action_required` -> "Next: <command>".
- `--help` shows `diff`, `check`, `verify`, `audit`, `init`, `doctor`; everything else stays available and documented under `--help-all`.
- `check --format text` (default in a TTY) reuses the #651 renderer.
- PR comment: stop escaping GitHub-flavoured Markdown; render each recommendation once.
- ROADMAP.md gains a 30-line "Where this project is, what to run, what is promised" block above the release record.

## Acceptance

- [ ] A guard test asserts that no headline or `control.reason` string contains engine vocabulary — `insufficient_evidence`, `control.state`, `binding graph`, `Route H`, `input_set_id`, `exclusion ledger` — across every sample and the three scenarios.
- [ ] `agents-shipgate --help` lists exactly the six commands; `--help-all` lists all.
- [ ] Generated `pr-comment.md` on `ai_generated_refund_pr` contains no backslash-escaped punctuation; recommendations appear once per finding.
- [ ] Surface-parity and positioning tests updated, not weakened.

Parent #644. Related #498, #329, #463.

The cold-reader test (five readers naming the widenings unaided) was an acceptance bullet here and is **deferred past v1.0** by owner ruling — recruiting readers means scheduling other people. #653 records the same recognition in the field, on the reviewer's own repository.

PR #673, opened 2026-09-11
feat(diff): show what a change does to the agent's authority (#651)

### LINKED ISSUE #651: `shipgate diff` v0 on the host route: before → after, direction, why, severity; `--base` synthesizes the baseline from the base tree
## Problem

`audit --host --drift` is the best moment in the product (0.8 s, 12 lines, four widenings, zero noise) and the least reachable: it needs `--save-baseline` on the base commit first, two checkouts, and the baseline must be reachable from the PR branch. Its output says only *what* changed: no why, no severity in the Markdown, and the only instruction is "re-record", which is acceptance, not a fix. `check --base --head` answers why and severity but in 449 lines of JSON with recommendations repeated four times.

## Proposal

`shipgate diff [--base <ref>] [--head <ref>]` (defaults per #649) on the host route:

- Synthesizes the baseline by running the host readers on the base tree; no committed `host-grants.json` required (it stays supported as an alternative source).
- Prints one table, human text first:

```
Host capability diff  base a1b2c3 -> head d4e5f6   (3 widened, 0 narrowed, 0 unchanged-risk)

subject                                before               after        direction  why it matters                                  severity
claude-code permissions.allow          Bash(pytest *)       Bash(*)      widened    any shell command runs without a prompt          high
claude-code mcpServers                 —                    postgres     added      new data authority; env DATABASE_URL             high
.github/workflows/ci.yml permissions   contents: read       write        widened    workflow can push; pull_request_target added     critical
```

- `--json` emits the same rows from the existing `capability_delta` / host-boundary facts (one projection, no second decision).
- Verdict, control envelope and exit code appear only as a footer, and only when a manifest or baseline exists; without them the footer says "advisory: no policy configured".
- Benign PRs print `no capability or permission change` and nothing else.

## Acceptance

- [ ] S1 scenario: `shipgate diff` from the branch prints the three rows above with direction and severity; five-question score >= 4/5 (what, why, severity, how to fix, is the fix safe).
- [ ] A benign PR (docs, code outside agent config) prints the one-line no-change message.
- [ ] `--json` rows are byte-derived from existing facts; goldens for the S1 fixture committed.
- [ ] `check`'s human text format (#652) reuses the same renderer.

## Out of scope

Route A (manifest / tool sources) — #655. Permission subsumption — #657.

Parent #644. Related #364 (subject rollup), #433, #468.

pengfei-threemoonslab:
Closes #651. **Stacked on #671** — base is `claude/issue-649-merge-base`, because `diff` shares that PR's base-ref resolver rather than carrying a second copy. Retarget to `main` once #671 merges.

## Why

The first-run measurement that opened this milestone took sixteen commands to reach its first actionable output. The single moment that worked was `audit --host --drift` — four widenings named in twelve lines, zero noise — and it was also the least reachable thing in the product: it needed a baseline recorded in advance, on the base ref, reachable from the branch. Two checkouts and a committed file before the first answer.

## What it does

```
Agent capability diff  main (17a958ba) -> working tree

⚠ critical  added    claude-code .claude/settings.json
                     Bash(*)
                     matches any command of this kind, without a prompt
⚠ critical  widened  github .github/workflows/ci.yml
                     read → write, contents: write, pull_request_target
                     runs with repository secrets on fork pull requests; the workflow can write to the repository
  high      removed  claude-code .claude/settings.json
                     Bash(pytest *) → gone
                     a permission the agent previously had here
⚠ high      added    claude-code .mcp.json
                     postgres
                     a new MCP tool surface the agent may call

5 change(s), 4 widening what the agent may do (⚠).
Static configuration only: this is what the files permit, not what the agent did. No verdict is implied.
```

No manifest. No committed baseline. No second checkout. A benign change prints `No change to what the agent may do.`

## Surface-discipline gate, answered

1. **Which headline metric?** Time-to-first-value. Sixteen commands to two is the milestone's exit criterion and this is the command it turns on.
2. **Could an existing surface carry it?** The *decision* logic is entirely reused — it adds no comparator, no severity model, no verdict. `audit --host --base <ref>` was the smaller alternative and I rejected it on one ground: `--help` is where a stranger finds the product's prominent commands, and a flag on a subcommand cannot be one. Discoverability is the metric being moved.
3. **Non-goals?** Respected. No second verdict, no new adapter, no network, no LLM.

## It is a projection, and the tests hold it to that

`risk` is the engine's severity. `expansion_signals` is the engine's word on widening. Neither is recomputed:

- a wildcard addition the engine did **not** signal stays `added`, not promoted to an expansion — pinned with a `low`-risk wildcard read the module would never rate that way itself;
- a change present on both sides stays `changed` until the engine calls it widening, because guessing *narrowing* needs the pattern lattice in #657;
- `expands` is a separate field from `direction` deliberately: presence is a fact both sides show, widening is a judgement only the engine may make.

## Registry

Registered in `docs/distribution-surfaces.md` under **rule 2** — a row with no claims and a note saying why: it restates none of the engine's answers, so there is nothing to drift. The parity test carries the matching empty-claim entry, so neither can be edited alone.

**Deliberately deferred:** contract `primary_commands` registration and `--help` prominence. Both belong to #652, which owns the six-command help; doing it there bumps the contract once instead of twice, which matters during the #654 freeze.

## Scope

Host route only. Tool-source subjects (`shipgate diff` over a manifest's tools) are #655, as the issue sequences them.

Twelve cases in `tests/test_capability_diff.py`.


pengfei-threemoonslab (commented):
Review: changes needed before merge.

1. The row projection suppresses all removed-grant expansion signals. Removing a deny or ask rule is explicitly an expansion in `host_grant_expansion_signals`, but this command displays it without the warning and describes it as a removed permission. Source-keyed signals such as hooks can also be lost by the signal-string parser.
2. An incomplete base inventory raises from `build_host_grants_baseline`; it never reaches the command's advertised incomparable-result output.
3. Base and head inventories use different roots when `--workspace` names a checkout subdirectory. Both must inspect the same repository scope. Missing default refs should also require an explicit comparison instead of silently reducing the question to uncommitted changes.
4. CI fails because the new workspace-taking command was omitted from the mandatory workspace-guard sweep.

I am addressing these issues, including focused regression tests, before rerunning the full suite and reconsidering merge.

pengfei-threemoonslab (commented):
Re-reviewed e6c158bb966d873be1bd9c614c8fd456f9a12563. The row projection now preserves deny/ask removals and source-keyed expansion signals using the engine’s own per-change mapping. Incomplete base inventories produce an explicit incomparable result, missing default refs require a named comparison, and subdirectory invocations compare the same repository root on both sides. The new command is covered by the full workspace-guard sweep. Full non-performance and separate performance suites pass; focused diff/workspace regression checks and Ruff pass. Exact-head verification and refreshed control report complete/passed. No remaining actionable findings; fresh CI is running before merge.

PR #667, opened 2026-09-11
docs(roadmap): add the roadmap to v1.0 as plan of record (#643)

pengfei-threemoonslab:
## What

Adds an **Adoption roadmap to v1.0** section to `ROADMAP.md`, placed ahead of the existing "Direction" release-qualification record (which is kept unchanged, including the 2026-09-08 no-go).

- v1.0 defined by external adoption: install, first plain-language diff within two commands, zero rows on benign PRs, repeat use across the next change — full exit criteria in #643.
- Two release lines: advisory pre-releases on a ≤14-day cadence (#648) and the qualified gate on evidence only.
- Three milestones with dated exits: #644 (M1 installable and navigable, to 2026-10-10), #645 (M2 diff first value, to 2026-11-10), #646 (M3 repeat use and team adoption, to 2026-12-10), each linked to its child issues #647–#666.
- Freeze rule until the M2 exit (#654): no new check ID, schema family or adapter.

## Why

The internal 2026-09-10 adoption study measured the cost of putting "make the gate trustworthy" before "let people see the diff": the published `v0.15.0` does not evaluate the host route, `check` without flags returns `allow` on a committed branch, the emitted next actions loop on host-only repositories, and the pilot ledger records zero attempts. The one verified value (host-grant drift on a baselined repository) is reachable only from the quickstart. This plan makes it the default experience.

## Checks

- `Latest release: \`v0.15.0\`` line untouched (pinned by `tests/test_public_surface_contract.py`).
- `tests/test_public_surface_contract.py` and `tests/test_distribution_surface_parity.py` pass locally.
- No forbidden name forms; no packet-schema mentions.

Refs #643, #644, #645, #646.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #648, opened 2026-09-11
Deliver the current advisory candidate and resolve its stable publication contract separately from qualified gates


pengfei-threemoonslab:
## Delivery disposition — 2026-09-14

The selected advisory publisher is implemented and ordinary v1.0.0 is published. [Release run 34902863526](https://github.com/ThreeMoonsLab/agents-shipgate/actions/runs/34902863526) published the exact wheel and signed advisory statement; [PR #777](https://github.com/ThreeMoonsLab/agents-shipgate/pull/777) updated the repository pins. The [post-publication record](https://github.com/ThreeMoonsLab/agents-shipgate/issues/648#issuecomment-5671726167) plus #777 completes this publication-delivery scope. **The literal post-publication hosted tag/Action observation remains owned by #570**; the local installed replay and pre-publication Action run are not relabeled as that observation. #781 owns the separately reproduced frozen optional-kit instructions inside the 1.0.0 wheel.

Product sequencing after publication is #778. #779 / #780 own post-release entry work; #653/#571 may start supported, advisory user research without waiting on qualification. #572's qualified obligations are unchanged. No external adopter, interview or independent human label is inferred from this closure.

## Original specification (historical)

The original acceptance and milestone record is retained below; the dated disposition above states what was delivered and which narrower exception or remaining obligation applies.

## Problem

Users need an installable current advisory workflow while qualified-gate evidence remains incomplete. A source-only fix does not satisfy first value. Release claims, version selection and the actual publisher must agree.

## Accepted publication scope — 2026-09-11

The owner's [latest decision](https://github.com/ThreeMoonsLab/agents-shipgate/issues/648#issuecomment-5640722248) supersedes the earlier preview proposal: publish ordinary **`v1.0.0` to PyPI when repository readiness holds**. #647 is closed; there is no outsider-facing prerelease or preview. Internal candidate builds remain the pre-publication validation vehicle. Qualified blocking claims retain their separate evidence obligations.

#676 delivered Amendment 4's separation of advisory and qualified cadence. A 14-day warning-only cadence target applies after the first advisory release, not as an extra first-release gate. The current `release.yml` still needs the explicit advisory publication contract; the owner decision is not an implemented publisher.

## Proposal

1. Build a current internal candidate from a CI-accepted source commit. Install the downloaded bytes in a clean environment; bind version, source, contract, digest and actual Action behavior. Reuse #570's distribution checks. No public preview dispatch is required.
[...]

Issue #647, opened 2026-09-11
Publish an installable advisory preview of the 0.16 engine (PEP 440 pre-release) and make the Action ref resolve


pengfei-threemoonslab:
## Problem

The newest published release is v0.15.0 (2026-07-08, contract 10). On a diff that widens `.claude/settings.json` to `Bash(*)` it returns `warn`/`none` and says "No Codex boundary rule fired". The evaluator that names the widening exists only in the untagged source tree (0.16.0, contract 33). README, quickstart and the pilot runbook describe the source tree. An outsider cannot install the product they read about. This is the single adoption blocker every other item depends on, and it is self-imposed: the v1.0 qualification bars (#512, #509, #510) are attached to the only tag that would carry the host evaluator.

## Proposal

- Publish `0.16.0rc1` (PEP 440 pre-release, so `pip install agents-shipgate` without `--pre` stays on 0.15.0 and the qualified `0.16.0` remains available) to PyPI and tag `v0.16.0rc1`.
- The README's install block shows the pre-release command first for the host route; the Action example pins the rc tag; `published_release.py` learns a distinct `LATEST_PREVIEW_VERSION` so the pin guards (#506) can express "latest installable" separately from "latest qualified".
- The release body states plainly: unqualified, advisory-only, no signing or rehearsal evidence.
- No release invariant is widened: `release_publication.build_manifest` still requires `tag == v<version>`; the rc is a real version.

## Acceptance

- [ ] `pipx install --pip-args=--pre agents-shipgate` (or the documented equivalent) installs 0.16.0rc1; `--version`, `doctor --json` and wheel METADATA agree (#505 guard).
- [ ] `check --base <main> --head HEAD` on the S1 scenario returns `block`/`critical` naming `Bash(*)`, `WebFetch(*)`, the added MCP server and the workflow permission change on the installed build.
- [ ] `uses: ThreeMoonsLab/agents-shipgate@v0.16.0rc1` resolves and the composite Action installs that exact version.
- [ ] `docs/design-partner-pilot-results.md` and `docs/quickstart.md` name the rc as the build to measure.
- [ ] Release published within 7 days of this issue opening.

## Out of scope

Signing, rehearsal, corpus qualification; those stay on the gate line (#648).

## Metric

Days from issue open to installable rc; `contract --json` on a clean `pipx install` >= 29.

Parent #644. Related #491, #506, #570, #572.

pengfei-threemoonslab:
[...]

PR #676, opened 2026-09-11
build(release): two lines, two promises, two cadences (#648)

pengfei-threemoonslab:
Related to #648.

## The problem

One release gate served two products. The advisory diff needs to be installable every couple of weeks to reach anyone; the blocking gate needs signing, rehearsals and a blind-labeled corpus before anyone should trust it to stop a merge. Tying them together produced **64 days without a tag while 130 pull requests merged**, and a README that has to warn readers the documented workflow is not in the published release.

## Two lines

| Line | Carries | Install | Promises | Cadence |
| **Advisory** | `diff`, `check`, `audit --host`, drift, advisory PR comments | unqualified preview pre-release | plain-language rows; **no blocking authority** | 14 days |
| **Qualified gate** | blocking verdicts, receipts, attestations | `v*` release tag | every bar in the evidence policy | on evidence only |

**Why an advisory pre-release needs no qualification.** Every bar in the evidence policy exists to justify one thing: a verdict that may stop a merge. The advisory line makes no such claim, so there is no authority for the qualification to underwrite. Requiring it would not make that line safer — it would only keep the engine out of reach, which is the state Amendment 2 was written to end.

**No gate-line bar moves.** The preview still sits outside `v*`, still carries a PEP 440 local segment a public index must refuse, still ships no qualification artifact, and still cannot satisfy `release_publication.build_manifest`. Recorded as **Amendment 4**.

## Measuring both

```
Release cadence: 65 days since v0.15.0 (2026-07-08) -- past 45 days; triage as a release defect (#491).
Advisory cadence: 8 days since preview-0.16.0+preview.20260903.gb61aca7 (2026-09-02) -- within the 14-day interval.
```

One renderer, two values, the label carried on the value. The separation is load-bearing rather than tidy: **counting previews in the release metric would let the channel that exists *because* the release cadence slipped report that cadence as kept.** That reasoning was already in `is_release_tag`'s docstring; this change is what makes it measurable rather than merely stated.

## Where I did not follow the issue

The issue asked for advisory staleness > 14 days to be **a red check on `main`**. I did not implement that. The reasoning already recorded for the release cadence applies unchanged, and is right: *"Going red here would fail whichever unrelated change happened to arrive after the interval lapsed, and that author is not the actor who can fix it."* Both lines warn; `--fail-when-overdue` now covers both, which is the red signal for an operator or a scheduled job. If you want a hard stop, the honest shape is a scheduled workflow that fails on its own, not a check attached to an innocent pull request — say the word and I will add it.

## Acceptance

- [x] Two-line table on README, ROADMAP (already landed in #667) and `docs/distribution.md`; surface-parity green.
- [x] Cadence distinguishes the lines.
- [x] Amendment 4 records why an advisory pre-release is admissible without qualification.
- [ ] **Two consecutive advisory pre-releases shipped on cadence** — cannot be met from a pull request. It needs a publish, which is #647 and yours to authorize.

Review scope: this PR establishes the two cadence measurements. The acceptance criterion requiring two consecutive published advisory releases remains unverified and must stay open in #648; merging source is not publication evidence.


pengfei-threemoonslab (commented):
Review found the advisory tag predicate accepts `preview-0.16.0` and arbitrary `+...` suffixes, although the preview publisher requires `+preview.<date>.g<sha>`. Such unrelated/malformed tags can reset the new advisory clock. I am tightening that predicate to the publisher’s format and adding coverage for both cadence boundaries, tag isolation, and the operator exit status. I am also preserving the existing top-level JSON fields for current script consumers while adding the advisory measurement.

The implementation does not demonstrate the two consecutive published advisory releases required by #648. This PR can establish the measurement, but should not claim that operational acceptance is complete.

pengfei-threemoonslab (commented):
Re-reviewed final head 4898bf0d4de26d271feccf5e520f6f1ad741730c after integration with current main. The advisory predicate validates the preview version/date/commit form and preserves legacy top-level release JSON fields. Lightweight tag creator dates are source commit dates, so advisory cadence now uses the embedded preview build date as an explicitly labelled offline proxy, never as claimed GitHub publication time. A real Git regression covers old source commits with newer preview builds. All 231 focused tests passed. The actual publication and scheduled failure acceptance work in #648 remains open.

Dependency-lock consistency checks, the release-pipeline and CLI/reader integration tests, Ruff, and all checks on this exact GitHub head passed. Current committed verification grants complete/merge permission. No remaining blocking findings in this PR.

Issue #658, opened 2026-09-11
MCP source reader precision: translated descriptions, source-level annotations, annotation contradiction on the source route, protocol defaults as coverage rows not findings


pengfei-threemoonslab:
## Problem

On `github/github-mcp-server` (114 tools): every description is wrapped in a translation helper (`Description: t("TOOL_ACTIONS_GET_DESCRIPTION", "Get details …")`, `pkg/github/actions.go:411`); the Go reader returns empty and emits 114 `SHIP-DOC-MISSING-DESCRIPTION`. `ReadOnlyHint: true` in source is not read, so 109/114 tools are projected `write` by protocol default. On a FastMCP server, `purge_account` declared `annotations={"readOnlyHint": True}` over a body that calls `_ITEMS.clear()` produced no finding: the #464 contradiction check runs only on the `tools.json` export route, and the tool was classified destructive by the `purge` keyword alone — rename it and it silently becomes `write`. Precision 1.0 on published findings is the precondition the MCP-channel strategy set for itself.

## Proposal

- Idiom registry additions: string-returning helper calls with a literal second argument (Go `t(key, default)`, TS `t()`), and source-level `annotations`/`ToolAnnotations` for Go, TS and Python FastMCP.
- Run the annotation-contradiction check on the source route with the same evidence basis rules as the export route.
- A protocol-default effect never produces a finding; it produces a coverage row ("effect not declared; treated as write for policy").
- Measure on ten top registry servers (github-mcp-server, mongodb-mcp-server, mcp-grafana, filesystem, postgres, …) with a per-server false-finding count committed as a fixture.

## Acceptance

- [ ] github-mcp-server: `SHIP-DOC-MISSING-DESCRIPTION` 114 -> 0; its literal `ReadOnlyHint`/`DestructiveHint` are read and challenged by `SHIP-MCP-ANNOTATION-CONTRADICTION`, never counted as effect evidence (amended by owner decision 2026-09-13; previously "read-only tools >= 90% projected `read`").
- [ ] S2 scenario: the `readOnlyHint` contradiction fires on `purge_account`. The renamed `process_account` staying quiet is a documented known limit, out of v1.0 scope (owner decision 2026-09-13; previously "renaming to `process_account` still fires").
- [ ] False-finding rate < 2% across the ten servers; the table is committed and CI-pinned.
- [ ] No new check ID (#654 was closed as not planned; reuse existing IDs to keep this repair bounded); changes are reader and rendering only.

[...]

PR #679, opened 2026-09-11
Read Go MCP descriptions from options and translated struct fields (#658)

pengfei-threemoonslab:
First increment of #658; the issue remains open for annotations and broader qualification.

The Go reader missed descriptions stored in option calls and translated struct fields. The latter is the actual shape reported in #658: `mcp.Tool{Description: t("KEY", "default")}`. The original PR added only option extraction, so it still returned no descriptions for all four registrations in the checked upstream Actions file.

The final change reads direct `WithDescription`/`WithToolDescription` options and direct struct `Description` fields. It accepts a whole string literal or a complete two-string call to a declared `TranslationHelperFunc` parameter. It does not infer fallback semantics from arbitrary function names or key-shaped arguments. Unbound, reassigned and shadowed helper parameters remain unread.

Option order follows the SDK: a later description replaces an earlier one, including empty or computed values. Nested unapplied options cannot document the parent tool, and a partial translated expression cannot stand in for its complete value. Go trailing commas and comments are supported. The package reader and zero-install detector run the same added source corpus.

Validation: 1,714 final integration tests passed with one skip; all four performance tests and Ruff passed. The new regression cases cover scope, expression boundaries, option precedence and translated struct fields. The pinned upstream [Actions source](https://github.com/github/github-mcp-server/blob/7d13a7ad6f2a17f351a6d77ce280c85ae1821f4d/pkg/github/actions.go) changes from 0/4 descriptions on the original PR code to 4/4 on the corrected reader. This is a representative source check, not a claim that the complete 114-tool acceptance criterion has been rerun.

Source annotations, contradiction detection and the ten-server qualification table are outside this increment. No idiom ID or diff token changes, so `IDIOM_REGISTRY_VERSION` is unchanged; the registry's published descriptions now state the supported reads.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,42 +2,17 @@
 
 ## Unreleased
 
-- Read a Go tool's description when it is an option, and when it is
-  translated. `mark3labs/mcp-go` registers a tool as a call carrying option
-  functions — `mcp.NewTool("get_job", mcp.WithDescription(...))` — and the
-  only Go description extractor looked for a `Description:` field in a
-  struct literal, which that shape does not have. So every tool registered
-  this way read `description=None`, plain literals included. On top of
-  that, `github/github-mcp-server` wraps all 114 of its descriptions in a
-  translation helper, `t("TOOL_GET_JOB_DESCRIPTION", "Get details ...")`,
-  so even a field reader would have found a call rather than a string. The
-  result was 114 `SHIP-DOC-MISSING-DESCRIPTION` findings against a
-  repository that documents every tool it ships, which is the kind of noise
-  that costs a reader's trust in everything else on the page.
+- Read Go MCP tool descriptions from struct `Description` fields and
+  direct `WithDescription`/`WithToolDescription` options. Later description options replace earlier
+  ones, including empty or computed values. Nested calls and partial
+  expressions cannot supply the parent tool's description.
 
-  `WithDescription`/`WithToolDescription` is now read for both Go
-  call-site idioms, and a translation helper is unwrapped to its default.
-  The unwrapping is deliberately narrow: the call must pass exactly two
-  string literals and the first must look like a lookup key — screaming
-  snake or dotted — so `wrap("some prose here", "invented")` is refused
-  rather than published as documentation its author never wrote. The key
-  itself is never read as the description; printing
-  `TOOL_GET_JOB_DESCRIPTION` at a human would be worse than printing
-  nothing. A computed description, a qualified call such as
-  `fmt.Sprintf(...)`, a three-argument call, a variable key, and a literal
-  concatenated with something else all still read nothing.
-
-  `tools/shipgate-detect.py` reads these idioms a second time, and the
-  shared conformance corpus caught the divergence the moment the engine
-  reader learned something the detector had not: both now carry the same
-  extractor, and the corpus keeps them honest.
-
-  `tests/test_go_tool_descriptions.py` pins three reads and seven refusals;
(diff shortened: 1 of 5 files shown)


pengfei-threemoonslab (commented):
Reviewed final head `e9fca6e15d165301b36d0fe402edd3368e262db9` after integrating the author's rebase. 1,708 integration tests passed with one skip; all four performance tests and Ruff passed. Fourteen of the 19 added shared regression cases fail on the original PR implementation. CI for this exact head is still running.

Technical review: the description-extraction increment is ready after the corrections below and final-head CI.

The original PR had concrete false-documentation cases: it searched nested calls, accepted only the default portion of a larger expression, retained the first of multiple description options, and treated arbitrary two-string calls as translation helpers. The fixes restrict extraction to direct fields/options and whole values, follow the SDK's last-option-wins behavior, and require a declared translation-helper parameter rather than guessing from a lookup-key shape. Empty/computed overrides clear an earlier description; comments and trailing commas remain supported.

The original code also missed the actual struct-field shape in #658. The shared extraction now covers that route. On the pinned [upstream Actions source](https://github.com/github/github-mcp-server/blob/7d13a7ad6f2a17f351a6d77ce280c85ae1821f4d/pkg/github/actions.go), the original PR reads 0/4 descriptions and the corrected reader reads 4/4. No full 114-tool result or annotation fix is claimed. #658 must remain open.

The SDK's [option application and description setter](https://github.com/mark3labs/mcp-go/blob/main/mcp/tools.go) and GitHub's [declared translation helper](https://github.com/github/github-mcp-server/blob/7d13a7ad6f2a17f351a6d77ce280c85ae1821f4d/pkg/translations/translations.go) support these extraction boundaries. The upstream Go source was inspected and parsed statically.

Current control: `complete`, release decision `passed`, merge permission `true`. No remaining blocking finding within this description-extraction increment.

pengfei-threemoonslab (commented):
Reviewed final head `5b0c75f1303132cba5832a74c5c5d82470cea3a7` after integrating the author's rebase. 1,714 integration tests passed with one skip; all four performance tests and Ruff passed. Sixteen of the 21 added shared regression cases fail on the original PR implementation. All GitHub CI checks passed on this exact head.

Technical review: the description-extraction increment is ready after the corrections below and passing final-head CI.

The original PR had concrete false-documentation cases: it searched nested calls, accepted only the default portion of a larger expression, retained the first of multiple description options, and treated arbitrary two-string calls as translation helpers. The fixes restrict extraction to direct fields/options and whole values, follow the SDK's last-option-wins behavior, and require a declared translation-helper parameter rather than guessing from a lookup-key shape. Parameters inside a function type do not bind names in the enclosing function, and a function-type declaration cannot borrow the body of a later declaration. Empty/computed overrides clear an earlier description; comments and trailing commas remain supported.

The original code also missed the actual struct-field shape in #658. The shared extraction now covers that route. On the pinned [upstream Actions source](https://github.com/github/github-mcp-server/blob/7d13a7ad6f2a17f351a6d77ce280c85ae1821f4d/pkg/github/actions.go), the original PR reads 0/4 descriptions and the corrected reader reads 4/4. No full 114-tool result or annotation fix is claimed. #658 must remain open.

The SDK's [option application and description setter](https://github.com/mark3labs/mcp-go/blob/main/mcp/tools.go) and GitHub's [declared translation helper](https://github.com/github/github-mcp-server/blob/7d13a7ad6f2a17f351a6d77ce280c85ae1821f4d/pkg/translations/translations.go) support these extraction boundaries. The upstream Go source was inspected and parsed statically.

Current control: `complete`, release decision `passed`, merge permission `true`. No remaining blocking finding within this description-extraction increment.

PR #682, opened 2026-09-11
Read a TypeScript tool's description in both SDK shapes (#680)

### LINKED ISSUE #680: TypeScript MCP tool descriptions are never read, so every tool in the reference SDK's shape is reported undocumented
Found while fixing the Go half of #658 ([PR #679](https://github.com/ThreeMoonsLab/agents-shipgate/pull/679)). Same defect class, larger blast radius: TypeScript is the reference MCP SDK.

## Measured

```ts
import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";

export function register(server: McpServer) {
  server.registerTool(
    "get_job",
    { description: "Get details of a specific workflow job.", inputSchema: {} },
    async () => ({ content: [] }),
  );

  server.tool("delete_job", "Delete one workflow job.", async () => ({ content: [] }));
}
```

```
'get_job'      description=None
'delete_job'   description=None
tools found: 2
```

Both tools are found. Both descriptions are written as plain string literals, in the two documented forms. Both read `None`.

## Cause

`ts_sdk_register_tool` is built by `_call_sites`, which resolves the *name* and nothing else:

```python
sites.extend(_call_sites(source, _TS_REGISTER_TOOL_RE, "ts_sdk_register_tool"))
```

There is no TypeScript description extractor at all. `_TS_DESCRIPTION_RE` exists but serves the class-property idiom (`this.description = "…"`), not the SDK registration shape.

This is exactly the gap #679 just closed on the Go side, where the description lives in an option rather than a struct field. Here it lives in an options **object** (`{ description: … }`) or in a **positional** second argument, and neither is read.

## Consequence

`SHIP-DOC-MISSING-DESCRIPTION` fires per undocumented tool, so a TypeScript server reports one false finding per tool it ships — the same shape as the 114 findings on `github/github-mcp-server` that motivated #658, against the more common SDK.

This is the noise that costs a reader's trust in everything else on the page, and it lands on the population the MCP wedge targets.

## What a fix has to cover

- `registerTool("name", { description: "…" }, handler)` — the current SDK shape; the description is a property of the options object, which must be the *registration's own* object and not a nested one.
- `tool("name", "description", handler)` — the older positional shape. Note `tool()` is also matched for lookups, so the existing one-argument and `+`-concatenation refusals in `_call_sites` still have to hold.
- Template literals and concatenation must read nothing rather than a fragment, as elsewhere in this reader.
- Translation helpers are a separate unresolved contract, tracked in #691 under #658. This issue covers the two demonstrated SDK literal shapes; arbitrary `t(key, value)` calls remain unresolved rather than asserting the second argument is documentation.

## Second implementation

`tools/shipgate-detect.py` reads these idioms independently and has the same gap. The shared conformance corpus (`tests/test_zero_install_detector.py`) will fail the moment one side learns this and the other does not — it caught exactly that during #679 — so both copies must move together.

## Verifying a fix

A guard here is only worth its name if the refusals are pinned alongside the reads: a reader that returns *some* string for every shape would pass a reads-only test while inventing documentation. #679's `tests/test_go_tool_descriptions.py` is the pattern — 3 reads, 7 refusals.

Related: #658 (the Go half, partially addressed by #679), #431 (idiom registry), #485 (why the second implementation is survivable).

## Review refinement — 2026-09-11

PR #682 review found that brace depth alone does not identify an own property, and returning the first literal ignores later JavaScript overrides. The fix must read direct object members in order, clear stale descriptions after an unknown spread/computed key or a later empty/computed value, and reject a ternary expression masquerading as a property. Both readers must receive these cases through `SOURCE_CASES`. The independently unresolved translation-helper item is deferred to #691 rather than counted as implemented here.

pengfei-threemoonslab:
Closes #680. Follows the merged Go description work in #679; parent #658 retains source annotation and measured precision work. The unresolved TypeScript translation-helper contract is separately deferred in #691.

## Problem and behavior

The TypeScript MCP SDK reader found tool names but discarded their literal descriptions, producing missing-description findings for documented tools. This change reads both `registerTool("name", { description: "…" }, handler)` and `tool("name", "…", handler)` in the installed package and zero-install detector.

The options reader processes direct object members in order. A later literal replaces an earlier description; empty/computed values, shorthand, accessors, spreads and computed keys cannot leave a stale description behind. A later explicit literal can establish the value again. Nested schema descriptions and colons inside another property's expression are never tool documentation. Concatenated strings, interpolated templates and options-object expressions are left unresolved.

The static reader does not infer a TypeScript translation helper's semantics from the name `t` or the presence of two strings. #691 owns establishing a useful bounded contract from pinned source evidence.

## Validation and review

- Original negative control: 6 of the initial 13 SDK-description tests failed before literal extraction; existing refusals passed by reading nothing.
- Independent GitHub review [5183738123](https://github.com/ThreeMoonsLab/agents-shipgate/pull/682#pullrequestreview-5183738123) found stale overrides, false property recognition and missing shared corpus coverage. Ten added counterexamples reproduced those failures on the original PR head.
- Both readers now receive all 33 TypeScript description cases through shared `SOURCE_CASES`, including every refusal and override. The package-level loader tests assert expected descriptions independently of reader parity.
- The second-round correction passed full regression (**9,696 passed, 5 skipped**). Final independent preparation then found undecodable quoted keys in supported CommonJS input; both readers now invalidate that unknown key instead of retaining an earlier literal. The final guard passed **1,288 package/zero-install/static tests, 1 skipped**; the additional real CommonJS loader regressions passed in the **38-case SDK-description suite**, and all four performance tests passed. Ruff/whitespace checks pass. Required CI validates the final published commit across the full suite.

No new check ID or idiom ID; the existing static literal boundary is made accurate. This is a reader repair, not a qualification or runtime behavior claim.

The latest correction also refuses async/escaped identifier overrides and undecodable quoted keys conservatively; it does not add a decoder or execute user JavaScript. A subsequent direct literal may re-establish the description.

### DIFF
--- a/src/agents_shipgate/inputs/mcp_idioms.py
+++ b/src/agents_shipgate/inputs/mcp_idioms.py
@@ -1625,49 +1625,46 @@ def _child_braces(masked: str, open_brace: int, close: int) -> list[tuple[int, i
 #:
 #:   server.registerTool("get_job", { description: "…", inputSchema: {} }, fn)
 #:   server.tool("get_job", "…", fn)
-_TS_OBJECT_DESCRIPTION_KEY_RE = re.compile(r"(?<![\w$.])description\s*:\s*")
+_TS_OBJECT_PROPERTY_RE = re.compile(r"[A-Za-z_$][\w$]*")
 
 
 def _ts_object_description(
     source: MaskedSource, open_brace: int, close: int
 ) -> str | None:
-    """The options object's own ``description``, never a nested one.
+    """Read direct members in order, preserving JavaScript override semantics.
 
-    The depth test is the whole point. A registration's options object
-    carries `inputSchema`, and a JSON Schema describes *every parameter*:
-
-        { description: "Get a job", inputSchema: { properties: {
-              job_id: { description: "The job to get" } } } }
-
-    Without the depth-1 restriction the first `description:` found inside a
-    schema would be published as the tool's, so a tool whose own description
-    is missing would be documented with one of its arguments' — an invented
-    answer rather than an absent one.
+    A colon inside another member's expression is not a property boundary.
+    Spreads and computed keys may replace description; a later explicit
+    literal can establish it again. Never retain a stale earlier value.
     """
 
-    candidates: list[int] = []
-    for match in _TS_OBJECT_DESCRIPTION_KEY_RE.finditer(
-        source.masked, open_brace + 1, close
-    ):
-        candidates.append(match.end())
-    # `{ "description": "…" }` is the same key; the masker hides its body, so
-    # a quoted key is found by value rather than by pattern.
-    for start, (value, end) in source.literals.items():
-        if not (open_brace < start < close) or value != "description":
+    description = None
+    # The balanced argument splitter also separates object members: strings,
+    # comments and nested expressions cannot introduce a member separator.
+    for start, end in _go_arguments(source, open_brace, close):
+        found, key, key_end = source.literal_at(start)
+        if not found:
+            match = _TS_OBJECT_PROPERTY_RE.match(source.masked, start, end)
+            if match is None:
+                description = None  # spread or computed key
+                continue
+            key, key_end = match.group(), match.end()
+        after = source.skip_space(key_end)
+        if key in {"get", "set"} and after < end and source.masked[after] != ":":
+            # Accessors can override the same property without a colon.
+            accessor = _TS_OBJECT_PROPERTY_RE.match(source.masked, after, end)
+            if accessor is None or accessor.group() == "description":
+                description = None
             continue
-        after = source.skip_space(end)
-        if after < len(source.masked) and source.masked[after] == ":":
-            candidates.append(after + 1)
-    for position in sorted(candidates):
-        if _brace_depth(source.masked, open_brace, position) != 1:
+        if key != "description":
             continue
-        found, value, end = source.literal_at(position)
-        if found and value and _literal_is_whole_value(source, end, ",}"):
-            return value
-        # The key is present at the right level but its value is not a plain
-        # literal. Reading further would be reading some other key's value.
-        return None
-    return None
+        description = None
+        if after >= end or source.masked[after] != ":":
+            continue  # shorthand or method: its value is not a known string
+        found, value, literal_end = source.literal_at(after + 1)
+        if found and value and literal_end == end:
+            description = value
+    return description
 
 
 def _ts_call_description(
@@ -1693,7 +1690,11 @@ def _ts_call_description(
         return value if value and _literal_is_whole_value(source, end, ",)") else None
     if second < len(source.masked) and source.masked[second] == "{":
         object_close = _matching_close(source.masked, second, "{", "}")
-        if object_close is None or object_close > close:
+        if (
+            object_close is None
+            or object_close > close
+            or not _literal_is_whole_value(source, object_close, ",)")
+        ):
             return None
         return _ts_object_description(source, second, object_close)
     return None
--- a/src/agents_shipgate/inputs/mcp_idioms.py
+++ b/src/agents_shipgate/inputs/mcp_idioms.py
@@ -1617,6 +1617,88 @@ def _child_braces(masked: str, open_brace: int, close: int) -> list[tuple[int, i
     return children
 
 
+#: The TypeScript SDK writes a tool's description in one of two places, and
+#: until #680 neither was read — `ts_sdk_register_tool` resolved the name and
+#: nothing else, so every tool on the *reference* MCP SDK was reported
+#: undocumented. `_TS_DESCRIPTION_RE` above looks like it covers this and does
+#: not: it serves the class-property idiom (`this.description = "…"`).
+#:
+#:   server.registerTool("get_job", { description: "…", inputSchema: {} }, fn)
+#:   server.tool("get_job", "…", fn)
+_TS_OBJECT_DESCRIPTION_KEY_RE = re.compile(r"(?<![\w$.])description\s*:\s*")
+
+
+def _ts_object_description(
+    source: MaskedSource, open_brace: int, close: int
+) -> str | None:
+    """The options object's own ``description``, never a nested one.
+
+    The depth test is the whole point. A registration's options object
+    carries `inputSchema`, and a JSON Schema describes *every parameter*:
+
+        { description: "Get a job", inputSchema: { properties: {
+              job_id: { description: "The job to get" } } } }
+
+    Without the depth-1 restriction the first `description:` found inside a
+    schema would be published as the tool's, so a tool whose own description
+    is missing would be documented with one of its arguments' — an invented
+    answer rather than an absent one.
+    """
+
+    candidates: list[int] = []
+    for match in _TS_OBJECT_DESCRIPTION_KEY_RE.finditer(
+        source.masked, open_brace + 1, close
+    ):
+        candidates.append(match.end())
+    # `{ "description": "…" }` is the same key; the masker hides its body, so
+    # a quoted key is found by value rather than by pattern.
+    for start, (value, end) in source.literals.items():
+        if not (open_brace < start < close) or value != "description":
+            continue
+        after = source.skip_space(end)
+        if after < len(source.masked) and source.masked[after] == ":":
+            candidates.append(after + 1)
+    for position in sorted(candidates):
+        if _brace_depth(source.masked, open_brace, position) != 1:
+            continue
+        found, value, end = source.literal_at(position)
+        if found and value and _literal_is_whole_value(source, end, ",}"):
+            return value
+        # The key is present at the right level but its value is not a plain
+        # literal. Reading further would be reading some other key's value.
+        return None
+    return None
+
+
+def _ts_call_description(
+    source: MaskedSource, open_paren: int, close: int
+) -> str | None:
+    """The description a `registerTool`/`tool` registration carries.
+
+    Both documented shapes put it in the *second* argument: a bare string in
+    the positional form, a `description` property in the options form.
+    Nothing else is read — `tool("name", handler)` is a registration with no
+    description, and must stay one.
+    """
+
+    found, _name, name_end = source.literal_at(open_paren + 1)
+    if not found:
+        return None
+    after = source.skip_space(name_end)
+    if after >= len(source.masked) or source.masked[after] != ",":
+        return None
+    second = source.skip_space(after + 1)
+    found, value, end = source.literal_at(second)
+    if found:
+        return value if value and _literal_is_whole_value(source, end, ",)") else None
+    if second < len(source.masked) and source.masked[second] == "{":
+        object_close = _matching_close(source.masked, second, "{", "}")
+        if object_close is None or object_close > close:
+            return None
+        return _ts_object_description(source, second, object_close)
+    return None
+
+
 #: The option-function shape: `NewTool("name", WithDescription("…"))`. This is
 #: how `mark3labs/mcp-go` declares a tool, which is the SDK behind
 #: `github/github-mcp-server` — so until #658 every one of its 114 tools read
@@ -3243,7 +3325,14 @@ def scan_source(
     sites: list[RegistrationSite] = []
     if language == "typescript":
         sites.extend(_ts_static_tool_name_sites(source))
-        sites.extend(_call_sites(source, _TS_REGISTER_TOOL_RE, "ts_sdk_register_tool"))
+        sites.extend(
+            _call_sites(
+                source,
+                _TS_REGISTER_TOOL_RE,
+                "ts_sdk_register_tool",
+                describe=_ts_call_description,
+            )
+        )
     else:
         sites.extend(
             _call_sites(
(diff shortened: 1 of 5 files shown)


src/agents_shipgate/inputs/mcp_idioms.py
```diff
@@ -1617,6 +1617,88 @@ def _child_braces(masked: str, open_brace: int, close: int) -> list[tuple[int, i
...
+            continue
+        after = source.skip_space(end)
+        if after < len(source.masked) and source.masked[after] == ":":
+            candidates.append(after + 1)
+    for position in sorted(candidates):
+        if _brace_depth(source.masked, open_brace, position) != 1:
+            continue
+        found, value, end = source.literal_at(position)
+        if found and value and _literal_is_whole_value(source, end, ",}"):
+            return value
```
pengfei-threemoonslab: [P2] Do not return before later options can replace the description

The first matching literal is returned immediately. At this exact head, both readers publish `Stale` for `server.registerTool("job", {description: "Stale", ...overrides}, fn)`, and also for a later `description: ""` or `description: buildDesc()`. JavaScript applies later properties/spreads after the earlier value, so this is not proven effective tool documentation and can hide a real missing-description finding. Process direct members in order, invalidating an earlier value when a later write/spread can override it, or conservatively refuse the ambiguous object. Cover these cases in the shared two-reader corpus; preserve ordinary plain-literal reads.

src/agents_shipgate/inputs/mcp_idioms.py
```diff
@@ -1617,6 +1617,88 @@ def _child_braces(masked: str, open_brace: int, close: int) -> list[tuple[int, i
...
+    schema would be published as the tool's, so a tool whose own description
+    is missing would be documented with one of its arguments' — an invented
+    answer rather than an absent one.
+    """
+
+    candidates: list[int] = []
+    for match in _TS_OBJECT_DESCRIPTION_KEY_RE.finditer(
+        source.masked, open_brace + 1, close
+    ):
+        candidates.append(match.end())
```
pengfei-threemoonslab: [P2] Require an object-member key boundary, not only brace depth

The regex matches `description:` anywhere at the same brace depth. For `server.registerTool("job", {title: enabled ? description : "Not tool documentation", inputSchema: {}}, fn)`, both exact-head readers publish `Not tool documentation`, although the options object has no description property. The ternary's variable and colon satisfy the regex and the depth check. Recognize direct object members/token boundaries or refuse ambiguous syntax before reading a value. Add this negative control to the shared corpus and fix the duplicated zero-install extractor as well.

pengfei-threemoonslab (commented):
Independent coding-agent review, round 1, on `396de3b1dd85434e7cea0692e0ca8bc7444ac0c7`.

Changes are required before merge. The two inline findings reproduce in both the packaged reader and `tools/shipgate-detect.py`: one returns stale documentation after a later property/spread can replace it; the other mistakes a ternary expression for the options object's own description. They undermine this PR's main promise of refusing invented documentation.

I inspected the exact GitHub diff, the shared conformance harness, issue #680, prior review/address history and current checks. There were no previous formal or inline reviews. All nine required checks on this head are green, but the branch is behind main; passing the existing cases does not cover the reproduced wrong answers. My probes loaded the two exact committed modules in memory with bytecode writing disabled; I did not modify the repository or run a test suite.

Two accompanying scope/test corrections belong in the address:

- #680 explicitly includes a translation helper such as `description: t("key", "Actual text")`. Both readers still return `None` for it. Either deliver the bounded helper case with refusals, or explicitly retain it under #658 / the remaining issue scope before using `Closes #680`; do not count it as implemented.
- The new cases live only in `test_ts_tool_descriptions.py` and exercise the package reader. `test_zero_install_detector.py` gets its comparisons from `tests/mcp_idiom_corpus.py:SOURCE_CASES`, which currently has no SDK-description cases. Add shared positive and refusal cases there, including these reproductions, so the claimed conformance protection actually exercises the new code in both implementations.

This COMMENT is independent coding-agent review, not human approval, release authorization or qualification evidence. One formal review round is recorded; the author should address it in this existing PR.

pengfei-threemoonslab:
## Review/address round 1

Addressed [independent review 5183738123](https://github.com/ThreeMoonsLab/agents-shipgate/pull/682#pullrequestreview-5183738123) in `a494b37b`; current PR head after the ROADMAP-only main merge is `d1af9b310984b3e2caef5f9109b2257fea90e8f9`.

1. **Stale overrides:** process the options object's direct members in source order. Later empty/computed/shorthand/accessor values, unknown spreads and computed keys invalidate the earlier description; a later explicit literal can re-establish it.
2. **False property recognition:** split balanced top-level members and identify the member's actual key. A ternary or nested schema description cannot become the tool description. Also refuse an options object that is only part of a larger expression.
3. **Unimplemented helper scope:** #680 now explicitly covers the demonstrated literal SDK forms. New deferred issue #691, linked from #658, owns establishing a TypeScript translation-helper contract from actual source evidence. An unbound `t(key, value)` remains unresolved; its second string is not silently asserted as documentation.
4. **Reader parity:** moved all original SDK-description cases and 13 additional cases into shared `TS_DESCRIPTION_CASES`/`SOURCE_CASES`. Package-level expected-value tests and zero-install parity exercise the same inputs.

Negative control on the original PR head: ten added cases failed. After addressing the findings: 772 targeted tests passed, one skipped; full regression 9,675 passed, five skipped; static-only 486 passed and performance four passed; Ruff/whitespace checks passed. Main integration afterward changed only ROADMAP.md. Required CI and a second independent review will evaluate the published head before merge. This response completes round 1; maximum three rounds remains in force.

src/agents_shipgate/inputs/mcp_idioms.py
```diff
@@ -1617,6 +1617,89 @@ def _child_braces(masked: str, open_brace: int, close: int) -> list[tuple[int, i
...
+            key, key_end = match.group(), match.end()
+        after = source.skip_space(key_end)
+        if key in {"get", "set"} and after < end and source.masked[after] != ":":
+            # Accessors can override the same property without a colon.
+            accessor = _TS_OBJECT_PROPERTY_RE.match(source.masked, after, end)
+            if accessor is None or accessor.group() == "description":
+                description = None
+            continue
+        if key != "description":
+            continue
```
pengfei-threemoonslab, line 1673: [P2] Prove the complete member key before ignoring an apparent unrelated property

`match()` can consume only a prefix, and this `continue` assumes that prefix is the object's actual key. At the reviewed head, both readers return `old` for `{description:"old", async description(){ return "actual"; }}` and `async *description(){...}`: `async` is a modifier, not the overwritten key. They also return `old` for `{description:"old", descrip\u0074ion:""}` and the same escaped key assigned `"actual"`; JavaScript decodes that identifier to `description`. I confirmed these object forms in Node (the resulting property is respectively a function, an empty string, or `actual`). Before retaining the earlier value, establish a complete known unrelated member shape; conservatively invalidate unrecognized modifiers/escaped-key syntax if it is outside the supported grammar. Add shared counterexamples for both reader copies. This can remain a narrow refusal change; no new description inference is needed.

pengfei-threemoonslab (commented):
Independent coding-agent review, round 2 of at most 3, on `d1af9b310984b3e2caef5f9109b2257fea90e8f9`.

The original round-1 counterexamples are addressed: later plain properties/spreads no longer leave stale text, ternary colons are not direct keys, options suffix expressions are refused, and all SDK-description cases now participate in shared-reader conformance. The unresolved translation-helper contract is explicitly deferred in #691 rather than guessed or claimed delivered. I read the author address #issuecomment-5641289202 and the new issue.

One P2 remains in the same stale-override boundary, recorded inline. An identifier prefix is still treated as a proven unrelated key, so async methods and escaped identifier keys can overwrite `description` while both readers retain the earlier literal. This needs a bounded refusal/key-boundary repair before merge, not a general TypeScript-parser expansion.

Independent validation: the selected package-description and shared-conformance tests exited 0 (261 cases). Separate read-only probes loaded both exact committed readers with bytecode writes disabled and reproduced the new cases; a small Node evaluation confirmed the corresponding JavaScript objects actually overwrite the property. The worktree remained clean, and I did not edit source or merge.

This is independent coding-agent COMMENT review, not human release approval or qualification. Two formal rounds have now been used. The author should address the remaining finding in this existing PR and preserve the final review budget.

pengfei-threemoonslab:
## Review/address round 2

Addressed [independent review 5183889853](https://github.com/ThreeMoonsLab/agents-shipgate/pull/682#pullrequestreview-5183889853). The remaining defect was a prefix match being treated as proof of an unrelated property: async methods and escaped identifier keys could overwrite `description` while the reader kept the earlier literal.

Both readers now require a complete direct field, shorthand or ordinary method key before skipping it as unrelated. Unrecognized modifiers/escaped keys invalidate the old description. Accessors likewise require a complete key followed by their parameter list. This is a conservative refusal of unsupported syntax, not an expansion into general TypeScript evaluation.

Five additional override counterexamples failed on the previously published head. The shared corpus now also retains positive controls for unrelated ordinary methods and getters. Targeted validation after the fix: **793 passed, 1 skipped**; Ruff and whitespace checks passed. Full regression and the exact new head are recorded with the final commit below. The next independent review is round 3, the requested maximum; no fourth formal review will be added.

Final correction commit: `1a2e35e7d52778b8eafe0922433b3cb3fcd4a76f`. Full regression before the final quoted-key guard passed **9,696 tests, five skipped**. Final-review preparation found one additional instance of the same defect in actual supported `.cjs` input: `literal_at` returning `found=True, key=None` was incorrectly treated as an unrelated key. Both readers now invalidate the unknown key. All 33 shared description cases include unknown-key refusal and subsequent direct-literal recovery; a real CommonJS loader regression covers both outcomes. Final package/zero-install/static selection: **1,288 passed, one skipped**; the SDK description suite including both CommonJS regressions: **38 passed**; performance: **four passed**. Required hosted CI will validate the final complete tree.

pengfei-threemoonslab (commented):
Independent coding-agent review, round 3 of 3, on `1a2e35e7d52778b8eafe0922433b3cb3fcd4a76f`.

No remaining actionable finding in the bounded description-extraction changes reviewed here. I read the round-2 author address (#issuecomment-5641428537), checked the published diff, and matched all five changed GitHub blob hashes to an isolated archive of this exact commit.

The round-2 finding (#discussion_r3993857786) is addressed: async methods, generators and escaped identifier/accessor prefixes cannot be mistaken for a proven unrelated key and leave stale documentation. The additional quoted-key case found during final preparation is also addressed: `literal_at(...)=found=True, key=None` now clears the earlier description in both readers. This matters for supported non-strict CommonJS input with a legacy-octal quoted key; it is a refusal of unknown decoding, not an invented interpretation. A later direct literal can establish the description again. Ordinary unrelated methods/getters and valid literal properties remain covered.

The round-1 stale-property/spread and false-ternary-property findings remain addressed. All 33 SDK-description cases are present in shared `SOURCE_CASES`, and the real CommonJS loader tests cover both invalidation and later-literal recovery. The unbound translation-helper contract remains explicitly deferred in #691; this review does not claim it implemented.

Independent validation on the isolated exact-commit archive: **436 tests passed, 0 failures/errors/skips, actual exit 0** (SDK-description tests and selected package/zero-install source-corpus/CRLF conformance tests). Ten additional syntax probes passed in both readers, including the quoted-key reproductions, escaped accessors/methods, known escaped string keys and recovery after an unknown override. No source or repository artifacts were changed. I did not rerun the full repository suite; the PR correctly distinguishes the author's full-suite run before the final guard from the final targeted tests. Hosted CI is a separate check and is not claimed complete by this review.

This is the third and final permitted coding-agent COMMENT review for this implementation pass, not human release approval, qualification evidence or merge authorization. The author can record the final address and resolve the confirmed findings within this loop; no fourth formal review is requested.

pengfei-threemoonslab:
## Review/address round 3 — complete

Addressed [final independent review 5183983215](https://github.com/ThreeMoonsLab/agents-shipgate/pull/682#pullrequestreview-5183983215), bound to `1a2e35e7d52778b8eafe0922433b3cb3fcd4a76f`. No remaining actionable finding was identified in this scope, so no further source edit is needed.

The reviewer verified all five published blobs, independently passed **436 tests** and rechecked ten syntax probes against both readers. The previous identifier/accessor override finding and the undecodable quoted-key case now invalidate stale descriptions; subsequent known literals correctly re-establish them. The real CommonJS loader regression also passed.

This completes **three independent GitHub review/address loops**, the requested maximum. No fourth formal review will be added. Required hosted CI must still pass on this final head before merge. #691 remains deferred, and #658 retains its broader annotation/precision work; this PR does not claim those items are delivered.

pengfei-threemoonslab:
## Required current-main integration

The repository requires strict up-to-date status checks. After #692 merged, this PR integrated main at `f57b49f3cf45ec0b8729f500563c711a577376a8`; the new head is `9b75e61311ce412f67b211fa80138fff4d338142`.

The four MCP implementation/test files are **byte-identical** to the third reviewed head `1a2e35e7`. The only conflict was in `CHANGELOG.md`: retain the shallow recovery note, describe the final TypeScript fix accurately, and remove a duplicate older Go-description paragraph whose maintained entry was already present. The incoming `diff.py` and capability-diff tests are exactly #692's independently reviewed changes.

No fourth review loop is introduced. The three completed loops still cover this PR's unchanged implementation. Required CI reruns on the integrated head. The committed verifier also reports `complete`/`passed` with merge permission; final merge will use fresh current evidence after CI. No release or whole-v1 readiness claim is implied.

Issue #694, opened 2026-09-11
check crashes when a new baseline and blocked host-capability changes share the working tree


pengfei-threemoonslab:
## Observed

A synthetic Route H repository contains `.claude/settings.json` and `.mcp.json`. Record `audit --host --save-baseline` on the base, leave that new `.agents-shipgate/host-grants.json` file untracked, then commit a branch changing `Bash(npm test)` / `Read(src/**)` to `Bash(*)` / `Read(**)` / `WebFetch(*)` and adding a remote MCP entry. Run `check --format agent-boundary-json` on the branch.

The command raises `ValidationError: a blocked result cannot authorize publication` from `core/agent_boundary.py` while constructing `AgentBoundaryResultV1`; no JSON answer is emitted. Reproduced on main's contract 33 implementation and the contract 34 #685 worktree. If the synthetic baseline is instead committed on the base before the capability change, the same fixture returns `block` / `critical` with four violations and valid JSON. No server was contacted and no agent action executed.

## Why it matters

A working tree can legitimately contain both a trust-root change and a blocked host-capability change. The machine envelope must combine these routes consistently. A Pydantic constructor failure gives an adopter neither the named violations nor an actionable control response.

## Proposed

Reconcile blocked host decisions with trust-root `review_publishable` routing before constructing the envelope. Retain the strictest required permissions and exact next action; do not loosen the schema invariant or turn the exception into an allowed/publication-ready answer. Audit the equivalent JSON/check projections for the same mixed input.

## Acceptance

- [ ] Add the mixed untracked-baseline + host-expansion regression through the actual check command.
- [ ] Emit one valid machine object carrying the violations and consistent control permissions, rather than a traceback/empty stdout.
- [ ] Keep committed-baseline and trust-root-only controls intact.
- [ ] Preserve the invariant that a blocked result cannot authorize publication; prove error handling does not grant merge/completion.

P2, no milestone; deferred separately from #685 as requested. Related #684 (default PR/check usefulness), #662 (surface consistency), #690 (workflow interruption). The current #685 workflow-diff repair does not change these control schemas.

Issue #693, opened 2026-09-11
Named reusable-workflow secret remapping is invisible to host capability diffs


pengfei-threemoonslab:
## Observed

While implementing #685, a static counterexample changed a reusable job from `secrets: {credential: "${{ secrets.STAGING_TOKEN }}"}` to `secrets: {credential: "${{ secrets.PRODUCTION_TOKEN }}"}`, keeping the same job, `uses: ./.github/workflows/deploy.yml`, triggers and token permissions. `_workflow_grant` → `diff_host_grants` returned `changes: []` and `expansion_signals: []` both on current main (`a3070edfb`) and the #685 implementation. The identifiers are synthetic; no credential was read or executed, and their names alone do not establish a widening.

On main, the generic config redactor drops the entire `secrets` container before hashing. #685 deliberately models target/ref and `secrets: inherit`, but named mappings are outside its bounded projection. Thus this is an existing blind spot, not a regression introduced to remove workflow noise.

## Why it matters

A reviewer cannot see that the same callee is now offered a different named credential source. Showing no selected change is narrower than saying the reusable workflow's secret authority is unchanged. The distinction must be documented and tested.

## Proposed

Add a bounded static representation of named secret forwarding to the existing reusable-workflow grant. A supported literal `${{ secrets.NAME }}` reference can identify which declared source is offered to which callee input without reading its value. Compare these typed facts before display redaction. Preserve destination identity and avoid silently merging distinct mappings whose display labels redact alike.

Show a source-reference addition, removal or change. Names such as `STAGING_TOKEN` and `PRODUCTION_TOKEN` do not establish credential privilege, caller availability, whether a step receives the secret, or downstream access. A reference change may be direction-unknown or non-expanding and still be useful review evidence.

Literal secret values, dynamic expressions and unsupported mappings need a privacy-safe limit, not a hash of the secret or a guessed reference. Use existing inventory/diff/PR surfaces. Do not add a runtime secret resolver, organization policy service or new command.

## Acceptance

- [ ] Pin the observed source-remapping counterexample and an unchanged-mapping control. The actual CLI names the changed forwarding relationship after the repair, while unchanged mappings and formatting-only edits remain quiet.
- [ ] Added/removed destinations and changed supported source references remain distinguishable. Two mappings are not collapsed by equal display redactions.
[...]

PR #695, opened 2026-09-11
fix: remove workflow diff noise and name inherited secret recipients

### LINKED ISSUE #685: Workflow rows are noise: any edit to a write-holding workflow prints `widened`, and a step's shell edit prints `changed read → read`
## Observed

Real first-parent history, `DanielLavrushin/b4`, `shipgate diff --base <parent>` at each step, 2026-09-11 on `main` at `396de3b1`:

**Step `a757e2a7`** — the only change to `.github/workflows/release.yml` is a new job that calls a reusable workflow (`uses: ./.github/workflows/sourceforge-mirror.yml`, `secrets: inherit`) and one `needs:` list. No `permissions:` block changed.

```
⚠ high      widened  github .github/workflows/release.yml
                     write, packages: write, contents: write → write, packages: write, contents: write
                     the workflow can write to the repository
```

The row says **widened** at **high** and its own transition text shows before and after identical.

**Step `2ea7e88c`** — the only change to `.github/workflows/sourceforge-mirror.yml` is inside a `run:` shell script and an input `description:` string.

```
  low       changed  github .github/workflows/sourceforge-mirror.yml
                     read → read
```

A row for a grant whose authority did not move.

Both steps are the only host-touching steps in the first 20 of that repository's history, and both rows are wrong. This is the "zero false `expanded` rows" bar in #657 and the Gate 1 bar in the v1.0 roadmap, failed on the first real repository the harness reached.

## Two causes, both in the workflow reader

1. **The grant's config is the whole document.** `_workflow_grant` (`core/host_grants.py:844`) passes `config=data` — the entire parsed YAML — to `_grant_base`. The grant fingerprint therefore changes whenever any byte of the workflow changes, so a shell edit is a `changed` grant even though `triggers`, `write_all`, `write_scopes` and `pull_request_target` are byte-identical. The config should be the authority projection, not the file.

2. **`workflow_write_changed` fires without comparing.** `host_grant_expansion_signals` emits `workflow_write_{prefix}` whenever the *after* grant has `write_all`, `write_scopes` or `pull_request_target` — regardless of whether the *before* grant had the same. `capability_diff_rows` then promotes `changed` to `widened` on that signal (`capability_diff_rows.py:147`). So every edit to a workflow that already holds write permission is a widening.

## What is actually a widening here, and is not modelled

`secrets: inherit` on a call to a reusable workflow passes every repository secret to the callee. That *is* a change in what the job can reach, and the reader does not see it at all — it printed the wrong row for the wrong reason. Modelling `secrets: inherit` and `uses:` of a reusable workflow as a workflow-grant attribute is in scope for this issue; without it the fix turns a false `widened` into a false silence on this exact commit.

## Proposed

- `_workflow_grant`: `config` becomes `{triggers, write_all, write_scopes, pull_request_target, reusable_calls: [{uses, secrets_inherit}]}`. A change anywhere else in the file produces no grant change and no row.
- `workflow_write_changed` is emitted only when the after surface is a strict superset: a new write scope, `write-all` newly present, `pull_request_target` newly present, or a reusable call newly inheriting secrets. Narrowings emit nothing (as for permission rules after #657).
- The transition text on a `widened` workflow row names what widened (`+ packages: write`, `+ secrets: inherit → sourceforge-mirror.yml`), never an identical before/after.
- Fixtures: the two real commits above, committed under `tests/fixtures/workflows/` with their expected rows; plus one genuine widening (`contents: read → write`) and one genuine narrowing.

## Acceptance

- [ ] Step `2ea7e88c` produces zero rows.
- [ ] Step `a757e2a7` produces exactly one **existing `release.yml`** row, direction `widened`, naming `secrets: inherit`. Its separately added `sourceforge-mirror.yml` retains an `added` row; the acceptance is not permission to omit that new file.
- [ ] Existing goldens with genuine workflow widenings are unchanged.
- [ ] Negative control: both real-commit fixtures fail on `main` at `396de3b1`.

Found by the zero-noise harness built for the v1.0 roadmap. Parent #645. Related #657, #651, #660, #683.

## Implementation evidence and scope (2026-09-11)

The exact upstream workflow files were replayed independently of the reduced committed fixtures. Before correction, `a757e2a7` shows the identical write-to-write widening and `2ea7e88c` shows read-to-read noise. With the repair, the former names the inherited-secret recipient and retains the separate new-workflow row; the latter yields zero capability rows. Full source files were read statically, never executed. Reduced fixtures preserve permissions, job identities, triggers and reusable calls, carry source-file hashes and use synthetic executable payloads.

Closed host schemas require a versioned workflow extension: v0.4 carries effective writes and reusable recipients, while v0.1–v0.3 remain readable. Legacy snapshots are not promoted to evidence that no recipient existed. Current Git-ref comparisons read both sides with the new reader, so first use requires no manifest or saved-baseline migration.

Named reusable secret mappings remain a demonstrated existing limitation, now deferred in #693; the initial release comparison covers `uses` and `secrets: inherit` without claiming arbitrary secret or callee semantics.

pengfei-threemoonslab:
Closes #685.

A script edit in a workflow with existing writes was reported as a high-severity widening whose before/after text was identical. Removing that noise alone would hide the real change in the same upstream example: a reusable job newly receives inherited secrets. This repair makes workflow grant rows follow effective job permissions and explicitly names the receiving workflow.

## What changes

- Keep artifact-content drift separate from capability changes. Effective permission context respects explicit job overrides, including `{}`, and treats equivalent inherited/explicit declarations as equal. Overridden workflow defaults do not become job authority.
- Preserve unknown repository defaults: removing an explicit restriction yields a named unknown change instead of silently claiming read-only access. Only demonstrated new writes, privileged triggers or inherited-secret recipients produce expansion signals.
- Render job identities, permission transitions, reusable target/ref and `secrets: inherit` in the existing rows. No callee is fetched or run, and no downstream effect is inferred.
- Advance the closed host inventory/baseline/drift schemas to 0.4 and advertise contract 34. Prior schemas/readers remain intact; legacy snapshots cannot acquire missing recipient evidence by default. Git-ref comparisons scan both sides with the current reader and need no manifest or saved baseline.

This moves the noise and first-value metrics on the existing `diff`/host-audit surface. It adds no command or verdict. The schema increment is required by the closed previous wire shape; operational-control semantics remain unchanged.

## Evidence

Pinned upstream workflow files from `DanielLavrushin/b4` were replayed with both old and new readers. At `a757e2a7`, the release row now names the inherited-secret recipient while the separate added workflow remains visible. At `2ea7e88c`, script-only edits yield zero capability rows. Reduced committed fixtures retain relevant configuration, source hashes and synthetic executable payloads; the upstream workflows were never executed.

The isolated original implementation fails nine of the first fourteen regression cases; all pass after correction. Additional tests cover overridden defaults, equivalent declarations, restoration of unknown defaults and legacy-baseline refusal. Existing 24 sample artifacts remain unchanged.

Local full regression passed **9,737 tests with 5 skips** before the final unknown-to-explicit direction guard. The final affected host/workflow/CLI selection passed **158 tests**; static-only **486 passed**, performance **4 passed**, schema generation/check, Ruff and whitespace checks passed. Exact-head GitHub CI and the formal review/address loop follow publication. The final guard keeps the visible high-risk write declaration but does not claim a proven widening against unknown prior defaults.

## Limits retained

Named secret remapping is an existing blind spot, separately reproduced and deferred in #693. A mixed untracked-baseline/blocked-host control crash found during route smoke is separately reproduced on the old implementation and deferred in #694. Neither is hidden by changing the schema invariant or growing this PR. Runtime secret availability, transitive callee behavior and external adoption are not claimed.


pengfei-threemoonslab (commented):
## Independent review — round 1

Reviewed published head `13beabece9e4c1a301e1305ed88eee0657b49ed7` against base `a3070edfb7e4a722b7a7f9c496ea36d28719042b`. I am the independent review subagent; I did not implement these changes. GitHub's published tree matches all 35 changed-file blob identities in my disposable review archive.

No actionable defect found within #685's scope. The final implementation separates workflow bytes from permission changes, compares actual job permission contexts, keeps unknown repository defaults visible without claiming proven widening, and names inherited-secret recipients. The earlier preparation findings about overridden top-level permissions and removed restrictions are addressed; preparation was not a formal review round.

Independent validation on this exact commit:

- 119 tests passed in 20.517 seconds: workflow capability comparison, host audit, capability diff and schema boundaries. This includes the final unknown-before guard and both reduced upstream scenarios through the real CLI.
- 49 independently constructed known-permission before/after pairs matched a set-based permission-expansion oracle.
- All seven previously published host-grant schema files remain byte-identical to the base commit. V4 is separate; legacy baseline loading preserves old evidence and comparison refuses unsupported proof.

Evidence: disposable archive `/private/tmp/shipgate-685-review1-fqrzq0s9`, with `review-identity.json`, `published-tree.json`, `review-tests.xml`, and `review-matrix.json`.

This review does not claim a full-suite rerun, runtime-effective permissions, named-secret mapping coverage (#693), or resolution of the independent control interaction (#694). Required CI and current merge-authority evidence remain separate obligations.

pengfei-threemoonslab:
## Address — round 1

Acknowledged the independent [round 1 review](https://github.com/ThreeMoonsLab/agents-shipgate/pull/695#pullrequestreview-5184305778) on `13beabece9e4c1a301e1305ed88eee0657b49ed7`. No actionable findings remain within #685, and no further code change was needed after review.

The final affected selection passed 158 tests; the independent reviewer additionally passed 119 tests and 49 permission oracle comparisons on the published commit. The local full-suite result of 9,737 passed / 5 skipped predates the final unknown-before guard; it is not presented as a full-suite result on this exact head. Current GitHub CI is running against the published head.

Named-secret remapping (#693) and the pre-existing mixed baseline/control failure (#694) remain explicitly deferred, without broadening this fix or claiming those gaps are resolved. This completes review/address loop 1. Merge still requires current required CI and fresh verifier authority.

Issue #688, opened 2026-09-11
A symlink or submodule anywhere in the base tree makes `shipgate diff` refuse the whole repository — 4 of 12 public repositories, including `CLAUDE.md -> AGENTS.md`


pengfei-threemoonslab:
## Observed

Zero-noise harness over 12 public repositories with committed host configuration, 2026-09-11, `main` at `396de3b1`, full clones. Four repositories failed on **every one of 40 first-parent steps** with the same shape:

| repository | refused at | what it is |
|---|---|---|
| `miantiao-me/bm.md` | `CLAUDE.md` | a symlink to `AGENTS.md` — the common one-instruction-file convention |
| `getsentry/sentry-mcp` | `.claude/skills` | a symlinked skills directory |
| `gtkx-org/gtkx` | `website/public/tasks-forms.png` | a symlinked image nowhere near any host surface |
| `awslabs/nx-plugin-for-aws` | `.git-secrets/git-secrets` | a symlinked git hook helper |

```
ConfigError: Git tree contains unsupported external binding at CLAUDE.md (mode 120000, type blob).
```

A third of the sample cannot run `diff` at all. Two of the four are refused because of a path the host reader would never open.

## Why

`_materialize_isolated_tree` (`cli/verify/git.py`, around line 1786) walks `git ls-tree -r -z` of the base commit and raises on any entry whose mode is `120000` (symlink) or `160000` (submodule), regardless of path. `archive_tree` calls it for every `diff`, so one symlink anywhere in the tree — a PNG under `website/` — refuses the entire comparison.

The refusal exists for a reason: a materialised symlink could point outside the scratch tree, and a submodule is content this checkout does not hold. Both are correct things to refuse *at the path the reader would read*. Refusing the repository because of an unrelated PNG is not that property; it is a repository-wide fail-closed on a per-path concern.

## Proposed

- **Materialise only what the reader reads.** Restrict the tree walk to the host-surface path set (the adapter exact paths and globs, plus their parent directories) instead of the whole tree. This also removes most of #686's cost, since the tree pack shrinks to a handful of files.
- **A symlink at a host-surface path is an inventory issue, not a refusal.** `CLAUDE.md -> AGENTS.md` is recorded as `symlinked_instruction_file` with the target path; if the target is inside the tree, read the target and attribute the grant to both paths; if it points outside, record the issue as blocking for that artifact only.
- A submodule at a host-surface path is recorded as `unsupported_submodule` for that artifact only.
[...]

Issue #683, opened 2026-09-11
`shipgate diff` dies with a traceback on any shallow clone; the integrity check names no next step


pengfei-threemoonslab:
## Observed

On a repository cloned with `--depth N` (any N), `shipgate diff --base <sha>` exits with a full Rich traceback ending in:

```
ConfigError: Copied Git object graph failed integrity validation: broken link
from  commit e26b65fddbad62f59e1a498508cc75c92783690e
              to  commit f81b2939b9804ea92a5217b7b712158799a56788
missing commit f81b2939b9804ea92a5217b7b712158799a56788
```

Reproduced on 12 of 12 public repositories in the zero-noise measurement run for the v1.0 roadmap (2026-09-11). `git fetch --unshallow` on the same clone makes the same command succeed.

## Why it happens

`run_capability_diff` materialises the base tree through `archive_tree`, which copies the object graph into an isolated directory and runs `git fsck`. A shallow clone has a grafted boundary, so `fsck` reports the missing parent as a broken link and the copy is refused. The fsck exists to catch a *tampered* graph; a shallow boundary is not tampering, and the two are indistinguishable to it.

## Why it matters for v1.0

- `actions/checkout` defaults to `fetch-depth: 1`. The docs require `fetch-depth: 0` in six places, but a user who forgets gets this traceback in CI rather than a sentence.
- Every developer who runs `git clone --depth 1` — the default in most CI images and many onboarding guides — gets a traceback on the very first `diff`.
- The message is engine vocabulary ("copied Git object graph", "integrity validation") with no next step. This is the #652 vocabulary class plus the #650 termination class: a terminal with nothing to follow.

## Proposed

- Detect a shallow repository before archiving (`git rev-parse --is-shallow-repository`) and either (a) skip the graph copy for a base that is reachable and unshallow-safe, or (b) fail closed with one sentence and a runnable next command: `This checkout is shallow; run git fetch --unshallow (or set fetch-depth: 0 in CI) and rerun`.
- Never print a traceback for this condition.
- Add the shallow case to `tests/test_capability_diff.py` with a negative control on the pre-fix behaviour.

## Acceptance

- [ ] `shipgate diff` on a `--depth 1` clone prints one sentence and a next command, exit code non-zero, no traceback.
- [ ] The same clone after `--unshallow` produces rows.
- [ ] The zero-noise harness (#660 driver) runs on shallow clones without a wrapper.

[...]

Issue #700, opened 2026-09-12
Read through an in-tree symlink at a boundary path: `CLAUDE.md -> AGENTS.md` and `.claude/skills -> ../.agents/skills` still leave the inventory incomplete


pengfei-threemoonslab:
## Observed

After #686/#688 scoped the base tree, an unrelated symlink no longer affects a comparison. A symlink **at** a boundary path still does. Measured 2026-09-11 on the twelve-repository sample:

| repository | link | effect |
|---|---|---|
| `miantiao-me/bm.md` | `CLAUDE.md -> AGENTS.md` | blocking `unreadable`; inventory incomplete; comparison refused |
| `awslabs/nx-plugin-for-aws` | `AGENTS.md -> …` | same |
| `getsentry/sentry-mcp` | `.claude/skills -> ../.agents/skills`, `.cursor/skills -> …` | same |

Three of twelve public repositories. `CLAUDE.md -> AGENTS.md` is the standard way to keep one instruction file under two names, so this is not an edge case.

The live worktree behaves identically — this is the reader, not the archive. `_repository_paths` matches the symlink by its own name, the read opens it with `O_NOFOLLOW`, the open fails, and the artifact is recorded `failed` with a blocking `unreadable` issue.

## Why it is not simply "follow the link"

`O_NOFOLLOW` is load-bearing. The identity-bound read exists so that a path validated during enumeration is the same object that is read, and so that a link cannot redirect a read outside the workspace between validation and use. Following links naively reintroduces both problems, and #614 landed a concealment guard in this area that must keep holding.

## Proposed

- At a boundary path, resolve the link **within the workspace**: read `readlink`, join it against the link's parent, and require the result to stay under the workspace root after normalization, with a bounded number of hops.
- Open the resolved target with the existing `O_NOFOLLOW` descriptor read, so the object actually read is still a regular file opened without following anything.
- Attribute the grant to the link's path, and record `resolved_through: <target>` on the artifact so a reviewer sees why two paths carry one grant. A server declared once under two names stays one capability, as in the duplicate-MCP-declaration rule.
- A link that escapes the workspace, dangles, or exceeds the hop bound keeps today's behaviour exactly: `failed` artifact, blocking `unreadable` issue on that artifact only.

## Acceptance

- [ ] The three repositories above produce rows or a correct covered no-change comparison; no blocking issue for the link itself.
[...]

Issue #698, opened 2026-09-12
Measure and reduce advisory diff first-run latency after the tree-reader fix


pengfei-threemoonslab:
## Observed

While implementing #686, the same pinned `DanielLavrushin/b4` comparison dropped from 16.622 s with the complete-history reader to five fresh-process samples of **2.152, 1.878, 1.884, 1.936, 1.955 s** with the tree reader (median 1.936 s). Both readers returned the same two rows. Filesystem caches were not flushed; this is not an installed-release or controlled cold-cache result.

Head: `a757e2a7b95327c296cae74d0b2e96da56e16dfc`; base: `93f38e988aaacdbca6b4a00cf9850b4b9b0dfbc9`. The checked-out tree has 1,149 files / 21,960,656 bytes. No measured-repository code was executed.

The first sample still exceeds #686's original ≤2 s per-step aspiration. That shortfall must not disappear behind a median or a fastest-run claim. The history-size bottleneck and unnecessary shallow-clone fetch are independently fixable; proving a strict first-run ceiling is additional work.

## Proposal

The two prerequisites for useful measurement now exist: #660 has the wider repository population, and #570 records a published 1.0.0 distribution. Measure that installed release outside a source checkout, then compare any proposed repair on identical pinned inputs. The old source-process samples remain historical; do not relabel them as installed or cold-cache data.

Preserve every sample, including failed and slow runs. Separate Python/dependency startup, Git base materialization, inventory/comparison and report rendering. State machine, operating system, Python, wheel digest, history/cache condition, sample count and whether filesystem caches were actually controlled. Distinguish a fresh process from a cold filesystem cache.

Measure the selected CLI/PR workflow before choosing an optimization. A slow install or CI environment setup is not an engine regression; a long local Stop path is separately covered by #661. Optimize only the dominant demonstrated phase, retaining object validation, coverage, output and determinism. Avoid new persistent caches or combined traversal rewrites without evidence that those mechanisms are needed.

## Acceptance

- [ ] Record all first-run and subsequent-run samples for installed 1.0.0 on the pinned b4 pair and representative #660 inputs, including refusals/failures and their time. Summaries include the full denominator and a clearly described distribution, not only the best run or median.
- [ ] Decompose the dominant phases and distinguish process startup, materialization and inventory before selecting an implementation. Record the installed wheel and environment so the comparison is reproducible.
[...]

PR #703, opened 2026-09-12
Read the base tree the host reader will read, and no more (#686, #688)

### LINKED ISSUE #686: `shipgate diff` repacks the repository's entire reachable history to read one base tree: 27 s per call on a 26 MB repository, 1 s on a toy one
## Observed

Zero-noise harness, 2026-09-11, `main` at `396de3b1`. `shipgate diff --base <parent>` per first-parent step:

| repository | working-tree size | per-step wall time |
|---|---:|---:|
| synthetic S1 scenario (3 files, 2 commits) | ~1 KB | **1 s** |
| `DanielLavrushin/b4` (full clone) | 26 MB | **27 s** |

`git fsck --connectivity-only` on the same clone takes 0.09 s, so the cost is not the integrity check.

## Why

`run_capability_diff` materialises the base tree through `archive_tree`, which calls `_copy_verified_commit_graph` (`cli/verify/git.py:1714`). That runs

```
git pack-objects --stdout --revs   <<< "<commit>"
```

with the base **commit** as the only rev. `--revs` packs everything reachable from it: every ancestor commit, every historical tree, every historical blob — the repository's whole history up to that point — then `index-pack`s it into a fresh bare store and `fsck --strict`s the result. `diff` needs one tree. It pays for the history.

The same walk is why #683 happens: reachability from a commit crosses the shallow boundary and `fsck` reports the grafted parent as a broken link.

## Why it matters for v1.0

- Gate 3 (#660) is "a non-empty row in ≤ 2 commands and ≤ 5 minutes". One `diff` at 27 s is a tenth of that budget on a *small* repository; on a repository with real history it will not fit.
- The local guard (#661) budgets 1.5 s for the Stop hook and would call this path.
- A command that takes half a minute to answer "did this PR widen anything" is a command people stop running.

## Proposed

Pack the **tree**, not the commit. `git pack-objects --revs` accepts `<commit>^{tree}` and then packs only the objects that tree references — no ancestors, no history. The isolation property `_copy_verified_commit_graph` exists for (verified objects in a store the working checkout cannot influence) is preserved; the `fsck` still runs, on a graph that has no parents to be missing. This closes #683 as a side effect, because a tree does not cross a shallow boundary.

`verify`, which materialises base trees for a release decision, may keep the commit-graph copy if its trust argument depends on commit ancestry; that is a separate question and should be answered on this issue before the two paths diverge.

## Acceptance

- [ ] `shipgate diff` on the b4 clone: ≤ 2 s per step (measured on the same machine).
- [ ] `shipgate diff` on a `--depth 1` clone produces rows (#683 acceptance) without a special case.
- [ ] `tests/test_capability_diff.py` gains a case asserting the scratch store contains no commit objects other than the base commit itself.
- [ ] `verify`'s materialisation is either unchanged with the reason recorded, or moved to the same tree pack with its goldens unchanged.

Found by the zero-noise harness built for the v1.0 roadmap. Parent #645. Related #683, #651, #660, #661.

## Implementation scope, reconciled 2026-09-12

The tree-reader change removes complete-history packing from advisory `diff` and the manifest-free PR comparison added in #684. Configured release verification retains its complete ancestry validation; its trust argument has not been weakened for latency. Tests assert the advisory scratch store contains **no commit objects**, while the configured release store still contains ancestry. Available shallow trees work directly; genuinely missing comparison history retains fetch recovery.

The original ≤2 s **every-step** aspiration is not yet established. Same-machine measurements against pinned b4 inputs were 16.622 s before, then 2.152 / 1.878 / 1.884 / 1.936 / 1.955 s after (median 1.936 s), with identical rows. #698 explicitly carries the remaining first-run latency measurement/optimization after readiness unless the broader #660 corpus demonstrates a selected first-value blocker. Do not check off the original two-second criterion as fully satisfied or infer #661's hook budget from this result.

Delivery of this issue now means removal of the history-size bottleneck with those integrity/shallow-clone tests and transparent timing evidence. #698 owns the residual performance criterion; #660 and #570 still own corpus and installed-candidate acceptance. This split follows adoption > completeness: ship the supported comparison improvement without adding a universal cold-start claim or a new readiness gate.

pengfei-threemoonslab:
Closes #686.
Related #688; remaining reader coverage is deferred under #700/#711.

## Why

`shipgate diff` and verify's host comparison materialized the whole base tree to read a handful of files. Two defects followed, both measured on twelve public repositories with committed host configuration.

**Cost.** Materialization runs one `git cat-file` per blob. On a 26 MB repository with 1,168 files whose host surface is four files, that is 1,168 subprocesses.

| | before | after |
|---|---:|---:|
| `DanielLavrushin/b4`, one `diff` | 25.3 s | **2.7 s** |
| the toy S1 scenario | 1 s | 1 s |

The pack was never the cost — 200 ms — the subprocesses were.

**Reach.** Every entry was checked for portability, name collision and external binding, so anything anywhere refused the whole repository. Four of twelve repositories were uncomparable on **every step** of their history:

| repository | refused for |
|---|---|
| `miantiao-me/bm.md` | `CLAUDE.md -> AGENTS.md` |
| `getsentry/sentry-mcp` | a symlinked `.claude/skills` |
| `gtkx-org/gtkx` | three symlinked PNGs under `website/public/` |
| `awslabs/nx-plugin-for-aws` | a symlinked git-hook helper |

Two of those four are refused for a path no reader would ever open.

## What changed

`archive_tree` takes a `scope` predicate. Scoped, it packs the tree rather than the commit — no ancestry, which is also why a shallow clone can be compared against a commit it already holds — and materializes only what the predicate accepts, plus symlinks. The predicate is `is_boundary_surface_path`, placed in the registry that already owns path classification, so the archive and the live reader cannot disagree about what the surface is.

Scoping narrows what is **materialized**, never what is **verified**: every blob written is still checked against its object ID and the isolated store is still `fsck`'d. An unscoped archive is unchanged in every respect, including refusing symlinks — verify materializes trees for the release decision, and narrowing the advisory host read must not loosen that.

A scoped archive recreates symlinks rather than refusing their tree. The reader still handles them conservatively; successful archive construction does not establish complete coverage. The attempted target-type optimization was withdrawn during independent round 2 because an unbound `os.stat` result could become stale and manufacture a complete inventory. #700/#711 retain identity-bound target handling as explicit later work.

## Review fixes

- A shallow merge can expose an older common ancestor around a graft that hides a newer one. Both CLI diff and manifest-free PR comparison now require proof that no remaining root lies outside the selected ancestor closure, while a visibly ancestral tip remains directly comparable. Real clone/fetch regressions cover the wrong-base case, known ancestor, HEAD and fetch recovery.
- A symlink-target replacement regression preserves incomplete coverage through enumeration and final validation. The old optimization-specific tests were replaced with explicit file/directory/dangling refusal checks; no safety assertion was weakened to keep the optimization.

## No guard was added where one could not fire

The recreated symlink invites a blob under a symlinked ancestor. A valid tree cannot express it: the name would be both a blob and a tree, a duplicate entry `fsck --strict` rejects before anything is written. An ancestor check in the materializer could therefore never fire, and an unreachable guard is not a guard. The invalid tree is built by hand in the tests and asserted to be refused; blobs are still written before links regardless.

## Two existing tests rewritten, not deleted

`test_shallow_comparison_retains_recovery_instead_of_init` asserted that a shallow clone is incomparable. It now asserts the comparison it can actually answer, and a second case covers a base the clone genuinely does not have, which still routes to `fetch_base` with `refs_missing`. Telling someone to fetch history in order to compare a commit they already have is a recovery for a problem they do not have.

Static-only allowlist line pins moved 2001 → 2092 and 2245 → 2336.

## Evidence

The original 13-case preparation included an unsafe symlink optimization withdrawn in review. The final tests retain scoped materialization, shallow trees, invalid trees, collisions and conservative link limits, with a new target-replacement regression. Final validation is recorded in the review/address response; the original negative-control count is not relabeled as final evidence.

The initial aggregate incorrectly counted `incomparable` as a benign zero-row result. The corrected combined-branch measurement is **8/12 usable repositories and 140 benign comparable steps with zero rows**, as recorded in [the correction](https://github.com/ThreeMoonsLab/agents-shipgate/pull/703#issuecomment-5644592850). That sample combines this PR with other changes; it is not standalone acceptance evidence for this head. The four incomplete repositories retain explicit symlink limits (#700/#711). Refusal becoming an honest incomplete result is useful diagnosis, not completed comparison coverage.

#698 retains the residual first-run latency criterion. The two-second every-run aspiration and #661's hook budget are not established by these measurements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+- Compare a shallow checkout against a base it already holds. `diff`
+  refused every shallow checkout, so `actions/checkout`'s default
+  (`fetch-depth: 1`) was sent to `git fetch --unshallow` for a comparison
+  the tool could already answer: the base tree is read through a scoped
+  archive that walks no ancestry. The recovery now fires when the base
+  ref or the merge base is genuinely unavailable, which is what Git
+  reports past a graft — measured on two diverged shallow clones, it
+  returns nothing rather than a wrong commit (#686).
+
 - Read the base tree the host reader will read, and no more. `shipgate diff`
   and verify's host comparison now materialize only boundary-surface paths
   (plus symlinks, which are what can conceal them),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ## Unreleased
 
+- Read the base tree the host reader will read, and no more. `shipgate diff`
+  and verify's host comparison now materialize only boundary-surface paths
+  (plus symlinks, which are what can conceal them),
+  packing the tree rather than the commit. A 26 MB repository with 1,168 files
+  and four host files took 25s and now takes 2.7s, and a shallow clone is
+  comparable for a commit it already holds. An unrelated symlink no longer
+  makes a repository uncomparable: one outside the surface is not
+  materialized, a symlink to a non-directory can conceal no boundary path,
+  and a scoped archive recreates links instead of refusing the tree. Unscoped
+  archives, which verify uses for the release decision, still refuse every
+  external binding (#686, #688).
+
 - Manifest-free host PR review now names capability changes in verify, PR comments,
   and check. Interactive check defaults to readable text; agent mode keeps JSON.
   Comparison evidence binds its input refs, excludes stale scan artifacts, and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,26 +2,16 @@
 
 ## Unreleased
 
-- Compare a shallow checkout against a base it already holds. `diff`
-  refused every shallow checkout, so `actions/checkout`'s default
-  (`fetch-depth: 1`) was sent to `git fetch --unshallow` for a comparison
-  the tool could already answer: the base tree is read through a scoped
-  archive that walks no ancestry. The recovery now fires when the base
-  ref or the merge base is genuinely unavailable, which is what Git
-  reports past a graft — measured on two diverged shallow clones, it
-  returns nothing rather than a wrong commit (#686).
-
-- Read the base tree the host reader will read, and no more. `shipgate diff`
-  and verify's host comparison now materialize only boundary-surface paths
-  (plus symlinks, which are what can conceal them),
-  packing the tree rather than the commit. A 26 MB repository with 1,168 files
-  and four host files took 25s and now takes 2.7s, and a shallow clone is
-  comparable for a commit it already holds. An unrelated symlink no longer
-  makes a repository uncomparable: one outside the surface is not
-  materialized, a symlink to a non-directory can conceal no boundary path,
-  and a scoped archive recreates links instead of refusing the tree. Unscoped
-  archives, which verify uses for the release decision, still refuse every
-  external binding (#686, #688).
+- Advisory `diff` and manifest-free PR review materialize boundary paths and
+  symlinks from a verified tree instead of copying commit ancestry and writing
+  every file. The original sample improved from 25s to 2.7s; residual latency
+  remains #698, not a universal two-second promise. Shallow comparisons verify
+  that a visible common ancestor does not hide a newer one beyond a graft;
+  unresolved ancestry requests a fetch, while HEAD and proven ancestor
+  comparisons remain usable. Configured release archives retain their existing
+  whole-history validation. Scoped archives preserve links, but the reader
+  remains conservative until link-target identity can be bound (#700/#711):
+  successful materialization is not complete inventory coverage (#686/#688).
 
 - Manifest-free host PR review now names capability changes in verify, PR comments,
   and check. Interactive check defaults to readable text; agent mode keeps JSON.
--- a/src/agents_shipgate/cli/diff.py
+++ b/src/agents_shipgate/cli/diff.py
@@ -45,6 +45,7 @@ def _resolve_base(workspace: Path, base: str | None) -> tuple[str, str]:
         commit_sha,
         detect_default_base,
         merge_base_sha,
+        shallow_merge_base_is_proven,
     )
 
     if base is not None and (not base.strip() or base.startswith("-")):
@@ -108,24 +109,19 @@ def _resolve_base(workspace: Path, base: str | None) -> tuple[str, str]:
             "common point to compare from.",
             param_hint="--base",
         )
-    # Shallow is refused for the comparisons it actually breaks, not for
-    # being shallow. The base tree is read through a scoped archive that
-    # packs the tree and walks no ancestry, so a base this clone already
-    # holds is readable at `--depth 1`. What shallowness can still break is
-    # the *choice* of base: `git merge-base` cannot see past a graft, so a
-    # merge base that is itself a graft may not be the real one, and
-    # comparing against the wrong base silently is worse than refusing.
-    # Refusing unconditionally sent `actions/checkout`'s default
-    # (`fetch-depth: 1`) to `git fetch --unshallow` for a comparison the
-    # tool could already answer (#686).
-    # No graft check here on purpose. `git merge-base` does not guess past a
-    # shallow boundary — it reports nothing, which the branch above turns
-    # into the `--unshallow` recovery. Verified on two diverged shallow
-    # clones whose true base lay past the graft: both returned empty, never
-    # a wrong commit. So a merge base that *is* returned was proven through
-    # commits this clone holds, and no absent commit can be a more recent
-    # common ancestor. A graft check could not be made to fire on a wrong
-    # answer, and a guard that cannot fire is not a guard.
+    if truncated:
+        # A merge can expose an older common ancestor along one parent while
+        # a shallow graft hides a newer one along another. A nonempty answer
+        # alone is therefore insufficient. After excluding the candidate and
+        # its ancestry, every visible path from both tips must terminate: any
+        # remaining root may be a graft hiding a better common ancestor.
+        # HEAD/self and fully visible paths to the candidate remain usable.
+        base_commit = commit_sha(workspace, requested)
+        head_commit = commit_sha(workspace, "HEAD")
+        if base_commit is None or head_commit is None:
+            refuse_shallow()
+        if not shallow_merge_base_is_proven(workspace, base_commit, head_commit, resolved):
+            refuse_shallow()
     return requested, resolved
 
 
--- a/src/agents_shipgate/cli/diff.py
+++ b/src/agents_shipgate/cli/diff.py
@@ -50,7 +50,7 @@ def _resolve_base(workspace: Path, base: str | None) -> tuple[str, str]:
     if base is not None and (not base.strip() or base.startswith("-")):
         raise typer.BadParameter("Base ref must be non-empty and cannot start with a dash.", param_hint="--base")
 
-    if _history_is_truncated(workspace) is True:
+    def refuse_shallow() -> None:
         from agents_shipgate.cli.agent_mode import emit_agent_mode_error_action
         from agents_shipgate.invocation import join_argv
         from agents_shipgate.schemas.diagnostics import NextAction
@@ -73,6 +73,9 @@ def _resolve_base(workspace: Path, base: str | None) -> tuple[str, str]:
         )
         raise typer.Exit(2)
 
+    truncated = _history_is_truncated(workspace) is True
+
+
     # Same resolver `check` uses (#649), including its narrow local
     # fallback: a local `main` is refused while a remote exists, because the
     # remote is the authority it might be stale against, and used only where
@@ -87,6 +90,10 @@ def _resolve_base(workspace: Path, base: str | None) -> tuple[str, str]:
             param_hint="--base",
         )
     if commit_sha(workspace, requested) is None:
+        if truncated:
+            # The ref is missing and history is cut: fetching is the repair,
+            # and it is a more useful answer than "fetch it first".
+            refuse_shallow()
         raise typer.BadParameter(
             f"Base ref {requested!r} is not available locally. Fetch it first; "
             "this command never fetches.",
@@ -94,11 +101,31 @@ def _resolve_base(workspace: Path, base: str | None) -> tuple[str, str]:
         )
     resolved = merge_base_sha(workspace, requested, "HEAD")
     if resolved is None:
+        if truncated:
+            refuse_shallow()
         raise typer.BadParameter(
             f"No merge base between {requested!r} and HEAD, so there is no "
             "common point to compare from.",
             param_hint="--base",
         )
+    # Shallow is refused for the comparisons it actually breaks, not for
+    # being shallow. The base tree is read through a scoped archive that
+    # packs the tree and walks no ancestry, so a base this clone already
+    # holds is readable at `--depth 1`. What shallowness can still break is
+    # the *choice* of base: `git merge-base` cannot see past a graft, so a
+    # merge base that is itself a graft may not be the real one, and
+    # comparing against the wrong base silently is worse than refusing.
+    # Refusing unconditionally sent `actions/checkout`'s default
+    # (`fetch-depth: 1`) to `git fetch --unshallow` for a comparison the
+    # tool could already answer (#686).
+    # No graft check here on purpose. `git merge-base` does not guess past a
+    # shallow boundary — it reports nothing, which the branch above turns
+    # into the `--unshallow` recovery. Verified on two diverged shallow
+    # clones whose true base lay past the graft: both returned empty, never
+    # a wrong commit. So a merge base that *is* returned was proven through
+    # commits this clone holds, and no absent commit can be a more recent
+    # common ancestor. A graft check could not be made to fire on a wrong
+    # answer, and a guard that cannot fire is not a guard.
     return requested, resolved
 
 
--- a/src/agents_shipgate/cli/diff.py
+++ b/src/agents_shipgate/cli/diff.py
@@ -20,6 +20,7 @@ from pathlib import Path
 import typer
 
 from agents_shipgate.cli.workspace_guard import require_workspace
+from agents_shipgate.core.boundary_registry import is_boundary_surface_path
 from agents_shipgate.core.capability_diff_rows import (
     ABSENT,
     CapabilityDiffRow,
@@ -151,7 +152,9 @@ def run_capability_diff(
 
         base_tree = Path(scratch) / "base"
         base_tree.mkdir()
-        archive_tree(workspace, base_commit, base_tree)
+        archive_tree(
+            workspace, base_commit, base_tree, scope=is_boundary_surface_path
+        )
         base_inventory = build_host_boundary_snapshot(
             base_tree, cache=HostStaticParseCache()
         ).inventory
--- a/src/agents_shipgate/cli/verify/git.py
+++ b/src/agents_shipgate/cli/verify/git.py
@@ -7,6 +7,7 @@ import subprocess
 import tempfile
 import threading
 import unicodedata
+from collections.abc import Callable
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Any, Literal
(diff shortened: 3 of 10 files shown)


pengfei-threemoonslab (commented):
Reviewed by replaying the twelve-repository sample this PR was measured on. The scoping design is right, the reasoning about what it does and does not weaken is right, and the "no guard was added where one could not fire" section is the correct call — I confirmed a valid tree cannot express a blob under a symlinked ancestor, and `_within` resolves through links so the second path is covered too.

Two things I checked and found fine, recorded so nobody re-derives them:

- The final `materialized` digest walk excludes symlinks (`path.is_file() and not path.is_symlink()`), so nothing is read *through* a link. I also checked whether `rglob` escapes through a symlinked directory on this interpreter — it does not (3.12 skips symlinked dirs when recursing), so the walk stays inside the scratch tree.
- `is_boundary_surface_path` accepts directory prefixes of `exact_paths` but not of `globs` (`.github/workflows`, `.agents/skills`, `.agents-shipgate` all return False). Harmless: `ls-tree -r` never emits a tree entry, parents are created by `mkdir(parents=True)` when a matching blob lands, and symlinks bypass the scope check entirely. Worth knowing before someone "fixes" the asymmetry.

## One blocking finding: `shipgate diff` cannot use what this PR built

`cli/verify/host_comparison.py` gets `scope=is_boundary_surface_path`. `cli/diff.py` does not, and it still carries the unconditional pre-emptive refusal from #683:

```python
if _history_is_truncated(workspace) is True:
    ...  # "This checkout is shallow; run git fetch --unshallow"
    raise typer.Exit(2)
```

That fires before any materialization is attempted, so it refuses a shallow clone **whose base commit and base tree are both present**. Measured just now on five shallow clones (`--depth 3`, base = `HEAD~1`, base object verified present with `cat-file -t`):

| repository | `shipgate diff` | `archive_tree(scope=...)` directly |
|---|---|---|
| `miantiao-me/bm.md` | refused: run `--unshallow` | **OK, 8 files, 0.5 s** |
| `DanielLavrushin/b4` | refused | **OK, 3 files, 1.0 s** |
| `gtkx-org/gtkx` | refused | **OK, 8 files, 1.4 s** |
| `getsentry/sentry-mcp` | refused | **OK, 36 files, 1.5 s** |
| `awslabs/nx-plugin-for-aws` | refused | **OK, 9 files, 2.1 s** |

The unscoped archive fails on all five (the shallow boundary), which is what the guard was written for. The scoped archive this PR adds succeeds on all five. So the capability landed and the command that the roadmap makes the product's default form still refuses.

This PR's own body states the principle: *"Telling someone to fetch history in order to compare a commit they already have is a recovery for a problem they do not have."* That is applied to `verify`'s test and not to `diff`.

It matters beyond tidiness: `actions/checkout` defaults to `fetch-depth: 1`, and #660's cold-start harness clones 30 public repositories where shallow is the natural clone. "≤ 2 commands, ≤ 5 minutes" does not survive `git fetch --unshallow` on a large repository.

**Suggested fix** — in `_resolve_base`, make the refusal conditional on the base actually being unavailable rather than on the repository being shallow. The information is already there: `commit_sha(workspace, base)` resolves, and `<commit>^{tree}` resolves, precisely when the scoped archive can proceed. Keep the existing message and `next_action` for the case where the base genuinely is missing — that recovery is correct and #692's tests should stay green on it.

Acceptance I would want: the five clones above produce rows or a correct empty comparison; a shallow clone whose base is genuinely absent still routes to `--unshallow`; a negative control showing the new case fails on this PR's head.

Everything else here looks good to me: full suite green on `0a4f12a6` (`pytest -n auto`, exit 0), ruff clean, and the 8-of-13 negative control on pre-fix `main` is the right shape.

pengfei-threemoonslab:
Addressed the finding on this branch (`9cd5f7e1`).

`_resolve_base` no longer refuses a shallow checkout for being shallow. It refuses when the base ref or the merge base is genuinely unavailable, which is the case the recovery was written for.

**What I found while writing the guard.** My first attempt kept a second check: refuse when the resolved merge base is a shallow graft, on the theory that `git merge-base` might return a graft as a best-effort answer past the boundary. Two things came out of testing it:

- `git rev-parse --show-shallow` **is not a real option**. Git 2.50 echoes it back as a literal with exit 0, so a helper built on it reports every boundary as empty — the guard would have been silently inert. The boundary is in the file Git keeps, located through `rev-parse --git-path shallow`.
- More importantly, the check could not be made to fire on a wrong answer. On two diverged shallow clones whose true merge base lay past the graft, `git merge-base` returned **empty**, not a wrong commit — which the `resolved is None` branch already turns into the recovery. And a graft that *is* returned is a genuine merge base: both tips reach it through commits the clone holds, so no absent commit can be a more recent common ancestor.

So I removed it, along with the helper and its tests. A guard I cannot make fire is not a guard, and this repository has paid for that class before. The reasoning is recorded as a comment where the check would have gone, so the next person does not re-add it.

**Measured**, five public repositories at `--depth 3`, base `HEAD~1`:

| | before | after |
|---|---|---|
| `DanielLavrushin/b4` | refused | 2.4 s, correct empty comparison |
| `miantiao-me/bm.md` | refused | 1.6 s |
| `getsentry/sentry-mcp` | refused | 2.8 s |
| `gtkx-org/gtkx` | refused | 3.0 s |
| `awslabs/nx-plugin-for-aws` | refused | 4.5 s |

Four of the five still report an input limit rather than rows — that is the symlink-at-a-boundary-path incompleteness you already filed as #700, unchanged by this.

**Tests.** `test_shallow_checkout_names_a_recovery_that_restores_the_diff` passed `--base HEAD`, which needs only HEAD's tree and is present at `--depth 1` — the same defect in test form. It now names a base the clone genuinely lacks (the graft's parent), keeps the full recovery-and-rerun walk, and keeps all eight parametrizations. `test_a_shallow_clone_compares_against_a_base_it_already_holds` is new and covers the case that was refused; both its parametrizations fail against this branch's previous head.

Full suite green (`pytest -n auto`, exit 0), ruff clean. Nothing in the scoping design changed — this only stops the command from refusing before it uses it.

pengfei-threemoonslab:
One more finding from replaying the full twelve-repository sample on the combined branch (#703 + #704 + #657), filed as **#711** rather than pushed here because the fix is a design choice that belongs to you.

`awslabs/nx-plugin-for-aws` returns `incomparable` / `base_inventory_incomplete` on every step. Two symlinks, and only one of them is #700:

```
.git-secrets/git-secrets -> ../packages/nx-plugin/src/preset/.../git-secrets
```

In the working tree that target is present, the link resolves to a file, `_symlink_may_be_directory` answers False, and the reader says nothing. In the scoped base archive the target is not host surface, so it is not materialized, the link dangles, the fail-closed branch treats it as possibly-concealing, the `O_NOFOLLOW` read fails, and the base carries an `unreadable` the head does not.

So scoping manufactures an incompleteness that does not exist in the tree being reviewed. The property this PR states — *"the base tree must present the reader the same object or the two sides disagree about what the file is"* — holds for the link and breaks for what the link resolves to, because `_symlink_may_be_directory` is a question about the target.

#711 lays out three directions and why I did not just pick one. The short version: materializing targets reintroduces the unbounded cost #686 removed; restricting which symlinks get materialized reopens the `**/`-glob asymmetry inverted; marking scope-induced dangling as distinct from repository dangling puts the knowledge where it actually exists, which reads best to me, but it is your design.

Everything else from the re-measurement is good, and it is a large change:

| | before | after |
|---|---:|---:|
| repositories producing a usable answer | 7 of 12 | **11 of 12** |
| benign steps measured | 174 | **281** |
| benign steps with rows (noise) | 0 | **0** |
| host-touching steps that produced rows | 60, nearly all false | **24** |
| median step | 27 s on the 26 MB repository | **2.2 s** (max 6.3 s) |

`getsentry/sentry-mcp`, `gtkx-org/gtkx` and `miantiao-me/bm.md` went from refused-on-every-step to 40, 40 and 40 clean steps. `Vortiago/mcp-outline` went from 31 of 31 host steps rowed to 3 of 31 — that repository was almost entirely the workflow false-widening, and #685 plus this PR account for the rest.

pengfei-threemoonslab:
**Correction to my previous comment.** I wrote "11 of 12 repositories producing a usable answer". That is wrong, and the error was mine: my aggregation script counted a step whose status is `incomparable` as a benign step with zero rows, because it only special-cased `exit*` and `skipped*`. So five repositories that answer `incomparable` on every step were being scored as clean.

The correct figures from the completed run, twelve repositories, 40 first-parent steps each:

| | first run (2026-09-11) | this branch |
|---|---:|---:|
| repositories producing a usable answer | 7 of 12 | **8 of 12** |
| benign steps measured | 174 | 140 |
| **benign steps with rows (noise)** | **0** | **0** |
| host-touching steps that produced rows | 60, nearly all false | **26** |
| median step / max | 27 s on the 26 MB repository | **1.8 s / 4.2 s** |

The eighth is `Doist/todoist-mcp`, and it is not this PR — it is #712, an unrelated frontmatter bug this measurement turned up. The four still incomparable are `getsentry/sentry-mcp`, `miantiao-me/bm.md`, `gtkx-org/gtkx` and `awslabs/nx-plugin-for-aws`, all `base_inventory_incomplete`, all symlink-related: #700 for a link at a boundary path, #711 for a link whose target scoping leaves out.

What this PR does change is real and I do not want the correction to bury it: those four went from **refusing with an object-integrity error** to **naming an incomplete inventory**, the workflow false-widening class is gone (`Vortiago/mcp-outline` from 31 of 31 host steps rowed to 3 of 31), and the median step went from 27 s to 1.8 s. Reach did not improve as much as I said, because refusing and answering "incomparable" are both unusable to a reviewer — the difference is that one is now honest about why.

The benign-noise number — the one the "ten real repositories, zero noise" judgment names — is 0 of 140, holding across both runs.

pengfei-threemoonslab (commented):
Independent coding-agent review — round 2 of at most 3, against `9cd5f7e156b8bc88435187e64211b46ef604605d`. This is a COMMENT review, not human release approval. Changes required before merge.

## P1 — A non-empty shallow merge base can still be wrong

Location: `src/agents_shipgate/cli/diff.py:102-129` (particularly the assertion at lines 121-128).

The removed unconditional refusal should not return, but the replacement assumes that Git either finds the correct merge base or returns nothing. A normal clone/fetch counterexample disproves that:

- A is the root; B descends from A; X from B; L from X; Q independently from A.
- H merges L and Q; R descends from B.
- Clone branch H at depth 3, then fetch branch R at depth 3 into `origin/base`.
- Full-history `merge-base H R` is B. In the shallow clone X is a graft, hiding H's path to B; the visible H → Q → A path makes Git return A, a non-empty but older common ancestor.

With narrow Claude permissions at A and unchanged wider permissions at B/X/L/H/R, the full `diff` correctly returns `comparable` with zero rows. The current shallow command returns `comparable` with a **critical Bash(*) addition** and a permission removal. No shallow metadata was hand-edited.

The existing depth-1/depth-2 tests cover self/HEAD and absent bases, not an incomplete merge graph with an older visible common ancestor. Add this negative control and retain direct self/known-ancestor comparisons. For genuinely divergent shallow tips, establish that no unexplored ancestry can hide a newer common ancestor; otherwise use the existing fetch recovery. A roots-outside-the-resolved-closure check catches this counterexample, but should preserve the trivially exact case where the resolved base equals either tip.

A runnable temporary-only reproducer was supplied to the implementer as `/private/tmp/shipgate703-shallow-merge-base-repro.py`. Its graph and commands above are sufficient to recreate it independently.

## Additional affected caller: manifest-free host comparison

The same real clone/fetch fixture reproduces the wrong base through `compare_host_refs(base="origin/base", head="HEAD")`: full history returns B/comparable/zero rows, while this head returns A/comparable/the critical addition. The shared `require_merge_base_sha` helper accepts the non-empty candidate, and this PR's switch to tree-only archives removes the old ancestry-validation refusal that previously prevented the comparison. Apply the certainty check to this newly enabled caller too; fixing only `diff` leaves the same regression in the PR's manifest-free comparison.

## P1 — Followed symlink target type is not bound to snapshot revalidation

Location: `src/agents_shipgate/core/host_grants.py:1098-1100,1165`.

The new `os.stat(candidate)` follows the link target and can omit the link from boundary candidates when it currently resolves to a file. The target/type fact is not recorded in `IdentityBoundReadSession`. In a deterministic schedule after enumeration and before collection/final validation, replace an external regular-file target with a directory containing `CLAUDE.md`. The link bytes, link lstat identity, and repository-root identity stay unchanged, so final validation succeeds even though the target now hides possible host inputs.

An independent script imports exact archived sources from both revisions: base `79280b91` reports incomplete with three blocking link issues during the change and on a fresh read; this PR reports **complete with zero issues** during the change, then incomplete with three blocking issues on a fresh read. This false-complete result is introduced by this PR, not inherited from main.

Bounded correction: retain the prior conservative symlink candidate treatment here and defer file-target exemptions until #700/#711 can establish both archive symmetry and identity-bound target evidence. The history/tree materialization improvement remains separable. Reproducer and captured comparison were supplied as `/private/tmp/shipgate703-symlink-race-repro.py` and `.json`.

## Evidence and scope

- Independently ran `tests/test_scoped_base_tree.py`, `tests/test_capability_diff.py`, and `tests/test_manifest_free_pr_rows.py`: **75 passed**.
- Independent real clone/fetch reproduction above fails the semantic comparison despite those green tests.
- #711 remains a separate known limitation: scoping removes the ordinary target of a tracked file symlink, making the archived link dangling and the base inventory incomplete. A minimal fixture reproduced live complete/0 issues versus archived incomplete/3 unreadable issues for the same link. Do not count these incomparable results as covered or clean comparisons.
- The initial “12 of 12 comparable / 347 benign steps” claim was corrected in the PR body before this review was posted. It now accurately attributes 8/12 usable and 140 benign comparable steps to the combined measurement, excludes incomparable runs from successful coverage, and links #700/#711. This resolves the documentation finding; it does not establish standalone reach acceptance or close those limitations.

pengfei-threemoonslab:
Addressed independent coding-agent review round 2 in `2b72eb964007da95e980dedd4a820e1ef81bb84e`.

Both advisory consumers now reject a nonempty but unproven shallow merge base. The shared ancestry check preserves HEAD and visibly ancestral tips, while a real clone/fetch graph with a hidden newer common ancestor requests fetch recovery in CLI and manifest-free PR comparison. The independent reviewer also exercised the actual preview route: incomparable/shallow_history with fetch_base, never incorrect rows.

The target-type symlink optimization was withdrawn: `host_grants.py` is byte-identical to main's conservative reader. A target that changes from file to directory cannot manufacture complete coverage. The three withdrawn optimization-specific tests are replaced by explicit file/directory/dangling limitation checks; a scheduling-point regression covers target replacement. #700/#711 retain identity-bound target handling and archive symmetry. This PR no longer claims to close #688 or make all symlink repositories comparable.

The PR measurement section now uses the corrected aggregate and distinguishes combined-branch observations from this head. No incomplete result counts as a clean comparison, and #698 retains the residual first-run latency criterion.

Validation: final local suite10,273 passed,5 skipped;2 failures/10 setup errors were isolated build-backend installation failures. Both affected modules rerun with network access passed27/27. The14 scoped-tree tests pass; independent shallow/race/preview address probes pass; Ruff and diff checks pass. Fresh committed-head control is complete (`sha256:35677f0fe6dfe363cf284eb85dfb3f7be69163d7855c0cdcffc4593799bea44c`). Required CI on this exact new head must still complete before merge.

This is the response to round2 (2/3 formal rounds), with independently verified address checks, not a fourth review or human qualification.

pengfei-threemoonslab:
Final integration for round 2 at `63eb3348e157a5f21ecd0d2088876f1d60d4b18d`: current main is included, retaining both reviewed changelog entries. The integration adds only the already reviewed dependency and Cursor-null fixes; the addressed shallow-history and conservative symlink behavior remain unchanged.

All nine required GitHub checks, including all three suite shards and coverage, now pass on this head. A fresh `agent control` read reports `complete` with merge permission. Round 2's independent findings and author fixes are recorded above; no additional formal review round was introduced for this integration. #688/#700/#711 remain open for bound symlink coverage and #698 retains residual latency. This merge does not claim v1 readiness or final corpus qualification.

Issue #702, opened 2026-09-12
A hook command that runs an in-repo script: editing the script changes what runs, with no row


pengfei-threemoonslab:
## Observed

`Archive228/loopkit` commits both halves of a Claude Code hook:

```
.claude/hooks/hooks.json   {"hooks": {"SessionStart": [... "command": "\"${CLAUDE_PLUGIN_ROOT:-.}/.claude/hooks/session-start\"" ...]}}
.claude/hooks/session-start   the shell script that command runs
```

#689 made the declaration readable, so adding or repointing the hook is now a row. Editing `session-start` is not. The command still says `session-start`, the declaration's digest is unchanged, and the referenced script bytes have been replaced. Whether that historical hook artifact was selected by a host still requires the loading evidence in #714.

This is deliberate today — `tests/test_claude_hooks_source.py::test_an_unrelated_edit_beside_the_hook_is_not_a_row` pins it — because the alternative as stated is unbounded: a hook command, a workflow `run:` step, and an MCP server's `command` all reference programs, and following every reference means reading the whole repository and calling it configuration.

## Why it is worth bounding rather than accepting

The reference is not arbitrary. A hook command that names a path **inside the repository** can identify a specific declared dependency, provided its path and selection are established. That dependency can matter to review without proving that a host loaded or executed it. The unbounded version is "follow every command"; the bounded version is "a hook command whose executable is an in-repo path contributes that file's digest to the hook grant".

That keeps the property the engine already promises — static, local, no execution — and it closes a gap where a reviewer reading a diff of `.claude/hooks/session-start` gets no signal that a declared hook references the changed file.

## Proposed

Build on #714's distinction between a discovered hook artifact and a supported selecting declaration. For a supported literal command path whose workspace interpretation is established, bind the selected in-repository regular file as a dependency and compare its bytes. Reuse existing input identity and host/diff projections; describe the event, declared path and changed dependency without claiming a semantic effect from arbitrary script text.

Textual shell fallback is not environment evidence. `${CLAUDE_PLUGIN_ROOT:-.}` names a conditional fallback; it does not prove the variable is unset, the current directory is the plugin root or that the host selected that plugin. Keep such bindings conditional/unresolved unless supported static evidence fixes the base. Do not execute a shell, expand the current process environment or invent deployed wiring.

[...]

Issue #701, opened 2026-09-12
Composite actions run in the caller's permission context and are unread: `.github/actions/*/action.yml` changed on 5 of 456 real steps with no row


pengfei-threemoonslab:
## Observed

The zero-noise harness (2026-09-11, 12 public repositories, 456 first-parent steps) found five steps that changed a composite action and produced no row, because no adapter reads the file:

| repository | actions changed |
|---|---|
| `gtkx-org/gtkx` | `run-in-container`, `compute-cache-env`, `pull-image` (3 steps) |
| `AndrewAltimit/template-repo` | `pr-review-fix`, `gemini-review` (2 steps) |

## Why it is host surface

A composite action invoked as `uses: ./.github/actions/x` runs **inside the calling workflow's job**, in that job's token-permission context and with the credentials actually supplied to its steps. So `runs.steps[].run` is a command executing with `contents: write` whenever the caller holds it. The engine reads the workflow's `permissions` block and step-level local-action `uses:` references are not modeled, so:

- adding a `run:` step to a composite action already invoked by a `contents: write` job adds a privileged command, and produces no row;
- pointing a workflow at a different composite action changes what runs with those permissions, and produces no row.

This is the same class as the reusable-workflow `secrets: inherit` edge that #685 modelled — authority reaching code the workflow itself does not contain — one level down, and in a file the reader does not open.

## Why it is not a one-line registry addition

There is no grant shape for it. A workflow grant is `{triggers, write_all, write_scopes, pull_request_target, reusable_calls}`; a composite action has no permissions of its own, only the caller's. The grant is therefore a *relation*: this action, invoked by these jobs, therefore holding their scopes. Modelling it means either a new `composite_action` grant kind carrying the resolved caller scopes, or extending the workflow grant with the actions its jobs invoke and bounded evidence for what those actions declare to run.

The second is likely better — it keeps one subject per workflow, so a reviewer reads which workflow reaches the changed action rather than an unbound row about a file they did not edit — and it avoids a new check ID. It needs a decision before implementation.

## Bounded implementation

Use the caller relation, not a repository-wide search for executable files. Select a local composite action through a supported workflow step, retain the calling job and its modeled token permissions, and compare the action's relevant declared `runs` content. A command change is a code/declaration change within that context, not proof of new scopes or effective access to every secret in the repository.

[...]

PR #704, opened 2026-09-12
Read Claude Code's hook file, and count what is still unread (#689)

### LINKED ISSUE #689: Host surface the reader does not read: `.claude/hooks/*`, `.github/actions/*/action.yml`, `.vscode/settings.json` — an independent census scored 8 of 456 real steps as host-touching that the engine scored benign
> Current review (2026-09-12): the historical totals below are diagnostic observations, not v1 candidate qualification. File discovery alone does not prove host loading; the original proposed blanket hook expansion and zero-disagreement acceptance are superseded by the scope below.

## Observed

The zero-noise harness (2026-09-11, `main` at `396de3b1`, 12 public repositories, 456 first-parent steps) originally classified each step as benign or host-touching using the engine's own `BOUNDARY_ADAPTERS`. PR #687 pointed out the flaw: a host file type the engine does not know looks like a zero-noise benign step. Re-classifying every step with an **independent census** — a path list written from the hosts' public documentation, not from the registry — gives:

| | steps |
|---|---:|
| classifications agree | 448 |
| census says host-touching, engine scored benign | **8** |
| engine says host-touching, census says benign | 0 |

The eight steps touched:

| path | steps | why it is host surface |
|---|---:|---|
| `.github/actions/*/action.yml` | 9 files | composite actions run in the workflow's permission context; a `uses:` of one is code the workflow executes |
| `.claude/hooks/*` | 2 files | executable hooks around the agent's actions — the exact surface #661 says the local guard must intercept |
| `.vscode/settings.json` | 1 file | carries `mcp` and Copilot agent settings in current VS Code |

The registry reads `.codex/hooks.json` but not `.claude/hooks/` files; it reads `.github/workflows/*` but not the composite actions those workflows call.

## Why it matters

Every one of those eight steps was scored "zero rows on a benign change" by the first run. That is silence from missing coverage, presented as correctness — the thing #687's "correct no-change output counts as value; omitted coverage does not" is written to prevent. At 8/456 the number is small; the class is not.

## Current scope

Keep the independent census honest before adding readers. PR #704 owns the bounded census reliability and configured-hook event-label increment:

- A Git clone/ref/read failure remains in the denominator, is counted as an input failure, and makes the measurement fail. Empty engine output is not agreement when Git supplied no observation.
- Include nested `.mcp.json` paths independently; concrete nested-path regressions must not strip the very nesting being tested.
- Compare per path. Report unknown disagreements, issue-owned gaps, and unresolved loading relationships separately. No omitted, incomparable, or unbound candidate becomes a benign/covered success.
- Name events for hooks declared through already supported settings/Codex surfaces.

Do not register every `.claude/hooks/*` file as an operative grant. The existing parser would label it execute/high without proving a settings/plugin binding; flagging all such examples as blocking unknowns would also interrupt users for inactive files. Preserve the historical files/refs as candidates for #714 to resolve. The concrete VS Code sample was editor formatting, not a capability change; content-conditional inputs require evidence, not a blanket path expansion.

Residual coverage stays explicit:

- #701: referenced composite actions and their relevant capability effects.
- #702: scripts referenced by hook commands; edits to a script are not proved by observing only its declaration.
- #714: static configured loading/binding provenance for hook candidates.
- #660: final fixed-corpus measurements after the implementation converges.

## Acceptance

- [ ] Failed Git reads and empty corpora cannot produce a successful measurement.
- [ ] Root and nested MCP configurations have independent census regressions.
- [ ] Per-path output retains issue-owned gaps and uncertain activation in the denominator; neither is labelled covered or benign.
- [ ] Configured hook rows name the event without expanding activation claims to unbound files.
- [ ] Re-run the pinned corpus and publish valid denominators, disagreement categories, and residual owners. Do not require a manufactured zero-disagreement count or reuse the historical aggregate as final evidence.

PR #704 delivers only the bounded implementation increment; this issue remains open until the corrected corpus measurement is recorded. Parent #645. Related #660, #661, #663, #685, #701, #702, #714.

pengfei-threemoonslab:
Closes #689.

**Stacked on #703** — review that first; this PR's diff is the second commit only.

## Why

Twelve public repositories, 456 first-parent steps, classified twice: once with the engine's own `BOUNDARY_ADAPTERS` and once with a census of host paths written from the hosts' documentation. Compared per path, the two agree on 441 steps.

Classifying history with the engine's own registry is what made this invisible: a path the engine does not read looks exactly like a change that did nothing. The disagreements were not one thing.

| disagreement | steps | verdict |
|---|---:|---|
| `.claude/hooks/hooks.json` | 2 | coverage gap — fixed here |
| `.github/actions/*/action.yml` | 20 | coverage gap — needs a grant kind that does not exist (#701) |
| `.claude/hooks/pre-compact` | 1 | the script a hook command runs, read by nothing (#702) |
| `.vscode/settings.json` | 1 | **census bug** |

The `.vscode/settings.json` file held `editor.formatOnSave` and `files.eol`. It is host surface only when it carries an `mcp` or agent key, which a census of *paths* cannot know, so the census entry is removed with the reason recorded rather than the reader extended.

## The fix was one registry entry

`.claude/hooks/hooks.json` is the shape the engine already reads at `.codex/hooks.json`:

```json
{"hooks": {"SessionStart": [{"matcher": "startup|clear|compact",
  "hooks": [{"type": "command", "command": "\"${CLAUDE_PLUGIN_ROOT:-.}/.claude/hooks/session-start\"", "async": false}]}]}}
```

`_source_kind` already routes any `hooks.json` to the hook reader and `_hooks_grants` already produces the grant. A command that runs at every session start was invisible for want of a path.

## The row said "hook"

`_grant_value` recognises a grant by `rule`, `server`, `name`, `value` or `permission`. A hook grant carries `event`, so every hook row rendered as its own kind — telling a reviewer that a hook changed and not which one. Rows now name the event.

## Registering a path is five files, and the guards said so

Adding one registry entry failed four existing tests, each for a real reason:

- the host-config decision table (`REGISTRY_ELIGIBILITY`) pins a yes/no per registry entry so a new one is a deliberate choice;
- `tools/shipgate-detect.py` is a second projection of the same registry and had to keep step. Its nesting rule was hardcoded as "`.codex/` or `.mcp.json`", which silently excludes any nesting path added later — that is now a named `HOST_CONFIG_NESTED` set, iterated sorted so the answer does not depend on set order;
- `docs/triggers.json` mirrors the registry for agents that fetch it;
- the pre-commit `files:` regex, in **three** copies (the hook, `docs/integrations.md`, `examples/pre-commit/README.md`). Without it CI never triggers on the path the engine now reads, so the check would exist and never run.

## The census is committed

`benchmark/cold-start/census.py`. It compares per path, not per step — a step where each side names a different file is two findings, not agreement — and separates three counts: unexplained gaps, gaps an issue already owns, and census bugs. It exits non-zero only on the first and third, so a known gap is a number someone is answerable for rather than a failure that trains people to ignore it.

A test asserts every registry path is named by the census. It immediately found ten it was not naming, including `AGENTS.override.md`, `.cursor/cli.json` and `policies/*.shipgate.yaml` — all blind spots that would have scored as benign on both sides.

## One limit is now recorded rather than unstated

The script a hook command names is not read, so editing `.claude/hooks/session-start` changes what runs with no row. That is #702; it is owned in the census, and the test pinning today's quiet behaviour says so instead of implying the silence is correct.

## Evidence

7 hook cases plus 2 census cases. **4 fail against the parent commit.** The 3 that pass on both are contract-preservation: `.codex/hooks.json` is still read, an unrelated edit beside the hook is still not a row, and every registered hook path still routes to the hook reader.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -61,7 +61,7 @@
       (.*/)?conductor/.*\.json|
       (.*/)?ai/examples/.*\.json|
       (.*/)?\.codex/(config\.toml|hooks\.json|requirements\.toml)|
-      (.*/)?\.claude/(settings(\.local)?\.json|commands(/.*)?)|
+      (.*/)?\.claude/(settings(\.local)?\.json|commands(/.*)?|hooks/hooks\.json)|
       (.*/)?\.cursor/(cli\.json|mcp\.json|rules(/.*)?)|
       (.*/)?\.vscode/mcp\.json|
       (.*/)?\.shipgate/agent-contract\.json|
@@ -109,7 +109,7 @@
       (.*/)?conductor/.*\.json|
       (.*/)?ai/examples/.*\.json|
       (.*/)?\.codex/(config\.toml|hooks\.json|requirements\.toml)|
-      (.*/)?\.claude/(settings(\.local)?\.json|commands(/.*)?)|
+      (.*/)?\.claude/(settings(\.local)?\.json|commands(/.*)?|hooks/hooks\.json)|
       (.*/)?\.cursor/(cli\.json|mcp\.json|rules(/.*)?)|
       (.*/)?\.vscode/mcp\.json|
       (.*/)?\.shipgate/agent-contract\.json|
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## Unreleased
 
+- Read `.claude/hooks/hooks.json`. A `SessionStart` command is executable code
+  around the agent and was invisible; the document is the same shape as
+  `.codex/hooks.json`, which has been read since the Codex adapter landed, so
+  only the registry entry was missing. Hook rows now name their event rather
+  than rendering as "hook". Found by counting disagreements between the
+  adapter registry and an independent census of host paths, now committed at
+  `benchmark/cold-start/census.py`: it prints unexplained coverage gaps,
+  issue-owned gaps and census bugs on every run, so a path nobody registered
+  can no longer look like a change that did nothing (#689).
+
 - Keep a directory at a recognized host configuration path visible as a failed
   input. Host audits, worktree diffs and committed-ref comparisons no longer
   mistake `.mcp.json/` for an absent configuration and report complete coverage.
--- a/benchmark/cold-start/census.py
+++ b/benchmark/cold-start/census.py
@@ -1,0 +1,163 @@
+"""An independent census of host-configuration paths.
+
+Written from the hosts' own published configuration surfaces, deliberately
+NOT from ``agents_shipgate.core.boundary_registry``. That independence is
+the whole point: classifying a repository's history with the engine's own
+registry makes every path the engine does not read look like a benign
+change with nothing to report, so a coverage gap and a correct silence
+become the same measurement.
+
+Run it against a harness result to get the disagreement count:
+
+    python benchmark/cold-start/census.py steps.jsonl clones/
+
+Two kinds of disagreement, and they mean opposite things:
+
+* **a path the census names and the engine did not read** — a coverage
+  gap. Every such step's zero rows are silence, not correctness.
+* **a path the engine read and the census does not name** — a census bug.
+  The list below is behind the registry, or names something that is not
+  host surface.
+
+Compared per path rather than per step: a step where each side names a
+different file is two findings, not agreement.
+
+Both are printed. Neither is allowed to be invisible.
+"""
+
+from __future__ import annotations
+
+import fnmatch
+import json
+import subprocess
+import sys
+from collections import Counter
+from pathlib import Path
+
+#: Path patterns each host documents as configuration it reads. Sources are
+#: the hosts' own docs, not this engine.
+CENSUS: tuple[str, ...] = (
+    # Claude Code
+    ".claude/settings.json", ".claude/settings.local.json",
+    ".claude/hooks/hooks.json", "**/.claude/hooks/hooks.json",
+    ".claude/hooks/*",  # the scripts those commands run — see KNOWN_GAPS
+    ".claude/commands/*", ".claude/commands/**",
+    ".claude/agents/*", ".claude/skills/**",
+    "CLAUDE.md", "**/CLAUDE.md", ".claude/CLAUDE.md",
+    ".mcp.json", "**/mcp.json",
+    # Codex
+    ".codex/config.toml", ".codex/hooks.json", ".codex/requirements.toml",
+    "**/.codex/config.toml", "**/.codex/hooks.json",
+    "AGENTS.md", "**/AGENTS.md", "codex.md",
+    "AGENTS.override.md", "**/AGENTS.override.md",
+    # Cursor
+    ".cursor/mcp.json", ".cursor/cli-config.json", ".cursor/cli.json",
+    ".cursor/rules/*", ".cursor/rules/**", ".cursorrules",
+    # Portable skills
+    ".agents/skills/*/SKILL.md", ".agents/skills/**",
+    # VS Code / Copilot
+    ".vscode/mcp.json", ".github/copilot-instructions.md",
+    ".github/instructions/*", ".github/prompts/*",
+    # Windsurf, Gemini, Aider
+    ".windsurfrules", ".windsurf/**", ".gemini/*", "GEMINI.md", ".aider.conf.yml",
+    # GitHub Actions: the token permissions an agent's CI runs with
+    ".github/workflows/*.yml", ".github/workflows/*.yaml",
+    ".github/actions/*/action.yml", ".github/actions/*/action.yaml",
+    # Shipgate's own declared surface: a policy or contract that governs the
+    # review is as much part of the boundary as the host files it reads.
+    "shipgate.yaml", ".agents-shipgate/*",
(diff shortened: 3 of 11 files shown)


pengfei-threemoonslab:
Reviewed and verified. This corrected my own issue in a way I want to record: `.vscode/settings.json` was a **census bug**, not a coverage gap, and `.github/actions/*` needs a grant kind that does not exist rather than a registry line. Splitting one reported "gap" into three things with three different answers is the right read, and #689 as I filed it was wrong on two of the three.

**Verified end-to-end**, not only through the tests. On a fresh repository, adding a `SessionStart` hook:

```
⚠ high  added  claude-code .claude/hooks/hooks.json
              SessionStart
              changes what runs around the agent's actions
```

and writing an unrelated file beside it in the same directory gives `No static host-grant changes detected.` The event naming is the difference between a row a reviewer can act on and one that says only "a hook changed".

**The census actually runs.** I executed the committed script against a real steps file over five public repositories, 10 first-parent steps: `steps=10 agree=10 unexplained_gaps=0 known_gaps=0 census_bugs=0`, exit 0. The three-way split with only unexplained gaps and census bugs failing is the right call — a known gap that fails the run is a failure people learn to ignore.

**The rot guard is the part I checked hardest.** A census is worth nothing in six months if it depends on someone remembering to update it. `test_every_registry_path_is_named_by_the_census` closes that: a registry path the census does not name fails the suite. That is what makes the disagreement count meaningful rather than decorative.

Also confirmed each of the four guards that fired was load-bearing, not ceremony — `REGISTRY_ELIGIBILITY`, the `shipgate-detect.py` second projection, `docs/triggers.json`, and the pre-commit `files:` regex in three copies. Replacing the hardcoded `.codex/`-or-`.mcp.json` nesting rule with a named `HOST_CONFIG_NESTED` set is the fix that stops the *next* path from being silently excluded; without it this PR would have added a check that never triggers in CI.

## One minor note, not blocking

`known_gap` exempts by basename:

```python
if Path(path).name in {Path(p).name for p in READ_DESPITE_GAP_PATTERN}:
```

so *any* path ending in `hooks.json` is exempt from the `.claude/hooks/*` gap pattern, not only `.claude/hooks/hooks.json`. Not reachable today — every other `hooks.json` in `CENSUS` is read by the engine, so it never lands in `by_census - by_engine`. But if someone adds another host's `hooks.json` to the census without registering it, the gap is silently exempted instead of counted. Matching the full pattern would close it; a comment naming the assumption would do too.

## Restacked

This branch sat on the pre-fix head of #703. I rebased it onto `9cd5f7e1` — the shallow-checkout fix I pushed to #703 after reviewing it — and force-pushed. The CHANGELOG conflict was additive on both sides and both entries are kept. Full suite green on `dea35e55` (`pytest -n auto`, exit 0), ruff clean across `src`, `tests` and `benchmark`.

This looks good to me. Merge #703 first; #704's base is that branch, so its diff stays clean.

pengfei-threemoonslab (commented):
Independent coding-agent review — round 1, exact head `dea35e55973ba20c99767b6163075a3bbb600e54`, compared with its current parent `9cd5f7e156b8bc88435187e64211b46ef604605d`. This is not human approval, merge authority, or validation of a subsequently changed parent stack.

**[P1] Do not count a failed Git query as an agreeing census step.** `benchmark/cold-start/census.py:main` consumes `subprocess.run(...).stdout` without checking returncode. Reproduction with a steps record whose repo does not exist and `host_files=[]` prints `steps=1 agree=1 unexplained_gaps=0 known_gaps=0 census_bugs=0` and returns0. Missing refs/shallow history have the same failure shape. Require a successful bounded Git query before classifying its paths; report/retain failed attempts and return nonzero rather than treating missing evidence as benign agreement. Add missing-clone/missing-ref negative controls and a successful empty-diff control.

**[P2] Include the actual nested MCP path in the independent census.** `CENSUS` contains `**/mcp.json`, while the registry reads `**/.mcp.json`. At this head `census_paths(["packages/app/.mcp.json"]) == []`. The rot guard removes `**/` before probing, so its root `.mcp.json` case masks this gap. Correct the spelling and exercise both root and nested examples of supported nested patterns. Keep the census authored independently; parity tests must not derive its expected classifications from the production registry.

**[P2] Qualify the hook activation claim.** The new tests promote a standalone `.claude/hooks/hooks.json` directly to a high/execute expanding grant without proving a loader reference. Official Claude documentation distinguishes project settings hooks from plugin-root `hooks/hooks.json` and explicit plugin-manifest paths. A discovered declaration is useful evidence, but is not proof it runs in every session. The current cited public fixture's plugin manifest has no custom hooks reference, while its settings declare a different event; this does not establish historical wiring either way. Deferred #714 records exact observations and official links. Keep immediate row/census claims to declared artifacts unless loading is proven; do not expand this PR into runtime activation detection.

The existing basename-exemption review note is lower priority: returning None from `known_gap` causes an unexplained failure, not a silent pass. Full-path matching is cleaner, but it is not the dangerous false-agreement case above.

Validation: `tests/test_claude_hooks_source.py` and `tests/test_host_discovery.py` completed with73passing tests on this exact checkout. Independently reproduced both census failures using disposable input, inspected registry/detector/trigger/pre-commit projections and hook-event rendering. No full-suite rerun, repository edits, or merge. Parent #703 is under separate review and must be integrated/rechecked before merge.

Issue #690, opened 2026-09-11
Revisit the legacy repository skill after readiness: prove its value before restoring default invocation


pengfei-threemoonslab:
## Problem

The legacy repository Agents Shipgate skill imposes workflow steps before its benefit to the maintainer's current task has been established. During the ROADMAP-only work in #687, local check/verify/control reads recurred around Git and PR operations. The owner explicitly asked to stop using that skill at the current stage and minimize interruption of the user's workflow.

This is observed workflow friction, not a controlled latency benchmark. Ordinary tests, required GitHub CI and skill-driven command overhead must be measured separately; the total PR duration cannot be attributed to the skill.

## Current disposition

**P2, deferred until after v1.0 readiness.** Do not invoke the legacy repository skill in the current implementation workflow. Revisiting or re-enabling it is not a v1.0 acceptance criterion. No skill file, published integration or GitHub protection is removed by recording this decision.

## Proposal

- Inventory which current instructions add a useful decision, duplicate another check, trigger irrelevant work or require the user to repeat an already supplied authorization. Distinguish repository maintenance from downstream agent-capability review.
- Replay representative documentation-only, ordinary source, capability-change and required-human-decision workflows. Record user interruptions, repeated questions, commands, added wall time and the actionable information delivered. Keep tests/CI and engine runtime separate from instruction overhead.
- Prefer relevant triggers, concise results, batched findings and one actionable handoff. A genuine missing decision should name the change, consequence and precise decision needed; internal routing vocabulary should not become work for the user.
- Decide whether to simplify, replace, make opt-in or retire the legacy skill. Restore default use only if the replay demonstrates useful value with lower workflow cost. Reuse the maintained integration work in #662 and interruption semantics in #661 rather than creating a second instruction stack.

## Acceptance

- [ ] The current-stage suspension and the eventual keep/change/retire decision are explicit; readiness never waits for this review.
- [ ] Baseline and candidate traces distinguish unnecessary human interruptions from necessary decisions, and separate verification latency from tests/CI.
- [ ] Routine already-authorized documentation/source workflows require zero redundant permission questions; a repeated settled question cannot be presented as a new decision without explaining what changed.
- [ ] Positive controls still expose materially changed authority, incomplete evidence and genuinely missing user decisions. Suppressing useful information is not counted as a UX improvement.
[...]

Issue #684, opened 2026-09-11
PR-time `verify` is blind to host widenings on a repository with no manifest: names none of `Bash(*)`, `WebFetch(*)`, a new MCP server and `contents: write`, and routes to `init --write`


pengfei-threemoonslab:
## Observed

On the S1 scenario (a repository with `.claude/settings.json`, `.mcp.json` and a CI workflow, no `shipgate.yaml`) whose branch adds `Bash(*)`, `WebFetch(*)`, a `postgres` MCP server, and widens the workflow from `contents: read` to `contents: write`:

| command | result |
|---|---|
| `shipgate diff` | 5 rows, each named, with direction, why and severity — 1 second |
| `shipgate check` (default output) | JSON envelope: "5 coding-agent boundary change(s) block local continuation" — **names none of them** |
| `shipgate verify --preview --base main --format text` | "Trigger: 4 run_shipgate rule(s) matched" … "Agents Shipgate verify: **not evaluated**" … "Next: initialize the local Shipgate contract" → `init --write` — **names none of them** |

`verify` is what the Action runs and what the PR comment is rendered from. It *saw* the files (four rules matched) and then evaluated nothing, because there is no manifest — and `init --write` on a host-only repository refuses and routes back to `audit --host` (#650), so the chain terminates without ever reaching the one command that answers the question.

## Why it matters for v1.0

The v1.0 definition (#643) is: *a platform engineer who installs Agents Shipgate sees what a PR changed in agent permissions and capabilities in plain language within two commands.* On the PR surface — the one a team actually installs — the product currently says "not evaluated" about the exact change it exists to catch. The capability that makes the product real (`diff`, #651) is not reachable from the PR path at all.

This is the measured form of the roadmap judgment "把已存在的 drift/diff 能力变成产品的默认形态": the capability exists, and no default entry reaches it.

## Proposed

- `verify` on a repository with host surface and no manifest computes the host-grant diff (the same `build_host_drift_payload` path `diff` uses) and prints the rows in its text output and in the PR comment, above any verdict line.
- The verdict remains a footer, and on a manifest-free repository reads `advisory: no policy configured` (#655's rule), never `not evaluated` when rows exist.
- `check`'s human-readable form prints the rows before the control block; the JSON envelope carries `rows[]` with subject, direction, severity so an agent that reads only JSON can name the change (#662).
- Next action after rows on a host-only repository is `audit --host --save-baseline` or nothing — never `init --write`.

[...]

Issue #662, opened 2026-09-11
Coding-agent workflow parity: CLI, PR and maintained JSON name the same host-capability changes without setup


pengfei-threemoonslab:
## Problem

The target user runs coding agents, and the agent may be the one that installs and runs Agents Shipgate. The original study found that the skill and plugin emphasized a compact control answer without the named change, or a large boundary JSON document. A human then received "a human must review" without learning what changed. The W24 harness also found activation problems: agents did not run verification in adversarial cells.

These are two different obligations. This repository can make the installation, instructions, machine output and rendering correct and reproducible. It cannot establish that a real Claude Code, Codex or Cursor session will choose to use them without running that provider-dependent experiment. #643 makes the former a readiness requirement and measures the latter after readiness.

## Current scope

The owner suspended the legacy Agents Shipgate skill for this implementation phase (#690). v1.0 readiness exercises the maintained CLI, PR, MCP/JSON and generated instruction contracts directly. Invoking, rewriting or restoring that skill is deferred; it cannot become a prerequisite through a parity checkbox. Minimize interruption: show the named change first, and request a decision only when evidence identifies a genuinely unresolved user-owned choice.

## Proposal

- For a supported host-configuration change, the maintained CLI/PR/JSON first-value route names the subject, direction, severity, evidence and coverage limits. #684 brings the existing host comparator into manifest-free `verify`/PR and named `check` rows; build on that route rather than duplicating it. Keep the existing current-control and verifier instructions for operational authority: a displayed row, successful diff or empty table never grants edit, push, merge or completion permission.
- Expose the same bounded capability rows through the existing machine-readable handoff/result surfaces so a JSON consumer can name the change without interpreting control prose. Reuse the canonical row model and renderer; retain omitted-row counts or an artifact reference when display is bounded. Reuse existing fields where sufficient; any necessary additive successor follows `STABILITY.md`, with the contract and consumer tests updated together.
- Keep the maintained CLI/MCP and Cursor instruction surfaces consistent through their existing generators and consumer tests. Use the actual supported paths rather than adding a new instruction file merely to match an old example. Legacy skill activation and bundle revision belong to deferred #690; retain its current behavior without treating it as a v1.0 success claim. The installation recipe uses the selected, installable advisory artifact; deferred PyPI publication #647 is not a prerequisite. The local guard work remains in #661; this issue does not replace its hook semantics or silently remove existing plugin capabilities.
[...]

PR #710, opened 2026-09-12
Lead the Cursor instructions with the named rows (#662)

pengfei-threemoonslab:
Part of #662. **Stacked on #704** (which is stacked on #703) — this PR's diff is its last commit only.

## Why

The Cursor rule opened with the control envelope:

```
For local agent control, run:
  shipgate check --agent cursor --workspace . --format agent-boundary-json
Read the check stdout JSON only... switch on `control.state`
```

An agent following that reports `human_review_required` to its user without naming what changed — the original finding behind #662. `AGENTS.md` has led with `shipgate diff` and the row table since #651; the generated Cursor surface never caught up. Measured on the three-widening scenario: `AGENTS.md`'s route names all four widenings, the Cursor route names none of them before the control block.

## What changed

The Cursor instructions now open with `shipgate diff --workspace .`, name the row fields an agent should quote (subject, before, after, direction, severity, why), and say two things the control envelope alone does not:

- a covered comparison with **no rows is a real answer**, not a missing one — otherwise an agent treats quiet as failure and escalates;
- a row is a description and **never a permission**: showing one, or seeing an empty table, grants no authority to edit, commit, push, merge or report complete.

The existing control-envelope instructions are unchanged and still follow. This adds the named change in front of them; it does not replace operational routing.

## The guards did the work

Editing the renderer failed two existing tests, both correctly:

- `test_committed_cursor_rule_matches_renderer` — the checked-in `.cursor/rules/agents-shipgate.mdc`;
- `test_target_repo_cursor_snippet_matches_renderer` — the copyable block in `docs/target-repo-agent-snippets.md`.

Both are regenerated from the one renderer rather than hand-edited, so the three copies cannot drift. This is the same "second projection" class the registry work hit in #704.

No render-hash bump: those pin the Claude Code and Codex **skill** bundles, which this does not touch. The legacy skill stays suspended under #690 and is not repaired here.

## Not in this PR

The rest of #662 — the 12 deterministic entry-contract cases, and the after-readiness provider experiment — is separate. What is already true and was verified while measuring this: `check --format agent-boundary-json` carries `rows` (5 on the scenario, naming every widening) and `verify --preview --base main --format text` renders the full table with an `advisory: no application release policy configured` footer. Both arrived with #684.

## Evidence

Full suite green (`pytest -n auto`, exit 0), ruff clean.

### DIFF
--- a/.cursor/rules/agents-shipgate.mdc
+++ b/.cursor/rules/agents-shipgate.mdc
@@ -38,6 +38,21 @@ When a change affects agent tools, MCP exports, OpenAPI specs, prompts,
 permissions, approval policies, or release gates, run Agents Shipgate.
 Default to advisory verification while adopting the gate.
 
+First, name what the change did to the agent's authority. This needs no
+manifest and no committed baseline:
+
+  shipgate diff --workspace .
+
+One row per host grant, each carrying subject, before, after, direction,
+severity and why it matters; `⚠` marks a row the engine read as an
+expansion of authority. Quote those rows to the user, and put them in the
+pull request body. A covered comparison with no rows is a real answer, not
+a missing one. `--json` emits the same rows under `rows`.
+
+A row is a description, never a permission: showing one, or seeing an empty
+table, grants no authority to edit, commit, push, merge or report the work
+complete.
+
 For local agent control, run:
 
   shipgate check --agent cursor --workspace . --format agent-boundary-json
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+- Lead the Cursor instruction surface with `shipgate diff`. It opened with
+  the control envelope, so an agent following it reported "a human must
+  review" without naming what changed, while `AGENTS.md` had led with the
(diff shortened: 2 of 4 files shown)


pengfei-threemoonslab:
Fixed a contamination in this PR before merge. The branch was built on my Gate 1 measurement branch, which had #678's lattice commit (`fix(host): tell a widened permission rule from a narrowed one (#657)`) cherry-picked in. The published diff therefore carried the whole #657 change: `permission_lattice.py`, `host_boundary.py`, the host-boundary policy, and its tests, plus stale #703/#704 commits after the base branch was force-pushed. Merging would have landed #657 here, bypassing #678's own review and duplicating it.

Rebuilt at `767977cb` on #704's rebased head. It is now one commit and four files: `cursor.py`, the regenerated `.cursor/rules/agents-shipgate.mdc`, the snippet in `docs/target-repo-agent-snippets.md`, and one CHANGELOG entry. The #657 changelog entry that the conflict resolution had carried over is removed too, so #678 remains the only place that change is recorded. Renderer tests pass (exit 0) and ruff is clean. #657 stays in #678.

Issue #711, opened 2026-09-12
Scoping manufactures incompleteness: an in-tree symlink whose target is out of scope dangles in the base archive but resolves in the working tree


pengfei-threemoonslab:
## Observed

`awslabs/nx-plugin-for-aws`, every first-parent step: `comparison_status: incomparable`, `incomparable_reasons: ["base_inventory_incomplete"]`. Measured on the combined Gate 1 branch (#703 + #704 + #657), 2026-09-12.

The repository has two symlinks:

```
120000 .git-secrets/git-secrets -> ../packages/nx-plugin/src/preset/git-secrets-files/git-secrets-dir/git-secrets
120000 AGENTS.md                -> (an in-tree file)
```

The two sides disagree about the first one:

| | `.git-secrets/git-secrets` |
|---|---|
| working tree (live reader) | target present, link resolves to a **file**, `_symlink_may_be_directory` is False, **no issue** |
| scoped base archive | target is not host surface so it is not materialized, link **dangles**, `_symlink_may_be_directory` fails closed to True, boundary candidate, `O_NOFOLLOW` read fails, **`unreadable` issue** |

Live-tree issues: `AGENTS.md` only. Base-archive issues: `AGENTS.md` **and** `.git-secrets/git-secrets`. The extra one makes the base inventory incomplete and the whole comparison incomparable.

`AGENTS.md` is #700 and is not this. This is the second symlink — one no reader would ever open, in a repository where the live tree is perfectly happy.

## Why

#703 materializes every symlink regardless of scope, deliberately: a linked directory is what can conceal a scoped path, so both sides must see the same links. That is right. What it does not carry across is what each link *resolves to* — and `_symlink_may_be_directory` is a question about the target, not the link. Out of scope, the target is absent; absent means dangling; dangling fails closed. So scoping creates an `unreadable` that does not exist in the tree being reviewed.

The property #703 states — "the base tree must present the reader the same object or the two sides disagree about what the file is" — holds for the link and breaks for its target.

## Why it matters

[...]

PR #712, opened 2026-09-12
An empty frontmatter value means absent, not invalid

pengfei-threemoonslab:
Found by the Gate 1 harness while re-measuring the twelve-repository sample. Independent of #703/#704 — branched off `main`.

## Why

`globs:` with nothing after it is how Cursor writes a rule that is not glob-scoped. This is the file Cursor generates:

```
---
description: 
globs: 
alwaysApply: true
---
```

YAML reads those empty values as `None`, and `_valid_metadata` rejected `None` for every field. So the canonical Cursor rule classified as `frontmatter_invalid_structure`.

That reason is a **blocking** inventory issue. One blocking issue makes the base inventory incomplete. An incomplete base inventory makes the whole comparison `incomparable`.

Measured on `Doist/todoist-mcp` — eleven readable host files (`.claude/settings.json`, `.mcp.json`, `.cursor/mcp.json`, five workflows, `AGENTS.md`, `CLAUDE.md`) and two Cursor rules in Cursor's own format:

| | before | after |
|---|---|---|
| `shipgate diff --base HEAD~1` | `incomparable`, `base_inventory_incomplete`, **0 rows** | `comparable`, 0 rows |
| across its history | all 40 first-parent steps incomparable | comparable |

A reviewer got nothing about eleven files that were read perfectly, because two files used a format we mis-read. This is the same shape as #688 — one unreadable path refusing the whole repository — arriving through a different door.

## The fix

An explicit null is the same statement as an absent key: the field is not set. `_valid_metadata` now skips `None` and lets the profile's own required-field checks speak.

## What still fails, and is pinned

Forgiving `None` in a shared type check would be a hole if identity were only enforced there, so the tests say what must still fire:

- a wrong **type** is still wrong — `globs: 7`, `alwaysApply: yes-please` → `frontmatter_invalid_structure`;
- an unknown key is still unknown, empty or not → `frontmatter_unknown_fields`;
- a skill with an empty `name` or `description` still fails → `skill_identity_missing`. (On pre-fix `main` it failed as `frontmatter_invalid_structure`; the specific reason is the better answer and this pins that it is reached.)

## Evidence

Eight cases in `tests/test_instruction_structure.py`. **Five fail against pre-fix `main`**; the three that pass on both are the contract-preservation half — a fully populated Cursor rule still resolves, and the two real-structure-problem cases still refuse.

Full suite green (`pytest -n auto`, exit 0), ruff clean.

## Scope

`.cursor/rules/*.mdc` is the case that was measured, but the fix is in the shared metadata check, so `claude_command/v1` and `skill_instruction/v1` gain the same reading. That is deliberate: an empty optional field means the same thing in all three, and the profile-specific required-field checks are what enforce the differences.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
   every step of its history. An explicit null is now read as an absent key,
   which is what it means. Wrong types are still wrong, unknown keys are
   still unknown, and a skill with an empty `name` or `description` still
-  fails on identity rather than sliding through (#711).
+  fails on identity rather than sliding through (#712).
 
 - Manifest-free host PR review now names capability changes in verify, PR comments,
   and check. Interactive check defaults to readable text; agent mode keeps JSON.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ## Unreleased
 
+- Read the Cursor rule format Cursor actually writes. `globs:` with nothing
+  after it — how Cursor spells a rule that is not glob-scoped — reads as
+  YAML null, and the frontmatter type check rejected null for every field.
+  That made the canonical `.cursor/rules/*.mdc` an unresolved instruction
+  structure, which is a blocking inventory issue, which made the whole
+  repository incomparable: `Doist/todoist-mcp` has eleven readable host
+  files and two such rules, and `diff` produced no rows for any of them on
+  every step of its history. An explicit null is now read as an absent key,
+  which is what it means. Wrong types are still wrong, unknown keys are
+  still unknown, and a skill with an empty `name` or `description` still
+  fails on identity rather than sliding through (#711).
+
 - Manifest-free host PR review now names capability changes in verify, PR comments,
   and check. Interactive check defaults to readable text; agent mode keeps JSON.
   Comparison evidence binds its input refs, excludes stale scan artifacts, and
--- a/src/agents_shipgate/core/instruction_structure.py
+++ b/src/agents_shipgate/core/instruction_structure.py
@@ -75,6 +75,18 @@ def _digest(value: object) -> str:
 
 def _valid_metadata(metadata: dict) -> bool:
     for key, value in metadata.items():
+        if value is None:
+            # `globs:` with nothing after it is how Cursor writes a rule that
+            # is not glob-scoped, and YAML reads that as None. Rejecting it
+            # made the canonical Cursor rule file an unresolved structure,
+            # which is a *blocking* inventory issue, which made every
+            # repository carrying one incomparable in full — `Doist/todoist-mcp`
+            # produced no rows for eleven readable host files because two
+            # `.cursor/rules/*.mdc` used the format Cursor itself generates.
+            # An explicit null is the same statement as an absent key: the
+            # field is not set. Where a field is genuinely required, the
+            # profile checks below still say so and still fire.
+            continue
         if key in {"disable-model-invocation", "user-invocable", "alwaysApply"}:
             if not isinstance(value, bool):
                 return False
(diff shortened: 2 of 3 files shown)


pengfei-threemoonslab (commented):
Independent coding-agent review — round 1, exact commit `2dcc85a95452cf7e85bae5b5ec2efde70bc3ead1`. This is an automated independent review, not human approval or merge authority.

The Cursor parser correction is useful, and unknown keys, wrong types, missing skill identity, bounded YAML handling, and execution-bearing structure retain their separate checks. Two changes should be addressed before merging:

1. **[P2] Apply the advertised null/absent equivalence to the structure digest.** Skipping a value in `_valid_metadata` leaves the key in `metadata`, which is later hashed. For `.cursor/rules/demo.mdc`, changing `alwaysApply: true` to the same header plus empty `globs:` yields two structured classifications but different digests. `compare_host_inventories` consequently emits a medium `changed` row saying “changes instructions the agent is given” for this no-op. The same digest discrepancy occurs for empty optional `allowed-tools` in both command and skill profiles. Normalize only accepted top-level null fields before hashing, after unknown-key/type/required-identity checks, and pin absent↔null equality plus zero host comparison rows. Keep actual non-null changes, unknown-null keys, and null required skill identity reviewable/refused.
2. **[P3] Fix the changelog reference.** The new entry cites #711 (scoped symlinks), while this change is PR #712.

Validation on this exact checkout: five focused modules (`test_instruction_structure`, `test_capability_diff`, `test_manifest_free_pr_rows`, `test_instruction_structure_contracts`, `test_instruction_structure_workflow`) completed with exit 0; Ruff for both changed Python files passed. Independently reproduced the unwanted medium Cursor row and the three-profile digest mismatch with disposable local files. No repository edits or merge performed. The PR author's full-suite claim was not independently re-run as part of this bounded review.

pengfei-threemoonslab:
Addressed independent coding-agent review round 1 in `f73425a54a3e1e7552d393ecb44d2e7766e85f3f`.

Optional null values are normalized only after unknown-key, type and required-identity validation. An absent key and explicit null now yield the same structure; actual permission changes remain distinct. Six added regressions cover Cursor, Claude commands and skill metadata. The unrelated CHANGELOG reference is corrected to #712. The independent reviewer replayed the original host-inventory case: comparable with zero rows, with invalid/required-field controls still refused.

Validation:45 focused tests and Ruff pass. Full local suite:10,269 passed,5 skipped;2 failures and10 setup errors all came from isolated wheel build-backend installation in the restricted environment. Re-running both affected modules with network access passed all27 tests. Schema and24golden checks pass. Fresh committed-head verification/current control is complete (`sha256:3542fad1679708b8ecf2647f932a04ff33c0338cd5ea4a21f8ab7d2c555ccb34`), with publication permissions. Required CI on this new head remains the final merge prerequisite.

This is the author response and independently verified address within round1, not a new review round or human qualification claim.

pengfei-threemoonslab:
Final head `f73425a54a3e1e7552d393ecb44d2e7766e85f3f` now passes all nine required checks, including aggregate coverage. Verification was refreshed against main after #708/#709; current control is complete and permits merge (`sha256:556c6ce8c8e568b7b002e1b09d42829321516fc9bd0fccad92bbef7e4f8bc1ac`). One independent GitHub review/address round is complete.

pengfei-threemoonslab:
GitHub required the branch to include current main despite all prior-head checks passing. Integrated main in `df2bb6cbd9916273de6e5472dcfc751eff4bf9a7`; the merge changes only the two already reviewed dependency lock entries from #708/#709. Instruction source and tests are byte-identical to reviewed `f73425a5`. Fresh verification/current control is complete and permits publication. Required CI will rerun on this integrated head before merge; no branch protection is bypassed and no additional review round is claimed.

PR #728, opened 2026-09-13
fix(boundary): route a blocked result to human review instead of crashing the envelope

### LINKED ISSUE #694: check crashes when a new baseline and blocked host-capability changes share the working tree

wangzhengzhuo05:
## Summary

- Fixes a crash in every projection of `AgentBoundaryResultV1` (including `check --format agent-boundary-json`) when the working tree contains **both** an untracked gate-governing baseline (`.agents-shipgate/host-grants.json`, as recorded by `audit --host --save-baseline`) **and** a blocked host-capability expansion (e.g. `.claude/settings.json` widened to `Bash(*)` / `Read(**)` / `WebFetch(*)`). Instead of a machine envelope the caller got a Pydantic `ValidationError: a blocked result cannot authorize publication` and no JSON.
- Root cause: the untracked trust root routes the projected control to `review_publishable` (`permissions.publishes=True`, `completion_allowed=False`), while the wildcard expansion makes the aggregate decision `block`. `schemas/agent_result.py` correctly refuses that pair; `build_agent_boundary_result` constructed it anyway.
- Fix: reconcile before constructing the envelope. When the aggregate decision is `block` and the projected control authorizes publication, the control is re-projected onto the strictest non-publishing route through the existing block-branch derivation: `human_review_required`, `publishes=False`, `completion_allowed=False`, `verify_required` preserved, violation-derived human obligations and `required_reviewers` re-synced.
- No schema invariant was touched or relaxed: the invariant that a blocked result cannot authorize publication still holds, and the exception was not converted into an allowed/publication-ready answer.

## Type

- [ ] Input adapter change
- [ ] Documentation only

## Verification

Additional local checks run:

- `pytest tests/test_agent_boundary.py` — 87 passed, 1 skipped (darwin-only test)
- `pytest tests/test_codex_boundary_check.py tests/test_agent_control_envelope.py tests/test_agent_control_contract.py tests/test_local_contract.py tests/test_current_control.py` — 340 passed together with the file above
- `python -m ruff check src tests` — All checks passed; `python -m compileall -q src tests` — clean
- New regression test `test_untracked_baseline_with_blocked_host_expansion_routes_to_human_review` reproduces the issue fixture exactly (committed `.claude/settings.json` with `Bash(npm test)`/`Read(src/**)`, worktree copy widened to `Bash(*)`/`Read(**)`/`WebFetch(*)`, plus an untracked `.agents-shipgate/host-grants.json`).
- **Mutation check:** neutralising the reconciliation (early `return legacy`) makes exactly the new test fail with the issue's original error `ValidationError: a blocked result cannot authorize publication`; restoring the fix returns the file and the test to green. So the test guards the fix and not something incidental.
- Unchanged behaviour confirmed by the existing suite: the same diff without the untracked baseline is still `block` / `human_review_required` (4 violations), a trust-root-only change is still `require_review` / `review_publishable`, and a baseline committed on the base is still an unchanged comparison.

## Release-readiness notes

- [x] New or changed check IDs are documented in `docs/checks.md` (no check ID added or changed)
- [x] Report/schema changes are additive or documented in `STABILITY.md` (no Pydantic model field or schema version changed; the fix is confined to control projection and its regression test)

Fixes #694

---

Implemented with AI assistance (OpenCode + muse-spark 1.3) against the issue's stated acceptance criteria; the diff, the regression test and the mutation check above were reviewed and re-run by a human before submission.


pengfei-threemoonslab:
Thanks @wangzhengzhuo05 — this reproduced #694 exactly, and your regression test and mutation check made the review easy.

While auditing the other projections, as #694 asks, I found the crash was one symptom of a routing-order bug. So I pushed a follow-up commit onto this branch, since maintainer edits are enabled. Your commit is unchanged underneath it.

**What was wrong.** In `core/codex_boundary.py::_control_for_result`, the undeclared-surface branch runs before the policy-block branch. An untracked baseline is an undeclared surface, so a `block` decision got a coding-agent `configure` route that could publish. Every projection reads that control, so the formats failed in two different ways on the same input:

| Format | Result before the follow-up |
| --- | --- |
| `text` | raised the schema invariant, no output |
| `agent-boundary-json` | raised the schema invariant, no output |
| `agent-control-json` | raised the schema invariant, no output |
| `codex-boundary-json` | printed `decision: block` with `agent_action_required` and a coding-agent configure command |

Your reconciliation fixed the first three. The fourth said "blocked" and "keep going" at once.

**What the follow-up does.**
- Guards the route at its source: `if undeclared_gap and decision != "block"`. A block falls through to the existing block branch: human stop, the violations' reviewers, nothing published.
- Drops `_reconcile_blocked_control`. With the source fixed it can no longer fire, and a second reconciliation in only one projection is how the formats drifted apart.
- Keeps your `test_untracked_baseline_with_blocked_host_expansion_routes_to_human_review`.
- Adds `test_check_stops_an_untracked_baseline_with_a_blocked_expansion_in_every_format`. It follows the issue's route through the real `check` command (`audit --host --save-baseline`, baseline left untracked, a committed widening) for all four formats.

**Verification.**
- Removing the guard fails all five tests: your test and the four-format test.
- The affected boundary, control, envelope, current-control and MCP suites pass: 924 tests.
- A committed baseline, and a trust-root-only change, route exactly as before.

PR #743, opened 2026-09-13
Name the host capability change in the compact control envelope (#662)

pengfei-threemoonslab:
Part of #662: the "machine-readable handoff/result rows" acceptance item for the compact control envelope.

## Decision

Owner decisions on #662 (2026-09-13):
- **Option (a).** A bounded capability-rows block in `shipgate.agent_control/v1`.
- **In place, not a `v2`.** A later ruling after review found the freeze: `docs/agent-control-schema.v1.json` is SHA-pinned in `tests/test_human_review_request.py::FROZEN_CONTROL_SCHEMAS`, and the envelope is a closed object. The owner chose to widen v1 in place, a deliberate break taken for the long term, rather than publish `v2`.

## Change

- **`schemas/agent_control_envelope.py`.** Adds `EnvelopeCapabilityRow`, `EnvelopeCapabilityRows` and `MAX_ENVELOPE_CAPABILITY_ROWS = 5`, plus an optional `capability_rows` on the envelope base.
  - The block carries `comparison_status`, `incomparable_reasons`, `rows`, `omitted_rows` and `unchanged_limit_count`.
  - It is omitted, not `null`, when no comparison ran (`exclude_if`), so those envelopes are byte-identical.
  - One rule is enforced in both layers, the model validator and the published JSON Schema `allOf`:
    - an `incomparable` block carries no rows and zero counts;
    - a `comparable` block carries no reasons;
    - a non-zero `omitted_rows` requires a full prefix of rows.
- **`core/agent_control_envelope.py`.** `capability_rows_block` copies the producer's rows.
  - Rows with `expands: true` come first, each group in producer order, so the cap drops a narrowing before a widening.
  - Only `why` (prose) is capped. `subject`, `before` and `after` are never abridged.
  - Wired into `envelope_from_agent_result` (`check --format agent-control-json`) and `envelope_from_verifier` (`verify --format control`, `agent control`).
  - The deprecated codex v2 result has no comparison fields and publishes no block.
  - `project_agent_control_envelope` never reads the block, so state, permissions and route are unchanged by it.
- **Contract 38.**
  - `CONTRACT_VERSION`, with its rationale note.
  - `.well-known`, the contract-doc stamp, the `docs/architecture.md` stamp and `docs/passed-verdict-contract.md`.
  - Three test literals.
  - `minimum_control_contract_version` stays `21`: `AgentControl` is unchanged, as for v23–v25.
- **Pin.** `FROZEN_CONTROL_SCHEMAS` moves for `agent-control-schema.v1.json`, with a comment naming #662. Nothing else in that list changes.
- **Docs.** STABILITY and `agent-contract-current.md` lead paragraphs, and a STABILITY migration note: shape, cap and ordering, status rules, no authority, and exactly who breaks. Also a CHANGELOG entry, the regenerated schema and `llms-full.txt`.
- **Pilot ledger.** A contract bump requires re-running the Route H readiness dry run, not carrying it forward. I reran the source-tree column on this branch rebased onto `294d0443`. Every cell reproduced except the contract number:
  - `check`: `block` / `critical`, 4 violations, coverage visible
  - `init --write --ci`: `not_applicable_host_review`, no workflow written
  - manifest-free `verify`: exit 0, six rows
  - drift: four expansion signals

  `docs/design-partner-pilot-results.md` records that rerun.

## Who breaks

A reader that validates envelopes against the v37 schema rejects an envelope carrying `capability_rows`.
- That happens only when a host comparison ran, and the reader fails closed.
- Readers that parse without validating, or that ignore unknown members, are unaffected.
- `test_a_v37_validator_rejects_rows_and_still_accepts_an_envelope_without_them` pins both halves.

## Tests (`tests/test_agent_control_envelope_rows.py`, 15)

- **Route agreement.**
  - `check --format agent-control-json` rows equal `agent-boundary-json` rows.
  - `verify --preview --format control` rows equal `verify --json` `host_comparison`, including `unchanged_limit_count`.
  - Both validate against the published schema.
- **Incomparable and absent input.**
  - Malformed input gives an incomparable block with the boundary result's reasons and no rows.
  - No comparison omits the block.
  - The codex projection publishes none.
- **Bounds.**
  - The cap keeps widenings first and counts the cut.
  - Only prose is capped.
- **Authority.** Rows change nothing: the envelope is identical once the block is removed.
- **Both layers.** Pydantic and the JSON Schema both refuse four contradictory blocks.
- **Budget.** A full block beside the full artifact map and capped prose fits the 6,144-byte budget.
- **Compatibility.** A v37 validator rejects row-bearing envelopes and accepts row-less ones.

## Verification

See the mutation and full-suite results in the first comment.


pengfei-threemoonslab:
## Verification

- **Mutations.** 8/8 fire, each run against `tests/test_agent_control_envelope_rows.py`. The sources were restored and checked byte-identical with `cmp`.

  | Mutation | Failing tests |
  | --- | --- |
  | No widenings-first order | 2 |
  | Check route drops the block | 3 |
  | Verifier route drops the block | 1 |
  | `null` emitted when the block is absent | 3 |
  | Pydantic consistency rule removed | 1 |
  | Omitted count forced to 0 | 1 |
  | `why` left uncapped | 1 |
  | Unchanged-limit count dropped | 1 |

  The last one first **survived**: the verify fixture had no limits, so 0 == 0. I added `test_verify_control_envelope_counts_the_unchanged_limits_it_compared_past`, an unresolved skill that stays unchanged beside a widening, and the mutation now fails it.
- **Full suite on the final commit, rebased on `294d0443`.** 10892 passed, 5 skipped in 550.71s (0:09:10)

Issue #742, opened 2026-09-13
Changed-file paths in check and verify publish credential-shaped file names verbatim


pengfei-threemoonslab:
## Observed

This is the scope boundary found while fixing #590. #590 covers host inventory publication.

The fixture is a repository whose change touches files named with a synthetic GitHub-token-shaped string: `.claude/skills/deploy TOKEN=ghp_…/SKILL.md` and `.github/workflows/deploy TOKEN=ghp_….yml`. On the #590 branch, the host inventory family carries no sentinel: `audit --host --json`, the audit markdown, the saved baseline, `audit --host --drift --json`, `diff --json` and `verify --json`'s `host_comparison`. The sentinel is still published verbatim in these fields:

| Surface | Fields |
| --- | --- |
| `check --format agent-boundary-json` | `changed_files[]`, `trigger.changed_files[]`, `affected_files[].path`, `violations[].path`, `violated_rules[].path`, `host_coverage[].paths[]` |
| `check` text output | the same paths |
| `verify --preview --json`, and `verifier.json` / `agent-handoff.json` on disk | `changed_files[]`, `trigger.changed_files[]` |

No real credential was used, no authority changed, and every permission stayed as before. This is an output-privacy gap in how repository paths are published, not a bypass.

## Why it is separate from #590

These are the repository's changed-file paths, not host-inventory locations, and agents and CI use them as locators: they open, quote and review the named file. Redacting them changes what a consumer can act on. The fix therefore needs its own decision rather than riding along with the inventory change.

## Proposed

Decide the policy for published changed-file paths that contain credential-shaped bytes. Options:

1. Apply #590's component-wise `public_host_path` redaction, with its exact-source digest, to every published path field. Keep exact paths internal for reads and policy. Say explicitly that a redacted entry is a label, not a locator, and name what the consumer should do instead.
2. Keep exact paths, and document that file names are published as-is because they are already visible in the repository and the PR diff.

## Acceptance

- [ ] A decision is recorded, with the reasoning about locator use versus disclosure.
- [ ] If redacting: the fields above carry no sentinel. Two paths that redact alike stay distinct. Boundary policy classification and verify's reads still use the exact path. A regression test covers the direct projection and the public `check` / `verify` output.
- [ ] If not redacting: the privacy section of `STABILITY.md` states that changed-file paths are published verbatim.

Related: #590 (host inventory paths), #586 (source-omission redaction accounting).

[...]

PR #744, opened 2026-09-13
Keep credential-shaped path bytes out of host inventory output (#590)

### LINKED ISSUE #590: Redact sensitive host artifact paths without changing source identity or policy scope
Found during the independent review of #589 (#547); deferred as requested rather than expanding the recovery change.

## Problem

The host inventory sanitizes configuration values, but `_artifact` retains its raw `source` as `artifacts[].path`. A credential-like value embedded in a directory name can therefore enter an inventory artifact even when the corresponding failed-input diagnostic and boundary result redact it.

## Reproduction

Create a temporary repository directory whose name is `settings TOKEN=<synthetic GitHub-token-shaped sentinel>` and place an invalid-UTF-8 `.mcp.json` inside it. Run `build_host_boundary_snapshot` or the repository host audit. The failed artifact path contains the sentinel. On #589, the new `issues[].source`/recovery text and `check` boundary output redact it and every permission remains denied. This is an output-privacy defect, not an observed authority bypass or disclosure of a real credential.

The independent reviewer compared `_artifact` on base `1d4c0f841b1d11f5f8e822d60263aa55cd77ded5` and head `5ed4b4452254102d2ddd7a8ee92a376781be6883`: both publish `path: source` unchanged. The same function derives `artifact_id` from the unsanitized source. `host_coverage[].sources_observed` is projected from artifact paths, so that projection must be inventoried as part of a repair; replacing one presentation field alone may leave another raw copy.

## Required follow-up

Trace path-bearing host inventory, grant, coverage and drift projections, then define consistent privacy and identity semantics for a redacted location. Preserve internal exact-path associations for source reads and policy classification. Do not truncate or redact paths before authority classification, silently merge different sources after redaction, or claim that a redacted path is an exact executable locator.

## Acceptance

- [ ] Public audit JSON/human output, host coverage and drift contain no synthetic credential sentinel from the failed-input path; source contents remain absent.
- [ ] The same privacy policy covers successful and failed artifact paths without altering exact internal read or policy scope.
- [ ] Redaction collisions and long paths have explicit stable identity/locator behavior; one source cannot inherit another source's grants or recovery evidence.
- [ ] Existing incomplete-input denial and baseline/drift comparability rules remain unchanged, or any required format migration follows STABILITY.md.

Related: #547 confines the immediate failure-recovery extension; #586 concerns source-omission redaction accounting, not host artifact paths. This issue is separately deferred and does not establish a candidate release privacy result.

## Selected-path review — 2026-09-12

Independent replay of the current host inventory shows a synthetic credential-shaped directory name surviving in artifact paths. The selected CLI/PR/JSON route must preserve source identity without publishing sensitive path bytes; this is a current output-privacy defect, not optional qualified-gate expansion.

Re-selected as P1 under #645 after the active #703/#712 review fixes. Use the existing issue and acceptance; do not add a new surface or change corpus denominators. A regression must fail on the observed current behavior and cover both the direct reader and its public comparison output. This changes sequencing, not completion status; #659/#660 final candidate replay still follows.

pengfei-threemoonslab:
Closes #590.

## Problem, reproduced on main (`294d0443`)

The fixture repository names three host files with a synthetic GitHub-token-shaped string:
- a skill directory (parsed successfully),
- a workflow file (parsed successfully),
- a nested `.mcp.json` directory holding invalid UTF-8 (a failed read).

The sentinel was published verbatim in:

| Output | Fields |
| --- | --- |
| `audit --host --json` | `artifacts[].path`, `grants[].path`, `grants[].source`, `host_coverage[].sources_observed[]` |
| saved baseline | `.agents-shipgate/host-grants.json` |
| `audit --host --drift --json` | `artifact_changes[].{baseline,current}.path`, `changes[].{baseline,current}.{path,source}`, `expansion_signals[]` |
| `verify --json` | `host_comparison.paths[]` (in `verifier.json` and `agent-handoff.json`) |

Issue sources were already redacted, but by the whole-string sanitizers: a greedy assignment pattern cut `/.mcp.json` away. Two unreadable sources that differed only in the secret therefore printed as the same source.

## Change (`core/host_grants.py`)

- **`public_host_path(source)`.** The one published form of a host location.
  - Each path component is redacted on its own, with the same sanitizers used for values (`privacy.redact_text` and `_sanitize_sensitive_string`). A pattern can never swallow the separator and the file name after it, and instruction profiles and precedence ranks read that file name.
  - When anything was redacted, the marker carries a 12-hex digest of the exact source, so two sources that redact alike stay two locations.
  - A path with nothing to redact is returned byte-identical.
- **Where it applies.** Artifact `path`, grant `source`, and instruction-grant `path`. `sources_observed`, drift, expansion signals, `host_comparison.paths` and capability-row subjects are projected from those fields, so they follow.
- **Identity and reads are unchanged.** `artifact_id` and `grant_id` still hash the exact source, and discovery, reads, instruction classification and policy scope still use it. A redacted path is a label, not a locator.
- **Issue sources.**
  - The source is `_issue_source_label`: the public path with control characters replaced, bounded at 512 bytes, and ending in a digest when truncated. It is no longer re-run through the whole-string sanitizers.
  - `issue_id` binds the shown source when that equals the exact source, so ordinary ids do not move. It binds the exact source when the shown form is lossy, so one source cannot inherit another's recovery evidence.

## Compatibility

- No schema or field changes.
- A path with nothing to redact is unchanged, so existing ids, baselines and drift are unaffected.
- A baseline saved before this change, from a path that held credential-shaped bytes, shows that artifact as changed once.
- Incomplete-input denial, baseline refusal and comparability rules are untouched.
- An unchanged limit on a redacted source no longer matches its Git path. That comparison refuses, which is the conservative direction.

## Out of scope, filed

`check` and `verify` still publish the repository's changed-file paths verbatim (`changed_files[]`, `violations[].path`, `host_coverage[].paths[]`). Those paths are locators agents open, so redacting them needs its own decision: #742.

## Tests (`tests/test_host_path_privacy.py`, 7)

- **Direct reader.**
  - No sentinel appears, for successful and failed sources alike.
  - A skill under a redacted directory is still classified as a structured skill.
  - Redaction is component-wise, deterministic and collision-free.
  - Two sources that redact alike keep distinct paths, `artifact_id`s, `issue_id`s and issue sources.
- **Public outputs.**
  - `audit --host` JSON and markdown carry no sentinel.
  - The saved baseline carries none, and an unchanged drift reports no artifact changes.
  - Widened drift names expansion signals without the sentinel.
  - `diff --json` and `verify --json` `host_comparison` carry no sentinel and still name `Bash(*)`.

## Verification

Mutation and full-suite results are in the first comment.


pengfei-threemoonslab:
## Verification

- **Mutations.** Each ran against `tests/test_host_path_privacy.py`; the source was restored and checked byte-identical with `cmp`:

  - FIRED main's publication (raw paths and whole-string issue sources): 6 failed
  - FIRED artifact path raw: 5 failed
  - FIRED grant source raw: 5 failed
  - FIRED instruction grant path raw: 4 failed
  - FIRED whole-string redaction: 3 failed
  - SURVIVED no source digest
  - SURVIVED issue id binds the shown source
  - FIRED issue source re-sanitized: 1 failed

  The first mutation restores main's publication behavior. The regression fails on it, which covers the direct reader and its public outputs.
- **Full suite on the final tree.** 10883 passed, 5 skipped in 592.45s (0:09:52)

pengfei-threemoonslab:
**Correction.** This comment's first version described a follow-up commit that does not exist. The edit script failed with a syntax error before changing anything, so nothing was committed or pushed. `f8d3f0e0` is still this PR's original commit.

The only valid result from that run: on the original code, removing both redaction digest stamps (the marker digest and the fallback digest) fails 2 tests. So the first comment's `no source digest` survival was a weak mutation, not a test gap.

The label change is being redone. It adds a digest whenever control characters are replaced or the length bound applies, and it removes the untestable id rule. Its results will be posted as a new comment.

pengfei-threemoonslab:
## Follow-up commit 24953a40

- **Label rule.** Replacing control characters and bounding the length now also leave a digest of the exact source. Redaction already did, in `public_host_path`. So no two sources share a label, and the issue id binds the label. The untestable exact-source id rule is gone.
- **New test.** `test_issue_sources_that_would_display_alike_keep_their_own_label_and_id` covers a tab versus a space, and two long paths that differ past the bound.

Mutations on this commit, each run against `tests/test_host_path_privacy.py`; the source was restored and checked byte-identical with `cmp`:

- FIRED no redaction digest at all: 2 failed
- FIRED no digest for a lossy label: 1 failed
- FIRED control characters kept in the label: 1 failed
- FIRED whole-string redaction: 3 failed
- FIRED artifact path raw: 5 failed

Focused host suites: 271 passed in 20.31s

PR #678, opened 2026-09-11
Tell a widened permission rule from a narrowed one (#657)

pengfei-threemoonslab:
Partially addresses #657. **Keep #657 open:** precedence resolution, scope-aware risk calibration and the broader acceptance measurements remain separate work.

Replacing an allow rule appears as one removal plus one addition. This change identifies direction only for a proven one-to-one replacement within the same host, scope, settings source and disposition. Narrowed grants remain visible in audit/diff without an expansion signal; independent additions in another file or scope retain theirs. Claude Code and Cursor's actual local boundary evaluators now use the same replacement predicate, so a proven narrowing also stops producing `HOST-PERMISSION-ALLOW-EXPANDED` there.

The comparison remains partial and host-specific. Claude path rules distinguish relative, project, home and absolute anchors and wildcard depth; supported Bash prefixes and MCP patterns are handled separately. Cursor comparisons are limited to supported Read/Write paths. Unsupported syntax and multiple-rule replacements remain undecidable and retain the existing addition signals.

The final review's unbounded-read concern is addressed by withdrawing the blanket low-risk assumption and restoring the existing wildcard gate for read-class grants. Read tools can reach sensitive files outside the workspace; their class alone does not prove containment. The audit class projection distinguishes read access from arbitrary execution, but does not call arbitrary reads low-risk. Scope-aware lowering remains in #657. Policy comments, release notes and tests now state this boundary; the older contrary documentation is removed.

## Validation and review

- Three formal GitHub reviews are already recorded; the final request is [5176774919](https://github.com/ThreeMoonsLab/agents-shipgate/pull/678#pullrequestreview-5176774919). The final author response addresses that request; no fourth formal review is added.
- Eight new negative-control failures reproduced the actual local narrowing interruption, cross-file warning leakage and unsupported read-scope assumption. All eleven new real-boundary/risk regressions pass after the fix.
- Permission lattice, real boundary and capability-diff selection: **132 passed**.
- Full repository regression: **9,715 passed, 5 skipped**; static-only **487 passed**, performance **4 passed**. Ruff and whitespace checks pass. Incoming main changes receive strict integrated-head CI before merge.

No new check ID, permission declaration, waiver or lowered wildcard safety floor. No runtime enforcement equivalence is inferred for an unsupported host syntax. This repair does not establish whole-v1 readiness or qualification.

[...]

Issue #661, opened 2026-09-11
Quiet local guard for supported host changes, with explicit unknown coverage and correct human handoff


pengfei-threemoonslab:
## Problem

The historical in-loop study found repeated prompts on benign edits. The current host diff and graded guard provide a narrower path, but direction-blind changes still create noise (#657/#685). Quiet operation must mean a correct covered comparison, not failure to read a permission change.

The earlier proposal also used Claude Code Stop `block` to mean waiting for a person. In the existing implementation (`src/agents_shipgate/cli/install_hooks.py`), that response asks the agent to keep working. Human-owned decisions therefore need a handoff, not an instruction to continue until the agent has somehow approved itself.

## Proposal

- Use #657's direction and precedence semantics for supported host edits. Fully covered unrelated or benign/narrowing host changes can avoid redundant advisory prompts. Missing refs, unknown semantics, parse failures, incomplete coverage and protected instruction/policy changes retain their supported diagnostic or control route; they are never classified as benign merely because no widening row was emitted.
- Explain each actual widening using #651's named row, with evidence and coverage available. The existing session-to-protected-path cache is advisory and does not bind evidence digests. Any new diff-based presentation deduplication must bind the exact compared evidence and invalidate on changes; the old path cache cannot stand in for that proof. A row string is not an authorization identity, and no memory hit grants control permissions.
- Use the existing supported hook installation path. If ConfigChange is supported by the declared host version, route external edits through the same comparison; unsupported host versions get an accurate fallback and documented limitation. Do not claim runtime enforcement or universal interception from static repository observations.
- At Stop, obtain the current supported routing answer. A human-owned decision emits a concise handoff and allows the agent to stop. Request continued work only for the exact action the current control authorizes for a coding agent. A fast advisory diff may reduce redundant work, but its success, zero rows or remembered presentation approval cannot grant merge/completion authority or replace verifier evidence where that evidence is required.

## Acceptance

- [ ] The 50-turn benign-session fixture has zero redundant advisory interrupts for fully covered code/docs/tests edits and a supported narrowing edit. Separate unknown-input and protected-change controls remain visible and follow their actual control routes.
- [ ] A supported `Bash(*)` widening is named once with evidence; repeating the same unchanged event uses evidence-bound presentation deduplication, while changed input, stale evidence and changed scope reopen evaluation.
[...]

PR #745, opened 2026-09-13
Stop calling a narrowed allow rule an expanded allowlist in check and verify (#661)

pengfei-threemoonslab:
Refs #661. This fixes one cause of redundant interrupts that the benign-session acceptance item names. The issue stays open for the remaining items below.

## Problem, measured on main (`294d0443`)

The fixture is an adopted repository (`samples/clean_read_only_agent`) with `install-hooks --target claude-code` applied, driven through the real CLI with real PostToolUse and Stop payloads.

| Edit | Stop (manifest present, so `verify` runs) |
| --- | --- |
| docs, tests | quiet |
| `.claude/settings.json` allow `[Bash(*), Read(**)]` → `[Bash(git status), Read(**)]` (a pure narrowing) | `review_required`: **`SHIP-HOST-BOUNDARY-PERMISSION-ALLOW-EXPANDED`** ("Coding-agent permission allowlist expanded") + `SHIP-VERIFY-TRUST-ROOT-TOUCHED` |
| same file widened to `Bash(*)`, `WebFetch(*)` | `blocked`: `SHIP-HOST-BOUNDARY-PERMISSION-WILDCARD-ALLOW` |

`core/host_boundary.py` took the set difference of allow rules: any rule text not in the old list was an expansion. #678 moved the drift reader and the audit table onto the permission lattice (#657), but not this evaluator, which `check`, `verify` and the PR route use.

## Change

- **`_widens_allow(rule, old_rules)`.** An added allow rule that some old allow rule subsumes grants nothing new and raises nothing. Only a decided `True` from `permission_lattice.subsumes` excuses a rule, so a pair the lattice declines stays an expansion.
- **Where it applies.** Claude Code settings (`.claude/settings.json`) and the Cursor CLI config (`.cursor/cli.json`), the two evaluators that compared allow lists by text.
- **Unchanged.** Deny removals, hooks, permission modes, unknown keys, and all MCP and workflow rules.

## Tests (`tests/test_permission_lattice.py`)

- **`test_direction_on_the_boundary_check`.** The existing 20 widen/narrow/unchanged pairs, which already run through the drift reader for both hosts, now also run through the boundary evaluator:
  - every widening raises an expansion;
  - no decided narrowing does;
  - the pair the lattice declines still does.
- **A narrower rule beside a retained wildcard** is not an expansion.
- **A rule no old rule covers** still is.
- **The Cursor CLI config** reads direction the same way.

## After this change, on the same fixture

The narrowing's review items are `SHIP-AGENT-BOUNDARY-PROTECTED-SURFACE-UNCLASSIFIED` and `SHIP-VERIFY-TRUST-ROOT-TOUCHED`, so the Stop hook still hands off to a human. The false "expanded" label is gone, but in an adopted repository `.claude/settings.json` is a release trust root, and any change to it is reviewed. Whether a lattice-decided narrowing of a trust root should still go to review is an owner question, raised on #661, not something this PR changes.

## Still open on #661

- the 50-turn benign-session fixture
- the trust-root narrowing question above
- the ≤ 1.5 s fast-path Stop budget: the manifest-free host diff route measured 1.9–2.1 s here


pengfei-threemoonslab:
## Verification

- **Mutations.** Each ran against `tests/test_permission_lattice.py`; the source was restored and checked byte-identical with `cmp`:

  - FIRED Claude settings keep set difference: 8 failed
  - FIRED Cursor CLI config keeps set difference: 1 failed
  - FIRED an undecided pair excuses the rule: 1 failed
- **Focused suites.** Boundary, lattice, agent-boundary, codex-boundary, route-parity, install-hooks, manifest-free PR rows, verify and samples: 488 passed in 53.52s
- **Tooling.** `ruff check .` is clean. CI is authoritative for the full suite.

PR #747, opened 2026-09-13
Read Go MCP servers' literal tool hints for the contradiction check (#658)

pengfei-threemoonslab:
Refs #658. This is the second annotation increment: Go. The issue stays open for the rest; see the end of this description.

## Problem

`SHIP-MCP-ANNOTATION-CONTRADICTION` can challenge a FastMCP `readOnlyHint` read from source (#732). It could not challenge a Go server's, because neither Go spelling was read. Both appear in real servers:

| Shape | SDK | Example |
| --- | --- | --- |
| `mcp.Tool{Name: "list_gists", Annotations: &mcp.ToolAnnotations{Title: t(...), ReadOnlyHint: true}}` | `modelcontextprotocol/go-sdk` | `github/github-mcp-server`, `pkg/github/gists.go` |
| `mcp.NewTool("delete_team", ..., mcp.WithReadOnlyHintAnnotation(false), mcp.WithDestructiveHintAnnotation(true))` | `mark3labs/mcp-go` | `hashicorp/terraform-mcp-server`, `pkg/tools/tfe/delete_team.go` |

## Change

The same code lands in both readers: `inputs/mcp_idioms.py` and the zero-install port `tools/shipgate-detect.py`.

- **`_go_struct_annotations`.** For a `go_tool_struct` site, it reads the literal's own `Annotations:` field, an `mcp.ToolAnnotations{...}` literal with or without `&` and with or without a package qualifier. It keeps `ReadOnlyHint` and `DestructiveHint` when each is exactly `true` or `false`.
- **`_go_option_annotations`.** For `go_new_tool` and `go_must_tool` sites, it reads `With{ReadOnly,Destructive}HintAnnotation(<bool>)` among the call's own arguments, in SDK order, so a later option overrides an earlier one.
- **`annotate=` hook.** `_call_sites` takes it alongside `describe=`.
- **Refused.** Each of these sets `annotations_unresolved` and reads nothing, as #732 does for Python:
  - a pointer helper such as `jsonschema.Ptr(true)` (go-sdk's `DestructiveHint` is a `*bool`) or `mcp.ToBoolPtr(true)`;
  - `WithToolAnnotation(...)`, a variable or computed value, positional fields, or a second argument.
- **Ignored.** `Title`, `IdempotentHint` and `OpenWorldHint`, options nested inside a func literal, and anything not written as one of the idiom's own arguments.
- **Unchanged.** The source-to-tool bridge and the gate. A hint is a claim `SHIP-MCP-ANNOTATION-CONTRADICTION` may challenge, and `annotation_hints_are_effect_evidence` keeps it out of every effect reading.

## Tests

- **Shared corpus.** `tests/mcp_idiom_corpus.py` gains `GO_ANNOTATION_CASES`: 19 cases, 10 reads and 9 refusals, across both spellings. They run through both readers in the existing parity sweep, including the CRLF variant.
- **Reader test.** `test_the_go_reader_keeps_literal_hints_and_refuses_the_rest` pins each case's `(hints, unresolved)`.
- **End to end.** `test_a_go_source_hint_earns_the_same_contradictions_and_nothing_else` scans a Go server with both spellings, with and without the hints.
  - `delete_team` and `purge_account` gain the contradiction, with `published_annotations` naming the hint.
  - `list_teams` does not.
  - No other finding appears, and none of the new findings blocks release.

## Still open on #658

- **Owner question.** #658's acceptance asks for "read-only tools >= 90% projected `read`" on github-mcp-server. #732 established that a source-read hint is a claim, never evidence: without that gate, a hint lowered `transfer_funds` to `read` with no finding. Reading github-mcp-server's `ReadOnlyHint: true` now lets the contradiction check challenge it, but by design it does not project those tools as `read`. That acceptance item needs a ruling: drop it, or restate it as "the hint is read and checked".
- **Not yet read.** TypeScript `annotations: { readOnlyHint: ... }`. Only one literal use turned up in a code search, so it is lower value.
- **S2.** The renamed `process_account` still needs body evidence.
- **The ten-server false-finding table** is still unmeasured.

## Verification

Mutation and full-suite results are in the first comment.


pengfei-threemoonslab:
## Verification

- **Mutations.** Each ran against `tests/test_mcp_source_annotations.py` and the zero-install parity sweep in `tests/test_zero_install_detector.py`. The sources were restored and checked byte-identical with `cmp`:

  - FIRED struct hints not read: 24 failed
  - FIRED option hints not passed for NewTool: 22 failed
  - FIRED a pointer helper is skipped instead of refused: 6 failed
  - FIRED WithToolAnnotation is ignored: 3 failed
  - FIRED a later option does not override: 3 failed
  - FIRED positional fields are tolerated: 3 failed
  - FIRED the port reads no struct hints: 16 failed
- **Full suite on `84c4ce1c`.** 10981 passed, 5 skipped in 743.78s (0:12:23)

Issue #643, opened 2026-09-11
Epic: advisory v1.0 delivery record — finish the published Action replay


pengfei-threemoonslab:
## Published advisory scope — 2026-09-14

`v1.0.0` is published from `bace7c1871834e0b3eb98e6f60c0627725c53a59`; its wheel is `d34012cad8164a8b9c092ad3512e4b15ec36989b330f7ce6c60c13156babd771`. [Release](https://github.com/ThreeMoonsLab/agents-shipgate/releases/tag/v1.0.0), [publication run](https://github.com/ThreeMoonsLab/agents-shipgate/actions/runs/34902863526), and #777's current pins replace the pre-publication state below. This is advisory, with no qualification claim.

The selected product has shipped. The remaining checklist work here is **#570's literal post-publication tagged-Action replay**: the published wheel and ordinary generated pin have been checked, but a local replay or pre-publication Action run is not a post-publication hosted Action observation. Keep that obligation visible; do not manufacture evidence to close it. #780 carries the host-only integration recipe and can supply that run.

This is a publication-evidence follow-up, not a stop on #653/#571's supported, nonblocking user research. #778 owns the next execution sequence. #572 retains qualified-gate obligations separately. #779 and #781 own concrete post-release entry defects; they do not reopen the entire v1.0 feature scope.

The original recall and benign bars were **not met**: the current engineering record is row precision 70/70, widening recall 55/65, benign zero-row 5/6, and 41/50 comparable. The [owner's retained-refusal decision](https://github.com/ThreeMoonsLab/agents-shipgate/issues/643#issuecomment-5657575213) explains the advisory scope; it does not convert those numbers into passes. Cold-start acceptance is 26/30; the ten-server table has 76 findings, one false. These are engineering measurements, not human qualification or adoption observations.

## Remaining acceptance

- [ ] #570 records the published tag/Action execution and engine parity on the actual downloaded release.
- [ ] Close this delivery record with links to that evidence and the retained metric exception; external participation and completion of #572 are not additional closure requirements.

## Original readiness specification (historical)

The record below describes the implementation milestones before publication. Its future-tense sequencing is superseded by the dated status above; its evidence requirements are not silently marked passed.

## Why this epic exists

#572 owns the release decision and qualified-gate evidence. This epic owns the product-readiness definition: what a developer or platform/DevEx reviewer can install and use to understand an agent capability change.

## The rule this epic is built on

[...]

PR #773, opened 2026-09-14
Build the MCP server on the SDK 2.x MCPServer (#713)

### LINKED ISSUE #713: Installed MCP 2.x extra cannot build the local server: FastMCP was removed
## Problem

An adopter installing `agents-shipgate[mcp]` receives MCP 2.x under the current declared range, but the local stdio server still imports `mcp.server.fastmcp.FastMCP`. MCP 2.2.0 deliberately raises `ModuleNotFoundError` from that module: the public entry point is now `mcp.server.mcpserver.MCPServer`.

Shipgate catches the exception and tells the user to install the same extra again. Reinstallation cannot repair this API mismatch. Default static scanning is unaffected.

## Evidence

Independently reproduced while reviewing #705 at `b272b6acef7f6e00c5c260d0ebfbeaf6fb140b90`, with exactly `mcp==2.2.0` installed in a disposable environment. `tests/test_mcp_server.py` produced 10 passed, 1 skipped, 1 failed. The failing test was `test_build_server_registers_read_only_tools`; the import fails at `src/agents_shipgate/mcp_server/server.py:209` before registering any tool.

The current main range already allows MCP 2.x (`>=2.1.1,<3`), so #705 did not introduce the original incompatibility. The locked development closure does not install the MCP extra; its successful suite is not installed-extra compatibility evidence.

Official source: [MCP v2 migration](https://py.sdk.modelcontextprotocol.io/v2/migration/#fastmcp-renamed-to-mcpserver), [2.2.0 package metadata](https://pypi.org/pypi/mcp/2.2.0/json).

## Suggested implementation

Choose the supported SDK contract explicitly: either migrate the adapter to MCP 2.x and its relevant API changes, or bound the optional requirement to a version whose FastMCP contract is actually supported. Do not add a fallback that hides an incompatible installed SDK as a missing dependency. Verify the five read-only tools and their annotations with the real installed extra.

## Acceptance criteria

- [ ] A fresh installation of the declared MCP extra builds the server and lists all five expected tools.
- [ ] Every tool retains `readOnlyHint=True` and `openWorldHint=False`; the adapter remains local/read-only and starts only on explicit invocation.
- [ ] An incompatible SDK produces an actionable version/API diagnostic rather than an ineffective reinstall instruction.
- [ ] A focused installed-extra check exercises the selected minimum version and supported range; the optional-free static suite remains independent.

## Scope and sequencing

Deferred, no v1.0 milestone added by this review. Keep #705 unmerged until its dependency/runtime contract is addressed. Reconsider priority if the v1.0 readiness path specifically requires this optional MCP server; this issue alone does not expand the repository-only release scope or require live agent/provider execution.

pengfei-threemoonslab:
Closes #713. Refs #643: one of the two tag blockers found in the 2026-09-14 readiness review.

## Problem

The `[mcp]` extra has installed MCP Python SDK 2.x since #354, but the server imported `mcp.server.fastmcp.FastMCP`. SDK 2.0 renamed it to `MCPServer` and ships `mcp.server.fastmcp` as a module that raises. So `pip install "agents-shipgate[mcp]"` followed by `agents-shipgate mcp-serve` could not start, and the error asked the user to install the extra they already had. No CI job installed the extra, so the one test that builds the server was skipped everywhere.

## Change

- Build the server on `mcp.server.mcpserver.MCPServer`, with snake_case `ToolAnnotations`.
- Tell a missing SDK apart from an SDK without `MCPServer`. The second names the installed version and the required range instead of repeating the install step.
- Declare the floor as `2.0.0`, the first release with `MCPServer`. A test holds the range the server names equal to the one `pyproject.toml` declares.
- Add a `mcp-extra` CI job with two legs. It installs the extra's SDK at its floor and at the newest release the range admits, reading both from `pyproject.toml`. It runs `tests/test_mcp_server.py` against the real SDK, then lists the tools over stdio from the installed `agents-shipgate mcp-serve`.

The new job is not a required check; it runs on every PR.

## Verification

- **Before:** on SDK 2.0.0 with `main`'s source, `test_build_server_registers_read_only_tools` failed at the import.
- **After, SDK 2.0.0 and 2.2.0:** `tests/test_mcp_server.py` exits 0 with the registration test running. The stdio smoke prints `mcp-serve answered over stdio with 5 read-only tools` on both.
- **SDK 1.27.2:** the registration test is skipped, and building the server raises the new diagnostic naming `mcp 1.27.2` and `mcp>=2.0.0,<3`.
- `tests/test_release_pipeline.py`, `tests/test_ci_recipes.py`, `tests/test_distribution_surface_parity.py` and `tests/test_docs_links.py` pass. Ruff passes on the changed Python files.
- The local full suite is still running; its result will be added as a comment.


pengfei-threemoonslab:
**Local results, and one follow-up commit.**

- **Full suite on `3d63108d`:** 10,680 passed, 6 skipped, 0 failed, exit 0. It ran CI's split: `-n auto -m "not perf" --ignore=tests/test_adapter_static_only.py`.
- **CI `test` failed at the trust-model invariant lint.** That step runs `tests/test_adapter_static_only.py`, which my local run excluded. The lint rejects `from importlib import metadata` as an import from the forbidden `importlib` module. `b92d2e05` respells it as `from importlib.metadata import PackageNotFoundError, version`, the form `environment.py` already uses.
- **On `b92d2e05`:** `tests/test_adapter_static_only.py` exit 0; `tests/test_mcp_server.py` exit 0 on SDK 1.27.2, 2.0.0 and 2.2.0; ruff passes.

Issue #781, opened 2026-09-15
A published 1.0.0 adoption kit installs 0.15.0 and falsely stops at the contract floor


pengfei-threemoonslab:
## Problem

An adopter can install Agents Shipgate `1.0.0`, ask it to render its existing optional adoption kit, and receive instructions that install `0.15.0` and stop because no published build supposedly supports runtime contract `21`. The running released package actually reports contract `39`.

This is an executable first-adoption failure, not just stale release prose. It can move the user back to an older engine or interrupt a workflow the installed engine supports. Updating the pins on `main` in #777 does not change the already published wheel.

## Evidence — published 1.0.0, 2026-09-14

The [PyPI release](https://pypi.org/project/agents-shipgate/1.0.0/) and [GitHub Release](https://github.com/ThreeMoonsLab/agents-shipgate/releases/tag/v1.0.0) serve the same wheel, SHA-256 `d34012cad8164a8b9c092ad3512e4b15ec36989b330f7ce6c60c13156babd771`, built from `bace7c1871834e0b3eb98e6f60c0627725c53a59`. An independent clean virtualenv installed those bytes outside a checkout and reported version `1.0.0`, runtime contract `39`, report schema `1.0`.

In an empty temporary directory, this real installed command exits `0` and reproduces the problem without writing or invoking a skill:

```sh
agents-shipgate init --workspace /path/to/empty-temp --minimal \
  --agent-instructions=claude-code-skill,codex-skill --json
```

The rendered output includes:

| Generated target | Observed behavior |
|---|---|
| Codex `assets/advisory-pr-comment.yml` | Action `@v0.15.0` and `shipgate_version: '0.15.0'` |
| Claude `ci-recipes/advisory-pr-comment.yml` | The same previous-release Action/package pins |
| Claude `prompts/add-shipgate-to-repo.md` | `uvx agents-shipgate@0.15.0` / `pipx run agents-shipgate==0.15.0`; says no published release reports contract `21` and directs the caller to stop at that boundary |
| Claude `prompts/decide-shipgate-relevance.md` | The old runners, Action ref and false contract-floor claim |
| Claude `prompts/stabilize-strict-mode.md` | Previous-release Action/package pins |

The files are rendered beneath `.agents/skills/agents-shipgate/` and `.claude/skills/agents-shipgate/` respectively. The optional targets, rather than the default instructions, are the reproduced path.

[...]

Issue #779, opened 2026-09-15
Make the published advisory host diff the first review: reconcile current release claims and entry instructions


pengfei-threemoonslab:
## Observed

The 2026-09-14 post-release review inspected `main` at `e2ab00078b2ca5b9e73405109a6f09b820f5fc9e`, after 1.0.0 publication and the pin refresh in #777.

- `README.md:98–101` says the published release predates `host_boundary_candidates` (contract 33), while the installed 1.0.0 reports contract 39. `README.md:144` still says pre-1.0 beta, and its channel table treats advisory as preview-only while the following paragraph identifies 1.0.0 as advisory.
- `docs/distribution.md:9–27` still maps `v*` to qualified releases and describes attached independently qualified wheel/safety-qualification artifacts. The actual [1.0.0 advisory statement](https://github.com/ThreeMoonsLab/agents-shipgate/releases/download/v1.0.0/advisory-statement.json) declares advisory defaults and no qualification claim.
- The README's first task is a constructed refund-tool gate. The selected job in `ROADMAP.md:24–39` is a reviewer understanding repository-declared host-configuration changes.
- `src/agents_shipgate/cli/diff.py:3–11` already uses Git to supply the base without a manifest or committed baseline. The 737-line quickstart never demonstrates `shipgate diff`; its Route H section starts at line 390 with audit, baseline creation and control JSON.
- The pilot's 1.0.0 remeasurement is current, but its runbook still describes an ahead-of-release loop and the old host-only `init` dead end. #571 now supplies a dated successor to the obsolete preview-only invitation decision.

These are conflicting instructions about what a user can install and do now, not a request for a new onboarding subsystem. Updating main does not rewrite the description embedded in already published PyPI 1.0.0 metadata.

## Why it matters

The intended first user has a PR and wants to know what changed. They can obtain that answer before authoring policy, learning the control protocol or deciding to add a gate. The present entry makes those heavier concepts prominent and asks the user to reconcile release history before trying the simplest supported workflow.

The selected promise is advisory review of **repository-declared static host changes**. It does not answer every runtime-effective permission or decide whether an arbitrary agent is safe to merge. A concise useful example and an honest incomplete result are better entry evidence than more supported-framework names.

## Proposed

- Make one current human entry in README/quickstart: an existing permission or MCP-configuration PR → the published CLI → `shipgate diff` on the intended Git base → a reviewer-understandable result. No manifest, policy, saved baseline, skill or account is required for this route.
[...]

Issue #570, opened 2026-09-08
Complete the published v1.0.0 tag/Action replay after the ordinary CI pin repair


pengfei-threemoonslab:
## Remaining work after publication — 2026-09-14

The original ordinary-CI pin-selection defect is fixed by #616 and shipped in `v1.0.0`. Installing the public wheel (`d34012cad8164a8b9c092ad3512e4b15ec36989b330f7ce6c60c13156babd771`) and generating normal CI selects `bace7c1871834e0b3eb98e6f60c0627725c53a59` and package `1.0.0`. #777 updated current source/documentation pins.

What is still unproven against the literal acceptance below is the **post-publication execution of the real tag/Action**. [Smoke 34900819498](https://github.com/ThreeMoonsLab/agents-shipgate/actions/runs/34900819498) exercised the candidate before publication; the [post-publication comment](https://github.com/ThreeMoonsLab/agents-shipgate/issues/570#issuecomment-5671727824) records the PyPI/local replay and pin resolution. Both are useful; neither is evidence of a new post-publication hosted Action run.

## Acceptance for closure

- [ ] Execute the real published `v1.0.0` composite Action with the actual public release in a disposable downstream/controlled GitHub fixture; record tag resolution, source SHA, installed version, runtime contract, wheel digest and the run URL. This may be the same hosted run used by #780; no duplicate harness is required.
- [ ] Compare its supported capability-change result with the installed CLI and distinguish the tag convenience reference from the generated immutable-SHA default. Do not rewrite or republish the existing wheel.
- [ ] Link the result from #646/#643 and close the distribution follow-up. Any newly reproduced product defect gets its own bounded issue rather than an invented pass.

P1 in the first post-release integration wave under #778. This observation does not block the supported local, nonblocking #653/#571 pilot. #781 is a distinct defect in optional bundled prompts/recipes; it does not reopen the fixed ordinary CI renderer.

## Selected candidate scope — 2026-09-11

The owner selected repository-independent advisory host-review readiness (#643) before adoption expansion. This issue remains required: validate the exact candidate in its real distribution channel, including post-publication download and actual Action replay. Qualification and independent signing are additional obligations for the qualified channel, not evidence that a synthetic preview smoke can manufacture. Preserve applicable publication integrity controls in either channel.

## Problem

[...]

PR #783, opened 2026-09-15
Render the release engine into the optional adoption kits (#781)

pengfei-threemoonslab:
Refs #781 — its release-time acceptance item (checking the next patch wheel through its candidate path and then the downloaded published bytes) stays open on the issue after this merges.

## Problem

A released wheel is built before the published-release constants move, so it carries the *previous* release in `LATEST_PUBLISHED_VERSION` / `LATEST_PUBLISHED_CONTRACT_VERSION`. The published `1.0.0` wheel still named `0.15.0` / contract `10` there. Two paths of the optional skill kits turned that into executable instructions:

1. The bundled CI recipes (`claude-code-skill/ci-recipes/advisory-pr-comment.yml`, `codex-skill/assets/advisory-pr-comment.yml`) and two prompts carried **hand-written literal pins**, bumped by hand after each tag. The wheel shipped whatever they said at build time.
2. The runner pins and the contract-floor sentence were rendered from the published constants, so the released `1.0.0` told adopters to install `0.15.0` and said no published release reports contract `21`, while it reports `39`.

`init --ci` did not have this bug: it already reads `_meta/release-source.json` (#570).

## Change

- **One selection rule.** `release_source.release_engine()` returns the engine that every pin `init` writes selects: a valid release-source record gives this wheel's own version, its immutable Action source SHA and the contract it emits; no record gives the published fallback (#506); a malformed record raises. `init --ci` (`ci_workflow._engine_selection`) and the adoption-kit renderer both use it, so they cannot disagree.
- **No hand-written pins left in the kit templates.** The four literal pins become `{{ shipgate_action_ref }}` / `{{ shipgate_version }}`. In a source checkout they render to exactly what was there before, so only the Claude recipe's explanatory comment changed rendered bytes (hash updated, outgoing hash appended to `prior_render_sha256`, mirrors re-rendered).
- **The floor is judged against the selected engine.** `contract_floor_prose(floor, engine)` describes a stamped wheel as "This release, `agents-shipgate` `X`, reports it"; the published-fallback wording is unchanged.
- **`AGENTS_SHIPGATE_WORKFLOW_REF` keeps its scope.** It still selects the Action source of the workflow `init --ci` writes (and still cannot hide a malformed record). It does not reach the kit's runner pins, which cannot name an Action source; this is documented and tested.
- **Candidate-path check.** `scripts/release_engine_smoke.py prepare` now renders both kits from the installed candidate and refuses any Action ref other than the source commit, any package version other than the candidate's, or the "no published release reports" sentence. The pins are recorded as `generated_kit_pins` in the smoke evidence.

## Tests

- `tests/test_release_source.py`: stamped wheel with the pre-publication constants (`0.15.0`/`10`) renders only its own SHA/version and the "This release" floor; the unstamped control keeps the published fallback; `init --ci` and the kit recipes name one engine; the override does not reach the kit; a malformed record refuses the kit even with an override; stamped-engine gap prose.
- `tests/test_packaging.py::test_a_stamped_installed_wheel_kit_does_not_downgrade_or_stop_at_the_floor`: builds the wheel, installs it outside the checkout, rewrites the installed constants to the pre-publication state, and runs the real `init --workspace <empty> --minimal --agent-instructions=claude-code-skill,codex-skill --json` route with and without a release-source record. It asserts the import came from the installed copy and that the dry run wrote nothing.
- Negative control: the new `test_release_source.py` kit tests fail against `origin/main` (`({'v1.0.0'}, {'1.0.0'}) != ({SHA}, {'1.0.0'})`).
- Local: static-only guard passed; full suite with CI's split (`-n auto -m "not perf" --ignore=tests/test_adapter_static_only.py`) — see the validation comment.

## Not in this PR

- Already-published `1.0.0` bytes are unchanged; the repair reaches adopters with the next patch release. The acceptance item that checks the final patch wheel through its candidate path and then the downloaded published bytes is release-time evidence for that patch; the smoke check above is what performs the candidate half.
- No default skill invocation or new kit surface (#690 stays suspended). Public entry prose is #779.


pengfei-threemoonslab:
**Local validation on `e97c9d9f`**

- Full suite, CI's split (`-n auto -m "not perf" --ignore=tests/test_adapter_static_only.py`): **10,712 passed, 7 skipped, 0 failed**, exit 0.
- `tests/test_adapter_static_only.py` (separate CI step): 495 passed, exit 0. Its allowlist pin for `adoption_kit.py`'s `importlib.resources.files` call moved from line 359 to 357 because the import block shrank.
- `ruff check src scripts` and the changed tests: clean; `git diff --check`: clean.
- Installed-wheel regression (`tests/test_packaging.py::test_a_stamped_installed_wheel_kit_does_not_downgrade_or_stop_at_the_floor`): passed against a freshly built wheel installed outside the checkout.
- Negative control: the new `tests/test_release_source.py` kit tests run against `origin/main`'s code fail with `({'v1.0.0'}, {'1.0.0'}) != ({SHA}, {'1.0.0'})`, i.e. the stamped wheel's kit did not name its own source before this change.

pengfei-threemoonslab (commented):
Independent coding-agent review/address cycle 1, reviewed at `e97c9d9f`.

Scope: the full diff, #781's acceptance, every target in `SPECS`, the smoke script's real CLI path, and malformed-record behaviour on the installed route. No P0. Findings:

1. **P1** — `add-shipgate-to-repo.md` (all four copies) told agents the generated workflow references `ThreeMoonsLab/agents-shipgate@v…`, while a stamped wheel's `init --ci` writes a 40-hex SHA. The sweep, the kit tests and the smoke check all exempted `v…` as a reader blank, so nothing caught it.
2. **P2** — With a malformed record, `init --agent-instructions` for a skill kit crashed with a traceback, and with `--write` it did so after writing `shipgate.yaml`/`AGENTS.md`.
3. **P2** — `Closes #781` would close the issue before its release-time item (patch candidate plus downloaded published bytes).
4. **P2** — The smoke's new `_kit_pins` had no unit test.
5. **P2** — `docs/distribution-surfaces.md`: the release-channels row did not name the kit, and the registry did not register the stamped proofs.
6. Nit — the engine was re-resolved per file. Nit — `upgrade-shipgate-version.md` has no SHA-pin guidance (predates this PR; not changed here).

Verified correct: ordinary-build renders byte-identical apart from the Claude recipe; `init --ci` output unchanged; override semantics; no import cycle; upgrade recognition uses sidecar/bootstrap hashes, not `prior_render_sha256`; the smoke regexes accept the stamped shape and refuse unstamped and malformed kits.

pengfei-threemoonslab:
Addressed review cycle 1 in `3fbccf0b`:

1. **P1** — The verification line now renders `@{{ shipgate_action_ref }}` (or the override's ref). The four copies were re-rendered; the render hash is updated and the outgoing hash appended to `prior_render_sha256`. `v…` is removed from `READER_BLANK_REFS`, the kit tests and the smoke check, so only the `upgrade-shipgate-version.md` blank `v<NEW>` stays exempt. The sweep's "still in use" guard holds.
2. **P2** — `init` calls `release_engine()` before any write whenever `--ci` or a file-tree kit is requested. A malformed record exits 2, prints the record error, writes nothing, and emits an agent-mode `config_error` routed to a reinstall. `tests/test_release_source.py::test_init_refuses_a_malformed_record_before_writing_anything` covers the dry-run kit, `--write` with `agents-md,claude-code-skill`, and `--write --ci`; a control shows a non-pin target ignores the record.
3. **P2** — The PR body now says `Refs #781`; the release-time item stays on the issue.
4. **P2** — `tests/test_release_engine_smoke.py::test_kit_pin_check_accepts_only_the_candidates_own_engine` covers acceptance and each refusal (tag ref, `v…`, stale package version, gap sentence, unrendered template).
5. **P2** — The release-channels row names the kit's stamped selection. The `adoption_kits` row and `Surface` register the stamped proofs under `executable_pin`/`contract_floor`, and the harness resolves `tests/test_packaging.py` names. The mirrors stay on source-rendering proofs, because only `init`'s render is stamped.
6. Nit — the engine is resolved once per `render_adoption_kit`. `upgrade-shipgate-version.md` SHA guidance is left as a pre-existing gap.

Validation on `3fbccf0b`: the targeted set exits 0 (release source, renderers, adopter pins, surface parity, engine smoke, init CI, prompt parity, plugin package, public surface, agent-instructions apply, first adoption). The static-only guard exits 0 after the allowlist line for `adoption_kit.py` moved to 359. The installed-wheel stamped-kit regression passes. Ruff and `git diff --check` are clean. An independent cycle-2 review follows.

pengfei-threemoonslab (commented):
Independent coding-agent review/address cycle 2, reviewed at `3fbccf0b`.

Verified with fake installed sites and the real CLI in subprocesses. Every cycle-1 fix resolves its finding, and no regression was found:
- The stamped kit names only the source SHA; `init --write --ci` agrees.
- Eleven malformed-record flag combinations, with and without agent mode: pin-rendering targets refuse with exit 2, write nothing and print no traceback; non-pin targets and `--local-review` are unaffected.
- `_AI_SPECS[name]` is safe for every parsed target.
- The one `_render_template` caller outside the renderer still works.
- Valid-record JSON is unchanged, and the upgrade from kit files rendered with `0.15.0` migrates cleanly.

No P0.

- **P1** — The checked-in copies of `add-shipgate-to-repo.md` (`prompts/`, `skills/`, `plugins/`) render from source and now say the workflow references `@v1.0.0`. The engine they pin, the published `1.0.0` wheel, writes a 40-hex source SHA from `init --ci`, so an agent following those copies would fail its own check.
- **P2** — The malformed-record refusal test did not assert the agent-mode structured `config_error` line.
- **Nits** — The refusal is checked before a scope refusal (both exit 2 and write nothing; acceptable). The refusal text repeated the reinstall instruction. The smoke `_kit_pins` would accept a kit that kept an unrendered `{{ shipgate_action_ref }}` beside the SHA. `upgrade-shipgate-version.md` has no SHA-pin guidance (pre-existing).

pengfei-threemoonslab:
Addressed review cycle 2 in `7dc3c432`:

- **P1** — The verification line is now engine-agnostic: the workflow's `uses:` line names the Action "at a release tag, or at the full 40-character source commit a release wheel records" (or the override's ref). The line carries no `@ref`, so it is true for both builds, and the template and all four copies render identically. The render hash is updated and the outgoing one appended to `prior_render_sha256`. `test_the_prompts_workflow_check_accepts_the_ref_init_writes` writes the workflow from an unstamped and a stamped build, checks each written ref is a tag or a 40-hex SHA, and checks the rendered line names neither concrete ref.
- **P2** — The refusal test runs with `AGENTS_SHIPGATE_AGENT_MODE=1` and asserts exactly one structured line: `error == "config_error"`, `exit_code == 2`, the record named, and `setup_incomplete`.
- **Nits** — `_kit_pins` refuses an unrendered `{{` (GitHub `${{ }}` expressions excepted), with unit cases for both. The refusal text is trimmed. Scope-refusal ordering is left as reviewed. SHA-pin guidance in `upgrade-shipgate-version.md` stays a pre-existing gap.

Validation on `7dc3c432`: the targeted set exits 0 (release source, engine smoke, renderers, adopter pins, surface parity, prompt parity, plugin package, public surface, init CI, agent-instructions apply, first adoption, static-only). Ruff and `git diff --check` are clean. An independent cycle-3 review follows.

pengfei-threemoonslab (commented):
Independent coding-agent review/address cycle 3 (final), reviewed at `7dc3c432`.

**No actionable findings remain at P0, P1 or P2.**

Verified:
- **Verification line.** Line 199 of `add-shipgate-to-repo.md` is identical in the template and all three checked-in copies, and true for both builds. Unstamped `init --ci` writes `@v1.0.0`; stamped writes a 40-hex SHA plus `shipgate_version`; with the override set, the override's ref is written.
- **Mirrors.** All 16 rendered kit files match all four mirrors byte for byte.
- **Sweep.** No other claim about the ref `init --ci` writes is false for a stamped wheel, across the four kit/prompt families, `docs/`, `AGENTS.md`, `llms*.txt` and `.claude/commands`.
- **P2 refusal test.** Mutation-tested: removing `_emit_agent_mode_error_routing` from the refusal branch fails it, in a full repo copy, since `conftest.py` puts `src/` first.
- **Smoke guard.** The `{{` guard never false-fails on the real rendered kits.
- **Test runs.** The targeted set and the installed-wheel stamped test exit 0.

Nits, all pre-existing:
- `docs/distribution-surfaces.md` #506 gap rows described only ordinary-build pins.
- `benchmark/runner.md:80` reads `@v…`.
- Line 199 names `agents-shipgate.yml` where a scoped manifest gets `agents-shipgate-<project>.yml`; the prompt already says to read `workflow.path`.
- `upgrade-shipgate-version.md` has no SHA guidance.

pengfei-threemoonslab:
Cycle 3 closed in `5d9fa150`. The one nit in a document this PR already edits is fixed: the two #506 gap rows in `docs/distribution-surfaces.md` now name the stamped-wheel pins (#570, #781); `tests/test_distribution_surface_parity.py` exits 0. The other nits are pre-existing and outside #781's kit scope (`benchmark/runner.md` rubric, scoped workflow file names, upgrade-prompt SHA guidance) and are left as they are. Three independent review/address cycles are complete with no remaining P0/P1/P2 findings.

PR #788, opened 2026-09-15
Read workflow step action references (#771)

pengfei-threemoonslab:
Closes #771

A workflow step's `uses:` never reached the workflow grant, so moving `actions/checkout` from a pinned SHA to `@main` gave a comparable result with no row and no limit. Each step's remote action reference is now read and compared as declared text. A change shows up as a `changed` row that names the job and step, with `expands: false`, in `diff`, `check`, manifest-free `verify`, the PR comment and the control envelope.

## Reproduction before the repair (published `agents-shipgate==1.0.0`, clean venv, scratch git repo outside any checkout)

The base commit pins `actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683`. The working tree changes it to `actions/checkout@main`; nothing else changes (`on: pull_request`, `permissions: contents: read`).

```
$ agents-shipgate --version
Agents Shipgate 1.0.0
$ agents-shipgate diff --workspace <repro> --base HEAD --json
{
  "base_commit": "a57e9045c8fa598880289a28859c6099b0259e38",
  "base_ref": "HEAD",
  "capability_diff_schema_version": "0.2",
  "comparison_status": "comparable",
  "incomparable_reasons": [],
  "rows": [],
  "static_analysis_only": true,
  "unchanged_limits": [],
  ...
}
$ agents-shipgate diff --workspace <repro> --base HEAD
Agent capability diff  HEAD (a57e9045) -> working tree

No static host-grant changes detected. No verdict is implied.
```

## After the repair (same repo, `./shipgate` from this branch)

```
$ ./shipgate diff --workspace <repro> --base HEAD --json
  "comparison_status": "comparable",
  "rows": [
    {
      "after": "read, test: contents: read, on: pull_request, test/steps[0]: uses actions/checkout@main",
      "before": "read, test: contents: read, on: pull_request, test/steps[0]: uses actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683",
      "direction": "changed",
      "expands": false,
      "severity": "low",
      "subject": "github .github/workflows/ci.yml",
      "why": "a step's action reference changed (test/steps[0]); it names different code to run with that job's existing token permissions and adds no scope"
    }
  ],
  "unchanged_limits": [],
$ ./shipgate diff --workspace <repro> --base HEAD
  low  changed  github .github/workflows/ci.yml
               read, test: contents: read, on: pull_request, test/steps[0]: uses actions/checkout@11bd…683 → read, …, test/steps[0]: uses actions/checkout@main
[...]

Issue #771, opened 2026-09-14
A workflow step's action reference is unread: moving `uses:` from a pinned SHA to `@main` produces no row


pengfei-threemoonslab:
## Observed

On `main` at `0bab265d`, a workflow step moved from a pinned commit to a branch:

```yaml
on: pull_request
permissions:
  contents: read
jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683   # base
      - uses: actions/checkout@main                                        # head
```

`shipgate diff --workspace . --base HEAD --json` returns `comparison_status: comparable`, no `incomparable_reasons`, no `unchanged_limits`, and no rows. The human output says no static host-grant changes were detected.

## Why

`_workflow_grant` (`src/agents_shipgate/core/host_grants.py`) projects triggers, token permissions, and a job-level reusable-workflow `uses` with `secrets: inherit`. A step-level `uses` never enters the grant. The step's code runs in the job's `GITHUB_TOKEN` context and can use credentials actually supplied to it, so moving it from an immutable SHA to a mutable branch changes what runs with that authority. This was found by an adversarial probe during the 1.0 readiness review, beside #701 (composite actions) and #702 (hook-run scripts).

## Proposed

Extend the existing workflow grant with the action references its steps declare. Preserve the calling job and step identity sufficiently to explain a change without reading or executing the action. Use the existing redacted subject/diff projection and retain one workflow subject; do not add a command or a second verdict.

A SHA-to-branch change names a different declared code reference. It does not establish additional token scopes, that the action ran, or what the fetched code will do. Compute change identity before display redaction, and keep changes visible even when their permission direction is unknown or non-expanding. A redacted display collision must not erase a distinct change (#767).

Start with remote action references. Keep local-action references explicitly outside this increment and owned by #701; do not silently count a local reference as fully inspected or double-count it when that reader lands. Expressions, invalid values and unresolvable references need an explicit supported boundary rather than guessed action identities.

## Acceptance

- [ ] Reproduce the pinned-SHA-to-branch case through the real installed CLI before the repair. Afterward a single changed remote reference produces a row naming its workflow and the declared reference change; an unchanged reference produces no row.
- [ ] Added, removed and changed references retain their calling job/step evidence. A reorder that changes no modeled reference is quiet unless a separately documented execution-order dependency is actually being evaluated.
[...]

PR #786, opened 2026-09-15
Add a host-only advisory PR recipe with no manifest or baseline (#780)

pengfei-threemoonslab:
Refs #780 — the hosted-run acceptance items stay open (see *Not in this PR*).

## Problem

A team that got value from a local `agents-shipgate diff` had no copyable PR workflow for the same job. `examples/github-actions/README.md` said every recipe needs a root `shipgate.yaml`, `docs/integrations.md` opened with `config: shipgate.yaml`, and the only host recipe (`12-host-grant-drift.yml`) is a scheduled baseline gate — a different job from reviewing the current PR against Git history.

## What was checked before writing the recipe

The composite Action already takes the manifest-free route: with no `shipgate.yaml`, `verify --workspace . --config shipgate.yaml --out agents-shipgate-reports …` returns `host_comparison` rows and a host-capability PR comment. On 2026-09-14, `agents-shipgate==1.0.0` from PyPI, run with the Action's exact arguments in a clone of a manifest-free repository, gave:

| PR branch | `comparison_status` | rows | PR comment summary |
| --- | --- | --- | --- |
| widens `Bash(npm test:*)`→`Bash(npm *)`, drops a denial, adds an MCP server | `comparable` | 4 | `Repository-declared host capability changes:` + four entries |
| edits README only | `comparable` | 0 | `Repository-declared host capability changes:` + `No static host-grant changes detected in the covered comparison. No verdict is implied.` |
| base has an unreadable `.cursor/mcp.json`, PR edits README | `comparable` | 0 | the no-change lines plus `Not compared: … .cursor/mcp.json — parse_failed` |
| leaves `.mcp.json` truncated | `incomparable` (`head_inventory_incomplete`) | 0 | `Host capability comparison unavailable: head_inventory_incomplete` |

Every summary carries `Advisory: no application release policy configured. This comparison grants no merge authority.`, followed by an `Evidence:` line (and, when the comparison is unavailable, next-actor/action/command lines); `can_merge_without_human` is false.

While isolating that, a separate defect turned up: the same `verify` with `--out` **outside** the workspace reports `comparison_input_unavailable:ValueError` and no rows. The Action writes inside the workspace, so the recipe is unaffected; it is filed as #785 rather than folded in here.

## Change

[...]

PR #789, opened 2026-09-15
Distinguish discovered Claude hook files from proven host loading (#714)

### LINKED ISSUE #714: Distinguish discovered Claude hook artifacts from proven host loading
## Problem

Discovering `.claude/hooks/hooks.json` establishes that a hook-shaped artifact exists. It does not by itself establish that Claude Code loads that file or that its commands run in the current project. #704 promotes this path, including nested instances, directly to a high-risk `execute` host grant. The advisory 1.0 support page now names this limitation (#775); the reader distinction remains unresolved.

The distinction matters to adoption: a reviewer should not be interrupted as though an inactive example or unreferenced configuration is an active permission expansion.

## Evidence and limits

At #704 head `dea35e55973ba20c99767b6163075a3bbb600e54`, the added fixture creates only `.claude/settings.json` with ordinary permissions, then adds `.claude/hooks/hooks.json`. The test expects an expanding hook row without any loader reference or plugin activation evidence.

The [official hook locations](https://code.claude.com/docs/en/hooks#hook-locations) list project settings, managed settings, plugin `hooks/hooks.json`, and skill/subagent frontmatter. The [plugin reference](https://code.claude.com/docs/en/plugins-reference) defines default hook configuration relative to the plugin root and allows explicit custom `hooks` paths in the plugin manifest.

Read-only inspection of the cited `Archive228/loopkit` repository on 2026-09-12 found `.claude-plugin/plugin.json` with no `hooks` reference; its `.claude/settings.json` declares a separate `PostToolUse` command. This is a current-source observation, not proof about every historical benchmark commit or external installation. No Claude session or hook was executed. The original benchmark's exact historical wiring still needs verification.

## Suggested implementation

Separate three facts the static reader can encounter: a parsed hook artifact, a supported declaration/reference selecting that artifact, and unavailable installation or session activation evidence. The first does not prove the second; neither proves runtime execution. Preserve the hook event and command-change evidence while making the basis of the row understandable.

Start with the settings and plugin-reference forms the repository can read deterministically. Where external installation or enabled state is unavailable, publish a declared-only/activation-unknown result with the known source and missing evidence. Do not ask the user to assert that a host loaded it merely to make the scan complete. Do not discard a declared artifact as harmless because activation is unknown.

Before changing a fixture's expectation, inspect the original pinned historical wiring independently of the engine. Current `loopkit` source cannot establish whether a historical plugin was installed or referenced. Record any corrected oracle provenance without relabeling past measurements as though the engine had always had that evidence.

## Acceptance criteria

- [ ] A standalone, unreferenced `.claude/hooks/hooks.json`, including a nested example, does not claim proven active execution authority solely from its path.
- [ ] Supported explicit settings/plugin references retain the selected artifact, event and command-change evidence. An unavailable external installation or enabled state remains unknown; a static reference is never presented as proof that a hook ran.
- [ ] Declared-only and supported-reference rows make the distinction visible in CLI, maintained JSON and advisory PR output. Unknown activation neither disappears into a clean claim nor becomes a fabricated permission expansion.
- [ ] The historical public fixture has pinned refs and the actual available selection/loading evidence, or its oracle is corrected with a dated reason and preserved prior provenance.
- [ ] Positive and negative controls distinguish an unreferenced artifact, a supported referenced configuration, unavailable activation and malformed references without starting a host or executing hooks.
- [ ] The support page describes the implemented distinction and remaining limit. Existing supported settings-hook changes retain their review/control behavior unless a separately reviewed semantic correction establishes otherwise.

## Sequencing

P1, trust and interruption repair under #778. Design/implementation can proceed alongside the workflow-reference work and #772. #702 consumes this distinction before it extends a hook row to referenced script bytes; it must not turn a declared-only artifact into proven active authority by finding a script beside it.

This is bounded static provenance work. It does not wait for or build a runtime/plugin-installation control plane, does not restore the suspended legacy skill (#690), and does not delay advisory user research (#653/#571).

## Product delivery boundary — 2026-09-16

Selected as the provenance foundation for #811's plugin-related follow-ons (F4). First establish supported **local** settings/manifest references and their plugin-root interpretation. Feed those facts to #812 and #795, so an unfamiliar reviewer can distinguish a discovered artifact from a selected declaration.

This issue continues to own hook loading/selection semantics, not every Claude/Cursor/Codex plugin format. An absent reader for a local MCP/plugin input is reported by #812; adding that reader requires its own bounded supported-format scope. External marketplace fetching, installation and runtime activation discovery remain outside this delivery.

Acceptance of a local reference must preserve three independent facts: artifact discovered, supported declaration selecting it, and runtime activation/execution unknown unless independently established. A missing target must name its declaration and unresolved source. #702 can consume a proven local reference; #808 can use it to establish the affected scope. Do not delay the first supported shell-permission experience on broad plugin coverage.

pengfei-threemoonslab:
Closes #714.

Head: `523aa7fd`. Four commits on `main` `42110535`, rebased onto #788 (#771): `4ebb175b`, the change; `f09ff734`, review cycle 1; `16263d84`, review cycle 2; `523aa7fd`, review cycle 3.

## Why

Parsing `.claude/hooks/hooks.json` proves a hook-shaped file exists. It does not prove Claude Code loads it. Claude Code's [hook locations](https://code.claude.com/docs/en/hooks#hook-locations) list settings layers, managed policy, a plugin's `hooks/hooks.json` when the plugin is enabled, and skill or subagent frontmatter. They do not list a project `.claude/hooks/hooks.json`. #704 turned every such file, nested copies included, into an `execute`/`high` grant and an expansion from its path alone. That interrupted reviewers, and the Stop hook, as though an unselected example were active authority.

## Reproduction on published `agents-shipgate==1.0.0`

Run in a clean venv, outside any checkout. Each repo has a base commit with only `.claude/settings.json` (`Read(**)`) and a change adding the loopkit-shaped `SessionStart` hook file. No settings file, plugin manifest or marketplace entry references it.

**Standalone `.claude/hooks/hooks.json`**, `diff --base main`:
```
⚠ high  added  claude-code .claude/hooks/hooks.json
              SessionStart
              changes what runs around the agent's actions

1 change(s), 1 widening what the agent may do (⚠).
```
- `diff --json` row: `"expands": true, "severity": "high"`.
- `audit --host --json` grant: `"access": "execute", "risk": "high"`.
- `check --format agent-boundary-json`: `require_review` via `BOUNDARY-PROTECTED-SURFACE-UNCLASSIFIED`, and an expanding `high` row.

**Nested `examples/demo/.claude/hooks/hooks.json`**: the identical expanding `high` row.

On this branch the same inputs give an `unknown`-severity, non-expanding row whose `why` says no settings file, plugin manifest or marketplace entry selects the file.

## Design

No schema, contract or `minimum_control_contract_version` change: this PR sits on contract 40 and host-grants `0.6` from #771 and moves neither. Hook grants keep their identity (`grant_id` and `config_sha256` don't depend on the basis). Each carries a **loading basis**, which one engine helper, `host_grants.hook_loading_basis`, reads back from the published `source`, `access` and `risk`.

| Basis | When | `access` / `risk` | Expansion signal | Row `why` |
| `host_configuration` | Claude Code settings layers; Codex `.codex/hooks.json` | `execute` / `high` (unchanged) | yes (unchanged) | unchanged |
| `project_enabled_plugin` | A plugin selects it, and the repository's project settings enable that plugin from an in-repository marketplace (see below) | `execute` / `high` (as in `1.0.0`) | yes (as in `1.0.0`) | "changes what runs around the agent's actions; this repository's project settings enable the plugin that selects this hook" |
| `plugin_selected` | A plugin in the repository selects it, without that enablement | `execute` / `medium` | **no** | "changes a hook a plugin in this repository selects; whether that plugin is installed or enabled is not established" |
| `declared_only` | A Claude Code hook file nothing selects | `unknown` / `unknown` | **no** | "declares a hook that no settings file, plugin manifest or marketplace entry in this repository selects; whether a host loads it is not established" |
| `unestablished` | A plugin-reference hook grant with none of those pairs; never produced by this reader | as recorded | no | claims no selection |

**Basis from exact pairs.** A Claude Code hook whose source is a hook file, a plugin manifest or a marketplace entry is classified only by those exact `access`/`risk` pairs. Nothing is inferred from a name plus `execute`. A plugin hook's `source` is never a settings layer, which separates `project_enabled_plugin` from `host_configuration` although they share a pair. Every removal outside settings reads "removes a hook declared in this file; whether a host loaded it is not established", because a removal is described from the baseline's grant.

**One pair is shared with history, deliberately.** `1.0.0` recorded every hook file as `execute`/`high`, which is now the `project_enabled_plugin` pair. No published field is free to tell them apart without a schema change. The engine reads a basis only from the current side of a change (the wording of an added or changed row, and the expansion signal), and a removal names no basis, so nothing a `1.0.0` baseline holds is ever described as selected or enabled. `test_a_removal_against_a_1_0_0_baseline_claims_no_selection` pins this, and `test_a_1_0_0_baseline_of_an_enabled_plugin_hook_drifts_once_with_no_grant_change` shows the enabled case re-reading with the same pair and the same `grant_id` against a baseline published `1.0.0` actually saved.

### Enablement the repository proves (review cycle 2, P1)

Checked against the settings reference (`enabledPlugins`, `extraKnownMarketplaces`), [discover plugins](https://code.claude.com/docs/en/discover-plugins) and [plugin marketplaces](https://code.claude.com/docs/en/plugin-marketplaces) on 2026-09-15:
- Only a plugin from an **external** source (a GitHub repository or an npm package, for example) that the project's settings enable waits for each user to install it. Marketplaces the project settings register are added once the folder is trusted.
- `extraKnownMarketplaces` sources are `github`, `git`, `url`, `file` (a path to a `marketplace.json`), `directory` (a local path, documented as development-only) and `settings` (an inline marketplace whose plugins must use external sources).
- A relative `directory` or `file` path resolves against the repository's main checkout.
- `enabledPlugins` keys are `<plugin>@<marketplace>`, and the plugin name is the marketplace entry's `name`. It may be set in user, project, local or managed settings; a local `false` is how one user opts out of a project-enabled plugin.

So a hook is `project_enabled_plugin` when all of these hold:
1. `.claude/settings.json` or `.claude/settings.local.json` sets `enabledPlugins` `<plugin>@<marketplace>` to exactly `true`;
2. either file registers `<marketplace>` in `extraKnownMarketplaces` (or its `additionalMarketplaces` alias) as a `directory` source whose relative `path` holds a `.claude-plugin/marketplace.json`, or a `file` source whose relative `path` is one;
3. that marketplace lists `<plugin>` with a `./` source or a bare name under `metadata.pluginRoot`. `<marketplace>` matches the `extraKnownMarketplaces` key or the registered `marketplace.json`'s own `name` (review cycle 3, nit 1);
4. the plugin selects the hook, by any route below.

Everything else establishes nothing and stays `plugin_selected`: `github`, `git`, `url` and `settings` sources (a GitHub source naming this very repository cannot be proved to be this checkout), absolute, `~`, drive, backslash or escaping paths, an unlisted plugin, and any value other than `true`.

Decisions:
- **`.claude/settings.local.json` counts.** It is usually uncommitted, but Claude Code reads it whenever it exists, so reading it errs toward showing a loaded hook. When it is absent from the compared tree it contributes nothing.
- **A `true` in either project file counts**, even when the other says `false`. A local `false` is one machine's opt-out; it does not change what the committed file enables for everyone else. (Two disagreeing layers already raise the existing `unresolved_precedence` issue for the `plugin_or_app` grant.)
- **Installation state, user settings and workspace trust are never read.**

A/B, one repository that is its own marketplace (`.claude-plugin/marketplace.json` lists `demo` with `"source": "./"`, `.claude-plugin/plugin.json` has `"hooks": "./.claude/hooks/hooks.json"`, `.claude/settings.json` registers `market` as `{"source": "directory", "path": "."}`), where the head changes the hook command:

| Settings | Published `1.0.0` | `b989b5e3` | This branch |
| --- | --- | --- | --- |
| `demo@market: true` | `⚠ high widened`, "changes what runs around the agent's actions" | `medium changed`, no ⚠ | `⚠ high widened`, "changes what runs around the agent's actions; this repository's project settings enable the plugin that selects this hook" |
| `demo@market: false` | `⚠ high widened` | `medium changed` | `medium changed` |
| `market` as a `github` source | `⚠ high widened` | `medium changed` | `medium changed` |
| `ghost@market: true` (not listed) | `⚠ high widened` | `medium changed` | `medium changed` (the manifest still selects the file) |

For the enabled case, the other consumers match `1.0.0` too:
- `verify` writes the identical PR comment line, `` - ` high ` / ` widened ` — ` claude-code .claude/hooks/hooks.json ` ``, and a `widened`/`expands: true`/`high` row in `verifier.json`.
- `check` gives `require_review` in all three formats, with the row first in `capability_rows` as `widened`/`expands: true`.
- The Stop hook announces it ("These rows widen what the agent can do").

The other three rows differ from `1.0.0` by design: `1.0.0` showed every hook file as `high`.

**What selects a hook.** Per the [plugins reference](https://code.claude.com/docs/en/plugins-reference) and [marketplace reference](https://code.claude.com/docs/en/plugin-marketplaces):
- **Plugin roots** are recognised by `<root>/.claude-plugin/plugin.json`, or by a `.claude-plugin/marketplace.json` entry whose `source` is a `./` path or a bare name (one directory name, no `/`) under `metadata.pluginRoot`. A remote `source` names nothing in the repository.
- **A plugin selects** `<root>/hooks/hooks.json`, and adds a `./` path or array from `hooks` in its manifest or marketplace entry.
- **Inline `hooks` objects** in either place are read too, including `strict: false` entries. An inline marketplace hook's source is `<marketplace>#plugins.<name>`.
- **Followed names:** a reference is followed only to `hooks.json`, or to a name ending in `-hooks.json`, `_hooks.json` or `.hooks.json`. `webhooks.json` is not followed.
- **Case:** a reference matches a file case-insensitively when exactly one file matches, erring toward showing a hook that a case-insensitive filesystem would load.

**How unknown activation is shown.** Without the enablement above, a static reference is never presented as loaded. A plugin-selected hook keeps its event and every command change, and its row names the missing evidence. An unselected file stays visible and is never discarded. Neither earns `hook_added`/`hook_changed`, so neither is marked ⚠ in `diff`, announced by the Stop hook, or sorted first in the envelope's capability rows.

**Comparison on both sides.** `diff` and `verify` build the base tree from paths alone (#686). So plugin manifests, marketplaces and hook-named files are added to `is_boundary_surface_path` for materialization only. They do not join the adapter registry: `check`'s protected-surface route, the triggers, `docs/triggers.json`, the pre-commit regex and the detector projection are unchanged. A manifest or marketplace enters the inventory only when it declares `hooks`, and only that part is digested, so a version bump is not drift.

**Limits.**

| Problem | Where | In `audit --host` / `diff` / `verify` |
| --- | --- | --- |
| Unparsable or non-object manifest; wrong-type `hooks`; a reference that is not a `./` path inside the plugin; a non-hook-named target | plugin manifest | Blocking. `diff`/`verify` name a parse or shape limit (`parse_failed`, `unsupported`) both sides share on an unchanged file, and refuse otherwise. The issue names the exact target file where there is one. |
| A read limit of a hook file only a plugin selects (for example over 1 MiB) | plugin manifest | Blocking, and never an unchanged limit: an untouched oversize plugin hook file makes `diff`/`verify` incomparable even on a README-only change, as an oversize settings file already does in `1.0.0`. |
| The same problems; a `metadata.pluginRoot` that is not a `./` path inside the marketplace (`./..`, `/plugins`) | marketplace entry | not blocking, named |
| Reference to a missing file | either | not blocking, named |
| Reference beneath a link that leaves the workspace | either | only the existing blocking limit on that link; no second "not in the repository" message |
| Reference into a walk-skipped directory (`node_modules`, `.venv`, …) | either | not blocking; named as "a directory the static reader does not walk; its hooks were not read", not as missing |
| Selected file with no `hooks` object | either | not blocking, named |

### `check` and plugin-reference limits (review cycle 1 P1, review cycle 2 P2)

`check`'s boundary result cannot name a limit, and it routes no plugin manifest, marketplace or plugin-selected hook file. The reader records plugin-reference issue ids on the snapshot.
- **Decision.** `agent_boundary` leaves those issues out of `check`'s input completeness, so a plugin-reference limit never changes a `check` decision from what `1.0.0` gave.
- **Host comparison.** `check` calls `compare_host_refs(..., exclude_plugin_reference_limits=True)`. That now drops only a blocking plugin-reference limit **both sides share on an untouched source**, plus non-blocking plugin-reference issues. A limit only one side carries, or one on a source the change touched, stays, so the existing comparator refuses with `base_inventory_incomplete` or `head_inventory_incomplete`. No new reason string and no new surface: this is the engine's existing refusal.

**Decision for a head that breaks a plugin manifest: `allow`, `control_state: complete`, and an incomparable host comparison with no rows.**
- The decision matches `1.0.0`, which never read plugin files, and `check` still routes no plugin file.
- The comparison no longer builds a "removed" row from a manifest it could not read.
- This is consistent with a head-broken `.claude/settings.json`, which stays `require_review`/`human_review_required` exactly as in `1.0.0`, because `check` routes that file.
- It is not fail-open relative to `1.0.0`: `1.0.0` gave `allow` with no rows, and a hook edit under that plugin was equally invisible to its `check`.

Setup: base `plugins/demo/.claude-plugin/plugin.json` → `./cfg/hooks.json`, head replaces the manifest with `{not json`:

| Head | Published `1.0.0` | `b989b5e3` | This branch |
| --- | --- | --- | --- |
| manifest broken | `check`: `allow`, `complete`, `comparable`, no rows | `check`: `allow`, `complete`, `comparable`, `medium / removed — plugins/demo/cfg/hooks.json` | `check` (text, agent-boundary-json, agent-control-json): `allow`, `complete`, `incomparable` (`head_inventory_incomplete`), no rows; `diff` and `verify`: `incomparable` (`head_inventory_incomplete`) |
| manifest broken and hook command edited | same as above | same "removed" row | same as above |
| `.claude/settings.json` broken | `require_review`, `human_review_required`, `BOUNDARY-INPUT-INCOMPLETE` + `HOST-CONFIG-PARSE-FAILED`, `incomparable` | same | same |

Cycle 1's controls still hold, for an untouched `examples/bad/.claude-plugin/plugin.json` containing `{not json`, in all three formats:
- **README-only head:** `allow`, `complete`, `comparable`.
- **Deny rule removed:** `require_review` with a comparable `HOST-PERMISSION-DENY-REMOVED` row.

A changed `.claude/hooks/hooks.json` still routes to protected-surface review whatever its basis; only its rows change.

**A hook file is no longer read as a settings file.** A `permissions`/`enabledPlugins`/`sandbox` key inside a hook file published grants no host grants: a wildcard allow from a path alone, the same bug class as this issue. A hook file now contributes hooks only. Settings files are untouched.

**Baseline effects** (see the STABILITY.md migration note), corrected in review cycle 3 against a baseline published `1.0.0` actually saved:
- A saved baseline that recorded an unselected hook file as `execute`/`high` stays comparable. It reports one `changed` row per event, with `expands: false` and no expansion signal; `--fail-on-drift` exits 20 once until the baseline is re-saved.
- **Every repository whose plugin manifest or marketplace declares `hooks` drifts once**, because that file is a new artifact and a new observed source. Two shapes:
  - A hook file `1.0.0` never read — one a plugin selects outside the Claude Code registry paths, such as `plugins/demo/hooks/hooks.json` — is also a new grant, and carries `hook_added` only where the project settings enable the plugin, because that hook was loaded all along.
  - A hook file `1.0.0` did read, such as a `.claude/hooks/hooks.json` an enabled plugin selects, keeps `1.0.0`'s `execute`/`high` and the same `grant_id`. **The grant does not change, and the gate still exits 20**, reporting `Drift detected — 0 typed grant change(s)` with no rows and no expansion signal. Read `artifact_changes` and `coverage_changes` in `audit --host --drift --json`; the Markdown summary counts typed grant changes only.
- In both shapes, re-save from a checkout of the reviewed default branch.
- A baseline holding permissions read from a hook file reports them removed.
- Separately, #771 makes a `0.4`/`0.5` baseline holding a workflow grant incomparable. `docs/baseline.md` and `examples/github-actions/12-host-grant-drift.yml` say both.

Surface discipline: no new CLI, schema, report block, reason string or adapter. The headline metric moved is noise rate: unselected hook files and untouched plugin limits no longer produce expansion rows, Stop-hook interruptions or `check` reviews. Enabled in-repository plugins keep the signal `1.0.0` gave them.

## Historical fixture (#704): oracle corrected with dated provenance

The #704 fixture (`tests/test_claude_hooks_source.py`) creates only `.claude/settings.json`, then adds `.claude/hooks/hooks.json`. At #704 head `dea35e55973ba20c99767b6163075a3bbb600e54` it asserted `("execute", "high")` and `expands`.

I inspected loopkit's pinned wiring on 2026-09-14 through the GitHub contents and trees APIs, independently of the engine, and ran no Claude session or hook. The cycle-1 and cycle-2 reviewers checked it independently. At both steps that touched the file, `b3e551474bbf7311a54d0f8722817c131d092968` (added) and `5ae033e63e698bc7cedacdea1483d8af677f52e1` (added `PreCompact`):
- `.claude-plugin/plugin.json` has no `hooks`;
- `.claude-plugin/marketplace.json` lists `loopkit` with `"source": "./"` and no `hooks` or `strict`;
- `hooks/hooks.json` does not exist at that root (only `hooks/pre-commit`);
- `.claude/settings.json` declares its own `PostToolUse` command and does not name the file.

The oracle is corrected to declared-only. The refs and evidence are pinned as `LOOPKIT_HOOK_STEPS` / `LOOPKIT_SELECTION_EVIDENCE`, the docstrings keep #704's prior assertion and commit, and a loopkit-shaped marketplace fixture reproduces the result. No benchmark results file was edited, and past measurements are not relabeled.

## Acceptance

- [x] A standalone or nested unreferenced `.claude/hooks/hooks.json` claims no execution authority from its path: `unknown`/`unknown`, no expansion signal, and a row stating the basis.
- [x] Supported settings, plugin-manifest and marketplace references keep the selected artifact, event and command-change evidence. Plugin installation or enablement is stated as unknown unless the repository's own project settings enable the plugin from an in-repository marketplace, in which case the hook keeps `1.0.0`'s `execute`/`high` and expansion.
- [x] The distinction is visible in:
  - `diff` (text and `--json`);
  - `audit --host --json`;
  - `check` rows in all three formats;
  - the Stop-hook advisory (quiet on a declared-only row, announcing an enabled plugin's row);
  - PR comments (`why` via `report/host_comparison.py`);
  - the envelope's capability rows.

  Unknown activation is neither dropped nor turned into an expansion, and plugin limits leave the `check` decision as 1.0.0 had it, with no row built from an unread manifest.
- [x] Historical fixture: loopkit refs and selection evidence pinned; the oracle is corrected with a dated reason and #704's provenance preserved.
- [x] Positive and negative controls. Every hook command is a `touch` marker asserted never to exist, and no host is started.
  - **Unreferenced files:** root and nested.
  - **Host-loaded hooks:** a settings reference and a Codex hook file.
  - **Plugin manifests:**
    - default hooks, at the repo root and in a nested plugin;
    - a custom path given as a string or an array, and a custom path added to the default;
    - inline hooks;
    - a manifest that starts selecting an already-declared file;
    - a version bump that is not drift.
  - **Marketplaces:**
    - an entry `hooks` path;
    - `strict: false` inline hooks;
    - manifest-less default hooks via a `./` source and via `pluginRoot`;
    - a remote source that selects nothing;
    - the loopkit shape;
    - a source with a `/` that is not a bare name, and three unusable `pluginRoot` values named without blocking.
  - **Project enablement (cycle 2):**
    - enabled through `.claude/settings.json` or `.claude/settings.local.json`;
    - `directory` and `file` sources, with and without `./`;
    - eleven non-establishing variants: `false`, a non-boolean, an unknown marketplace, `github`, `git`, `url`, absolute, escaping, home-relative, a non-marketplace `file`, and a `settings` source;
    - an unlisted plugin selecting nothing, and only the enabled plugin of two being loaded, inline hooks included;
    - a baseline published `1.0.0` saved, checked in as a fixture: its shape, the one-time drift with no grant change, and the gate exiting 20 through the CLI;
    - either name of a registered marketplace identifying it, and a `name` alone registering nothing (not registered, registered remotely, non-string `name`);
    - `diff` text and JSON, `verify` rows and PR comment, `check` in three formats, and the Stop hook announcing the real engine row.
  - **Malformed and unread references:**
    - enablement without an in-repository marketplace not read as activation;
    - nine malformed-manifest shapes (blocking);
    - five malformed-marketplace shapes (non-blocking);
    - a missing target;
    - a reference beneath a link leaving the workspace, keeping only the link's limit;
    - `node_modules` and `.venv` references;
    - a selected file without `hooks`;
    - the `webhooks.json` name boundary.
  - **Baselines:** a 1.0.0-shaped baseline as a changed row and as a neutral removal, and each basis from its exact pair.
  - **Comparison and consumers:**
    - base-tree plugin and marketplace hooks compared on both sides;
    - `check` cycle-1 controls in three formats;
    - `check` cycle-2 controls: a head-broken manifest, with and without a hook edit, in three formats; a base-only limit; a limit changed on both sides;
    - `diff` naming an unchanged malformed manifest and refusing a newly malformed one;
    - the Stop hook.
- [x] The support page describes the four bases, the enablement rules, marketplace selection, the neutral removal, subagent-frontmatter hooks as unread, the remaining limits including read limits, and the final `check` behaviour. Its pinning test is updated.

## Remaining limits (recorded on the support page)

- Installation state, user settings and workspace trust are never read; enablement is read only from project settings and only for an in-repository marketplace.
- Hook-named files only: any other referenced name is a limit.
- `check` still routes an unselected `.claude/hooks/hooks.json` to protected-surface review. It does not route plugin manifests, marketplaces or plugin-selected hook files.
- Subagent frontmatter hooks (`.claude/agents/*.md`) are unread. A skill's `hooks` frontmatter is type-checked with the skill, but never read as a hook grant.
- #702 (script bytes) must consume the basis, and must not promote a declared-only or `unestablished` file.
- An enabled plugin's hook at a non-registry path (`plugins/demo/cfg/hooks.json`) gets an expanding `high` row under a `check` `allow`, because `check` routes no plugin-selected hook file. `1.0.0` allowed the same change with no row. Follow-up #809.
- A blocking plugin-reference limit only one side carries refuses the whole host comparison, so an unrelated row is withheld too. Every `check` decision still equals `1.0.0`. Follow-up #808.
- Pre-existing and out of scope: `check` allows a hook added in `.claude/settings.local.json`, and crashes with a `ValidationError` when settings only add `enabledPlugins` (#810, reproduced on `1.0.0`, `main` and this branch).

## Review cycle 3

**P2: the "does not drift" claim was wrong.** Reproduced against a baseline that published
`agents-shipgate==1.0.0` saved for a repository that is its own marketplace
(`.claude-plugin/plugin.json` names `./.claude/hooks/hooks.json`; `.claude/settings.json`
registers a `directory` marketplace at `.` and sets `demo@market: true`):

| `audit --host --drift --fail-on-drift` | Exit | Output |
| --- | --- | --- |
| Published `1.0.0` | 0 | `No drift — current host grants match the acknowledged baseline.` |
| `main` `42110535` | 0 | same |
| `16263d84` (previous head) | **20** | `**Drift detected** — 0 typed grant change(s).` |

`--json` on the branch: `comparison_status: comparable`, `has_drift: true`, `changes: []`,
`expansion_signals: []`, `artifact_changes` adding `.claude-plugin/plugin.json`, and
`coverage_changes` adding it to claude-code `sources_observed`. The hook grant is unchanged.

Fixed:
- STABILITY.md, `docs/baseline.md`, `examples/github-actions/12-host-grant-drift.yml`, the
  CHANGELOG and the "Baseline effects" section above now describe the one-time exit 20 with
  zero typed grant changes, and point at `artifact_changes`/`coverage_changes`.
- `test_a_1_0_0_baseline_of_an_enabled_plugin_hook_does_not_drift` is replaced.
  `_recorded_by_1_0_0` rebuilt its "1.0.0" baseline from this reader's own inventory, so it
  held artifacts `1.0.0` never read and could never fail; it now refuses an inventory holding
  a plugin manifest or marketplace. The enabled case uses
  `tests/fixtures/host-grants-1.0.0-enabled-plugin.json`, saved by the published `1.0.0`
  venv, with the reproduction commands recorded beside it. Three tests replace the one that
  could not fail: what makes the fixture a `1.0.0` artifact, the drift payload (comparable,
  `has_drift: true`, no typed change, no signal, the named artifact and coverage change, and
  the hook grant identical on both sides), and the gate exiting 20 through the CLI.
- The helper's two other callers build workspaces with no plugin file, where the synthesis is
  faithful; they are unchanged.

**Nit 1: `enabledPlugins` matched a marketplace only by its `extraKnownMarketplaces` key.**
Claude Code's marketplace schema calls the `name` in `marketplace.json` the identifier users
see after the `@` ([plugin marketplaces](https://code.claude.com/docs/en/plugin-marketplaces)),
and the [settings reference](https://code.claude.com/docs/en/settings) only ever shows a key
equal to that `name`, so neither is documented as the one the host matches. Both are now
accepted, on the same principle as case-insensitive reference matching: err toward showing a
hook. A `name` alone still registers nothing.

With key `market`, `marketplace.json` `name` `real-market` and `demo@real-market: true`, on a
head that changes the hook command:

| `diff --base main` | Row |
| --- | --- |
| Published `1.0.0` | `⚠ high widened claude-code .claude/hooks/hooks.json` |
| `16263d84` | `medium changed`, no ⚠ |
| `523aa7fd` | `⚠ high widened`, "…; this repository's project settings enable the plugin that selects this hook" |

**Follow-ups filed:** #808 (a newly broken plugin manifest withholds unrelated rows), #809 (an
enabled plugin's hook at a non-registry path gets an expanding row under `allow`) and #810
(pre-existing: `check` crashes with a `ValidationError` when settings only add
`enabledPlugins`; reproduced on `1.0.0`, `main` and this branch). #808 and #809 are recorded on
the support page.

**Rebase:** none needed. `origin/main` is still `42110535`, the merge base of this branch.

## Rebase onto #771 (PR #788)

#788 merged first. This branch is rebased onto `42110535`, and each of the three commits was resolved by hand:
- **`capability_diff_rows.py`:** the import line now carries both `hook_loading_basis` and `step_action_key`, and #771's `_why` signature and step-action code sit beside the hook branch.
- **`host_grants.py`:** merged cleanly.
- **STABILITY.md:** the intro paragraph now reads "still contract v40", beside #771's v40 paragraph. The #714 migration note sits below #771's.
- **`docs/distribution-surfaces.md`:** the `capability_diff` row keeps #771's step-reference sentence and this PR's basis sentences.
- **Support page:** keeps #771's step-reference paragraph and replaces the pre-#714 hook paragraph.
- **CHANGELOG:** `## Unreleased` holds main's seven bullets once and this PR's #714 bullet once.
- **Stale wording:** the drift notes that said #771 "may add a separate one-time drift" now state its incomparable-baseline rule.

## Tests (on `523aa7fd`)

- `python -m ruff check .` over the whole repository: **exit 0**.
- `tests/test_adapter_static_only.py`, run alone as CI does: **exit 0**.
- `test_claude_hook_loading_evidence.py`, `test_claude_hooks_source.py`,
  `test_host_boundary_unread_surfaces.py`, `test_install_hooks.py`,
  `test_distribution_surface_parity.py`, `test_workflow_step_action_references.py`:
  **exit 0**.
- The 20 host audit, check and comparison files (`test_host_*.py`, `test_capability_diff.py`,
  `test_manifest_free_pr_rows.py`, `test_scoped_base_tree.py`,
  `test_unchanged_host_limits.py`): **exit 0**.
- **Full suite** (`-n auto -m "not perf" --ignore=tests/test_adapter_static_only.py`) in ten
  foreground chunks of literal paths, together covering all 307 top-level `tests/test_*.py`
  files except the static-only one, plus `tests/harness` and `tests/integration`.

  | Test files | Exit | Passed | Skipped |
  | --- | --- | --- | --- |
  | `a*` | 0 | 1,219 | 3 |
  | `b*`, `c*` | 0 | 1,373 | 0 |
  | `d*`–`g*` | 0 | 1,964 | 1 |
  | `h*` | 0 | 608 | 0 |
  | `i*`–`l*` | 0 | 454 | 0 |
  | `m*`–`o*` | 0 | 802 | 2 |
  | `p*` | 0 | 1,192 | 0 |
  | `q*`, `r*` | 0 | 812 | 0 |
  | `s*`, `t*` | 0 | 1,075 | 0 |
  | `u*`–`z*` + harness + integration | 0 | 1,497 | 1 |
  | **Total** | all 0 | **10,996** | **7** |

  No failures and no errors. Counts come from the progress output, because this suite prints
  no summary line.
- The published-`1.0.0` comparisons in this description were run against an
  `agents-shipgate==1.0.0` venv in scratch repositories on 2026-09-15, at this head for the
  cycle-3 rows.


pengfei-threemoonslab:
Local full suite result on `545de467` (the PR head), as promised in the description:

- Command: `pytest -q -p no:cacheprovider -n auto -m "not perf" --ignore=tests/test_adapter_static_only.py`, with `PYTHONPATH=<worktree>/src` and the main checkout venv.
- **Exit code 0.**
- 10,742 results, counted from the progress output (this suite prints no summary line): **10,735 passed, 7 skipped, 0 failed, 0 errors.**
- With `tests/test_adapter_static_only.py` (exit 0) and `ruff check .` (exit 0), both CI-style checks pass locally on this commit.

Not run against the current `main`. The branch is behind `67a3a5e7` (#783, #784), and the trial merge showed no conflicted paths; required CI covers the merged state.

pengfei-threemoonslab (commented):
Independent coding-agent review/address cycle 1, reviewed at `545de467`.

Verified: 30 touched/related test files plus 18 benchmark/corpus/envelope files exit 0; CI green; the author's full suite passed. Published `1.0.0` was compared with this branch in scratch repositories.

- **Loopkit provenance** — independently checked at `b3e55147`/`5ae033e6`: the marketplace entry has no `hooks` or `strict`, `plugin.json` has no `hooks`, and `hooks/` holds only `pre-commit`.
- **Hook file no longer read as settings** — correct.
- **Genuinely loaded hooks** — none turned `unknown`.
- **Registry, triggers and `check` routing** — unchanged.
- **Interaction with #788** — textual only; both PRs cause a one-time baseline drift, which the migration notes should say.

**P1**
1. **New blocking plugin-manifest limits flip `check` to `require_review` even when nothing about the plugin changed.**
   - The issues at `host_grants.py:1435/1470/1489/1499` are `blocking=True`, and `check` refuses any unchanged limit (`check.py:491`, #721).
   - With an untouched malformed `examples/bad/.claude-plugin/plugin.json`, a README-only head gives `1.0.0: allow` and `branch: require_review` (`BOUNDARY-INPUT-INCOMPLETE`).
   - A deny-rule change loses its row, and `capability_rows` becomes `incomparable`.
   - This contradicts STABILITY.md:128 ("`check` decisions are unchanged").
   - No malformed-reference control runs `check`.

**P2**
1. **`hook_loading_basis` infers `plugin_manifest` from `access: execute` plus a `*hooks.json` name**, so a removal row against a `1.0.0` baseline says a plugin manifest selects `.claude/hooks/hooks.json`.
2. **Marketplace entries select hooks, but the reader ignores them.** A `.claude-plugin/marketplace.json` entry with a `hooks` path, or `strict: false` inline hooks, reads as "selected by nothing"; so does a manifest-less plugin's auto-discovered `hooks/hooks.json`. That would permanently block #702 for truly selected hooks.
3. **A reference into a walk-skipped directory** (`.venv`, `node_modules`) is reported as missing rather than not read.
4. **Scheduled drift users cannot see why they drifted.** `--fail-on-drift` exits 20 with `expansion_signals: []`, and the Markdown shows only a count. `examples/github-actions/12-host-grant-drift.yml` and `docs/baseline.md` do not mention the one-time change.

**Nits**
- Case-insensitive reference matching selects files Linux would not load.
- `is_hook_declaration_file_name` matches `webhooks.json`.
- A selected file with wrong-shape `hooks` yields nothing and no issue.
- Skill/subagent frontmatter hooks are neither read nor listed as unread.
- "External enablement means no expansion" applies only to hooks, not `enabledPlugins` or `.mcp.json` additions (owner's call).

pengfei-threemoonslab:
**Addressed review cycle 1.** New head `b989b5e36f346727e13f9cba2a233db00ec3d237`, rebased onto `3131c9c3`. There is still no schema or contract change.

**P1: `check` refused when a plugin manifest nobody touched was broken**

The reader now tags issues that come from plugin references. `check` leaves those out of its boundary completeness and its host comparison. `diff` and `verify` are unchanged: `diff` still lists an untouched malformed manifest under `unchanged_limits`, and still refuses a manifest the change itself broke (`head_inventory_incomplete`).

With an untouched `examples/bad/.claude-plugin/plugin.json` containing `{not json`, this branch matches published 1.0.0 in text, `agent-boundary-json` and `agent-control-json`:
- **README-only head:** `allow`, `complete`, `comparable`. The previous head gave `require_review` (`BOUNDARY-INPUT-INCOMPLETE`).
- **Deny rule removed:** `require_review` with a `HOST-PERMISSION-DENY-REMOVED` row that is present and comparable.

Six `check` tests cover this. Each first asserts that the manifest issue is a blocking plugin-reference issue. STABILITY.md and the support page now describe the final `check` behaviour.

**P2**
- **P2-1:**
  - Plugin-selected hooks are published as `execute`/`medium`, and the basis is read only from exact `access`/`risk` pairs.
  - A 1.0.0-shaped `execute`/`high` hook grant is `unestablished` and claims no selection.
  - Non-settings removal rows now read "removes a hook declared in this file; whether a host loaded it is not established".
- **P2-2:**
  - Marketplace entries go through the same resolver: a `./` source or a bare name under `metadata.pluginRoot`; entry `hooks`; inline hooks including `strict: false`; and the default `hooks/hooks.json` of a plugin without a manifest.
  - Remote sources select nothing in the repository, and marketplace issues never block.
  - `LOOPKIT_SELECTION_EVIDENCE` records that loopkit's marketplace entry has no `hooks` or `strict`.
- **P2-3:** a reference into a walk-skipped directory (`node_modules`, `.venv`, …) says the directory is not walked and its hooks were not read. It does not block.
- **P2-4:** `examples/github-actions/12-host-grant-drift.yml` and `docs/baseline.md` explain the one-time `changed` hook rows after upgrading, and the possible separate drift from #771 (PR #788).

**Nits**
- `hooks.json` now matches only at a name boundary, so `webhooks.json` does not.
- A selected file with no `hooks` object gets a named, non-blocking issue.
- Subagent frontmatter hooks are listed under Known unread surfaces.
- Case-insensitive matching is kept and documented, because it errs toward showing a hook.
- The enablement-principle nit is left to the owner.

**Tests on `b989b5e3`**
| Run | Result |
|---|---|
| 21 related test files (including #786's recipe and Action install tests) | exit 0 |
| `ruff check .` | exit 0 |
| `tests/test_adapter_static_only.py` | exit 0 |
| Full suite (`-n auto -m "not perf"`), ten foreground chunks covering every test | each exit 0 (10,837 passed, 7 skipped) |

pengfei-threemoonslab (commented):
Independent coding-agent review/address cycle 2 at `b989b5e3`. **Not mergeable yet: one P1 and one P2.** No P0. The cycle-1 fixes hold, and CI is green (11 pass, 1 skipping).

**P1: a plugin hook the repository itself enables lost its 1.0.0 widening signal** (`host_grants.py:911` sets the `medium` pair; `:2892` builds expansion signals).

Setup — the repository is its own marketplace:
- `.claude-plugin/marketplace.json` lists `demo` with `"source": "./"`;
- `.claude-plugin/plugin.json` has `"hooks": "./.claude/hooks/hooks.json"`;
- `.claude/settings.json` sets `extraKnownMarketplaces` (directory source) and `enabledPlugins: {"demo@market": true}`;
- the head changes the hook command.

Result:
- **1.0.0 `diff`:** `⚠ high widened … changes what runs around the agent's actions`.
- **Branch `diff`:** `medium changed … whether that plugin is installed or enabled is not established`, with no ⚠.
- The Stop hook announces only expanding rows (`install_hooks.py:1279`), and the envelope sorts expanding rows first (`agent_control_envelope.py:342`), so both lose this change.
- `check` still gives `require_review` through the protected-surface route.

The Claude Code docs say only plugins from *external* sources wait for a manual install, so this in-repo plugin, enabled by the project's own settings, can load. Nothing else keys on `high`. Without enablement evidence, `medium` with no expansion is justified and documented.

Fix: when committed project settings enable `<plugin>@<marketplace>`, and that marketplace resolves to an in-repo `marketplace.json` listing the plugin with an in-repo source, publish its selected hooks as `execute`/`high` with expansion. Otherwise, record a Known parity gap with an owning issue.

**P2: `check` shows a comparable "removed" row when the head breaks a plugin manifest** (`cli/verify/host_comparison.py:122` drops the head-only limit).

With base `plugins/demo/.claude-plugin/plugin.json` → `./cfg/hooks.json`:
- the head replaces the manifest with `{not json` (and, in a second case, also edits the hook command);
- the branch's `check` gives `allow`, `complete`, `comparable`, plus `medium / removed — plugins/demo/cfg/hooks.json`, even where the command actually changed;
- 1.0.0 gives `allow` with no rows, and `diff` correctly refuses.

The decision is unchanged, but a comparable row built from missing evidence contradicts the `capability_diff` registry row.

Fix: exclude only plugin-reference limits present on **both** sides. A limit on one side only makes the comparison incomparable, with a reason. Add a `check` test where the head breaks the manifest and edits the hook.

**Nits**
- `host_grants.py:1645` treats a source containing `/` as a bare name. A `pluginRoot` of `./..` silently selects nothing, with no issue.
- An untouched plugin hook file over 1 MiB makes `diff`/`verify` incomparable on a README-only change. That matches 1.0.0 for an oversize settings file, but the PR body and STABILITY.md wording "refuse, or name the limit when it is unchanged" is not true for read limits.
- A reference through a symlink that leaves the workspace also gets a non-blocking "not in the repository" message. The blocking link issue is still present.
- Pre-existing, not this PR: `check` allows a hook added in `.claude/settings.local.json`.

**Verified with no finding**
- **Cycle-1 P1:** all three `check` formats match 1.0.0 for a malformed or unsupported manifest, a malformed marketplace, a `node_modules` reference, and wrong-shape hook files, with both README-only and deny-rule heads.
- **Nothing hidden:** a broken `.claude/settings.json` and a settings hook change behave the same as 1.0.0, and plugin hook edits are never quieter.
- **1.0.0-saved baselines:** neutral declared-only wording, and no plugin claim on removal.
- **Marketplace resolution:** matches the docs. `../`, absolute and backslash references are refused, and nothing outside the workspace is read.
- **Fail-open hunt:** settings and `settings.local.json` inline hooks stay high and expanding; a root manifest without `hooks` still selects `hooks/hooks.json`; Codex hooks are unchanged.
- **Rebase and provenance:** CHANGELOG and the parity test are consistent. Loopkit at `b3e55147`/`5ae033e6` was re-checked through the GitHub API.

**Runs:** the 5 changed test files, `test_adapter_static_only.py`, and `ruff` on the 10 changed files all exit 0, plus 1.0.0-vs-branch A/B harnesses.

pengfei-threemoonslab:
**Addressed review cycle 2.** New head `16263d8401a7fcda6c02d5b1c71a72d89d3b9cb3`, rebased onto `42110535` (#771).

**P1: a plugin the repository itself enables**
The docs checked were the settings reference, discover-plugins and plugin-marketplaces. Only external-source plugins wait for a manual install; relative `directory`/`file` marketplace sources resolve against the checkout; a `settings` marketplace lists only external plugins.

A plugin's selected hooks are now `execute`/`high` with expansion (basis `project_enabled_plugin`) only when all three hold:
1. `.claude/settings.json` or `.claude/settings.local.json` sets `<plugin>@<marketplace>` to exactly `true`;
2. that marketplace is registered as a relative `directory`/`file` source resolving to an in-repo `.claude-plugin/marketplace.json`;
3. that marketplace lists the plugin with an in-repo source.

Everything else stays `medium` with no expansion: `false`, an unknown marketplace, `github`/`git`/`url`/`settings` sources, and absolute, `~` or escaping paths. An unlisted plugin selects nothing.

| Settings | 1.0.0 `diff` | `b989b5e3` | `16263d84` |
|---|---|---|---|
| `demo@market: true` | ⚠ high widened | medium changed | **⚠ high widened** |
| `false` / github source / unlisted | ⚠ high widened | medium changed | medium changed |

In the enabled case, the `verify` PR-comment line is identical to 1.0.0, `check` puts the row first as `expands: true`, and the Stop hook announces it.

Decisions:
- `settings.local.json` counts (the host reads it).
- `true` in either file counts.
- The basis is read only on the current side of a change, and removal rows name no basis, so a 1.0.0 baseline's `execute`/`high` hook grant is never described as enabled. Tests pin the neutral removal and no-drift.

**P2: `check` with a manifest the head broke**
`check` now drops only a blocking plugin-reference limit present on **both** sides for an unchanged file. A one-sided limit, or one on a changed file, refuses the comparison with the existing `head_inventory_incomplete`/`base_inventory_incomplete` reason.

With a head-broken manifest (with or without a hook edit):
- 1.0.0: `allow`, comparable, no rows;
- `b989b5e3`: `allow`, comparable, plus a `medium / removed` row;
- `16263d84`: `allow`, `control_state: complete`, **incomparable**, no rows, in all three formats.

The decision stays `allow`, as in 1.0.0, because `check` does not route plugin files. A head-broken `.claude/settings.json` stays `require_review`, identical to 1.0.0. The cycle-1 untouched-manifest tests still pass.

**Nits**
- **(a)** A source containing `/` is not a bare name. An escaping or non-`./` `metadata.pluginRoot` gets a named, non-blocking issue.
- **(b)** STABILITY.md, the support page and the PR body say read limits (such as an untouched plugin hook file over 1 MiB) make `diff`/`verify` incomparable, as an oversize settings file does.
- **(c)** A reference beneath a symlink leaving the workspace keeps only the blocking link issue.

**Rebase:** conflicts in CHANGELOG, STABILITY.md, distribution-surfaces, the support page and `capability_diff_rows.py` were resolved by hand. The #714 note says "still contract v40" and follows #771's migration note. `## Unreleased` has main's 7 bullets plus #714, each once.

**Tests on `16263d84`**
| Run | Exit |
|---|---|
| `ruff check .` | 0 |
| `test_adapter_static_only.py` | 0 |
| 18 touched/related files | 0 |
| Full suite, 10 foreground chunks | all 0 (10,989 passed, 7 skipped) |

The published-1.0.0 comparisons ran before the second rebase; the new tests cover the same scenarios on the final commit.

pengfei-threemoonslab (commented):
Independent coding-agent review/address cycle 3 at `16263d84`. **No P0/P1. One P2, which is a docs and test correction.** CI is green (11 pass).

**P2: STABILITY.md:182 (and the PR body's "Baseline effects") says a 1.0.0 baseline of an enabled plugin's hook "does not drift". Against a baseline actually saved by 1.0.0, it does.**

`test_a_1_0_0_baseline_of_an_enabled_plugin_hook_does_not_drift` (`tests/test_claude_hook_loading_evidence.py:1037`) builds its "1.0.0" baseline with `_recorded_by_1_0_0` (`:583`). That helper rebuilds the baseline from the branch's own inventory, so it already contains the new artifacts and sources.

Setup: a repository that is its own marketplace; `plugin.json` names `./.claude/hooks/hooks.json`; settings register a `directory` marketplace at `.` with `demo@market: true`. Saved with published 1.0.0:
- 1.0.0 and main: `--drift --fail-on-drift` exit 0, no drift;
- branch: exit 20, "Drift detected — 0 typed grant change(s)", no signals. `--json` shows `artifact_changes` adding `.claude-plugin/plugin.json` and `coverage_changes` adding it to claude-code `sources_observed`; the hook grant itself is unchanged.

Only repositories whose manifest or marketplace declares `hooks` are affected.

Fix:
- Reword STABILITY.md, the PR body, `docs/baseline.md` and `examples/github-actions/12-host-grant-drift.yml`: the gate exits 20 once with 0 typed grant changes; read `artifact_changes`/`coverage_changes`.
- Replace the synthetic test with a real 1.0.0-shaped baseline that lacks that artifact and source.

**Nits and follow-ups**
1. **Possible fail-open, unproven.** `host_grants.py:1758/:1766` match `enabledPlugins` only against the `extraKnownMarketplaces` key. With key `market`, marketplace.json `name` `real-market` and `demo@real-market` enabled: 1.0.0 shows `⚠ high widened`, the branch shows `medium changed`. The docs don't say which name Claude Code uses; suggest also accepting the in-repo marketplace.json `name`.
2. **A newly broken plugin manifest makes every host row incomparable** (`head_inventory_incomplete`, no rows) in `check`, `diff`, `verify` and the Stop hook. Every `check` decision still equals 1.0.0, and the Stop hook says "treat as unreviewed". 1.0.0 behaves the same for other broken files but also escalates `check` with `BOUNDARY-INPUT-INCOMPLETE`. Follow-up: refuse only the grants under the affected plugin root.
3. **An enabled plugin's new hook at a non-registry path** (`cfg/hooks.json`) gets `check` `allow` alongside a `high`, `expands: true` row. 1.0.0 also allowed it, with no row, and the PR documents the limit. Follow-up: route these like settings hooks.
4. **Pre-existing (1.0.0, main and branch):** `check` crashes with a `ValidationError` ("publication authority requires complete boundary input coverage") when settings only add `enabledPlugins: {"x@y": true}`.

**Verified with no finding**
- **Enablement rule vs the docs, 18 probes.** Every case Claude Code would load stays `⚠ high widened`, matching 1.0.0:
  - registration and enablement split across `settings.local.json` and `settings.json`, in either direction;
  - array `hooks`, the manifest-less default, and an entry `hooks` path;
  - nested `directory`/`file` marketplaces, and paths without `./` or with a trailing slash;
  - subdirectory plugins, and a `pluginRoot` bare name;
  - enablement added in the head.
- **Shared `execute`/`high` pair:** removal rows are neutral (the removal check comes first, `capability_diff_rows.py:247`), and drift renders no `why`, so nothing saved by 1.0.0 is described as enabled. The declared-only, selected-not-enabled and newly-enabled drift cases match the docs.
- **`check` with a head-broken manifest:** `allow` + `complete` + incomparable rows is consistent with STABILITY.md:257–261 and with `unchanged_limits_not_representable`. A head-broken `.claude/settings.json` matches 1.0.0 (`require_review`, `BOUNDARY-INPUT-INCOMPLETE`).
- **Untouched limits:** 11 scenarios give 1.0.0-identical `check` output in all three formats. Nits (a)–(c) are done.
- **Rebase:** `## Unreleased` has main's 7 bullets plus #714; the #771 STABILITY note is intact; the parity test passes; #771's step-action code is intact.

**Runs:** 18 hook, host, check, comparison and boundary test files, static-only and `ruff` all exit 0.

pengfei-threemoonslab:
**Addressed review cycle 3.** New head `523aa7fd39fc94c9a97dac2c2ce33eab13b5c360` (no rebase needed; `origin/main` is still `42110535`). CI: 11 pass, 1 skipping.

**P2: the "does not drift" claim**

Reproduced first against a baseline saved by **published 1.0.0** (venv on a 3.12 interpreter), on a repository that is its own marketplace:
| `audit --host --drift --fail-on-drift` | Exit | Output |
|---|---|---|
| published 1.0.0 | 0 | no drift |
| main `42110535` | 0 | no drift |
| `16263d84` | **20** | "Drift detected — 0 typed grant change(s)" |

`--json` on the branch: comparable, `has_drift: true`, `changes: []`, `expansion_signals: []`, `artifact_changes` adding `.claude-plugin/plugin.json`, `coverage_changes` adding it to claude-code `sources_observed`. The hook grant is identical on both sides.

- STABILITY.md:182, `docs/baseline.md`, `examples/github-actions/12-host-grant-drift.yml`, the CHANGELOG and the PR body now say: a repository whose plugin manifest or marketplace declares `hooks` exits 20 once with zero typed grant changes; read `artifact_changes`/`coverage_changes`; re-save from the reviewed default branch.
- `tests/fixtures/host-grants-1.0.0-enabled-plugin.json` is the byte-identical baseline 1.0.0 wrote, with its reproduction commands recorded beside it.
- `_recorded_by_1_0_0` now **refuses** an inventory holding a plugin manifest or marketplace, instead of rebuilding the baseline from the branch's own inventory. Three tests replace the one that could not fail: the fixture's 1.0.0 shape, the drift payload, and the CLI gate exiting 20.
- The helper's two other callers build declared-only workspaces. Checked against a real 1.0.0 baseline: one changed grant (`execute/high` → `unknown/unknown`), no artifact or coverage change, no signals — so what they pin is correct.

**Nit 1: marketplace name**

The marketplace schema calls `marketplace.json`'s `name` the identifier users type after `@`; the settings reference only shows an `extraKnownMarketplaces` key equal to that name. Neither is documented as the one `enabledPlugins` matches, so both are accepted now, on the same "err toward showing a hook" principle as case-insensitive matching. A `name` alone still registers nothing.

Probe (key `market`, `name` `real-market`, `demo@real-market: true`):
| `diff --base main` | Row |
|---|---|
| 1.0.0 | ⚠ high widened |
| `16263d84` | medium changed |
| `523aa7fd` | **⚠ high widened**, "this repository's project settings enable the plugin that selects this hook" |

Negatives pinned: not registered, registered remotely, non-string `name` → `plugin_selected`/`medium`.

**Follow-ups filed**
- **#808** a newly broken plugin manifest makes every host row incomparable. Evidence strengthened: with a head that breaks a manifest *and* drops a deny rule, 1.0.0 `diff` shows `⚠ low removed … Bash(curl:*) → gone`, the branch refuses with `head_inventory_incomplete`; `check` keeps the violation and `require_review` but drops the row.
- **#809** an enabled in-repo plugin's hook at a non-registry path gets `check` `allow`.
- **#810** `check` crashes with a `ValidationError` when settings only add `enabledPlugins` — reproduced on 1.0.0, main and the branch (`schemas/agent_boundary.py:136-140`); `diff` is fine, and agent mode prints no `next_action`.

#808 and #809 are recorded on the support page and in the PR body's "Remaining limits".

**Tests on `523aa7fd`**
| Run | Exit |
|---|---|
| `ruff check .` | 0 |
| hook evidence, hooks source, unread surfaces, install hooks, surface parity, step actions | 0 |
| 20 host audit/check/comparison files | 0 |
| Full suite, 10 foreground chunks | all 0 (10,996 passed, 7 skipped) |

pengfei-threemoonslab (commented):
Independent coding-agent review/address cycle 4 (focused) at `523aa7fd`. **No P0/P1. One P2, documentation only, now fixed in `3d27ee59`.**

**P2: the documented re-save recovery failed on a real 1.0.0 baseline.** STABILITY.md:181 and :186, `docs/baseline.md:67-70` and `examples/github-actions/12-host-grant-drift.yml:16,29` all ended at "re-save the baseline". Against a baseline published 1.0.0 actually saved (`host_grants_schema_version: "0.5"`):
```
audit --host --drift --fail-on-drift   # exit 20, "0 typed grant change(s)"  (as documented)
audit --host --save-baseline           # exit 2, unsupported_baseline_schema, file unchanged
audit --host --drift --fail-on-drift   # still exit 20
```
Recovery completes only after moving the old file aside. The refusal is inherited from #771 and is documented at STABILITY.md:172, but both cross-references scope it to *a baseline holding a workflow grant*, and the #714 paragraphs describe a workflow-grant-free 1.0.0 baseline.

Fixed in `3d27ee59`: all three places now name the move-aside step and the refusal. Verified with the checked-in fixture — save over it exits 2 (`unsupported_baseline_schema`), and after `git mv` the re-save exits 0 and the gate reports "No drift". Doc-pinning and hook tests plus the static-only test exit 0.

**Nits (not fixed, recorded)**
1. **`enabled_entries` (`host_grants.py:1769-1790`) over-reports when two registered marketplaces answer to the same string.** With keys `a`→`mkt-a` and `b`→`mkt-b` whose `marketplace.json` files are both named `shared`, `demo@shared: true` marks **both** plugins' hooks `execute`/`high`; likewise when one marketplace's own name equals another's settings key. Claude Code resolves the name to one marketplace, so one row is a false positive. Every pinned negative holds, and **no case was found where a hook nothing selects becomes high or expanding** (`enabled_roots` is intersected with `roots_by_file`). Optional: prefer the `extraKnownMarketplaces` key on collision, or name the ambiguity as a non-blocking issue.
2. **`_recorded_by_1_0_0`'s new guard (`tests/…:618-627`) has a hole of the class it closes.** It refuses on a manifest or marketplace *artifact*, but those become artifacts only when they declare `hooks`; a plugin with no `hooks` member whose root holds `plugins/demo/hooks/hooks.json` slips through, and 1.0.0 never read that file. No current caller has that shape. Better: refuse any source outside 1.0.0's `sources_expected`.

**Verified with no finding**
- **The fixture is genuine.** Rebuilt the workspace and ran published 1.0.0's `--save-baseline` independently: byte-identical to the checked-in fixture (same sha256), no normalisation.
- **The replacement tests are not vacuous.** Three of them characterise behaviour `16263d84` already had, so they don't discriminate the commits; rewriting the fixture with a baseline this reader would write makes all three fail, while the old `…_does_not_drift` passed on a synthesized baseline. The marketplace-name test does discriminate (`medium` on `16263d84`, `high` here). The guard raises rather than skipping.
- **The helper's other callers hold** against a real 1.0.0 baseline: one `changed` row with declared-only wording, no artifact or coverage change; after deletion, the neutral removal wording with no mention of a plugin or enablement.
- **Every other documented sentence is true:** gate exit 20 with 0 typed grant changes and no rows; `--json` carries `changes`, `artifact_changes` and `coverage_changes`; a plugin hook outside the registry paths is a new grant with `hook_added` only when project settings enable the plugin; a manifest version bump is not drift; main reports no drift on the same repository.
- **Regression sweep vs 1.0.0, all identical:** untouched malformed manifest with a README-only head and with a deny-rule removal in all three `check` formats (byte-equal rows); a settings inline hook still `⚠ high added`; the enabled-plugin row `⚠ high widened`; the key/name split now `⚠ high widened` again.
- **#808, #809 and #810 exist, are open, and each reproduces** as described; #810 reproduces on 1.0.0 as well.
- **CHANGELOG** `## Unreleased` has main's 7 bullets plus #714, each once. CI on `523aa7fd`: 11 pass, 1 skipping.

**Runs:** the five hook/surface test files, the static-only test alone, `ruff` on 10 files, the A/B on `16263d84`, the mutation check, and the CLI sweeps — all as expected.

Issue #780, opened 2026-09-15
Add an exercised host-only advisory PR recipe with no manifest or saved baseline


pengfei-threemoonslab:
## Observed

The 2026-09-14 review at `e2ab00078b2ca5b9e73405109a6f09b820f5fc9e` found a gap between the shipped host-review engine and the integration a first team is told to copy:

- `examples/github-actions/README.md:3` describes its recipes as requiring a root `shipgate.yaml`; the recommended first recipe is presented as adding a gate.
- `docs/integrations.md:22–31` starts with `config: shipgate.yaml`, although the released Route H measurement in `docs/design-partner-pilot-results.md:95–97` records successful manifest-free `verify` and six advisory change rows.
- The host-grant recipe is scheduled baseline/drift, a different job from reviewing the current PR against Git history. Local `diff` already needs neither a manifest nor a committed baseline.

The product has a manifest-free per-change path, but a team that got value locally has no equally clear, exercised first PR recipe. A GitHub App is not required to fix that gap.

## Why it matters

Repeated use should arrive on the next relevant PR without a developer remembering another audit command. Requiring declarations or baseline maintenance before that happens creates adoption work unrelated to the selected host-review job. The first integration must show what the current PR changed, explain unread coverage and let the team's reviewer decide.

## Proposed

Add one copyable **host-only advisory PR** recipe to the existing GitHub Actions examples, using the current composite Action and existing manifest-free verifier/PR projection. Pin the Action source ref and package version from the tested released provenance rather than maintaining another handwritten version constant. Do not introduce a new renderer, Action wrapper, hosted service or gate vocabulary.

Make the base/head checkout explicit and available locally (`fetch-depth: 0` or an equivalently exercised bounded fetch). The engine never fetches. Exercise the actual composite Action, not a script that merely reconstructs its expected CLI arguments. A manifest or pre-existing host baseline must be absent in the tested repository.

Default to advisory output, with no required check, fail-on-verdict policy or generated host instruction. Show one human PR summary naming changed capabilities, evidence, coverage limits and next action. Keep detailed machine output in the existing artifacts. Explain the minimum `contents`/PR-comment permissions, and preserve the workflow-summary/artifact fallback when a fork or token cannot post a comment. Do not ask for broad write permissions or use an untrusted-code execution path to make a fork appear supported.

[...]

Issue #778, opened 2026-09-15
Epic: earn the complete permission-review journey, from a real change to voluntary reuse


pengfei-threemoonslab:
## Selected execution — 2026-09-24

Days 1–5 execution is selected and recorded in PR #869. Application-agent proof (#868, milestone 13) is primary; host correctness is bounded maintenance. Accountable product owner: @pengfei-threemoonslab; technical executor: Codex current task. #580/#655 design precedes #867 and the reproduced #864/#865/#866 readers. #610 is reproduced; #787 is the bounded recipe repair. Milestones 9–11 were reconciled, retaining Oct 14 / Nov 13 / Dec 13 as evidence checkpoints. #830 and external outreach holds remain unchanged. Older recruitment-first and hosted-deferred text is superseded by this dated selection; no external value or ten-case result is claimed.

---

Earlier sequencing below is retained as history and superseded where it conflicts with this dated execution selection.

## Product outcome

A reviewer can turn a real agent-permission change into an explicit, evidence-grounded decision, carry out the next step, verify the selected correction and choose to use the workflow on the next eligible change.

**Real change → understand → decide → verify correction → voluntary next use.**

This is the sole product program epic. #811 owns the selected experience; its implementation children #839 and #840 close the fact-to-decision and correction-verification gaps. #791 remains discovery coordination. No separate dashboard, control plane, research cohort or approval engine is introduced.

## Current sequence — 2026-09-18

The owner's [2026-09-16 decision](https://github.com/ThreeMoonsLab/agents-shipgate/issues/778#issuecomment-5706318045) controls execution: build the selected work in milestones 10–12, run #830 on a released build, and begin external recruitment/outreach only after that gate passes and an explicit owner decision permits it. Internal preparation, fixture replay and product design may proceed now. This update performs no outreach and does not change #830's thresholds or per-note approval.

The original 2026-09-14 program clock and Oct 14 / Nov 13 / Dec 13 checkpoints remain. A missing participant is a dated shortfall, not value. Each participant's existing four-week observation window starts at first value; preserve any prior observations and never reset a window for this plan. #811's known-change decision-support hypothesis and #830's novel automatic author-actionable finding hypothesis retain separate measures and denominators.

## Why this is the next product work

The evidence chain has useful foundations, but a generic "Does the team intend this declared capability change?" leaves the reviewer to infer the available choices and the condition a rerun would verify. More findings or severity labels do not supply business intent.

[...]

PR #790, opened 2026-09-15
Refuse FIFOs in evidence readers before the open can block (#577)

### LINKED ISSUE #577: Evidence readers can block on a FIFO before rejecting non-regular inputs

pengfei-threemoonslab:
## Problem

`read_regular_file_beneath` opened the leaf with `O_RDONLY | O_NOFOLLOW` and only then checked `S_ISREG`. Opening a FIFO with no writer blocks in `open(2)`, so the reader never reached its refusal. The size limit bounded bytes, not time. The same open-before-type-check shape existed in two other readers. Two verify-side reads had no type check at all on paths an agent can place before the run.

This is a bounded repair under the post-1.0 reliability reserve (#778). It adds no new surface, schema, contract or check.

## Reproduction (macOS, Python 3.12)

Each case uses a FIFO with no writer and runs the reader in a child process with a 3-second timeout.

| Reader | `origin/main` 45d6ca7a | This branch |
|---|---|---|
| `read_regular_file_beneath(root, "input.json")` | timed out after 3 s | `receipt artifact is not a regular file: input.json` (1.03 s, incl. start-up) |
| `load_external_trust_policy(<fifo>)` | timed out after 3 s | `trust_policy_not_regular_file` (0.20 s) |
| `ensure_local_review_excludes` (`.git/info/exclude` is a FIFO) | timed out after 3 s | `ConfigError: Refusing to use a non-regular file: …` (1.19 s) |
| `_declaration_continuation_holds` (`sg-out/declaration-continuation.json` is a FIFO) | timed out after 3 s | `False` (1.47 s) |
| `_load_cached_capability_lock` (`capabilities.lock.json` is a FIFO) | timed out after 3 s | `None` + "not a regular file" note (1.09 s) |

A probe of `os.open` on macOS shows why:

| Open | Result |
|---|---|
| Leaf FIFO, old flags | blocks |
| Leaf FIFO, `+O_NONBLOCK` | opens at once as `prw-r--r--`, so `fstat` refuses it |
| FIFO as a parent, `O_DIRECTORY` | `ENOTDIR` |
| Directory leaf | opens, then `fstat` refuses it |
| Unix socket | `EOPNOTSUPP`, raised as `OSError` and reported as the named read failure |

## Change

- **`core/verification_identity.py::read_regular_file_beneath`**
  - The leaf now opens with `O_NONBLOCK`, plus `O_BINARY` following the `release_source.py` convention.
  - Unchanged: `O_NOFOLLOW` on every component, descriptor-relative containment, the `fstat` type check, the size limit, the before/after identity check and the returned bytes.
  - A FIFO, directory or device is refused by the existing `"<label> is not a regular file"` failure before any read. A socket fails at open with the existing `"could not safely read <label>"` failure.
  - Parent components still use `O_DIRECTORY`, so a FIFO parent fails with `ENOTDIR` without blocking.
- **`core/human_authorization.py::_read_trust_policy_no_follow`**: the leaf now opens with `O_NONBLOCK`, so a FIFO reaches `trust_policy_not_regular_file`.
- **`cli/discovery/local_review.py::_read_regular_file`**
  - Now opens with `O_NONBLOCK`, so a FIFO reaches `Refusing to use a non-regular file`.
  - Reads the descriptor with an `os.read` loop, like the other two readers, instead of a buffered stream. A buffered read over an `O_NONBLOCK` descriptor can return `None` instead of raising.
  - It still reads to EOF. No size limit was added, and every error message is unchanged.
- **`cli/verify/orchestrator.py`**: two reads that had no type check now go through `read_regular_file_beneath`. That is the reader `current_control` already uses for receipts in the same output directory.
  - **Declaration-continuation receipt:** previously `receipt_path.read_bytes()`. Now capped at 1 MiB. A refusal still returns `False`, so the publish-only route is simply not offered. That includes a receipt over the cap, which fails closed.
  - **Cached base capability lock:** previously `exists()` then `read_text()`. Now capped at 64 MiB, the same cap as the cached `report.json` beside it. A refusal still returns the existing "invalid; capability diff may fall back" note.
  - **Side effect, symlinks:** a *symlinked* receipt or cached lock is now refused too. On `main` both were followed. Both outcomes fail closed. Shipgate writes both files as regular files (`apply-patches`, `_write_capability_lock_to_cache`), and the sibling `_cache_report_valid` already refuses symlinks.
  - **Side effect, non-UTF-8 cached lock:** on `main`, `read_text(encoding="utf-8")` raised `UnicodeDecodeError`, which is not an `OSError` or `InputParseError`, so it escaped `_load_cached_capability_lock`. It is a `ValueError`, so the handler now catches it and returns the fall-back note.

### Platform notes

- **`O_NONBLOCK` stays set on regular files.** POSIX gives the flag no effect on regular-file reads. The one historical exception is Linux mandatory locking (removed in 5.15), where a read could return `EAGAIN`. All three readers now read through `os.read`, so that case raises `OSError` and fails closed. The precedent readers (`release_source`, `trust_roots`, `host_grants`, `agent_result`, `install_hooks`) also leave the flag set.
- **Windows:**
  - Every flag is read through `getattr(os, …, 0)`, and Windows has no filesystem FIFOs.
  - `read_regular_file_beneath` and the trust-policy reader open their root with `os.open(<directory>, O_RDONLY)` and then use `dir_fd`. Windows supports neither, so both readers are POSIX-only. This PR does not change them in that respect.
  - **The two rerouted `orchestrator.py` reads now share that POSIX-only reader.** On Windows the cached base capability lock always falls back with the "invalid" note, and `_declaration_continuation_holds` always returns `False`.
  - Nothing visible changes today: `verify` publishes current control through the same reader, so it already exits 4 on Windows before either outcome matters.
  - `O_BINARY` is added where bytes are hashed.
  - `O_CLOEXEC` is not added: `os.open` already creates non-inheritable descriptors (PEP 446), and the repo does not use the flag elsewhere.

## Readers swept

I grepped every `os.open(` in `src/agents_shipgate`, every `S_ISREG` site, and the plain `read_bytes`/`read_text`/`open` reads in `cli/verify/*`, `core/current_control.py`, `core/verification_identity.py`, `core/host_grants.py`, `cli/host_audit.py`, `core/human_authorization.py` and `core/authorization_execution.py`.

**Fixed.** A non-regular input present before the run reached a blocking open:
- `verification_identity.read_regular_file_beneath`. This covers every `current_control` read and `validated_receipt_snapshot`, plus `scripts/run_safety_qualification.py`.
- `human_authorization._read_trust_policy_no_follow`
- `local_review._read_regular_file`
- `orchestrator._declaration_continuation_holds` (receipt read)
- `orchestrator._load_cached_capability_lock`

**Already non-blocking.** Unchanged; `release_source.py` is the precedent:
- `release_source.candidate_action_ref`
- `trust_roots.IdentityBoundReadSession.read_bytes`, both the `dir_fd` and portable paths. This covers `StaticInputSnapshot`, `_cache_report_valid` and the changed-worktree capture.
- `trust_roots.read_identity_bound_bytes` and its portable variant
- `host_grants._read_exact_baseline_text`
- `agent_result._read_exact_untracked_file`
- The untracked-diff reader in `install_hooks`

**Not applicable, with reason:**
- **Directory-only opens.**
  - `host_audit._fsync_directory` uses `O_DIRECTORY`, so a FIFO fails with `ENOTDIR`.
  - `current_control._fsync_directory` fsyncs the directory it has just `os.replace`d into. It is a writer, not an evidence reader.
  - Parent components opened with `O_DIRECTORY` fail with `ENOTDIR` rather than blocking.
- **Validators that never open the file:** `--config` checks in `agent_result`, `install_hooks`, `preflight` and `orchestrator`; `host_grants._exact_baseline_read_path`; the baseline check in `host_audit`; `trust_roots._directory_member_kind`.
- **Shipgate-created private paths:** `verify/git.py` pack file and materialized tree; `authorization_execution` pack file; the `load_validated_receipt_artifacts` snapshot temp directory.
- **Type-checked before the open, so only a concurrent swap during the run could block:**
  - `static_inputs.read_static_input_bytes` fallback (`stat` + `S_ISREG`)
  - `orchestrator._load_external_human_authorization` (`is_file`)
  - The base-report, plan, packet and report re-reads in `orchestrator` (`is_file`)
  - `_sha256_file` (`is_file`)
  - The manifest re-read in `_declaration_continuation_holds` (config validated with `lstat` earlier)
  - `current_control._current_control_id` (`is_file`)
  - `verification_identity.build_blob`/`sha256_file` fallbacks, `_manifest_declared_input_paths` and `validate_receipt_artifacts` (`is_file` or a validated config)

  A FIFO placed before the run is refused by those checks. Making them swap-proof is a wider change than this issue asks for, so they are left as they are.

## Tests

New file: `tests/test_nonregular_evidence_readers.py`, 16 tests.
- Every probe that could block runs in a child process with a hard 30-second `subprocess.run` timeout, so a regression fails the test instead of hanging the suite.
- "Promptly" is asserted separately: the reader call alone, timed inside the child, must take under 10 s.

Platform skips, one marker per test:
- `needs_fifo` (skips without `os.mkfifo`, i.e. Windows): the six FIFO tests.
- `needs_dir_fd` (skips without `dir_fd` support for `os.open`, i.e. Windows): the directory and regular-file tests for `read_regular_file_beneath`, the trust-policy directory test, the continuation-receipt and cached-lock directory tests, and the three symlink/UTF-8 tests.
- `needs_posix` (skips when `os.name != "posix"`): the local-review directory test.

What it covers:
- **`read_regular_file_beneath`:**
  - FIFO with no writer: the exact named failure
  - FIFO with a writer and bytes in the pipe: refused
  - FIFO as a parent component: refused
  - Directory: refused
  - Regular file: bytes including CR LF, `0x1A` and NUL come back byte-identical with the same SHA-256; the size limit holds; `..` and absolute paths are refused; a symlinked leaf or parent is refused
- **Trust policy:** FIFO and directory give `trust_policy_not_regular_file`.
- **Local review:** a FIFO or directory at `.git/info/exclude` is refused.
- **Declaration-continuation receipt:**
  - A FIFO or directory gives `False`.
  - A symlinked receipt gives `False`. The same receipt as a regular file first gives `True`, on a first-adoption repository with the introduction proof.
- **Cached capability lock:**
  - A FIFO or directory gives no lock plus the fall-back note.
  - A symlinked lock gives no lock plus the fall-back note. The same bytes as a regular file first load.
  - A lock that is not UTF-8 gives no lock plus the fall-back note, instead of raising.

Results on 58f7f903 (rebased onto `origin/main` 3131c9c3; macOS, Python 3.12):
- `tests/test_nonregular_evidence_readers.py`: passed (16 tests), exit 0.
- **A/B, FIFO tests:** at 112db542, the file run against an `origin/main` tree gave 6 failed and 7 passed.
  - The six failures: the five no-writer FIFO tests hit the 30 s limit (one per fixed reader), and the cached-lock directory test failed only on the note text.
  - The seven passes were expected: FIFO with a writer, FIFO parent, directories and regular files already behaved.
- **A/B, the three new tests**, run from inside an `origin/main` 67a3a5e7 tree so its own `src/` is imported (the repository's `pythonpath`/`conftest.py` otherwise put this checkout's `src/` first). All three fail on `main`:
  - Symlinked cached lock: `main` follows the link and loads the lock.
  - Non-UTF-8 cached lock: `main` raises `UnicodeDecodeError` from `read_text`.
  - Symlinked receipt: `main` returns `True`.

  The regular-file control step in each test passes on `main`.
- Related modules, run with `-n auto`: `test_local_review`, `test_declaration_confirmation_route`. Both passed on this head. At 112db542 the wider set also passed: `test_human_authorization`, `test_current_control`, `test_current_control_closure`, `test_current_control_auxiliary_inputs`, `test_verify_orchestrator`, `test_safety_qualification`, `test_human_review_decision`, `test_capability_lock`.
- `pytest -q -p no:cacheprovider tests/test_adapter_static_only.py`: passed, exit 0. No allowlisted file was touched.
- `pytest -q -p no:cacheprovider -n auto -m "not perf" --ignore=tests/test_adapter_static_only.py`: passed, exit 0.
- `ruff check` on the changed files: clean, exit 0.

CHANGELOG: one bullet under `## Unreleased`, kept after `main`'s #781, #779/#571 and two #780 bullets. It now also mentions the symlink and non-UTF-8 outcomes.

Closes #577


pengfei-threemoonslab (commented):
Independent coding-agent review/address cycle 1, at `112db542`.

No P0. The code change is correct. Merge is blocked by one P1; two P2s and three nits remain.

**Verified:**
- `O_NONBLOCK` is added only to the leaf open, in all three readers.
- `fstat`/`S_ISREG` still runs before any read; the size limit, identity tuple and returned bytes are unchanged.
- No new exception type escapes.
- The "already non-blocking" sweep claims hold.
- A/B against the merge base `45d6ca7a`: the new file gives 6 failed / 7 passed there, matching the PR.
- The new file plus the static-only test, 20 related modules (786 tests), the full suite (`-n auto -m "not perf"`) and `ruff` all pass.

**P1**
1. **Conflicting with `main` (CHANGELOG only), so CI has never run.** The FIFO tests have run only on macOS and never on the Ubuntu shards. Fix: rebase, keep all three Unreleased bullets, push, and confirm green CI.

**P2**
1. **The Windows claim in the PR body is wrong for the two rerouted `orchestrator.py` reads** (`_load_cached_capability_lock`, `_declaration_continuation_holds`).
   - `read_regular_file_beneath` opens the root with `os.open(root, O_RDONLY)`, which fails on Windows. The cached lock now always falls back, and continuation always returns `False`.
   - Nothing visible changes today, because verify publishes current control through the same POSIX-only reader and already exits 4 on Windows.
   - Fix: correct the body.
2. **Two behaviour changes have no test:**
   - a symlinked cached lock or receipt: `main` follows it, the branch refuses it;
   - a non-UTF-8 cached lock: `UnicodeDecodeError` escaped `(OSError, InputParseError)` on `main`, and the branch returns the fall-back note.

   Fix: add both tests and mention the UTF-8 fix in the body.

**Nits**
- The three directory tests have no platform skip, contrary to the body's claim. Add `@needs_dir_fd` or a POSIX skip.
- The 1 MiB receipt cap fails closed; a short comment would explain it.
- `local_review._read_regular_file` reads through a buffered stream, which could return `None` under `O_NONBLOCK` (Linux mandatory locking only). An `os.read` loop like the other readers would remove that.

pengfei-threemoonslab:
**Addressed review cycle 1.** New head `58f7f903db0abeb3b71b676fe694c64803f1f2e1`, rebased onto `3131c9c3`.

**P1 (conflict)**
- Rebased twice, because #786 landed in between; both conflicts were CHANGELOG only.
- `## Unreleased` keeps every bullet from `main` plus this PR's bullet, each exactly once.
- Source and test files are identical to the pre-rebase fix commit.
- CI is now running.

**P2 (Windows claim)**
- The PR body now says both rerouted `verify` reads share current control's POSIX-only reader.
- On Windows the cached lock always falls back and the continuation check returns `False`; `verify` already exits 4 there.

**P2 (untested behaviour)** — three new tests, each with a regular-file control:
- a symlinked cached lock falls back with the note;
- a symlinked continuation receipt returns `False`;
- a non-UTF-8 cached lock falls back instead of raising.

All three fail on `origin/main`, run from a fresh `origin/main` tree: the lock loads, the receipt returns `True`, and the UTF-8 case raises `UnicodeDecodeError`. The root `conftest.py` puts the checkout's own `src/` first, so an A/B run with only `PYTHONPATH` changed does not test `main`. The body and the CHANGELOG bullet mention the UTF-8 fix.

**Nits**
- Added `@needs_dir_fd` to the lock and continuation directory tests, and a POSIX skip to the local-review directory test.
- Added a fail-closed comment at the 1 MiB receipt cap.
- `local_review._read_regular_file` now uses an `os.read` loop; error messages are unchanged and it still reads to EOF.

**Tests on `58f7f903`**
| Run | Exit |
|---|---|
| `tests/test_nonregular_evidence_readers.py` (16 tests) | 0 |
| `ruff` on changed files | 0 |
| `test_local_review` + `test_declaration_confirmation_route` | 0 |
| Full suite (`-n auto -m "not perf"`) | 0 |

pengfei-threemoonslab (commented):
Independent coding-agent review/address cycle 2 at `58f7f903`: **no P0/P1/P2, mergeable.**

**Cycle-1 items resolved**
- **CHANGELOG:** `## Unreleased` gains only this PR's bullet (5 → 6, none duplicated), and its wording matches the behaviour.
- **Windows statement:** correct. `bind_current_control_artifacts` (`current_control.py:358`) already uses the same POSIX-only reader, so the rerouted reads change nothing visible on Windows.
- **The three new tests are meaningful.** A/B from a separate `git worktree add … origin/main` tree:
  - on main, the symlinked lock loads (the regular-file control step passes first), the non-UTF-8 lock raises `UnicodeDecodeError`, and the symlinked receipt returns `True`;
  - the whole file gives 9 failed / 7 passed on main and 16 passed on the branch.
- **Skip markers:** `needs_dir_fd`, `needs_posix` and `needs_fifo` are each on the right tests.
- **1 MiB cap comment:** present at `orchestrator.py:1726-1727`.
- **`os.read` loop:** it stops at EOF and closes the fd in `finally`. `EAGAIN` surfaces as an `OSError`, the same as before. The `fstat`/`S_ISREG`/`st_nlink` checks are unchanged, and the error messages are byte-identical to main.

**Fail-open re-check**
- `O_NONBLOCK` is on the leaf open only.
- Regular-file bytes and SHA-256 are unchanged.
- The wider `ValueError` catch adds only the decode error and the reader's own refusals, and both lead to the existing fall-back note.
- The continuation receipt has a single writer (`apply_patches.py:459`) and a single reader.

**Nothing newly broken**
- No doc is left stale.
- `-n auto` is clean, and no child processes are left over from the FIFO timeouts.

**Nit (pre-existing, optional):** `local_review.py:310` has no `O_BINARY`, which is the same as main. It matters only if local review is meant to support Windows.

**Runs**
- `test_nonregular_evidence_readers.py`, serial and `-n auto`: exit 0.
- `test_local_review.py` + `test_declaration_confirmation_route.py`: exit 0.
- `test_adapter_static_only.py`: exit 0.
- `ruff`: exit 0.
- A/B against main: exit 1, as expected.

CI: 11 pass, 1 skipping; `MERGEABLE`/`CLEAN`.

Issue #627, opened 2026-09-10
Current-control refresh omits live currency checks for declared input blobs outside dependency_inputs


pengfei-threemoonslab:
## Problem

A current-control read can accept a stale worktree answer after a declared OpenAPI input changes without a Git-visible overlay. The input is already hashed in the verification plan; the missing step is reconfirming that blob during the live read.

## Reproduction and evidence — 2026-09-10

Reproduced in an isolated synthetic GET workspace while implementing #607 on base `61d21527124ed35d006c1a1b0157ba2beab820c1`:

1. Commit a literal OpenAPI GET operation and its synthetic fixture binding; run worktree verification (`archive_head=False`).
2. `read_current_control` accepts the current `review_publishable` answer. `verification-plan.json.inputs.tool_sources` contains `api.json`, its SHA-256 and `source: worktree`.
3. Mark the tracked input `assume-unchanged`, then change only its literal server URL. `git status --porcelain` remains empty.
4. The next current-control read still accepts the previous answer and its edit/commit/push/update-PR permissions. No claim is made that this reproduction returned merge authority; merge remained false.

`current_control.py` reconfirms `options.dependency_inputs` via `validate_dependency_inputs`. That helper does not visit `plan.inputs.tool_sources`. The separate `validate_plan_inputs` checks the broader input set, but cannot simply be called against the checkout for every mode: archived inputs, generated diff artifacts and external inputs have different origins.

## Required implementation

Unify live currency coverage with the actual declared and reader-selected inputs the verifier used. Preserve committed-tree versus worktree meaning, external/generated artifact origins, bounded identity reads, absent/refused input handling and the final second-observation race check. An intact receipt and a clean Git status are insufficient evidence that a worktree source is unchanged.

## Acceptance

- [x] The concrete tracked-but-assume-unchanged OpenAPI input change refuses the old control on the first read; an unchanged input remains readable.
- [x] Applicable manifest/configured source/framework/policy/baseline inputs have one explicit currency rule, including untracked/ignored inputs where the engine read them.
- [x] Committed snapshots do not incorrectly compare their historical source bytes against an unrelated current checkout; worktree snapshots do not acquire committed semantics.
- [x] Missing, moved, unreadable, alias/special-file and concurrently changed inputs retain bounded refusal behavior, without reopening unsafe paths or inferring absence from Git status.
[...]

PR #801, opened 2026-09-15
Resolve agent control's default reports directory under --workspace (#575)

### LINKED ISSUE #575: Control refresh outside the workspace defaults to the caller's reports directory
## Problem

An agent invoking verification from outside a repository can select it with `--workspace`, but the corresponding control refresh still looks for its default reports under the caller's current directory. A valid completed run can therefore appear missing until the caller supplies a second absolute path.

## Evidence

Observed while implementing #567 on main `0aac28671`, using the existing clean-read-only test fixture:

1. From a different current directory, run `agents-shipgate verify --workspace <repo> --config shipgate.yaml --format control`. The fixture completes and publishes under `<repo>/agents-shipgate-reports`.
2. Run `agents-shipgate agent control --workspace <repo>` from the same caller directory. It refuses with `missing`, naming the caller-relative `agents-shipgate-reports/current-control.json`.
3. The same refresh with `--reports-dir <repo>/agents-shipgate-reports` reaches the intended current-control read.

`src/agents_shipgate/cli/agent_interface.py:125–187` validates `workspace` but passes the separate default relative `reports_dir` straight to the reader. This is an entry/recovery usability gap; no incorrect merge authority was observed.

## Resolution direction

Decide and document the default path relationship consistently with `verify --workspace`. Preserve intentional explicit `--reports-dir` behavior and custom output directories. Do not silently retarget an explicit path or search arbitrary repositories for a passing pointer.

## Acceptance

- [ ] A verified repository can be refreshed from another cwd using the documented workspace/default-output flow.
- [ ] Explicit relative/absolute report directories and custom verify outputs retain an unambiguous supported meaning.
- [ ] Missing reports name the actual searched path and an exact usable recovery route.
- [ ] Tests cover a caller directory that also contains another repository's reports; those artifacts cannot be substituted for the requested workspace.
- [ ] Any compatibility change to relative-path defaults is called out before rollout.

## Sequencing

P2, deferred follow-up discovered during #567. Related #328 and #572. The current #567 PR uses explicit report paths to test live-currency behavior and does not change this independent CLI default.

### LINKED ISSUE #785: Keep host comparisons reproducible across worktrees and report output locations
## Observed

Found while exercising #780's host-only advisory route on 2026-09-14. A repository with supported host configuration and no `shipgate.yaml`: `main` allows `Bash(npm test:*)`, denies `Bash(rm -rf:*)` and configures one MCP server; the PR branch widens the rule, drops the denial and adds a server.

With `agents-shipgate==1.0.0` from PyPI (clean virtualenv, outside any checkout) and with this repository's `main` (`45d6ca7a`), run from the PR branch:

| `--out` | `--head` | `host_comparison.comparison_status` | reasons | rows |
| `agents-shipgate-reports` (inside the workspace, as `action.yml` passes it) | full SHA | `comparable` | — | 4 |
| `agents-shipgate-reports` | `HEAD` | `comparable` | — | 4 |
| an absolute directory outside the workspace | full SHA | `incomparable` | `comparison_input_unavailable:ValueError` | 0 |
| outside the workspace | `HEAD` | `incomparable` | `comparison_input_unavailable:ValueError` | 0 |
| outside the workspace | omitted | `incomparable` | `comparison_input_unavailable:ValueError` | 0 |

Command shape (the Action's arguments, with only `--out` varied):

```sh
agents-shipgate verify --workspace . --config shipgate.yaml --ci-mode advisory \
  --out <dir> --format json --pr-comment-style capability-review \
  --base main --head <sha|HEAD>
```

Every run exits 0. `agents-shipgate diff` on the same repository is unaffected. The PR comment for the outside-the-workspace runs reads `Host capability comparison unavailable: comparison_input_unavailable:ValueError` and routes to `audit --host`.

## Why it matters

It fails closed: the result is `incomparable`, not a quiet no-change, and the documented GitHub Action path is unaffected. But a local user or script that writes reports to a temporary directory — a common way to keep a repository clean — loses the manifest-free comparison entirely, and the reason names a Python exception class rather than an input the user can repair. `src/agents_shipgate/cli/verify/host_comparison.py:host_comparison_failure` renders any exception as `comparison_input_unavailable:<ExceptionType>`, so the underlying cause is not visible.

## Proposed

Find the `ValueError` (likely a path relation between the output directory and the workspace in the manifest-free verify/host-comparison path) and either support an output directory outside the workspace or refuse it up front with a named, actionable reason. Keep the comparison fail-closed; do not turn an unreadable input into an empty comparable result.

## Acceptance

- [ ] A regression reproduces the outside-the-workspace `--out` case above through the real CLI.
- [ ] Afterwards the comparison either succeeds with the same rows as the in-workspace run, or refuses before comparing with a reason that names the output location (no exception class name as the only reason).
- [ ] The in-workspace Action path, `diff`, and existing incomparable reasons are unchanged.

## Sequencing

P1 reliability repair for #811's local reproduction path. The working in-workspace #780 Action path remains usable; do not make unrelated fixes a prerequisite to trying it. Keep #575's control-refresh path handling and #804's tracked-output exclusion defect separately owned.
## Additional F1/F7 evidence and acceptance — 2026-09-16

The September 15 1.0.0 product-feedback run compared a plain clone and linked worktree on the same compucorp#31 refs. Plain-clone preview produced 20 rows, linked-worktree preview returned `comparison_input_unavailable:ValueError`, and standalone diff produced 20 rows on both. The report includes a no-head variant. This is an additional symptom to investigate here; it does not establish that .git-file layout alone is the cause or that it shares the original --out root cause.

- [ ] Exercise a matrix of plain clone/linked worktree, caller inside/outside workspace, report output inside/outside workspace, and explicit/omitted head. Keep logical inputs identical and record absolute resolved paths, refs and version.
- [ ] Support the documented valid combinations with the same semantic rows. A genuinely invalid output/input path is rejected with the actual location and actionable reason; a bare exception class or unrelated setup route is insufficient.
- [ ] Keep standalone diff as a control; verify/preview do not become quiet no-change on failure. If the two symptoms have distinct causes, record bounded follow-ups and which matrix cells each fixes.
- [ ] Document how verify --out resolves explicit relative paths and its default. Correct help and emitted reproduction commands; do not silently switch resolution to cwd. Exercise emitted commands from another directory.
- [ ] Coordinate with #575 so follow-up control reads target the intended reports. Preserve #804's requirement that output exclusion cannot hide tracked source changes.
- [ ] Preserve the existing Action behavior, source identity, redaction and refusal semantics. F7 is a small path-contract/help repair in this issue, not a new feature subsystem.

pengfei-threemoonslab:
## Problem

`agents-shipgate verify --workspace <repo>` publishes under `<repo>/agents-shipgate-reports` from any current directory, but `agents-shipgate agent control --workspace <repo>` checked `workspace` and then read its default `--reports-dir` (`agents-shipgate-reports`) against the **caller's** cwd. From anywhere else, a valid run looked `missing`. From inside another verified repository, the reader opened *that* repository's pointer and refused it as `workspace_changed` against the requested one.

A second defect of the same kind is recorded on the issue (the #627 case), and #785 is the same class: an output directory **outside** the repository, such as a sibling `verify --out`, was handed to Git as a change-set exclusion. `_worktree_pathspec` rejects an exclusion outside the workspace ("must remain inside workspace"), and **every** caller swallowed that refusal:

| Where the exclusion was built | What the swallowed refusal became |
| --- | --- |
| `verify` worktree read (`orchestrator.py`, no `--head`) | the diff read failed, the run exited early with no input census: exit 2, "input directory capture is unavailable" |
| `verify --preview` worktree read | "Preview could not read the working tree": an empty change set |
| the pointer's worktree overlay (`_safe_worktree_overlay`) | no overlay bound, so an untracked `agent.py` the preview was run on read as "an uncommitted change this decision never saw", on every rerun |
| manifest-free host comparison (`host_comparison.py`, via the overlay) | `incomparable`, `comparison_input_unavailable:ValueError`, no rows (#785) |
| `agent control` / `verify --format control` refresh (`live_workspace`) | `workspace_unverifiable`, exit 4, with a recovery repeating the same `--out` |

## Rule chosen, and why

**Without `--reports-dir`, the reader looks where a default `verify --workspace` writes: `agents-shipgate-reports` under the requested `--workspace`. It is not the Git root and not the caller's cwd.**

- **One rule shared by both sides.** `cli/current_workspace.py::default_reports_dir(workspace)` defines the default. `verify`'s `_resolve_out_dir` writes there when `--out` is omitted, and `agent control` reads there when `--reports-dir` is omitted. `DEFAULT_OUT_DIR` is the same `DEFAULT_REPORTS_DIR` object, and a test ties that spelling to the published `DEFAULT_PATHS["reports_dir"]`. The rule follows the *requested* workspace, not the Git root, because that is what `verify` already did for nested projects (#363). Using the Git root would disagree with `verify` for a monorepo project.
- **Explicit `--reports-dir` keeps its meaning.** An absolute path is used as given, and a relative path still resolves against the current directory. It is never retargeted to the workspace. If it holds no pointer, the read refuses and nothing else is searched. The caller's own reports cannot stand in for the requested workspace.
- **Refusals name a path that resolves for the caller.** The directory is absolute internally, so the `missing` message, `why`, and the recovery command's `--out` all name the directory that was actually searched. The recovery uses an absolute `--out` because `verify` resolves a relative `--out` against the Git root, not the caller, so a relative spelling would name a different directory. The refusal guidance spells `verify` through the invocation policy (#322), like the recovery command.
- **Artifact paths in the envelope keep their spelling wherever they already worked.** An explicit `--reports-dir` keeps the caller's own spelling. The default is printed relative to the cwd when it lies beneath it, so `--workspace .` from the repository prints the same `agents-shipgate-reports/...` paths as before, and absolute otherwise.

**An output directory outside the repository is not a Git pathspec — one rule at every `exclude=`.**

- `cli/current_workspace.py::worktree_exclusion(root, reports_dir)` is the single containment rule, used at every place an output directory becomes a Git exclusion: `verify`'s two worktree reads, the preview's worktree read, `_safe_worktree_overlay` (which the host comparison goes through) and `live_workspace`. It lives beside `live_workspace` rather than in `verify/git.py` so the line-pinned subprocess allowlist there does not move.
- **Containment is a property of the directory, not of how it was spelled.** The reports path is resolved and its ancestors are compared with the repository root's `(st_dev, st_ino)`, so `verify`'s already-resolved `--out` and the refresh's cwd-anchored `--reports-dir` cannot reach different answers for one place, and a spelling differing only in case where the filesystem folds case is the directory it physically is. Inside the repository the exclusion is returned respelled beneath the root — the one spelling the Git helpers can turn into a pathspec. It is dropped **only** when the resolved directory is disjoint from the repository: an outside spelling resolving inside keeps its exclusion, an in-repository symlink resolving outside loses it, and the root itself or an ancestor is still handed over, still refused by the helper, and still withholds authority (tested directly). The walk is per directory, so `repo-reports` is never inside `repo`.
- **Defence in depth in `core/current_control.py`.** A currency test that could not be run refuses for **every** pointer state, not only for one that authorizes completion. `changed_paths=None` now raises `workspace_unverifiable` on the plan-bound worktree path, on the plan-less preview path (`_validate_live_overlay`, where the change set is the whole test) and on the committed-tree path; an overlay that cannot be recomputed no longer returns silently from the preview path; and a live workspace carrying no ref resolver refuses the base check the same way. So even if a directory could not be classified, a stale pointer cannot answer.
- Dropping an exclusion can only surface more changed paths, never hide one. A change before the run is part of what it decided; a change after it is refused as `workspace_changed` naming the file, identically for reports inside and outside (tested for tracked and untracked files).

No schema or contract version changes. `--reports-dir` keeps its name, and its default moved from a literal to `None` with documented resolution.

### Compatibility (called out per the issue's acceptance)

- The directory read changes only when `--reports-dir` is omitted **and** `--workspace` is not the current directory. A portable `verification prepare`/`assemble` chain that publishes into `agents-shipgate-reports` under some directory other than the workspace now needs `--reports-dir`.
- `--workspace .` reads the same directory and prints the same artifact paths. Its refusal text and recovery `--out` now use the absolute spelling instead of the relative one.
- An in-repository `--out`, and `--out` at the repository root or a parent, behave exactly as before.

## Sweep: other `--workspace` commands with a default relative reports/output path

| Command | What it does with reports | Finding | Change |
| --- | --- | --- | --- |
| `agent control` | reads `current-control.json` (default `agents-shipgate-reports`) | defect: default resolved against the cwd; outside directory refused | **fixed** |
| `verify` / `verify --format control` (with or without `--head`) | writes the run, reads the pointer it just published | directory resolution already right; an outside `--out` failed as in the table above | **fixed** by `worktree_exclusion` |
| `verify --preview` | worktree read and pointer overlay | outside `--out`: empty change set, unbound overlay | **fixed** by `worktree_exclusion` |
| manifest-free `verify` host comparison | overlay capture | outside `--out`: `incomparable`, no rows (#785) | **fixed** by `worktree_exclusion`; default spelled with `default_reports_dir` |
| `verify` (`fix_task.verification_command`, reused as `agent control`'s currency recovery) | records the rerun command | already records absolute `--workspace <git root>` and a root-relative `--out` when non-default; the non-default test now uses `default_reports_dir(git_root)` (the rerun's own default) | spelling only |
| `org bundle` | `--from agents-shipgate-reports/verifier.json` | already resolves `--from` under `--workspace` | none |
| `authorization execute` | `--receipt` / `--artifacts-root` default to the cwd | enforces artifacts inside the workspace and refuses otherwise. The documented route is the absolute command `authorization_execute_command` emits, and `authorization request` (no `--workspace`) shares the same defaults. Retargeting a guarded broker surface is out of scope | none (refuses, never substitutes) |
| `verification prepare` / `worker` / `assemble` | write plan, unit result, receipt and pointer to the cwd / `--artifacts-root` | writers, not pointer readers; the chain is consistently cwd-relative and `assemble` has no `--workspace` | none (compatibility note above) |
| `core/human_review_decision._current_review` | reads the pointer through `live_workspace` | no CLI entry point; takes an explicit reports dir; inherits `worktree_exclusion` | none |
| `bootstrap` | locates scan's `report.json` | `workspace / "agents-shipgate-reports"` literal | uses the shared `DEFAULT_REPORTS_DIR` name, spelling unchanged (it becomes `apply-patches --from`) |
| `skill` runner | `workspace / "agents-shipgate-reports"` default | the `skill` package does not import `cli`, and the path is recorded in its report artifacts, so resolving it would change their bytes | none (left deliberately) |
| `scan`, `audit --host`, `check`, `detect`, `diff`, `trigger`, `preflight`, `doctor`, `mcp`, `install-hooks` | none read a published pointer by a cwd-relative default | n/a | none |

The reader never searches any other directory or repository for a pointer.

## Before / after

Default reports directory, cwd = `<caller>` (issue triage on `main` `45d6ca7a`, reproduced on `3131c9c3`):

```
$ shipgate verify --workspace <repo> --config shipgate.yaml --format control   -> exit 0 (publishes <repo>/agents-shipgate-reports)
$ shipgate agent control --workspace <repo>                                     -> exit 3 (before) / exit 0 complete (after)
Current control is unavailable (missing): No current-control.json is present in agents-shipgate-reports
# caller is itself a verified repo:
$ shipgate agent control --workspace <repo>                                     -> exit 4 workspace_changed (before) / exit 0 (after)
$ shipgate agent control --workspace <unverified>                               -> exit 3 (after)
Current control is unavailable (missing): No current-control.json is present in <tmp>/unverified/agents-shipgate-reports; ...
  next_actions[0].command: shipgate verify --workspace <tmp>/unverified --out <tmp>/unverified/agents-shipgate-reports --json
```

Sibling `--out`, cwd = the repository's parent, real `./shipgate` CLI; `origin/main` `421105359` vs this branch:

| Scenario | `origin/main` | Branch |
| --- | --- | --- |
| `verify --workspace repo --config shipgate.yaml --out <sib> --format control`, clean / tracked edit before / untracked file before | exit 2, `human_review_required`, "input directory capture is unavailable" | exit 0, `complete` — identical to the in-repository run |
| then `agent control --workspace repo --reports-dir <sib>` | exit 4 `workspace_unverifiable` | exit 0 `complete`, same `current_control_id` |
| then an edit after the run | exit 4 `workspace_unverifiable` | exit 4 `workspace_changed` naming the file — identical to in-repository |
| `verify --preview` into `<sib>` with an untracked `agent.py` | preview read no working tree; PR head at cycle 0 then reported "never saw (agent.py)" | reads `agent.py`; envelope, `changed_files`, refresh and after-edit refusal identical to the in-repository preview |
| manifest-free `verify ... --base main --out <outside>` with a full SHA, `HEAD`, or no `--head` (#785) | `incomparable`, `comparison_input_unavailable:ValueError`, 0 rows | `comparable`, the same 4 rows as `--out agents-shipgate-reports` |

### Review cycle 2: the writer and the reader disagreeing (fixed here)

cwd = the repository's parent; `repo/.gitignore` names the link, committed;
`ln -s <parent> repo/via`; `verify --workspace repo --config shipgate.yaml
--format control --out <parent>/sib`, then a tracked edit outside the declared
inputs plus a new untracked file, then a refresh.

| Step | `c685856a` (cycle 1) | This head | `origin/main` `42110535` |
| --- | --- | --- | --- |
| verify into `<parent>/sib` | exit 0, `review_publishable` (commit/push/update_pr true) | exit 0, `review_publishable` | exit 2, "input directory capture is unavailable" |
| refresh through `repo/via/sib` after the edits | **exit 0, still `review_publishable`** | exit 4 `workspace_changed (new_tool.py, notes.txt)` | exit 4 |
| refresh through `<parent>/sib` after the edits | exit 4 `workspace_changed` | exit 4 `workspace_changed`, byte-identical to the line above | exit 4 |
| one string for `--out` and `--reports-dir` through `repo/lnkdir -> <outside>` | **exit 0, still `review_publishable`** | exit 4 `workspace_changed` | exit 2 |
| case-variant `--out <parent>/REPO/rpt` (nit 1) | exit 3, "path changed identity while it was read"; pointer left `in_progress` | exit 0 `complete`, refresh `complete`, later edit refused | exit 2 |

Cycle-1 behaviour is unchanged, re-checked through the same script: a plain
sibling `--out` and an in-repository `--out` both give `complete`, a clean
refresh `complete`, and a later edit `workspace_changed` naming the files;
`--out` at the repository root is still "Verifier `--out` cannot be the
workspace root" and at its parent still "overlaps the config input", on all
three builds.

A preview of a repository **with** a manifest publishes `human_review_required` ("input directory capture is unavailable") and its refresh exits 4 on `main` and on this branch alike, for an in-repository `--out` too. That is existing preview behaviour, not the output location, and it is not changed here; the branch only makes the outside case match the inside one.

## Tests

`tests/test_agent_control_reports_dir.py` (16 test functions, 20 cases with parameters; the caller fixture is itself a *different verified repository*, so a reader resolving against the cwd finds a real, valid pointer, just not the requested one):

- default follows the workspace from another cwd; `current_control_id` is the requested repo's, never the caller's; every printed artifact path opens
- `--workspace .` inside the repo keeps relative artifact spelling
- missing default names the absolute searched path; recovery `--out`/`--workspace` resolve to it; caller's reports never mentioned
- explicit **relative** `--reports-dir` still resolves against the cwd (found from the parent; refused, not retargeted, from the caller)
- explicit **absolute** `--reports-dir` used as given
- `--head` run into an outside directory is read, and drift still refuses; `verify --format control` with an outside `--out` returns `complete`
- **new:** plain `verify` into a sibling, from the parent cwd, for clean / tracked-before / untracked-before: `complete`, refresh `complete`, the pre-run change is in `changed_files`, a post-run change refuses as `workspace_changed` naming it — and every answer equals the in-repository run's
- **new:** `verify --preview` into a sibling, clean and with an untracked `agent.py`: the working tree is read, and envelope, `changed_files`, refresh and after-edit refusal equal the in-repository preview's; "never saw" never appears
- `worktree_exclusion`: inside and `..`-folded spellings kept, respelled beneath the root; a prefix-sharing sibling dropped; root and parent still passed and still refused by the Git helper; a dropped exclusion still reports a dirty file
- **cycle 2:** symlinked spellings (skipped on Windows): outside-resolving-inside keeps its exclusion; **inside-resolving-outside now drops it** — the flipped assertion — for the link, its resolved form and a child of each
- **cycle 2:** a case-variant spelling is the directory it physically is (skipped when the temporary filesystem is case-sensitive)
- **cycle 2:** a refresh spelled through an in-repository link, after a tracked edit and a new untracked file, refuses with exit 4 and gives the *same* line as the direct spelling; the fixture asserts the pointer is not `complete`, so a completion pointer (which failed closed already) cannot make it vacuous
- **cycle 2:** one string used for both `--out` and `--reports-dir` through `repo/lnkdir -> <outside>`: refused as `workspace_changed` naming both files
- **cycle 2:** a case-variant `--out` end to end: `complete`, refresh `complete`, later edit `workspace_changed`
- **cycle 2:** unit — `changed_paths=None` refuses as `workspace_unverifiable` for a `review_publishable` pointer (plan-bound) *and* a plan-less preview pointer, while the same read over a readable clean tree succeeds
- `verify` and `agent control` share one default rule

`tests/test_manifest_free_pr_rows.py`, **new** `test_reports_outside_the_workspace_keep_the_host_comparison[HEAD|None]` (#785): the issue's command shape with `--out` outside the workspace is `comparable` with no reasons, the PR comment carries every change, and the rows equal the in-workspace run's.

**A/B for cycle 2** on a separate worktree at the previous head `c685856a`, with the test file copied in under a new name (pytest's root conftest puts each checkout's own `src` first, so only a separate tree runs the old code): the 7 changed/new tests all fail there — `assert 0 == 4` with `review_publishable` for both symlink refreshes, `assert 3 == 0` for the case-variant `--out`, `{'review_publishable': 'answered anyway', 'preview': 'answered anyway'}` for the unit test, and the three rule assertions. On `origin/main` `42110535` every one of those CLI setups exits 2 at the writer, so the fail-open is reachable only on the cycle-1 head. The other 13 tests in the file pass on both trees.

**A/B from cycle 1, on a separate `git worktree add <scratch> origin/main` (`421105359`)** with both test files copied in under new names, run from that tree (pytest's root conftest puts each checkout's own `src` first, so only a separate tree runs main's code): 15 tests fail there — the 9 new ones (3 plain-verify parity on behaviour, 2 preview parity on "Preview could not read the working tree", 2 #785 on `incomparable`, 2 rule tests on the missing helper) and the 6 earlier behaviour tests. The 2 unchanged-behaviour tests and every pre-existing manifest-free test pass on both trees.

Verification on the final commit `be7441c3` (`origin/main` `42110535` is already an ancestor, so no rebase was needed):

| Command | Exit |
| --- | --- |
| `pytest tests/test_agent_control_reports_dir.py tests/test_manifest_free_pr_rows.py` | 0 |
| `pytest tests/test_adapter_static_only.py` (line-pinned allowlist) | 0 |
| `ruff check` on changed files | 0 |
| `scripts/build-llms-full.py` (no diff) | 0 |
| `pytest -n auto` over `test_current_control*.py`, `test_agent_control_envelope*.py`, `test_workspace_input_guard.py`, `test_adoption_walk.py`, `test_bootstrap.py` (+ the two above) | 0 |
| `pytest -n auto` over `test_verify*.py` (7 files) | 0 |
| `pytest -n auto` over `test_host_*.py` (17 files) | 0 |
| Full suite `pytest -q -p no:cacheprovider -n auto -m "not perf" --ignore=tests/test_adapter_static_only.py`, 10 literal-path chunks (322 files incl. `tests/harness` and `tests/integration`, via pytest `@argsfile`) | `01` 0, `02` 0, `03` 0, `04` 0, `05` 0, `06` 0, `07` 0, `08` 0, `09` 0, `10` 0 |

Docs: `docs/agent-contract-current.md` (+ `llms-full.txt`, regenerated with no diff) and `agent control --help`. One CHANGELOG `## Unreleased` bullet ending `(#575, #785)`; cycle 2 extends that same bullet with the physical-identity containment rule and the refusal that no longer depends on the pointer's state.

Not in this PR, both pre-existing and confirmed on `1.0.0`, `main` and this head: `--out` set to a *tracked* directory leaves that directory out of both the decision and the currency check (#804); and `verify --preview --format control` in a configured repository publishes `human_review_required` ("input directory capture is unavailable") whose refresh exits 4, for the default, an in-repository and a sibling `--out` (#807). Neither is touched here.

Closes #575
Closes #785


pengfei-threemoonslab (commented):
Independent coding-agent review/address cycle 1 at `42795bae`. **No P0 or P1, and no fail-open in the Git exclusion change. One P2.**

**P2: an `--out` outside the repository still fails for plain `verify` and `verify --preview`** — the runs without `--head`, which is the normal local agent loop.

The writing run still passes the outside directory to Git as an exclusion:
- `cli/verify/orchestrator.py:707`, `:728` (`working_tree_context(git_root, exclude=out_dir…)`), `:4845` (`_safe_worktree_overlay`) and `:6110` (preview);
- `cli/verify/host_comparison.py:71`.

`git.py:1589-1591` raises "must remain inside workspace", `orchestrator.py:758` swallows the error, and no input-directory list is recorded. The read then fails at `core/input_directory_identity.py:23`.

Reproduction (clean repo, cwd is its parent, `verify --workspace repo --config shipgate.yaml --out <tmp>/sib --format control`):
| Scenario | 1.0.0 / main | Branch |
|---|---|---|
| clean worktree, sibling `--out` | exit 2, "input directory capture is unavailable" | same |
| then `agent control --reports-dir <tmp>/sib` | exit 4 | exit 4; the recovery repeats the same `--out` |
| `--preview`, sibling `--out`, untracked `agent.py` | "input directory capture is unavailable" | "uncommitted change(s) this decision never saw (agent.py)", which is misleading and loops forever |

This is not a regression, but the CHANGELOG says a sibling `--out` generation "refreshes" and "no longer withholds authority" without scoping that to `--head` runs, and the #627 comment on #575 asks for real support or an early refusal with a usable recovery.

Fix: apply the same containment rule (`_exclusion_within`) at those call sites and add tests for plain `verify` and `--preview` with a sibling `--out`. Alternatively, narrow the claims, file a follow-up, and don't close #575 here.

**Nits**
- `orchestrator.py:1897` (`git_root / DEFAULT_OUT_DIR`, correct for nested projects), `skill/runner.py:86`, `bootstrap.py:413` and `host_comparison.py:71` still build the default path themselves.
- The recovery command's `--workspace` keeps the lexical path while `--out` is resolved through symlinks. It works, but the two are spelled differently.
- `agent_interface.py:212` hard-codes "Re-run `agents-shipgate verify`" instead of following the invocation-policy spelling.

**Pre-existing, outside this PR**
- `--out` set to a tracked directory (`<repo>/docs`, or reached through a link into the repo) leaves that directory out of both the decision and the currency check. An edit to `docs/a.txt` stays `complete`, identically on 1.0.0, main and branch. To be filed separately.
- On 1.0.0, `--workspace ..` from a nested project accepted the nested project's pointer. The branch reports `missing`, which matches the documented rule.

**Verified with no finding**
- **Exclusion change:** dropping an exclusion can only surface more paths. `relative_to` compares whole components, so `repo-reports` is not "inside" `repo`. A helper error is refused as unverifiable.
- **Dirty-tree A/B across 1.0.0, main and branch:**
  - a dirty or untracked file before or after a `--head` verify with a sibling `--out` is refused (`workspace_changed`);
  - `--out` at the repo root or a parent is refused;
  - in-repo symlinks show as untracked;
  - an outside path resolving into the repo keeps its exclusion.
  - `complete` never appears on a dirty tree.
- **Default rule:** holds for nested workspaces, `--workspace .`, a relative workspace with spaces from another cwd, a symlinked workspace, and a cwd inside the reports directory.
- **`verify` default output:** every path-bearing string in all 11 JSON artifacts, `pr-comment.md` and the file lists is identical to main, for both the spaces/relative case and the nested case.
- **Recovery commands:** paths are shell-quoted, the launcher spelling is kept, and the printed recovery run from a third directory exits 0, with the refresh after it also exiting 0.
- **Sweep:** agreed with the PR's "left alone" decisions.
- **Docs and stability:** no pinned relative refusal text, no `agent control` surface row, `message`/`next_actions` are prose, and the CHANGELOG has one `(#575)` bullet.

**Runs:** `test_agent_control_reports_dir.py` (8 tests), about 930 related current-control/envelope/workspace-guard/adoption-walk tests plus static-only, and `ruff` all exit 0; A/B scripts compared branch, main and 1.0.0.

**CI:** all pass; release-tag-consistency skipped.

pengfei-threemoonslab:
**Addressed review cycle 1.** New head `c685856ae099abfb2abf548fc68d0a43552d2292`, rebased onto `42110535` (#771). The CHANGELOG was resolved by hand in each commit; every bullet appears once.

**P2: an `--out` outside the repository**
Every place that hands the output directory to Git as an exclusion now goes through one helper, `worktree_exclusion(root, reports_dir)` in `cli/current_workspace.py`:
- the two `verify` worktree reads, the preview read and `_safe_worktree_overlay` (which `host_comparison` also uses);
- the `agent control` refresh (`live_workspace`).

The exclusion is dropped only when the directory is fully outside the repository, checked on both the given and the resolved path. In-repo directories, the repository root and parents are unchanged.

From the parent directory, real CLI:
| Scenario | main | branch |
|---|---|---|
| Plain `verify --out <sibling>`: clean / tracked edit before / untracked before | exit 2, "input directory capture is unavailable" | exit 0, `complete` (same as an in-repo `--out`) |
| `agent control --reports-dir <sibling>` | exit 4, `workspace_unverifiable` | exit 0, same id |
| Edit after the run | `workspace_unverifiable` | exit 4, `workspace_changed` naming the file |
| `--preview` into a sibling with an untracked `agent.py` | preview could not read the worktree | identical to an in-repo `--out` preview |

A remaining preview behaviour (`human_review_required` with "input directory capture is unavailable" in a repository with `shipgate.yaml`) happens identically with an in-repo `--out` on main and on the branch. The claims here say only that outside now matches inside.

**#785 folded in.** A manifest-free `verify` with an outside `--out` (full SHA, `HEAD` or no `--head`) was `incomparable` with `comparison_input_unavailable:ValueError` and 0 rows. It is now `comparable` with the same 4 rows as the in-workspace run. The PR body has `Closes #575` and `Closes #785` on separate lines.

**New tests**
- plain-verify parity (3 cases) and preview parity (2);
- the containment rule: root and parent still refused, `repo-reports` not treated as inside, dirty files still visible;
- symlink spelling (POSIX);
- #785 (2 cases).

A/B from separate `origin/main` worktrees: all 9 new tests fail there, and the unchanged-behaviour tests pass on both.

**Nits**
- **(a)** The refusal guidance uses `render_command(['verify'])`.
- **(b)** `host_comparison.py` and `orchestrator.py:1897` use `default_reports_dir`; the latter is exactly that function for the git root, with a comment. `bootstrap.py` uses `DEFAULT_REPORTS_DIR`. `skill/runner.py` is left alone: `skill` does not import `cli`, and the change would alter the recorded artifact bytes.
- **(c)** Left as is.

**Checks on `c685856a`**
| Run | Exit |
|---|---|
| `test_agent_control_reports_dir.py` + `test_manifest_free_pr_rows.py` | 0 |
| `test_adapter_static_only.py` | 0 |
| `ruff` | 0 |
| `build-llms-full.py` (no diff) | 0 |
| 26 related files | 0 |
| Full suite, 10 foreground chunks covering 322 files | all 0 |

pengfei-threemoonslab (commented):
Independent coding-agent review/address cycle 2 at `c685856a`. **Not mergeable yet: one P2 fail-open.** No P0/P1. #785 is fixed, and CI passes.

**P2: `verify` and the `agent control` refresh can compute different exclusions for the same directory, so an edited worktree reads as current.**
- **Writer:** `verify` resolves `--out` before calling the helper (`orchestrator.py:5280-5287`), so an outside directory loses its exclusion.
- **Reader:** the refresh passes the unresolved, cwd-anchored path (`agent_interface.py:330`).
- **Helper:** `worktree_exclusion` (`current_workspace.py:131-138`) keeps the exclusion if *either* spelling is inside the repository. So an in-repo symlink pointing outside keeps it.
- **Result:** the Git helper raises "must remain inside workspace". `live_workspace` swallows the error and returns `changed_paths=None` (`current_workspace.py:89-90`). For a non-completion pointer, `current_control.py:864-874` then skips the unseen-change test; preview pointers do the same at `:890-893`.

Reproduction (cwd is the repo's parent; `repo/.gitignore` contains `via`, committed; `ln -s <parent> repo/via`):
```
verify --workspace repo --config shipgate.yaml --format control --out <parent>/sib
echo edit >> repo/notes.txt; echo "x = 1" > repo/new_tool.py
agent control --workspace repo --reports-dir <parent>/repo/via/sib
agent control --workspace repo --reports-dir <parent>/sib
```
| Step | Branch | main |
|---|---|---|
| verify | exit 0, `review_publishable` (commit/push/update_pr true) | exit 2, `human_review_required` |
| refresh through `repo/via/sib` after edits | **exit 0, still `review_publishable`** | exit 4 |
| refresh through `<parent>/sib` after edits | exit 4, `workspace_changed (new_tool.py, notes.txt)` | exit 4 |

- A gitignored or committed symlink `repo/lnkdir -> <outside>`, with the same string passed to both `--out` and `--reports-dir`, gives the same stale `review_publishable`.
- `complete` pointers fail closed (`workspace_unverifiable`).
- This is newly reachable: main and 1.0.0 exit 2 at the writer in every one of these setups.
- It contradicts the CHANGELOG line "a change made after the run is still refused", and `tests/test_agent_control_reports_dir.py:501-503` asserts the behaviour that causes it.

Fix:
1. Decide containment on the resolved spelling only (or by file identity), identically for writer and reader.
2. Flip the assertion at `:503`.
3. Add a CLI regression: sibling `--out`, `review_publishable` pointer, refresh through an in-repo symlink after an edit → `workspace_changed`.

Separately, an unverifiable worktree view (`changed_paths=None`) should never skip the unseen-change test for `review_publishable` or preview pointers.

**Nits**
1. **Case variant on macOS (fail-closed).** On case-insensitive APFS, `--out <parent>/REPO/rpt` is physically inside the repository but classified outside. The branch exits 3 ("path changed identity while it was read") and leaves the pointer `in_progress`; main exits 2. Containment by file identity (`st_dev`/`st_ino` ancestor walk) fixes this and the P2 together.
2. **Docstring.** `current_workspace.py:104-128` overstates what the in-repo-resolves-outside spelling does.

**Pre-existing, confirmed on 1.0.0, main and branch (to be filed separately).** In a configured repository, `verify --preview --format control` exits 0 with `human_review_required` ("input directory capture is unavailable"), and the refresh exits 4, for the default, an in-repo and a sibling `--out`. The docs say preview emits `agent_action_required` with the `verify` command (`docs/agent-contract-current.md:1059-1071`, `:788-791`). The cause: the preview pointer's `verification_plan.inputs.options` has no `input_directories`, and the reader raises at `core/input_directory_identity.py:23`.

**Verified with no finding**
- **Fail-open matrix across branch, main and 1.0.0**, each followed by a tracked edit, an edit under `docs/` and a new untracked file.
  - Now `complete` on the branch, with every later edit refused: sibling, `repo/../sib`, a trailing slash, relative `--out`/`--reports-dir`, a workspace reached through a symlink, an outside directory with a symlink back into the repo, and a nested repo.
  - Identical on all three builds: `--out repo/.git/rpt`, outside spellings that resolve inside, `sib/../repo/rpt`, and an outside symlink into tracked `docs/` (#804 not worsened).
- **#785:** an outside `--out` with a head SHA, `HEAD` or no `--head`, and `--base main` or a base SHA, is `comparable` with 4 rows (main: incomparable, 0 rows). An inside `--out` and `diff` are identical to main.
- **Unaffected:** the Stop hook and `verify --format control`.
- **Cycle-1 nits:**
  - launcher spelling is correct for both entry points;
  - the nested-project artifact paths in all 15 artifacts are identical to main and 1.0.0;
  - `bootstrap.py` is unchanged.
- **CHANGELOG and PR body:** 8 Unreleased bullets, one ending `(#575, #785)`; the `Closes` lines are separate.

**Runs:** the reports-dir and manifest-free rows tests, current_control/envelope/workspace-guard/adoption-walk/bootstrap/host tests, the verify/diff-input/unchanged-limit/first-adoption tests, static-only and `ruff` all exit 0. Probe scripts compared branch, main and 1.0.0.

pengfei-threemoonslab:
**Addressed review cycle 2.** New head `be7441c334557404f017be7431556297dd3ae896` (no rebase needed; `origin/main` is still `42110535`).

**P2a — containment by physical identity.** `cli/current_workspace.py`: `worktree_exclusion` resolves the reports path and walks its ancestors comparing `(st_dev, st_ino)` against the resolved repository root (`_identity`, `_beneath_repository`, `_contains_repository`). Inside the repository it returns the path **respelled beneath the root**, the one spelling `_worktree_pathspec` can relativize. The root and its ancestors are still refused, and a genuinely disjoint directory still drops its exclusion. The lexical `_absolute` helper is gone. Writer (`_resolve_out_dir`, resolved) and reader (`_reports_location`, cwd-anchored) now get the same answer, because the result depends only on the physical directory.

**P2b — defence in depth.** `core/current_control.py`: `changed_paths is None` now raises `workspace_unverifiable` for **every** state, on all three paths (`_validate_worktree_currency`, `_validate_live_overlay`, `_require_clean_worktree`). A fourth skip of the same class was found and closed: an overlay that failed to recompute returned silently. `_validate_base_currency` with no ref resolver refuses instead of returning. `required` survives only where it means "a completion pointer must bind a plan".

**Evidence** (real `./shipgate`, cwd is the repository's parent):
| Scenario | `c685856a` | `be7441c3` | main |
|---|---|---|---|
| refresh through gitignored `repo/via/sib` after tracked + untracked edits | **exit 0, `review_publishable`** | exit 4, `workspace_changed (new_tool.py, notes.txt)` | exit 2 at the writer |
| refresh direct `<parent>/sib` | exit 4 | exit 4, identical | exit 2 |
| one string for `--out` and `--reports-dir` through `repo2/lnkdir -> outside` | **exit 0, `review_publishable`** | exit 4, `workspace_changed` | exit 2 |
| case-variant `--out <parent>/REPO3/rpt` | exit 3 ("path changed identity while it was read") | exit 0 `complete`; refresh `complete` | exit 2 |
| controls: plain sibling, in-repo, root, parent | unchanged | unchanged | unchanged |

**Tests.** The assertion at `tests/test_agent_control_reports_dir.py:501-503` is flipped (an in-repo spelling resolving outside now returns `None`; inside spellings assert the respelling beneath the root). Four regressions added: the link refresh, one string for both flags, the case-variant end to end (skipped on a case-sensitive filesystem), and a unit test that `changed_paths=None` refuses for `review_publishable` **and** plan-less preview pointers.

Fixture note: the tracked edit must not be a declared manifest input — editing one is caught by the recorded-input check and masks the skipped unseen-change test.

**A/B on `c685856a`** (separate worktree, exit 1): all 7 changed or new tests fail there — `assert 0 == 4` with `review_publishable` for the two link cases, `assert 3 == 0` for the case variant, `{'review_publishable': 'answered anyway', 'preview': 'answered anyway'}` for the unit test, plus the three rule assertions. The other 13 pass on both trees.

**Nit 2** done: the docstring states the final rule. Cycle-1 behaviour is preserved (all pre-existing tests in the file pass unchanged). #807 and #804 are untouched. The CHANGELOG keeps one `(#575, #785)` bullet, extended with the identity rule and the state-independent refusal.

**Verification on `be7441c3`** (all foreground): the reports-dir file; reports-dir + manifest-free rows + 5 current_control files + envelope + workspace guard + adoption walk + bootstrap; 7 verify files; 17 host files; the static-only test; `ruff`; `build-llms-full.py` (no diff); and the full suite (322 files in 10 chunks) — every run exit 0.

pengfei-threemoonslab (commented):
Independent coding-agent review/address cycle 3 (focused) at `be7441c3`: **no P0/P1/P2 — mergeable.** The cycle-2 fail-open is closed, both reproductions and the case-variant nit are fixed, and no new fail-open or merge-blocking over-refusal was found. CI: all required checks green.

**Over-refusal sweep (the new risk).** `live_workspace` returns `None` — not `changed_paths=None` — whenever the workspace is not a Git checkout or the root/HEAD cannot be resolved, and a non-completion pointer still returns early for that. So a non-Git workspace, a plan-less pointer in a non-Git directory, an unreadable Git directory and a repository with no commits are untouched. `changed_paths=None` is reachable only when `ensure_git_workspace` succeeds and `working_tree_context` raises.

Identical on branch, `c685856a`, main and 1.0.0: control, detached HEAD, `GIT_DIR` set, `core.fileMode` change (exit 4, `workspace_changed`), a non-Git workspace holding the reports directory, and a complete pointer with an unreadable tree (exit 4). The Stop hook never calls these paths, and `verify --format control` reading its own artifact is unchanged in kind.

**Identity rule.** An 18-case battery compared the helper's classification against `resolve()`-based ground truth and then handed each result to `working_tree_context`: all 18 matched on the branch — plain inside, non-existent deep components, `..`-folded, trailing slash, a reports directory that is itself a symlink to an inside target and a child of it, in-repo symlinks pointing outside, outside symlinks pointing inside, a repository under a symlinked parent, the prefix-sharing sibling `repo-reports`, the root, the parent, nested repositories, a dangling symlink, and the macOS `/tmp` → `/private/tmp` alias. On `c685856a` the in-repo-symlink cases are handed over and raise the swallowed `ConfigError` that produced the fail-open.

**Cycle-2 reproductions**
| Case | branch | `c685856a` | main / 1.0.0 |
|---|---|---|---|
| refresh through gitignored `repo/via/sib` | **4**, `workspace_changed (new_tool.py, notes.txt)` | **0, `review_publishable`** | 2 at the writer |
| refresh direct `<parent>/sib` | 4, byte-identical | 4 | 2 |
| one string for both flags through `repo/lnkdir` | 4, naming both files | **0, `review_publishable`** | 2 |
| case-variant `--out <parent>/REPO/rpt` | 0 `complete`, refresh `complete`, later edit 4 | exit 3 | 2 |

A/B: 7 of 20 tests fail on `c685856a`, including both link cases and the unit test; the other 13 pass. That matches the author's claim.

**Regressions:** in-repo `--out`, root and parent refusals, the refresh happy path and nested-project artifact paths are all unchanged (every path-bearing string across 15 artifacts byte-identical to main and 1.0.0). #785's test passes with 5 rows equal to the in-workspace run. Recorded rerun commands are identical to main. `--preview` remains #807 on all four builds.

**Nits**
1. **Borderline, not merge-blocking.** The `changed_paths is None` refusal also fires for causes unrelated to the reports directory. Concrete trigger: a repository configuring `filter.*.clean/smudge` (git-lfs) or `.gitattributes` with `filter=`, which `working_tree_context` rejects and `live_workspace` swallows. The branch then refuses `workspace_unverifiable` where main and 1.0.0 answered `review_publishable`. `verify` already fails in such repositories on every build, and refusing is the documented intent, but the message names neither the cause nor a fix. Suggested: carry the swallowed exception text on `LiveWorkspace` and append it to the three refusals. Filed as a follow-up.
2. `--out` naming the repository root through a case-variant spelling exits 2 with the generic reason instead of "cannot be the workspace root", because the overlap checks still compare lexically. Fail-closed, and main's message is no better.
3. The identity rule widens #804's reach to case-folding spellings: `--out <parent>/REPO/docs` now gives `complete`. Same class, recorded on #804.
4. `current_workspace.py:155-158` returns the unresolved `reports_dir` on the ancestor branch, the one spot that contradicts the docstring. Harmless.

**Verified with no finding:** writer/reader agreement for every spelling constructed; the exclusion can never hide a change outside the reports directory; root and ancestors still refused; dropping an exclusion only adds paths; exit codes and machine-readable error kinds unchanged; no schema, contract or check-ID surface touched; both CHANGELOG sentences literally true; docstrings accurate; `Closes #575` and `Closes #785` on separate lines.

**Runs:** reports-dir + manifest-free rows (54), current_control/envelope/workspace-guard/adoption-walk (318), verify files (245), static-only (423), `ruff` — all exit 0, plus about 40 real-CLI probes across four builds.

pengfei-threemoonslab:
**Rebased onto `6030e706`** (PR #789 / #714 merged) at head `81c18f65a6142f323cd6d567f0352b60f0ab9299`, with no code change: four CHANGELOG conflicts were resolved by hand, keeping main's #714 bullet once and this PR's `(#575, #785)` bullet in its cycle-2 wording. Against `origin/main`, `## Unreleased` gains exactly one line.

Re-verified on the rebased head (all foreground, exit 0): `test_agent_control_reports_dir.py`, `test_manifest_free_pr_rows.py`, `test_current_control*.py`, `test_workspace_input_guard.py`, `test_adoption_walk.py`, `test_bootstrap.py`, `test_verify*.py`, `test_agent_control_envelope*.py` and `tests/test_adapter_static_only.py`.

Cycle-3 nit 1 (an unreadable worktree refuses without naming the cause) is filed as #813; the case-variant spelling from nit 3 is recorded on #804.

Issue #803, opened 2026-09-15
Base-cache namespace checks are check-then-use on the Windows lexical fallback


pengfei-threemoonslab:
## Problem

#638 anchors every base-cache read, publication and prune at the metadata directory Git selected, and refuses a link, non-directory or unopenable component below it. On POSIX that lookup, publication and prune are descriptor-relative, so a component replaced after its directory is opened cannot redirect them.

One window remains, and it is the platform fallback:

**The Windows lexical fallback is check-then-use.** Where descriptor-relative operations are unavailable (`os.open`/`mkdir`/`rename`/`unlink`/`stat` without `dir_fd`, `scandir` without fd support, or `shutil.rmtree` that does not avoid symlink attacks), `_base_cache_directory` inspects each namespace component (`inspect_lexical_path_identity`: links, junctions, reparse points, aliases, non-directories) and then performs pathname operations. A component replaced between that inspection and the operation is not detected. Reads on that path go through `read_identity_bound_bytes` from the anchor, and pruning skips links and junctions, but neither closes the window.

Scope: this needs a local process that can write the Git metadata directory while `verify` runs. Such a writer can already rewrite a cache entry in place with a matching checksum. The cache claims no authenticated source provenance (`docs/verification-reproducibility.md`), so this is not a new trust boundary; it is the part of the namespace guarantee #638 does not provide on Windows. There is no settled-link, repository-controlled or remote reproduction.

## Closed by PR #805 (review cycle 1)

Both same-run pathname-reopen windows originally filed here are gone; the evidence below is kept because it is what the regressions now assert.

`_prepare_base_report` settles the cache in one no-follow open per run. A hit yields the validated `report.json` bytes themselves and the capability lock read through the *same* directory handle, so the report and the lock can no longer come from two directory generations. Those bytes are written into a directory private to the run, and the head scan's diff reference, the `verification-base-report.json` export, `_gap_provenance_note`, `_observed_subjects_from_report_json` and the capability review artifacts open only that copy. A cold run reads back its own copy rather than the entry it has just published.

The original reproduction, on `claude/issue-638-cache-namespace` at `72a46061` (macOS/POSIX), replaced `base-scans` with a link to a copy whose `report.json` says `effective_policy.ci_mode: advisory` (checksum recomputed), hooked after validation and, separately, after the cold store. The fixture is `tests/test_verify_weakening.py::_weakened_repo`, where an unraced run reports `ci_mode_weakened`.

[...]

PR #805, opened 2026-09-15
Keep base-cache access inside its Git metadata namespace (#638)

### LINKED ISSUE #638: Base-cache recovery can follow a linked parent directory outside its cache namespace
## Problem

While reviewing #596's cache recovery, an independent coding-agent probe confirmed that a base-cache **directory** link can redirect report/checksum publication outside the intended cache namespace. This is separate from engine compatibility and from the individual report/checksum file-link repair in #596.

The isolated fixture creates `metadata/agents-shipgate/base-scans/entry` as a directory symlink to another scratch directory. Current production `_cache_report_valid` refuses the linked cache root, but `_copy_report_to_cache` follows its parent while creating/replacing `report.json` and `report.sha256`. Both external files are replaced and the directory link remains.

The same probe executed the archived helpers from `604649e5fc259c058aaeca1c2829b7c89d5de174`: the old path resolution accepts a valid external entry, and a bad checksum then triggers the same external writes. This is a pre-existing cache namespace defect, not a consequence of keying by engine identity. Evidence is actual helper execution with successful process exit, **not** a full CLI/verifier reproduction. It requires a writable local Git-cache layout; no remote PR-controlled exploit has been demonstrated.

Relevant code: `cli/verify/orchestrator.py` (`_cache_report_path`, `_cache_report_valid`, `_copy_report_to_cache`, `_write_capability_lock_to_cache`, `_prune_base_scan_cache`) and `cli/verify/git.py::git_path`.

## Proposed direction

Anchor cache access at Git's selected metadata directory and preserve the lexical cache namespace through lookup, creation, publication and pruning. Refuse linked/aliased non-directory ancestors before writing; do not replace a refusal with a resolved external pathname. Review concurrent parent replacement separately from settled-link rejection and publish the supported guarantee precisely. Preserve legitimate worktree-specific Git directory selection and bounded cache pruning.

## Acceptance

- [ ] A full verifier regression with a linked cache entry/ancestor leaves the external directory and its contents unchanged and returns actionable cache-unavailable guidance.
- [ ] Report, checksum and capability-lock publication share the same cache-directory boundary; pruning cannot traverse an external target.
- [ ] Normal cold/warm caching and legitimate Git worktree layouts retain their expected behavior.
- [ ] Settled aliases, wrong kinds and unreadable parents have bounded refusal tests; any remaining concurrency limitation is explicit and separately tracked.

## Sequencing

P2, deferred from #596's selected v1.0 engine-compatibility repair as requested for newly discovered independent work. No claim of authenticated cached source provenance or of cache-directory containment should be added by #596. Reassess release priority if a repository-controlled or ordinary-install reproduction establishes broader reach. Refs #596, #572.

pengfei-threemoonslab:
## Root cause

The base-scan cache lives at `<git metadata>/agents-shipgate/base-scans/<key>/`. `_cache_report_path` resolved only Git's metadata directory. Every later operation then used pathnames below it, and those follow links. `_cache_report_valid` refused a linked *entry* (its identity session lstat'ed the entry root) but followed a linked *ancestor*. `_copy_report_to_cache` and `_write_capability_lock_to_cache` ran `mkdir(parents=True)` and wrote temporaries into `cache_report.parent`. `_prune_base_scan_cache` used `iterdir` and `rmtree` on `cache_report.parents[1]`. A directory link at the entry, `base-scans` or `agents-shipgate` therefore redirected publication and pruning outside the namespace. The worktree static snapshot then refused the linked path at artifact export, so verify also ended in an internal error.

## Anchoring rule

- **Anchor:** the metadata directory Git selects (`git rev-parse --git-path ""`, resolved once): `.git`, a linked worktree's own Git dir, or wherever a `.git` link or `gitdir:` file points. Links at or above it are Git's selection and keep working.
- **Namespace:** `agents-shipgate/base-scans/<key>/` below the anchor is lexical. No component is followed.
- **One boundary:** `_base_cache_directory(anchor, components, create=...)` is the only way cache files are touched. Report, checksum and capability-lock reads and publication, and pruning, all use its handle.
  - **POSIX:** each component is opened with `O_NOFOLLOW | O_DIRECTORY` relative to its parent descriptor. With `create`, `mkdir(dir_fd=)` runs inside that same parent first.
  - Files are read with `O_NOFOLLOW | O_NONBLOCK` relative to the entry descriptor. They must be regular and singly linked, with before/after identity checks.
  - Files are published through an `O_CREAT | O_EXCL` temporary, renamed with `os.replace(src_dir_fd=, dst_dir_fd=)` into the same descriptor.
  - Pruning uses `scandir(fd)`, skips links, and removes trees with `shutil.rmtree(name, dir_fd=)`.
- **Reuse, not a second implementation:** the directory walker and fd-relative reader are factored out of `read_regular_file_beneath` (`open_directory_beneath`, `read_regular_file_at`), which now composes them. `replace_file_at` is the fd-relative writer beside them.
- **Unavailable cache:** a namespace component that is a link, is not a directory (regular file, FIFO) or cannot be opened or created makes the cache unavailable for that run. The base is regenerated from Git. Nothing is read, written or pruned through the component, the comparison completes (`base_status: succeeded`), and a base note names the path:

  > Base-scan cache unavailable: `<path>` is a symbolic link. This run neither read nor wrote the cache and regenerated the base report from Git. Make that path a plain directory (or remove it so verification recreates it), then rerun verification to reuse cached base reports.

- **Unchanged:** a failure *inside* a usable entry, such as a directory at `report.json`, keeps the existing "Could not store the regenerated base report" route (`test_directory_at_cache_report_is_preserved_and_routes_to_repair`).

## Every path fixed

| Path | Before | After |
|---|---|---|
| Lookup / validity (`_cache_report_valid`, probe in `_prepare_base_report`) | `IdentityBoundReadSession(entry path)`: linked ancestor followed | one anchored `create=False` open (`_probe_base_cache`); report, checksum and lock read from the same entry descriptor, and the validated bytes are what it returns |
| Report + checksum publication (`_copy_report_to_cache`) | `mkdir(parents=True)` and temporaries through the path | `replace_file_at` inside the anchored entry descriptor |
| Capability-lock publication (`_write_capability_lock_to_cache`) | same | same handle as report and checksum |
| Capability-lock read (`_cached_capability_lock`) | `read_regular_file_beneath(entry path)`: ancestors followed | the same handle the report was validated through; missing lock note preserved |
| Pruning (`_prune_base_scan_cache`) | `iterdir` + `is_dir()` + `rmtree(path)` on `cache_report.parents[1]` | anchored `base-scans` descriptor, links skipped, `rmtree(dir_fd=)`; a linked or wrong-kind `base-scans` prunes nothing |
| Store after an unavailable probe | wrote through the link | no store and no prune |
| Every same-run read of the base report | reopened the cache pathname | one copy private to the run, written from the validated bytes; warm, cold and unavailable alike |

## A/B evidence

The reproduction test, `tests/test_base_cache_namespace.py`, drives `verify` through the CLI and uses no private helper. It was copied into a separate `git worktree add <scratch> origin/main` tree at `7ac1c456` and run there with that tree's `--rootdir`, `-c` and `PYTHONPATH`, so the root conftest imports main's `src`.

- **origin/main:** `test_linked_cache_component_is_never_read_or_written_through` failed for all five scenarios (four with `files outside the cache namespace changed`); controls passed (three linked leaves, linked worktree, linked `.git`). Pytest exit 1.
- **This branch:** the same file passes (exit 0).

A direct `run_verify` probe on each tree gave:

| Scenario | origin/main | this branch |
|---|---|---|
| entry → external copy, bad checksum | external `report.json`, `report.sha256`, `capabilities.lock.json` replaced; then `ValueError: path contains a symlink or reparse point` | exit 0, succeeded; external unchanged; link intact |
| `base-scans` → external (warm) | external entry reused as a cache hit; then the same `ValueError` | exit 0, succeeded, base rescanned; external unchanged |
| `base-scans` → external with 16 foreign dirs (cold) | external `<key>/` entry created **and `foreign-00` deleted by prune**; then `ValueError` | exit 0, succeeded; external unchanged |
| `agents-shipgate` → external (cold) | external `base-scans/<key>/` tree created; then `ValueError` | exit 0, succeeded; external unchanged |

## Acceptance (#638)

- [x] **Full verifier regression.** `verify` runs through `CliRunner` with a linked entry (bad checksum; valid report), a linked `base-scans` (warm; cold with foreign entries) and a linked `agents-shipgate` (cold). The external tree is unchanged by kind, inode, link count, mtime and bytes, and the link survives two runs. The run completes by recomputing the base (`ci_mode_weakened` still found, one more base scan per run), and the note names `<path> is a symbolic link` and says to rerun verification. Two of the scenarios also run through the lexical (no-dir-fd) boundary.
- [x] **One boundary for report, checksum and lock.** All three use one `_base_cache_directory` handle. Pruning never traverses an external target: `test_prune_never_traverses_a_linked_entry_or_namespace` covers a linked entry and a linked `base-scans`, for both boundaries.
- [x] **Normal caching and worktree layouts unchanged.**
  - Cold and warm runs are unchanged: `test_warm_hit_answers_exactly_like_the_cold_run` shows identical `verification-base-report.json` bytes, `request_id`, `input_set_id`, `decision_id`, `base_tree_sha` and `merge_verdict`, and entry files and checksum as before.
  - Existing `test_base_cache_engine_identity.py` and `test_verify.py` prune tests pass.
  - `test_linked_worktree_caches_under_its_own_git_dir`: `git worktree add`, Git dir ≠ common dir, warm hit, entry under the worktree Git dir.
  - `test_git_selected_metadata_directory_may_itself_be_a_link`: symlinked `.git`.
- [x] **Bounded refusal tests.**
  - Settled aliases: directory links, file links for `report.json`, `report.sha256` and `capabilities.lock.json`, and a hard-linked `report.json`.
  - Wrong kinds: a regular file at `agents-shipgate`, `base-scans` or the entry; a FIFO at `agents-shipgate` or `base-scans`, in a bounded child process, both lookup and create.
  - An unreadable parent: `base-scans` mode 0, skipped as root.
  - The remaining concurrency limitation is stated below and in the docs, and tracked in #803.

## One directory generation per run (review cycle 1)

The anchoring rule above covers a link in place when a cache directory is opened. Review cycle 1 raised the other half as a P2: after validation the run went back to the *pathname*. `_prepare_base_report` returned `cache_entry.report`, and the head scan's `diff_from_path`, the `verification-base-report.json` export, `_gap_provenance_note`, `_observed_subjects_from_report_json` and the capability review artifacts all reopened it; the capability lock was read in a second anchored open, so the report and the lock could come from two directory generations. Swapping `base-scans` for a link to a forged copy after that point removed `ci_mode_weakened` in a `git worktree add` layout and raised an internal error in an ordinary checkout — on cold runs as well as warm.

The cache is now settled by **one** no-follow open per run:

- That open decides between a hit, a cold run and an unavailable cache, and a hit returns the validated `report.json` **bytes** together with the capability lock read through the same directory handle (`_probe_base_cache` → `_BaseCacheProbe`). The report and the lock cannot disagree about which directory they came from.
- Those bytes are written into a directory private to the run (`run_report_dir()`, owned and cleaned up by `run_verify`), and every later consumer opens only that copy. A cold run reads back its own copy rather than the entry it has just published — the entry is a pathname like any other.
- No namespace pathname is opened after the cache step, so a component replaced *during* the run cannot be read back.

### The published reference stays stable

`run_scan` takes a `diff_from_display_path` alongside `diff_from_path`; `_load_diff_references` spells it with a new `_anchored_display_path`, which does not resolve an already-anchored path (resolving would follow the very link the boundary refused). `verify` passes the entry's cache location. So `tool_surface_diff.base.path` and `action_surface_diff.base.path` name the cache entry for a warm run, a cold run and a run the cache was unavailable to alike, and never a temporary directory.

**`report.json` is byte-identical to main for ordinary warm and cold runs.** One weakened repository was built once, then run against a `42110535` worktree, a `72a46061` worktree and this head at the same path. Masking the three engine-derived identity fields (`engine_requirement_id`, `request_id`, `decision_id`) and the cache key derived from them — all necessarily different between two source trees — the diff is **0 lines** for both cold and warm, against main and against `72a46061`. Cold and warm are identical within each tree.

### Race regressions

`tests/test_base_cache_namespace.py` gains, each parametrized over an ordinary checkout and a `git worktree add` layout:

- `test_namespace_swapped_after_validation_is_never_consumed` — hooks `_prepare_base_report`, so the swap lands after validation and before the head scan.
- `test_namespace_swapped_after_the_cold_store_is_never_consumed` — hooks `_prune_base_scan_cache`, which runs immediately after publication, so the swap lands between the cold store and every read-back.

Both forge a copy of the cache whose base says `ci_mode: advisory` with a recomputed checksum, then assert the run exits 0 with no internal error, still finds `ci_mode_weakened`, matches an unraced control on `request_id`, `input_set_id`, `decision_id`, `base_tree_sha`, `merge_verdict` and `verification-base-report.json` bytes, and names neither the forged tree nor a temporary directory in `report.json`.

| Case | `72a46061` | This head |
|---|---|---|
| ordinary, after validation | exit 4, `path contains a symlink or reparse point` | exit 0, `ci_mode_weakened`, ids match |
| ordinary, after the cold store | exit 4, same | exit 0, `ci_mode_weakened`, ids match |
| `git worktree add`, after validation | exit 0, findings `[]` — forged base consumed | exit 0, `ci_mode_weakened`, ids match |
| `git worktree add`, after the cold store | exit 0, findings `[]` — forged base consumed | exit 0, `ci_mode_weakened`, ids match |

All four fail on `72a46061`, run from a separate worktree with that tree's `--rootdir`, `-c` and `PYTHONPATH`.

### The other four review notes

1. **A cache-unavailable run published a random temporary path.** Covered by the display-path change. `test_an_unavailable_cache_publishes_the_same_report_every_run` asserts two such runs produce identical `report.json` bytes and that the published reference is the cache location, not the external target of the refused link and not a temporary directory. Fails on `72a46061` (`report.json` bytes differ).
2. **Modes.** `replace_file_at` takes the creation mode. `report.sha256` and `capabilities.lock.json` are published 0600 again, `report.json` 0644 — what main produces. Observed end to end: main `0600/0644/0600`, `72a46061` `0644/0644/0644`, this head `0600/0644/0600`. `test_published_cache_files_keep_their_modes` pins them for both boundaries (with the umask pinned, since a creation mode is masked), and the boundary unit test asserts the mode it asked for is the mode it got. Fails on `72a46061`.
3. **`_cache_report_valid` is gone.** Production and tests use the one boundary, which now returns bytes: `_probe_base_cache(entry).report is not None` in `test_base_cache_engine_identity.py` and `test_operation_attribution_verification.py`. The lock reader takes the open handle (`_cached_capability_lock(cache, entry)`); `test_nonregular_evidence_readers.py` opens the same boundary itself, so its FIFO, directory, symlink and non-UTF-8 refusals are unchanged.
4. **#803 is rescoped** to the Windows lexical fallback's check-then-use window only, with the race evidence kept as the description of what the new regressions assert, and retitled to match.

One behaviour change worth naming: `tests/test_verify.py::test_verify_successful_base_scan_feeds_head_diff_from` asserted the base report still existed *after* the run. It is run-private now, so the stub records whether the scan could read it at the moment it was called.

## Remaining limitation (tracked in #803)

One window is left, and it is the platform fallback: where descriptor-relative operations are unavailable (Windows), `_base_cache_directory` inspects each namespace component and then uses pathnames, so a replacement in between is not detected. #803 now covers that and nothing else. This PR adds no claim of authenticated cached source provenance.

## Windows

The descriptor primitives are POSIX-only. Where `os.open` / `mkdir` / `rename` / `unlink` / `stat` lack `dir_fd`, `scandir` lacks fd support, or `rmtree` does not avoid symlink attacks (Windows), `_base_cache_directory` inspects each namespace component with `inspect_lexical_path_identity`. It refuses links, junctions, reparse points, aliases and non-directories, then uses pathname operations as before.

- Reads use `read_identity_bound_bytes` from the anchor (lexical checks and singly-linked files).
- Pruning skips links and junctions.
- Settled links and junctions are now refused where they were followed. Caching still works.
- The cached capability lock now loads on Windows through that reader; after #790 it always fell back.
- Nothing is worse than before. No Windows run was made: CI runs only the launcher job on Windows. The lexical boundary is exercised on POSIX by forcing it in tests.
- New POSIX-only tests carry the file's `os.name` skip, `needs_fifo` and not-root markers.

## Other notes

- No schema, contract, check-ID or JSON-field change.
- `read_regular_file_beneath` keeps its messages; its symlinked-intermediate refusal now reads `<path> is a symbolic link` or `is not a directory` after the same `could not safely read …:` prefix.
- The report read no longer counts a 64-entry directory budget, because it reads descriptor-relative instead of scanning the entry directory. The 64 MiB and 128-byte bounds and the singly-linked-file rule are kept.
- Docs: `docs/verification-reproducibility.md` has a new § Base-scan cache namespace; `STABILITY.md` has a one-sentence guarantee.
- CHANGELOG: one `## Unreleased` bullet ending `(#638)`.

## Tests

The branch was rebased onto `42110535` (#788); `origin/main` had not moved when review cycle 1 landed, so no second rebase was needed. Review-cycle runs are on head `083b2625`.

| Command | Head | Exit |
|---|---|---|
| `ruff check` on the 10 changed Python files | `083b2625` | 0 |
| `pytest tests/test_adapter_static_only.py` (line-number pins) | `083b2625` | 0 |
| full suite `-n auto -m "not perf" --ignore=tests/test_adapter_static_only.py`, 308 files in six literal-path chunks | `083b2625` | 0, 0, 0, 0, 0, 0 |
| `test_base_cache_namespace.py test_base_cache_engine_identity.py test_nonregular_evidence_readers.py test_current_control.py test_verify_weakening.py test_operation_attribution_verification.py test_verify.py` plus the scan diff tests (`test_action_surface_diff.py test_tool_surface_diff.py test_capability_diff.py test_diff_input_status.py`) | `083b2625` | 0 |
| A/B: the seven new tests on a separate `72a46061` worktree | `72a46061` | 1 (all seven fail) |
| A/B: `tests/test_base_cache_namespace.py` on a separate `origin/main` worktree | `42110535` | 1 (expected) |

The earlier cycle-0 runs on `72a46061` (`ruff`, static-only, six full-suite chunks) all exited 0.

Closes #638


pengfei-threemoonslab (commented):
Independent coding-agent review/address cycle 1 at `72a46061`. **One P2, four nits.** No fail-open in the new lookup, publication or pruning boundary.

**P2: the base report is reopened by pathname later in the run, on cold runs as well as warm** (`cli/verify/orchestrator.py`)

- `_prepare_base_report` returns `cache_entry.report` as a pathname at the warm hit (1478–1485), the operation-callback warm hit (1553–1556) and the cold store (1677).
- The head scan's `diff_from_path`, the `_write_artifacts` export, `_gap_provenance_note` and `_observed_subjects_from_report_json` all reopen it by path.
- The lock is read in a second anchored open (`_load_cached_capability_lock`, 2028), so the report and the lock can come from different directory states.

Probe using the `_weakened_repo` fixture: `base-scans` is swapped for a link to a copy with `ci_mode: advisory` and a matching checksum.
| Layout | Result |
|---|---|
| warm, ordinary checkout | exit 4 (internal error) |
| warm, `git worktree add` | exit 0, `base_status: succeeded`, findings `[]` (the control run had `ci_mode_weakened`); `tool_surface_diff.base.path` points at the forged tree |
| cold, `git worktree add` (swap after the store) | exit 0, findings `[]`; not mentioned in #803 or the docs |

The fix is small. The prototype kept the `report.json` bytes that validation already read and wrote them into the run-owned `uncached_report_dir()`. With that, both layouts kept `ci_mode_weakened`, and `request_id`, `input_set_id`, `decision_id` and the `verification-base-report.json` bytes matched the unraced control.

Suggested fix:
1. Return the validated bytes, and load the lock from the same handle.
2. Warm and cold runs both use a run-private copy.
3. Pass a stable display path through `run_scan` into `cli/scan/diffs.py:40`; `load_tool_surface_diff_reference` already accepts `display_path`.
4. Add a race regression for both layouts.
5. Trim #803 to the Windows check-then-use fallback.

Rated P2 rather than P1: the race needs a local writer active during the run, which could already rewrite the entry in place, and no source provenance is claimed.

**Nits**
1. The cache-unavailable route writes a random, already-deleted temp path into `tool_surface_diff.base.path` and `action_surface_diff.base.path`, so `report.json` bytes differ between runs while the ids stay stable. The display-path change fixes this.
2. `report.sha256` and `capabilities.lock.json` went from 0600 on main to 0644, because `replace_file_at` creates files with 0o644.
3. `_cache_report_valid` (orchestrator.py:1980) is now used only by tests.
4. #803 should name the cold store path.

**Verified with no finding**
- **Reader refactor** (`open_directory_beneath` + `read_regular_file_at`): same behaviour as main.
  - `O_NOFOLLOW` on root and parents; `O_NONBLOCK`/`O_BINARY` on the leaf;
  - size cap before and after, fstat identity, `S_ISREG`; `single_link` off for existing callers;
  - fds closed on every path; a FIFO parent fails without blocking.
  - The only change is the documented message for a linked intermediate directory.
- **`replace_file_at`:** `O_CREAT|O_EXCL|O_NOFOLLOW` in the dir fd, partial writes handled, rename within the same fd, temporary unlinked on failure, no fsync on either side.
- **Pruning:** links and files are skipped; `rmtree(dir_fd=)` is used only when `avoids_symlink_attacks` is true, and a link swapped in after listing is not followed. Mount points are crossed as on main.
- **Git layouts:** `--git-path` resolves correctly for the main checkout, a worktree, a submodule `gitdir:` file and `GIT_DIR`.
- **Cold vs warm:** ids and report bytes are identical.
- **Windows fallback (reasoned, not run):** a cold create works; links, junctions and reparse points are refused.
- **Docs:** one `(#638)` CHANGELOG bullet and no provenance claim.

**Runs:** base cache namespace, engine identity, nonregular readers, current_control, verify weakening, operation attribution, `test_verify`, static-only and `ruff` all exit 0. A/B from a separate `42110535` worktree: all 7 `never_read_or_written_through` cases fail and the 7 controls pass.

**CI:** all pass; release-tag-consistency skipped.

pengfei-threemoonslab:
**Addressed review cycle 1.** New head `083b2625ca578e9b1ef6c6dc137d9b47db8b84dd` (no rebase needed; `origin/main` is still `42110535`). CI: 11 pass, 1 skipping.

**P2 — the base report is no longer reopened by pathname.**

`_prepare_base_report` settles the cache in **one** no-follow open (`_probe_base_cache` → `_BaseCacheProbe`). A hit returns the validated `report.json` bytes plus the capability lock read through the *same* directory handle (`_cached_capability_lock(cache, entry)` replaces the second anchored open). Those bytes go into a run-owned directory (`run_report_dir()`, renamed from `uncached_report_dir` because every route now uses it), and that copy is what the head scan's `diff_from_path`, the `_write_artifacts` export, `_gap_provenance_note`, `_observed_subjects_from_report_json` and the capability review open. The cold path reads back the bytes it just stored, never the entry it published.

Race regressions, each parametrized `ordinary`/`worktree` (one hooks `_prepare_base_report`, the other `_prune_base_scan_cache`, which runs right after publication):
| Case | `72a46061` | `083b2625` |
|---|---|---|
| ordinary, after validation | exit 4 ("path contains a symlink or reparse point") | exit 0, `ci_mode_weakened`, ids match the control |
| ordinary, after cold store | exit 4 | exit 0, ids match |
| `git worktree add`, after validation | exit 0, **`ci_mode_weakened` gone** (forged base consumed) | exit 0, `ci_mode_weakened`, ids match |
| `git worktree add`, after cold store | exit 0, forged base consumed | exit 0, ids match |

All four fail on `72a46061` from a separate worktree.

**Display path and `report.json` bytes.** `run_scan` takes `diff_from_display_path`; `_load_diff_references` spells it with a new `_anchored_display_path` that deliberately does **not** `resolve()` — resolving would follow the very link the boundary refused and would publish the forged tree. Running one weakened repository at the same path against `42110535`, `72a46061` and `083b2625`, and masking the three engine-derived fields plus the cache key derived from them, the `report.json` diff is **0 lines** for cold and warm, both main-vs-pre and main-vs-branch. Cold and warm are byte-identical within each tree.

**Nits**
1. **No random temp path:** a new test asserts two cache-unavailable runs produce identical `report.json` bytes and that the surface-diff base paths name the cache entry, not the link target or the temp directory. It fails on `72a46061` (bytes differ at offset 11493).
2. **Modes:** `replace_file_at` takes `mode`. Observed end to end (lock / report / checksum): main `0600/0644/0600`, `72a46061` `0644/0644/0644`, `083b2625` `0600/0644/0600`. A test pins both boundaries under a fixed umask; the lexical branch guards `getattr(os, "fchmod", None)` for Windows.
3. **`_cache_report_valid` removed:** `_validated_cache_report(cache)` returns bytes, and tests use `_probe_base_cache(entry).report is not None`. `test_nonregular_evidence_readers.py` opens the boundary itself and keeps its FIFO, directory, symlink and non-UTF-8 refusals.
4. **#803 rescoped** to the Windows lexical check-then-use window, retitled accordingly, with a "Closed by PR #805" section keeping the race evidence.

**Behaviour change worth noting:** `test_verify.py::test_verify_successful_base_scan_feeds_head_diff_from` asserted the base report still existed after the run. It is run-private now, so the stub records `diff_from_readable` at call time.

**Docs:** `docs/verification-reproducibility.md` § Base-scan cache namespace is rewritten (one directory generation per run, the run-private copy, the stable published reference, Windows as the single remaining #803 limit, and no provenance claim). The CHANGELOG keeps one `(#638)` bullet. STABILITY.md is unchanged and still correct.

**Tests on `083b2625`**
| Run | Exit |
|---|---|
| `ruff` on 10 changed files | 0 |
| cache namespace, engine identity, nonregular readers, current_control, verify weakening, operation attribution, verify, and 4 scan-diff files | 0 |
| Full suite, 308 files in 6 foreground chunks | 0, 0, 0, 0, 0, 0 |
| A/B: the 7 new tests on a separate `72a46061` worktree | 1 (all 7 fail) |

Windows was not executed; the lexical boundary is exercised on POSIX by forcing the descriptor-relative path off.

pengfei-threemoonslab (commented):
Independent coding-agent review/address cycle 2 (focused) at `083b2625`: **no P0/P1/P2 — mergeable.** The cycle-1 P2 is closed and all four cycle-1 nits are addressed. Five nits remain, all cosmetic or hardening. CI: 11 pass, 1 skipping.

**The race is closed, checked at four moments the cycle-1 tests did not target.** Independent probes, each parametrized ordinary / `git worktree add`, swapping `base-scans` for a link to a checksum-consistent copy whose `ci_mode` is `advisory`:
| Swap moment | `72a46061` | `083b2625` |
|---|---|---|
| between the report read and the lock read | n/a | pass, both layouts |
| between the probe and the run-private copy | n/a | pass, both layouts |
| during the cold base scan | **fail** ("temp path published") | pass |
| immediately before `_write_artifacts` | **fail** — ordinary: exit 4; worktree: exit 0 with `request_id` diverged, so the export re-read the forged tree and it reached run identity | pass; `verification-base-report.json` has `ci_mode: strict` |

The author's seven tests also fail on `72a46061` and pass here.

**One directory generation, confirmed by instrumenting `_base_cache_directory`:** three call sites, all inside the cache step; warm = 1 open, cold = 3 (probe, store, prune), and no read afterwards. `_cached_capability_lock` takes the open handle, so report and lock cannot come from two generations, and its `except` cannot be misreported as "cache unavailable".

**`run_report_dir()` is created safely:** `mkdtemp`, measured 0700, unique per run, removed on success and on a forced head-scan exception (not on SIGKILL, same as the pre-existing head temp).

**Nothing broke.** `verification-base-report.json` is still exported and still validated from the persisted copy; all 20 artifacts are produced; no consumer outside `tests/` reads the cache path as a file; **no temp path appears in any artifact** (zero hits, where `72a46061` published one); a warm hit still avoids re-scanning (3 runs, 1 base scan, identical `input_set_id`).

**`report.json` bytes are identical to main in every cell** across four layouts (ordinary, `git worktree add`, `GIT_DIR`, submodule), cold and warm, masking only time, engine version, 64-hex identifiers, the engine-derived cache key and the fixture root. Cold and warm are byte-identical within each tree, and the published `base.path` matches main in all four.

**Nits**
1. **`keep_for_this_run` (`orchestrator.py:1493`) is the one write left that follows a symlink:** `destination.write_bytes(data)`. Planting `report.json` as a link to another file clobbers that file. The directory is a fresh 0700 `mkdtemp`, so this needs a same-uid process racing an unguessable path — outside #803's model, and strictly harder than the `.git` entry this PR protects. Still worth closing in a change whose thesis is that nothing is written through a link: `os.open(dir_fd=…, O_CREAT|O_EXCL|O_NOFOLLOW)` or `replace_file_at`.
2. **The lexical (Windows) replace does not produce the descriptor boundary's mode.** `os.open(…, mode)` is umask-masked, `fchmod` is not: at umask 0o077 the descriptor path gives report 0600 while the lexical path gives 0644. The comment at `verification_identity.py:996` overstates it.
3. **`docs/verification-reproducibility.md:283`** says the cache is opened once per run; cold runs open it three times. The guarantee two sentences later (no namespace pathname is reopened after the cache step) is the true one.
4. **When the cache is unavailable, the published `base.path` resolves into the refused link.** The string is deliberately the ordinary one, which is what keeps bytes reproducible, but `.resolve()` lands in a tree the engine never read. Worth a clause saying the reference is a location, not a readable source.
5. `ruff format --check` would reformat `orchestrator.py`; not enforced in CI and main is unformatted too.

**Verified with no finding:** #577's refusals intact; pruning untouched and still bounded, and the just-stored entry is newest so the following prune cannot delete it; the forced lexical boundary creates the cache cold, keeps the warm hit and refuses a settled link with the external tree unchanged; `UnsafeDirectoryComponent` is an `OSError`, so the probe's typing is sound; the return-arity change reached every caller; the early-exit `_write_artifacts` call sites pass no base path; one `(#638)` bullet; STABILITY unchanged and correct; the documented `git rev-parse --git-path` command matches the implementation; #803 as rescoped is accurate and every row of its evidence table reproduces.

**Runs:** cache namespace + engine identity + nonregular readers (64), verify + weakening + operation attribution (155), surface/diff/static-only (599), `ruff`, `doctor`, the new race probes on both heads, and a 4-layout × 3-tree × cold/warm A/B — all as stated.

pengfei-threemoonslab:
**Addressed review cycle 2 and rebased onto `ba10b486`** (PRs #789/#714 and #801/#575 merged since). New head `6cb069c58203b1b4770ff507af40b3e72ea27599`; the rebase had no conflicts, and `## Unreleased` keeps one `(#638)` bullet beside main's.

**Nit 1 (fixed): the one write left that followed a link.** `keep_for_this_run` (`orchestrator.py`) published the validated bytes with `write_bytes`, so a link planted at `report.json` inside the run's own directory was followed and its target clobbered. Where descriptors allow it, the copy now goes through `replace_file_at` on an open handle for that directory, at the same `0600` the cache uses; the lexical fallback keeps its pathname write, as on Windows today.

A regression plants that link and asserts the target survives. It fails without the fix: `assert b'{\n  "schem…' == b'untouched\n'` — the victim is overwritten with the base report.

**Nit 2 (fixed, wording):** the lexical branch's comment claimed it publishes "the same mode the descriptor boundary would". `fchmod` ignores the umask that masks the descriptor creation, so under a restrictive umask it publishes a wider file. The comment now says it is a best-effort widening.

**Nit 3 (fixed, docs):** `docs/verification-reproducibility.md` said the cache is opened once per run; a cold run opens it three times (probe, store, prune). It now says the cache is opened once to decide the run, and keeps the true guarantee — no namespace pathname is reopened after the cache step.

**Nit 4 (fixed, docs):** the same section now says the published base reference is a location rather than a source, and that with an unavailable cache that location resolves through the refused component. That is what lets two such runs agree byte for byte.

**Nit 5** (`ruff format`) left alone: CI runs `ruff check`, and main is unformatted too.

**Verification on the rebased head** (all foreground, exit 0): `test_base_cache_namespace.py` (including the new test), `test_base_cache_engine_identity.py`, `test_nonregular_evidence_readers.py`, `test_verify.py`, `test_verify_weakening.py`, `test_operation_attribution_verification.py`, `test_current_control.py`, `test_agent_control_reports_dir.py`, `tests/test_adapter_static_only.py`, and `ruff check` on the changed files. The last two overlap with #801, which also touched the orchestrator.

Issue #802, opened 2026-09-15
Token-shaped workflow job names are published raw in host grants


pengfei-threemoonslab:
**What happens**

Host grants publish a GitHub workflow job id verbatim wherever they name a job. A job id shaped like a credential (`ghp_…`, `AKIA…`) therefore reaches the inventory, saved baselines, drift payloads, `diff`/`check`/`verify` rows and the PR comment unredacted.

GitHub accepts any job id that starts with a letter or `_` and contains only alphanumerics, `-` and `_`, so these are valid ids. PR #788 (#771) runs step `id`/`name` and step `uses:` through `privacy.redact_text`; the job id printed beside them is not.

**Evidence**

- **Published 1.0.0 already does this.** `_workflow_permissions(value, job)` returns `{"job": job, …}` with `job = str(job_name)`, so `permission_contexts[].job` is raw (`src/agents_shipgate/core/host_grants.py` at `v1.0.0`, lines 1098–1113, called at 1141). `reusable_calls[].job` (1152) and the `write_scopes` / `effective_write_scopes` prefixes are raw too.
- **PR #788 adds sinks with the same raw value.**
  - `step_actions[].job`, set in `_workflow_grant` → `_step_action` / `_unreadable_step` (`host_grants.py` ~1270–1280 on that branch).
  - The `job/step` row labels and the `why` text for changed and moved references (`core/capability_diff_rows.py`, `_step_action_value` and the `why` builder).
- **Probe on the PR branch**, job id `ghp_` followed by 36 `J`:

  ```
  permission_contexts: [{"job": "ghp_JJJJ…", "state": "explicit", "permissions": {"contents": "read"}}]
  step_actions: [{"job": "ghp_JJJJ…", "step": "steps[0]", "uses": "actions/checkout@main", "form": "remote", …}]
  row.why: a step's action reference changed (ghp_JJJJ…/steps[0]); it names different code to run …
  ```

**Acceptance**

- **Redacted everywhere, one way.** A job id goes through `privacy.redact_text` (then the host sanitizer, as step text does) in every field that publishes it:
  - `permission_contexts[].job`, `reusable_calls[].job`, `step_actions[].job`;
  - the `write_scopes` and `effective_write_scopes` prefixes;
  - row labels and `why` in `diff`, `check`, `verify`, drift and the PR comment.
[...]

Issue #772, opened 2026-09-14
An MCP server URL's path is outside its change digest: `/read` → `/admin` produces no row


pengfei-threemoonslab:
## Observed

On `main` at `0bab265d`, a remote MCP server's URL path changed and nothing else:

```json
{"mcpServers":{"remote":{"type":"http","url":"https://mcp.example.com/read"}}}
{"mcpServers":{"remote":{"type":"http","url":"https://mcp.example.com/admin"}}}
```

`shipgate diff --workspace . --base HEAD --json` returns `comparison_status: comparable`, no `incomparable_reasons`, no `unchanged_limits`, and no rows. A host change and a query change on the same server each produce a `widened` row.

## Why

#723 added the URL query to the change digest so removing `read_only=true` is visible. `_url_capability_parts` (`src/agents_shipgate/core/host_grants.py`) leaves the path out on purpose: a webhook-style path can itself be the secret, and rotating one must stay quiet (`test_drift_all_env_header_value_rotation_quiet_but_key_addition_fires`). A digest cannot tell a capability path from a secret one. At the time of that probe the trade-off was recorded only in the function's docstring and on the closed #723. #775 subsequently named the limit on the support page before advisory 1.0; it did not add path comparison.

## Decision before implementation

A static reader cannot tell whether a changed path encodes an authorization boundary, a resource choice or a rotated secret. The product cannot promise both that every capability-shaped path change is visible and that every secret-shaped path rotation is quiet without additional evidence. Record the choice before changing the current quiet-rotation invariant.

Evaluate a comparison-only observation such as “this server's URL path changed”, with no path text or reusable path digest. Such an observation states only that the declaration changed; it does not infer greater privilege from words such as `admin`. It may also appear for a secret rotation, and that noise cost must be explicit. If quiet rotation remains the chosen requirement, preserve the refusal to infer path semantics and document the resulting review limit instead of claiming the gap closed.

The design must account for both Git base/head comparisons and saved inventory/baseline comparisons. A sanitized baseline may not retain enough information to compare a path; do not republish a secret-derived identifier or claim path comparability merely to make the routes agree. Prefer a precise route-specific limit to a false no-change answer.

## Acceptance

- [ ] Record the selected privacy/noise behavior and its reason before implementation. Pin both the `/read` → `/admin` example and a secret-shaped path rotation; neither example by itself proves the runtime privilege effect.
[...]

PR #806, opened 2026-09-15
Show named reusable-workflow secret remapping in host diffs (#693)

### LINKED ISSUE #693: Named reusable-workflow secret remapping is invisible to host capability diffs

pengfei-threemoonslab:
Closes #693

A reusable job changed from `secrets: {credential: "${{ secrets.STAGING_TOKEN }}"}` to `${{ secrets.PRODUCTION_TOKEN }}`, with the same job, `uses:`, triggers and token permissions. That change gave `comparison_status: comparable` with no row and no limit. The job-level half of the #767 collision rule from the #788 review was silent the same way (`uses: …/x.yml@token=aaaaaaaa` → `@token=bbbbbbbb`). Both now show a named row or refuse.

## Design

**One reader, one comparator, no new surface.**
- The existing reusable-call entry on the workflow grant (the #685/#771 shared workflow reader) gains two members, both omitted when unset:
  - `secret_mappings[]`: `{destination, source, form, unresolved_reason}`;
  - `uses_redacted`.
- There is no new command, verdict, check ID, expansion signal or runtime secret resolver.
- Surface discipline: this moves widening/review recall on the host `diff` route, where a silent omission becomes a row. It extends the one comparator instead of adding a parallel one.

**Typed facts, not display text.**
- A whole-value `${{ secrets.NAME }}`, with any spacing or YAML quoting, is `form: secret`: `destination` is the callee's secret input and `source` is `NAME`.
- `${{ github.token }}` is the *same* source. GitHub's contexts reference calls it "functionally equivalent to the GITHUB_TOKEN secret", so both spellings are `source: "GITHUB_TOKEN"` and migrating between them is quiet.
- The typed facts are extracted from the parsed YAML before any display redaction, and compared as each call's set of facts.
- Destination identity is kept per job, so the same destination in two jobs stays two mappings.

**Case.** Source names compare **case-insensitively**: GitHub's secrets reference says names "Are case insensitive when referenced. GitHub stores secret names as uppercase regardless of how they are entered." So `staging_token` → `STAGING_TOKEN` is not a remap, while the name is still published as written. The **destination** is the callee's `on.workflow_call.secrets.<secret_id>`, which GitHub's workflow syntax describes only as "A string identifier to associate with the secret" — no case rule — so it is compared as written, and a case-only edit there reads as a removal plus an addition. `${{ SECRETS.X }}` stays an unread `expression`: GitHub's expressions reference states case-insensitivity only for string *comparison*, never for a context name. Both decisions are documented on the support page and in STABILITY.

**Rows.**
- A changed row lists only the mappings that differ, as `job: secret <destination> ← secrets.<source>`.
- `why` tells the cases apart: a remap between two named sources; an entry that became named or stopped being named; a re-formed unreadable entry; an added or removed destination. An entry whose value is not read is described as unread, never as a name that was never seen.
- Renaming a destination is a removal plus an addition, not a changed source.
- Every mapping row adds one sentence, **scoped to the mapping**: a name does not establish the secret's privilege, caller availability or downstream use, "so this mapping change is not itself counted as a widening". It no longer reads as a denial of a real widening on the same row.
- `host_grant_expansion_signals` is unchanged, so `expands: false` always. `access`/`risk` are unchanged. `secrets: inherit` keeps its own `workflow_secrets_inherited_*` widening.
- Reordering the entries, re-spacing or re-quoting is quiet.

**An unreadable value is a named, NON-blocking limit.**
- Every other value is `form: unresolved` with `source: null`, and nothing of the value is kept: a literal (`literal_value`); any other expression (`expression`); a non-string (`not_a_string`); a `secrets:` that is neither `inherit` nor a mapping (one entry, `destination: null`, `secrets_not_a_mapping`).
- Each records a **non-blocking** `unsupported` coverage issue naming its `job/destination`, the way an unread `envFile` does. GitHub coverage stays `complete`.
- The entries are still compared on `(destination, form, unresolved_reason)` with no value, so adding one, removing one, or moving one between forms is a row. Only an edit *between two values of the same unreadable form* is unreported, and the named limit says where.

**One redaction-collision rule for job and step level (#767), still blocking.** The step-level code from #788 became the shared helpers:
- `_published(text) -> (display, redacted)` is used by step `uses`, job-level reusable `uses`, and secret destination and source names.
- `_uncompared_workflow_text(grant)` is the single place that turns a redacted value into the blocking issue. The step-only message is byte-identical to #788's.
- Job-level `uses` was previously only `_sanitize_sensitive_string`-ed, which is why `@token=aaaaaaaa`/`@token=bbbbbbbb` collapsed, and why a `ghp_…` token in a reusable target was published verbatim on 1.0.0. Both now take the `uses_redacted` + blocking path.

**Consumers.** `diff` (JSON and text), `check` (text, `agent-boundary-json`, `agent-control-json` `capability_rows`), manifest-free `verify` text, `verifier.json` and `pr-comment.md` all carry the same row through the shared `capability_diff_rows` projection. `control.permissions.merge` stays false, and no text renderer changed.

## Measured against published 1.0.0

Scenario repos built on disk and run twice: once with `agents-shipgate==1.0.0` in a clean venv, once with `./shipgate` from this branch. `origin/main` (now `ff300e2d`) matches 1.0.0 on every line below. Unread spellings measured: a literal, `${{ secrets['X'] }}`, `${{ inputs.x }}`, `${{ vars.X }}`, `${{ secrets.A || secrets.B }}`, a non-string, plus `${{ github.token }}`.

| Scenario | 1.0.0 | Branch at `d80ce4f1` | Branch now |
|---|---|---|---|
| Unchanged call with an unread value, README-only head | `allow`, comparable, merge `true`, coverage complete, save-baseline 0, drift comparable | `require_review`, incomparable `unchanged_limits_not_representable`, merge `false`, coverage partial, save-baseline **2**, drift **2** | **identical to 1.0.0**, plus one non-blocking named issue in the inventory |
| The same, with `${{ github.token }}` | as above | as above | identical to 1.0.0, **and no issue at all** (typed as `GITHUB_TOKEN`) |
| Unrelated `.claude/settings.json` change | 1 row | incomparable, **0 rows** | 1 row |
| `contents: read`→`write` **plus** `Bash(curl:*)` (P1‑2) | 2 rows, both `expands: true` | incomparable `base_inventory_incomplete; head_inventory_incomplete`, **0 rows** | 2 rows, both `expands: true` |
| `${{ github.token }}` → `${{ secrets.GITHUB_TOKEN }}` migration | comparable, 0 rows | **incomparable**, 0 rows | comparable, 0 rows, no issue |
| `staging_token` → `STAGING_TOKEN` | 0 rows | **1 spurious `changed` row** | 0 rows |
| Literal A → literal B | comparable, 0 rows, silent | incomparable, 0 rows | comparable, 0 rows, **named non-blocking limit** naming `deploy/token`, no canary or sha256 prefix anywhere |
| The #693 remap (`STAGING_TOKEN`→`PRODUCTION_TOKEN`) | comparable, **0 rows** | 1 row | 1 row, `expands: false` |
| Literal → `${{ secrets.STAGING_TOKEN }}` | comparable, 0 rows | incomparable | 1 row: "now comes from a named source, where its value was not readable before" |

`audit --host` names the limit under **Coverage issues**, with GitHub coverage `complete`:

```
## Coverage issues

- `github` `.github/workflows/ci.yml` (declared exclusion): the secret a reusable
  workflow call passes at deploy/token is a literal value; its value is neither
  published nor compared, so an edit between two such values is not reported
```

## Schema / contract decision: extend 0.6 / contract 40 in place

- Host-grants `0.6` and runtime contract `40` came from #771 (#788) and **have not shipped in a tagged release**. `published_release.py` pins `LATEST_PUBLISHED_CONTRACT_VERSION = "39"`, and published `1.0.0` emits host-grants `0.5`.
- No test or rule freezes the `0.6` bytes: `docs/host-grants-*-schema.v0.6.json` is only validated against, never SHA-pinned; STABILITY's "keeps its bytes forever" applies to published schemas; the `0.5` schemas and models are untouched.
- The new members are successor models inside the 0.6 union: `HostReusableWorkflowSecretV6`, `HostReusableWorkflowCallV6`, `HostWorkflowGrantV6.reusable_calls`.
- `scripts/generate_schemas.py --check` passes with no schema byte change in this cycle. `.well-known`, the AGENTS.md table, the kit render hashes and `LATEST_PUBLISHED_CONTRACT_VERSION` are unchanged, since no version number moved. `llms-full.txt` was rebuilt for the contract-doc paragraph.

**Legacy vs inspected absence.**
- A `0.4`/`0.5` baseline holding a reusable call is still incomparable, and names `baseline_reusable_workflow_secret_mappings_unavailable` beside `baseline_workflow_step_actions_unavailable`. A `0.4`/`0.5` baseline without a workflow is unchanged.
- In a `0.6` grant, absence of `secret_mappings` means "read, none declared". The schema version, not the key, separates that from legacy, as #771 did for `step_actions`.
- A `0.6` baseline saved from a source tree between #788 and this PR has no `secret_mappings`, so a named mapping it held shows as an **added** destination: a row, never silence. STABILITY says so.

## Privacy argument

- **Values are never kept.** A secret's value is never read. A literal or expression value is dropped at the reader: it contributes no field text, no hash input and no issue text. The workflow's `config_sha256` is computed on the projection, whose `secrets*` keys the config redactor already replaces wholesale.
- **Names are published only when redaction leaves them unchanged.** Otherwise the entry is `redacted` and blocks comparison, so no digest of a credential-shaped name or target is ever needed to tell two values apart.
- **Row wording survives redaction.** The row and issue texts were probed through `redact_text`, `redact_data`, the host sanitizer and `safe_failure_text` and come out unchanged.
- **Canary sweep.** For six canaries (literal value, literal inside an expression, redacted source name, redacted destination name, job-level `@token=`, job-level `ghp_…`), each canary **and its sha256 (full, 24- and 12-hex prefixes)** is absent from inventory JSON and audit Markdown, drift JSON and Markdown, the baseline (refused where blocking, **and the saved bytes where it now saves**), `diff` JSON and text, manifest-free `verify` text, `pr-comment.md` and `verifier.json`. A separate case sweeps a literal A→B *edit* across the same surfaces.
- **Scope (#802).** Job ids stay raw, exactly as today, and the `job/destination` label reuses the job id the row's `before`/`after` already printed on 1.0.0. Destination names that match no credential pattern are likewise published verbatim — the same class as #802, tracked there, not implemented here.

## Acceptance

- [x] **Observed counterexample, via the real CLI.** `test_diff_names_the_remap_in_json_and_text` and `test_the_formerly_silent_named_secret_fixture_now_produces_its_row`. Quiet controls stay quiet: unchanged, `formatting-only[block-style|spacing-and-quotes|surrounding-whitespace|comment]`, reordering.
- [x] **Added/removed destinations distinguishable from a changed source**: `test_added_removed_and_changed_destinations_are_told_apart`, `test_the_same_destination_in_two_jobs_stays_two_mappings`.
- [x] **No collapse on equal display redaction, job-level `uses` and secret names.** `test_distinct_values_that_redact_alike_are_published_alike` proves the display collision for four canary pairs; `test_a_changed_workflow_whose_values_redact_alike_never_compares_as_unchanged` gets `incomparable` with no canary or digest; `test_an_unchanged_redacted_reusable_target_is_a_named_limit_beside_other_rows`.
- [x] **No credential value or digest in any output**: `test_no_canary_or_digest_reaches_json_markdown_baselines_errors_or_pr_output` (6 canaries, now parametrised over blocking/non-blocking) and `test_a_literal_value_edit_is_a_named_limit_and_publishes_no_canary`. Negative controls: `test_an_unsupported_value_publishes_nothing_of_itself` (9 forms), `test_secrets_that_are_neither_inherit_nor_a_mapping_are_one_unresolved_entry`, `test_an_empty_secrets_key_is_not_read_as_none_declared`. Over-redaction controls: 5 ordinary targets and 7 ordinary secret names.
- [x] **No widening inferred from a name**: `expands: false`, `expansion_signals == []`, `access`/`risk` unchanged.
- [x] **The same change on `diff`, `check` and `verify`, and in advisory PR output**: `test_check_names_the_same_remap` (three formats), `test_manifest_free_verify_and_its_pr_comment_name_the_remap[preview|non-preview]`, `merge` false.
- [x] **Existing cases still pass**: `secrets: inherit` both directions, reusable target/ref, token permissions, workflow-script noise; `tests/test_workflow_capability_diff.py` and `tests/test_workflow_step_action_references.py` unchanged and passing.
- [x] **Closed-schema compatibility and legacy vs inspected absence**: `[0.4|0.5]` legacy baselines, current-baseline schema validation, out-of-order mappings.
- [x] **Support page and its pin tests cover only the forms read**, including the `github.token` typing and both case decisions.

## Review cycle 1 (`d80ce4f1` → this head)

Rebased onto `origin/main` `ff300e2d` (through #789/#714, #801/#575 and #805/#638). The `## Unreleased` section keeps every bullet from `main` exactly once plus one `(#693)` bullet; `capability_diff_rows.py` keeps #714's `hook_loading_basis` import and `why` sentence beside the mapping rows, and `docs/distribution-surfaces.md`'s `capability_diff` row and `docs/host-boundary-support.md` keep both #714's and this change's text. `scripts/generate_schemas.py --check` and `scripts/build-llms-full.py` were rerun after the rebase with no diff.

- **P1‑1 / P1‑2 (one root cause) — fixed.** The four unresolved forms are non-blocking `unsupported` issues naming `job/destination` (`uncompared_secret_mapping_texts`), the `envFile` model; coverage counts only blocking issues. `_uncompared_workflow_text` now records only the redaction collisions, which stay blocking. New cases: `test_an_unchanged_call_with_an_unread_value_never_flips_check`, `test_an_unrelated_host_change_keeps_its_row_beside_an_unread_value`, `test_a_widening_workflow_edit_still_shows_every_row` (each over six unread spellings), `test_the_workflow_token_migration_is_quiet_and_comparable_on_every_route`, `test_a_literal_value_edit_is_a_named_limit_and_publishes_no_canary`, `test_every_unread_form_names_its_job_and_destination`, and `test_an_unsupported_mapping_is_a_named_non_blocking_github_limit`.
- **P2 — fixed.** `secret_mapping_key` casefolds the source. Destination case and the `SECRETS.` context case are decided against the GitHub docs above and documented; `test_a_case_only_source_edit_is_quiet`, `test_a_case_only_destination_edit_is_a_removal_and_an_addition`, `test_an_uppercase_secrets_context_stays_an_unread_expression`.
- **Nit 1 — fixed.** The sentence is scoped: "…so this mapping change is not itself counted as a widening", and its subject follows whether a name was involved.
- **Nit 2 — done.** STABILITY's migration note, the contract lead paragraph, `docs/agent-contract-current.md`, `llms-full.txt`, `docs/host-boundary-support.md`, `docs/distribution-surfaces.md` and the CHANGELOG bullet all state the non-blocking behaviour, the typed `github.token` and the case folding. The migration note's cost paragraph no longer claims refusals.
- **Nit 3 — left, referencing #802**, as the review directed.

**Not carried out.** The review's fix list asked for the literal A→B limit to be "a named non-blocking limit"; the dispatching instruction additionally asked for `diff`/`verify` to list it under **Not compared**. That is not possible without breaking a shipped contract, so it is named in the inventory (and in `audit --host`'s *Coverage issues*) instead: `unchanged_limits` is pinned by verifier `0.19` / contract v37 to limits whose "source is byte-identical at base and head" and to the three kinds `unsupported | parse_failed | experimental_coverage`, and `cli/check.py:485` refuses **any** non-empty `unchanged_limits` — so routing these limits there would either mis-describe a changed file or reintroduce P1‑1 verbatim.

## Tests and exit codes (on `e5f7fc2b`)

| Run | Exit |
|---|---|
| `tests/test_reusable_workflow_secret_mappings.py`, `tests/test_workflow_step_action_references.py`, `tests/test_workflow_capability_diff.py`, `tests/test_host_boundary_unread_surfaces.py` | 0 |
| Host/check/docs group: `test_host_audit`, `test_unchanged_host_limits`, `test_manifest_free_pr_rows`, `test_host_change_route_parity`, `test_capability_diff`, `test_host_path_privacy`, `test_host_boundary_check`, `test_check_default_comparison`, `test_host_diff_entry_docs`, `test_host_only_advisory_recipe`, `test_host_inventory_stability`, `test_distribution_surface_parity`, `test_public_surface_contract`, `test_instruction_structure_contracts`, `test_documentation_checks`, `test_agent_control_envelope_rows`, `test_host_input_recovery` | 0 |
| `tests/test_adapter_static_only.py` (line-number pins) | 0 |
| `scripts/generate_schemas.py --check` | 0 |
| `ruff check` on the changed Python files | 0 |
| Full suite `-m "not perf"`, all 310 other test files in 7 literal-path foreground chunks | 0, 0, 0, 0, 0, 0, 0 |

## Out of scope / not done

- #802 (raw job ids, and destination names published verbatim) is not implemented; no new raw-job sink beyond the row's existing job text was added.
- #701 (local composite actions), #702 and #772 keep their unread-surface entries.
- `${{ secrets['NAME'] }}` index-form support is still deliberately absent: it stays a named limit rather than a guess.


pengfei-threemoonslab (commented):
Independent coding-agent review/address cycle 1 at `d80ce4f1`. **Not mergeable yet: two P1s with one root cause, and one P2.** The reader, the rows and privacy all hold.

**P1-1: an unchanged reusable call passing `${{ github.token }}`, a literal or any other expression now forces `require_review` and blocks baselines.**

`_uncompared_workflow_text` (~1344) and `_collect_file` (~1562) record `blocking=True`, and `cli/check.py:485-493` refuses any unchanged blocking limit.

Setup: `ci.yml` calls `./.github/workflows/deploy.yml` with `secrets: {token: <value>}`, and the head changes only README. Results are identical for `${{ github.token }}`, a literal, `${{ secrets['X'] }}`, `${{ inputs.x }}`, `${{ vars.X }}` and `${{ secrets.A || secrets.B }}`:
| Surface | 1.0.0 / main | Branch |
|---|---|---|
| `check`, all three formats | `allow`, comparable, merge true | `require_review`, incomparable `unchanged_limits_not_representable`, merge false |
| `audit --host --save-baseline` | exit 0 | exit 2 (inventory incomplete) |
| GitHub coverage | `complete` | `partial` |
| `diff` / `verify` / PR comment | comparable | comparable, plus a "Not compared … unsupported" limit |

- With an unrelated `.claude/settings.json` change, `check` goes from 1 row to incomparable with **0 rows**.
- Drift against a 1.0.0 baseline adds `current_inventory_incomplete` and `--fail-on-drift` exits 20. The re-save STABILITY prescribes then exits 2, so these repositories can never get a comparable baseline.
- The `${{ secrets.GITHUB_TOKEN }}` control matches 1.0.0 everywhere.

Why P1: GitHub documents passing the workflow token to reusable workflows, and `secrets.A || secrets.B` fallbacks are common. #789 cycle 1 was the same class and was fixed with non-blocking issues. #693 asks for a privacy-safe *limit*, not a blocking one.

**P1-2: any edit to such a workflow hides every host row in the PR.** With a `github.token` call, a head that changes `contents: read` to `write` and adds `Bash(curl:*)` to `.claude/settings.json`:
- 1.0.0: two `expands: true` rows;
- branch: `diff`, `verify`, the PR comment and `check` are all incomparable (`base_inventory_incomplete`, `head_inventory_incomplete`) with 0 rows.

That fails closed, but a routine `actions/checkout` bump removes all host rows, and the migration STABILITY recommends (`github.token` → `secrets.GITHUB_TOKEN`) is itself refused.

Fix for both:
- Make `literal_value`, `expression`, `not_a_string` and `secrets_not_a_mapping` **non-blocking** `unsupported` issues naming `job/destination`, like `vscode` `envFile` at `:762-768` (coverage counts only blocking issues, `:1985`).
- Keep comparing on (destination, form, reason) without the value, so adding, removing or changing the form of an unresolved entry still gives a row, and a literal A→B edit is a named non-blocking limit.
- Keep `redacted` blocking (the #767 collision rule).
- Type `${{ github.token }}` as the same source as `secrets.GITHUB_TOKEN`, so switching between them is quiet.

**P2: a case-only change to a secret name makes a spurious row.** `secret_mapping_key` and `_SECRET_SOURCE_RE` compare names as written. `staging_token` → `STAGING_TOKEN` gives a `changed` row (1.0.0: none). GitHub's Secrets reference says names are case-insensitive when referenced and stored uppercase. Fix: casefold the source in the compare key. Lower confidence: a case-only destination edit shows as removed plus added, and `${{ SECRETS.X }}` is classed `expression`.

**Nits**
1. The mapping sentence "so this is not reported as a widening" is appended even to `widened` rows (named→inherit, or combined with `contents: write`), `capability_diff_rows.py` ~110-118. Suggested: "the name change itself is not counted as a widening".
2. If any blocking behaviour is kept, STABILITY and the support page must say that `check` refuses and that edits remove all rows.
3. Destination names that don't match a credential pattern are published verbatim; same class as #802, and GitHub rejects such ids.

**Verified with no finding**
- **Privacy, 18 value shapes:** literal, multi-line block, flow mapping, anchor/alias, int/bool keys, unicode, an expression embedding a canary, list/dict values, string/list `secrets:`, and a literal in `with:`. No canary or sha256 prefix appears anywhere. Credential-shaped destinations are `redacted` and blocking.
- **Rows:**
  - added, removed and changed wording is right, and reordering is quiet;
  - the same destination in two jobs stays distinct;
  - inherit→named is `changed` and non-expanding; named→inherit keeps the inherit widening;
  - `expands` never comes from a name, and merge is unchanged.
- **Legacy 0.5 baselines** with a reusable call give the new reason plus the step-actions reason, exit 20.
- **Extending 0.6 in place is safe:** PyPI 1.0.0 and the `v1.0.0` tag are host-grants 0.5 / contract 39; `LATEST_PUBLISHED_CONTRACT_VERSION` is 39; AGENTS.md, `.well-known`, `llms-full.txt` and `agent-contract-current.md` agree; STABILITY documents the side effect for source-saved 0.6 baselines.
- **CHANGELOG:** exactly one `(#693)` bullet; the distribution-surfaces row is accurate.

**Runs:** the secret-mapping, step-reference, workflow-diff, unread-surfaces and static-only tests, the host-audit, unchanged-limit, manifest-free-rows, route-parity, capability-diff, path-privacy, envelope-rows and parity tests, `generate_schemas.py --check` and `ruff` all exit 0.

**CI:** all pass.

pengfei-threemoonslab:
**Addressed review cycle 1.** New head `e5f7fc2ba2978fde04771d7b74513a58501eb5d3`, rebased onto `ff300e2d` (main moved three times mid-cycle: #789/#714, #801/#575, #805/#638). CI green.

**P1-1 and P1-2 (one root cause) — fixed.** `_uncompared_workflow_text` now records only the #767 redaction collisions, which stay blocking. The four unresolved forms go through a new non-blocking `uncompared_secret_mapping_texts()` naming `job/destination`, appended in `_collect_file` with `blocking=False` — the `envFile` model the review cited.

Measured by building scenario repositories and running published 1.0.0 in a clean venv against the branch:
| Scenario (6 unread spellings) | 1.0.0 | `d80ce4f1` | now |
|---|---|---|---|
| unchanged call, README-only head | allow, comparable, merge true, coverage complete, save 0, drift comparable | require_review, `unchanged_limits_not_representable`, merge false, partial, save 2, drift 2 | identical to 1.0.0, plus one non-blocking named issue |
| same with `${{ github.token }}` | as above | as above | identical to 1.0.0, **no issue at all** |
| unrelated `.claude/settings.json` change | 1 row | incomparable, 0 rows | 1 row |
| `contents: read`→`write` plus `Bash(curl:*)` | 2 rows, both expanding | incomparable, 0 rows | 2 rows, both expanding |
| `github.token`→`secrets.GITHUB_TOKEN` | comparable, 0 rows | incomparable | comparable, 0 rows, no issue |
| literal A→B | comparable, silent | incomparable | comparable, 0 rows, named non-blocking limit (`deploy/token`), no value or digest |
| the #693 remap itself | 0 rows | 1 row | 1 row, `expands: false` |

**P2 — fixed.** `secret_mapping_key` casefolds the source, so `staging_token`→`STAGING_TOKEN` goes from one spurious row to none, matching 1.0.0.

**Nit 1 — fixed.** The sentence is scoped to "…so this mapping change is not itself counted as a widening", and an entry moving between named and unread no longer claims a different named source.

**Nit 2 — done** across STABILITY, `agent-contract-current.md`, `llms-full.txt`, the support page, the surface registry, the contract comment and the CHANGELOG bullet; the cost paragraph no longer claims refusals. **Nit 3** left, referencing #802.

**Case decisions, with citations**
- **Source name casefolded:** GitHub's Secrets reference says names are case-insensitive when referenced and stored uppercase.
- **Destination compared as written:** the workflow syntax reference defines `on.workflow_call.secrets.<secret_id>` only as a string identifier, with no case rule, so a case-only destination edit stays a removal plus an addition.
- **`${{ SECRETS.X }}` stays an unread expression:** the expressions reference gives case-insensitivity for string comparison, never for a context name.
- **`${{ github.token }}` typed as `GITHUB_TOKEN`:** the contexts reference calls it functionally equivalent to the `GITHUB_TOKEN` secret.

**Deviation, reported deliberately.** The review asked for the literal A→B limit to appear under "Not compared" in `diff`/`verify`. That was not done: STABILITY's contract-v37 note pins `unchanged_limits` to limits whose source is byte-identical at base and head and to three `limit` kinds in published verifier `0.19`, and `cli/check.py:485` refuses any non-empty `unchanged_limits` — so routing these there would either mis-describe a changed file or reintroduce P1-1. The limit is named in the inventory and printed by `audit --host` under Coverage issues, with GitHub coverage `complete`.

**Tests on `e5f7fc2b`:** the mapping, step-reference, workflow-diff and unread-surface files plus the static-only test exit 0; a 17-file host/check/docs group exits 0; `generate_schemas.py --check` exits 0 with no schema byte change; `build-llms-full.py` exits 0 with no diff; `ruff` exits 0; and the full suite runs in 7 foreground chunks, all exit 0.

pengfei-threemoonslab (commented):
Independent coding-agent review/address cycle 2 (focused) at `e5f7fc2b`. **No P0/P1. Two P2s, both claim mismatches, now fixed in `6afa6cc4`.**

**P2-1: the `0.6` JSON schemas still asserted the pre-fix blocking rule.** The `HostReusableWorkflowSecretV6` docstring (`schemas/host_grants.py:516-517`) said "Every unresolved mapping records a blocking coverage issue", and Pydantic publishes it verbatim as the JSON-Schema `description`, so `docs/host-grants-inventory-schema.v0.6.json:1195` and `docs/host-grants-baseline-schema.v0.6.json:1137` said it too. `generate_schemas.py --check` passed precisely because the JSON mirrored the stale docstring. Fixed and regenerated: a `redacted` name blocks, and every other unresolved mapping records a non-blocking issue naming its `job/destination`.

**P2-2: `docs/distribution-surfaces.md:77` claimed `capability_diff` leaves the unreadable value "as a named non-blocking limit".** It names it nowhere: `core/host_comparison.py:37-44` builds `unchanged_limits` from blocking issues only, and the PR's own test reads the limit out of `audit --host --json`, not out of `diff`. Reproduction on the branch (`token: litA` → `litB`): `diff --json` gives `comparable`, `rows: []`, `unchanged_limits: []`, while `audit --host --json` carries one non-blocking `unsupported` issue. The row and `docs/host-boundary-support.md` now say the value is neither compared nor named on that surface, and name where the limit is carried.

**Nit 1 (fixed):** `audit --host` printed the new limit as a `declared exclusion` (`core/host_grants.py:3769`), though nothing was declared or excluded. An `unsupported` or `unreadable` non-blocking issue now prints `not compared`. **Nit 2 (fixed):** the support page now says where the limit is named. **Nit 3** carried to #802; the reviewer also noted a credential-shaped job name reaches the inventory verbatim on 1.0.0 and the branch alike, while this PR's new message redacts it.

**Verified with no finding**
- **Both P1s are gone, with no new silence.** 8 value spellings × 3 head variants (24 repositories), published 1.0.0 versus branch, across `check` in three formats, `diff`, `audit --host` with `--save-baseline` and `--drift --fail-on-drift`: byte-for-byte parity in every cell. `github.token` produces zero issues; each other unread form produces exactly one non-blocking named issue; coverage stayed `complete` in all 24.
- **23 silence hunts:** a destination added or removed beside an unreadable value, named ↔ unreadable, `inherit` ↔ literal (the inherit widening preserved), two jobs with one unreadable, a reusable target changed while a mapping is unreadable, a `secrets:` block deleted or emptied, and a form change all produce a row. The only silence is an edit between two values of the same unreadable form, where 1.0.0 is equally silent. Structurally the projection is 1.0.0's plus the mapping keys, so it can only add rows.
- **The deviation is correct.** STABILITY's contract-v37 note pins `unchanged_limits` to limits byte-identical at base and head, and `cli/check.py:485-493` turns any non-empty value into `unchanged_limits_not_representable` — so routing an unchanged unread mapping there would reintroduce P1-1.
- **`github.token` typing and casefolding:** `github.token` → a different named secret is a row; only `github.token` ↔ `secrets.GITHUB_TOKEN` is quiet, which GitHub documents as equivalent, and GitHub reserves the `GITHUB_` prefix so no competing secret exists. `_SECRET_SOURCE_RE` admits only ASCII, so casefolding has no Unicode hazard.
- **Privacy:** a canary literal, a changed literal and a canary inside an expression produce zero hits across `diff`, `check`, audit Markdown and JSON, saved baselines, drift, `verify`, `verifier.json` and `pr-comment.md`, with no digest prefix. Credential-shaped names still redact, block, make coverage partial and refuse `--save-baseline`.
- **Rebase integrity across #714, #575/#785 and #638:** `## Unreleased` has 11 bullets with `(#693)` once; schemas, `llms-full.txt`, contract 40 / schema 0.6 and `LATEST_PUBLISHED_CONTRACT_VERSION` 39 are consistent, and no kit claims 40 is published.

**Runs:** 232 mapping/step-reference/workflow-diff/unread-surface tests, 192 host-audit/limits/rows/check tests, 833 parity/static-only/documentation/public-surface tests, 71 more, `generate_schemas.py --check`, `build-llms-full.py` and `ruff` — all exit 0. CI green.

**Fix verification on `6afa6cc4`:** the stale sentence is gone from both schema JSON files, `generate_schemas.py --check` and `build-llms-full.py` are clean, and the mapping, host-audit, unchanged-limit, parity, documentation, public-surface, unread-surface, inventory-stability and static-only tests plus `ruff` all exit 0.

Issue #813, opened 2026-09-16
workspace_unverifiable does not say why the worktree could not be read


pengfei-threemoonslab:
## Problem

Found during the independent review of PR #801 (#575), as a non-blocking nit worth its own issue.

`core/current_control.py` (~867-876, ~898-907, ~947-956) refuses with `workspace_unverifiable` whenever the live worktree view is unavailable — the correct fail-closed behaviour, added in #575 so that an unreadable worktree can never let a pointer read as current. But `live_workspace` (`cli/current_workspace.py` ~88-89) swallows the underlying exception, so the refusal names neither the cause nor anything that clears it, and the emitted `next_action` just repeats the same `verify` command.

A concrete legitimate trigger is a repository that configures executable Git filters, which `working_tree_context` rejects (`cli/verify/git.py`, `_reject_executable_worktree_filters`):

```
git -C repo config filter.lfs.clean "git-lfs clean -- %f"
agents-shipgate agent control --workspace repo --reports-dir <dir>
```

**Observed:** exit 4, "The current set of uncommitted changes could not be determined, so it cannot be shown that this worktree verification still describes the workspace."

`verify` in such a repository already fails on published 1.0.0, on main and on the PR branch, so the loop was already blocked there — this is about the agent being told why.

## Acceptance

- [ ] `LiveWorkspace` carries the swallowed failure (for example `changed_paths_error: str | None`), and all three `workspace_unverifiable` refusals append it, so the reader sees the cause (e.g. "static worktree collection refuses executable Git filters (filter.lfs.clean)").
- [ ] The `next_action` names something that can change the answer, rather than repeating a command that fails the same way.
- [ ] Tests cover at least the executable-filter cause and one other, and pin that an unavailable worktree view still refuses for every pointer state.
- [ ] No raw file content or secret reaches the message.

Refs #575, #801.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Journey delivery - 2026-09-16

Program: #778. Delivery milestone: [Adoption Phase 1 - A real change the reviewer understands](https://github.com/ThreeMoonsLab/agents-shipgate/milestone/10). Accountable role: **Diagnostics/control engineering**.

[...]

Issue #807, opened 2026-09-15
verify --preview in a configured repo publishes human_review_required: input directory capture is unavailable


pengfei-threemoonslab:
## Problem

Found and confirmed during the independent review of PR #801 (#575). Reproduces identically on published `1.0.0`, on `main` `42110535`, and on the PR branch.

In a repository with a committed `shipgate.yaml`, run from the repository:

```
agents-shipgate verify --workspace . --config shipgate.yaml --preview --format control
agents-shipgate agent control --workspace . --reports-dir agents-shipgate-reports
```

**Observed**
- The preview exits 0 with `control_state=human_review_required` and the reason: "The recorded source and dependency inputs are no longer current: input directory capture is unavailable; re-run verification".
- The refresh exits 4 with `workspace_unverifiable`.
- The same happens with the default `--out`, an explicit in-repository `--out`, and a sibling `--out`.

**Comparison**
- `verify --preview --json` on the same repository gives `control.state=agent_action_required` ("Shipgate is configured; run verify on the PR to get a merge verdict.").
- A repository without a manifest gives `agent_action_required`, exit 0, on all three builds.

**Expected**
- `docs/agent-contract-current.md:1059-1071` says preview emits `control.state: "agent_action_required"` with the exact `verify` command for configured repositories.
- `:788-791` describes preview pointers.

The control format sends the agent to a human for a non-human reason, and re-running only reproduces it.

**Likely cause:** the preview pointer binds a `verification_plan` whose `inputs.options` has no `input_directories` (checked on all three builds). The current-control reader then raises at `core/input_directory_identity.py:23`.

## Acceptance

- [ ] In a configured repository, `verify --preview --format control` publishes the documented `agent_action_required` control with the `verify` command, matching `--preview --json`.
- [ ] The `agent control` refresh of that preview pointer is current while the worktree is unchanged, and refuses (`workspace_changed`) after a tracked or untracked edit. It never fails open when the input-directory list is missing.
- [ ] Tests cover the default, in-repository and sibling (#575) `--out`, with and without a manifest.
- [ ] Any contract documentation that changes is updated.

Refs #575, #801.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Journey delivery - 2026-09-16

Program: #778. Delivery milestone: [Adoption Phase 2 - Decide and verify the correction](https://github.com/ThreeMoonsLab/agents-shipgate/milestone/11). Accountable role: **Verification/control engineering**.

[...]

PR #814, opened 2026-09-16
Keep check coherent when settings add keys outside the host-boundary rule allow-list (#810)

pengfei-threemoonslab:
Closes #810

## Problem

On `main` (`9df307ab`, reports 1.0.0), `shipgate check` exits 1 with an unhandled pydantic `ValidationError` ("publication authority requires complete boundary input coverage") whenever a changed Claude Code settings file (`.claude/settings.json` or `.claude/settings.local.json`) or Cursor `.cursor/cli.json` sets a top-level key outside the host-boundary rule allow-list (`_evaluate_unknown_keys`). The trigger is wider than the issue title:

- `enabledPlugins` added, set to `false`, or switched from `true` to `false`
- `extraKnownMarketplaces`
- `outputStyle`
- `enableAllProjectMcpServers`
- `enabledMcpjsonServers`
- any unknown Cursor key
- any of these next to an ordinary shell allow rule

Some of these keys are modelled elsewhere: `core/host_grants.py` reads `enabledPlugins` and `extraKnownMarketplaces` as plugin grants and `enableAllProjectMcpServers` as a critical permission-mode grant, so inventory and diff rows describe them. Only the boundary rules lack them, and record the key instead.

It crashes in every maintained format (text, agent-boundary-json, agent-control-json), in both worktree and `--base/--head` modes, and for every `--agent`. `AGENTS_SHIPGATE_AGENT_MODE=1` prints no `next_action`. `diff`, `audit --host` and `verify` do not crash.

**Root cause.** Two modules classify the same row in opposite ways. The key produces a `HOST-CONFIG-PARSE-FAILED` row with evidence kind `unknown_host_config_key`:

- `core/codex_boundary.py` puts that kind in `_PARSEABLE_EVIDENCE_KINDS`. It counts as complete, read input, so the control may publish.
- `_coverage_for` in `core/agent_boundary.py` treated the same row as unread, so `input_coverage` became `partial`.

The v3 schema invariant then rejected the projection.

A second path fails the same way. The generic `BOUNDARY-INPUT-INCOMPLETE` suppressor also listed `unknown_host_config_key` as a row that explains an input issue. Take a repo with an unreadable legacy `policies/host-boundary.shipgate.yaml` (a warning-level issue with no row of its own) and change only such a key. `check` still crashed with only the coverage fix applied (measured). On `main` the same input crashes too, except in the deprecated `codex-boundary-json` format.

## Design

[...]

PR #831, opened 2026-09-17
Refuse an output directory that holds repository content (#804)

### LINKED ISSUE #804: verify --out set to a tracked directory hides that directory's changes from the decision and currency check
## Problem

Found during the independent review of PR #801 (#575). It reproduces identically on published `1.0.0`, on `main` `7ac1c456`, and on the PR branch, so it is pre-existing.

`verify` excludes its output directory from Git's working-tree view so that its own reports don't count as changes. When `--out` names a **tracked** directory inside the repository, that directory is excluded as well. Its tracked content drops out of both the decision's inputs and the later currency check:
- `--out <repo>/docs`;
- an outside path that resolves into the repository, such as `--out <link-to-repo>/docs`.

## Evidence (reviewer A/B)

1. Clean committed repository with a tracked `docs/a.txt`.
2. Run `agents-shipgate verify … --out <repo>/docs --format control`.
3. Edit `docs/a.txt`.
4. The `agent control` refresh still reports `complete`.

The same holds through a symlinked path into the repository. `verify` already refuses `--out` equal to the workspace root, a parent of it, or a directory overlapping the config input. A tracked subdirectory is not refused.

## Why it matters

The output directory is chosen by the caller, who may be the coding agent under verification. A reviewed decision should not be able to exclude tracked content through an output flag. Only `docs/` was demonstrated.

**Priority:** filed as P2. First confirm whether a decision-relevant directory (tool sources, host configuration, a manifest-referenced path outside the config-input overlap check) reproduces; if it does, raise this to P1.

## Acceptance

- [ ] `verify`, `--preview` and `--head` runs each either refuse an `--out` that contains tracked files, with an actionable message naming the directory and a safe default, or never exclude tracked paths. Pick one rule and apply it at every exclusion call site through one shared helper.
- [ ] The `agent control` refresh, the Stop hook and every other currency check apply the same rule. A pointer produced under a tracked output directory can never read as `complete` after a tracked change there.
- [ ] Tests cover: a tracked `docs/`, a tracked source directory named by a manifest, a symlink into the repository, a directory that is gitignored but inside the repository (must stay supported), the default `agents-shipgate-reports` (unchanged), and a sibling directory outside the repository (from #575).
- [ ] If the fix changes a documented flag contract, STABILITY.md and the CHANGELOG say so.

Refs #575, #801, #785.

## Journey delivery - 2026-09-16

Program: #778. Delivery milestone: [Adoption Phase 1 - A real change the reviewer understands](https://github.com/ThreeMoonsLab/agents-shipgate/milestone/10). Accountable role: **Verification identity engineering**.

Immediate bounded triage in the reliability reserve. The demonstrated scope is tracked docs; first test decision-relevant source/config paths, symlink aliases and existing overlap refusals. Preserve the P2 severity label until that evidence supports changing it; scheduling triage early does not assert an unproven bypass. Apply one reviewed output-exclusion rule to verification and current-control readers, preserving legitimate generated outputs. Bind the fix to changed tracked-input invalidation tests. Do not manufacture a broader security claim to justify priority.

Planning checkpoints preserve #778's original 2026-09-14 start: Oct 14 / Nov 13 / Dec 13. They are evidence-review dates, not promises about external replies. Phases overlap; a later milestone never delays supported research, an already-possible correction or an individual four-week opportunity window. A named accountable owner is required before execution; the role here is not an assignment to an uninvolved person.

pengfei-threemoonslab:
Closes #804

## Problem

`verify` leaves its output directory out of every working-tree read it makes: the Git change set, the worktree overlay its pointer binds, the manifest-free host comparison and the static input census. `agent control` does the same to the reports directory when it checks currency. The point is that a run's own reports never count as part of the change.

That only works while the directory holds nothing else. When `--out` names an in-repository directory with other content, that content drops out of the decision and out of every later currency check. The issue's P1 test is met: decision-relevant directories reproduce.

Measured on `origin/main` `9df307ab` (reports `1.0.0`), with throwaway Git repositories built from `samples/clean_read_only_agent` (`.gitignore` = `agents-shipgate-reports/`, branch `feature` off `main`):

| Scenario | Default output directory | Output directory = the affected directory |
| --- | --- | --- |
| `.claude/settings.json` gains `Bash(curl:*)`, `Bash(rm:*)`, uncommitted, `verify --base main` | `review_publishable`, merge=false, `merge_verdict: human_review_required`, `SHIP-HOST-BOUNDARY-PERMISSION-ALLOW-EXPANDED` + `SHIP-VERIFY-TRUST-ROOT-TOUCHED` | `--out .claude`: **`complete`, merge=true**, `passed`, `mergeable`, `changed_files: []`, no findings |
| Same widening, committed | same | `--out .claude`: **`complete`, merge=true** |
| New untracked `.claude/settings.json` | `review_publishable`, merge=false | `--out .claude`: **`complete`, merge=true** |
| `main` has deny rules in `.claude/settings.json`; the change `git rm`s the file, staged or committed | `review_publishable`, `SHIP-HOST-BOUNDARY-PERMISSION-DENY-REMOVED` | `--out .claude` (the directory no longer exists): **`complete`, merge=true, `mergeable`, no findings** |
| Tracked `docs/a.txt`: `verify`, then edit it, then `agent control` | refresh exit 4 `workspace_changed` | `--out docs`: refresh **exit 0 `complete`, merge=true** |
| Manifest tool source `tools/tools.json`: `verify`, then add a destructive tool | refresh exit 4 | `--out tools`: refresh **exit 0 `complete`, merge=true** |
| Committed widening through `<symlink-to-repo>/.claude` or `<parent>/REPO/.claude` | — | **`complete`, merge=true** |
| Committed widening with `--head HEAD` | `review_publishable` | `review_publishable` (CI's archived head was never affected) |

## Design (option A: refuse)

**One classifier:** `cli/current_workspace.py::classify_output_directory(root, reports_dir, *, compared=())` answers whether leaving a directory out of the change set can hide anything. It returns `None`, or an `OutputDirectoryRefusal(kind, reason)` naming what was found. `output_directory_refusal` is its string form for the readers.

It reads the directory through Git's own inventory: `verify/git.py::output_directory_inventory`, with `:(top,literal)` pathspecs, the same safe config and the same bounds as the change-set inventory. That function sits below `_run_git`, so the line-pinned subprocess allowlist is untouched. The inventory lists:
- `ls-tree` of `HEAD` and of each `compared` commit that resolves;
- `ls-files --cached`;
- `ls-files --others --exclude-standard`;
- `check-ignore --stdin -z` over the artifact paths a run would write there.

These hold repository content, and make the directory unsafe:
- anything committed beneath it at `HEAD`, whatever its name;
- anything a compared commit held there that `HEAD` no longer does, and any committed path the index and working tree no longer hold. A `git rm` leaves no directory to inspect, and was hidden just the same;
- any staged or untracked unignored path that is not a Shipgate artifact. An artifact is a name in `REPORTS_DIRECTORY_ARTIFACT_NAMES` directly beneath the directory (the pointer, verifier route, scan/packet renderers, capability-lock files, suggestion skeletons, the Action's payloads, and the `skill lint`/`security`/`review` reports), a file under `verification-inputs/`, or the pointer's in-flight temporary file;
- **any trust-root path, whatever its name** (`core/trust_roots.trust_root_class_for`): a name cannot tell a generated `packet.md` from `.claude/commands/packet.md`;
- **a directory inside a trust root that Git does not ignore**, even while empty (`.claude/reports`, `.github/workflows`), because every report a run wrote there would be a trust-root file;
- the repository root or an ancestor.

These leave the directory safe:
- it is outside the repository;
- it is gitignored, empty, or does not exist yet, with nothing committed there;
- it holds only uncommitted artifacts outside any trust root.

It never raises: a directory whose content cannot be listed is reported as not shown to be safe.

- **Containment** comes from #801's `worktree_exclusion` and its `(st_dev, st_ino)` walk, not a second spelling rule. A symlink into the repository and a case variant of the repository directory classify as the physical directory. `is_default_reports_dir` uses the same physical identity.
- **One extension to the walk:** below the root, each existing component is now spelled as its stored directory entry, matched by physical identity. Git matches pathspecs byte for byte, so on a case-folding filesystem `<repo>/.CLAUDE` excluded nothing and listed nothing while the run wrote into `.claude`. On main, that spelling of a safe directory exited 3 ("path changed identity while it was read"). Now it is refused when unsafe and works when safe.

**Writer:** `_reject_output_directory_content` runs right after the existing `_reject_output_input_overlap`, before `begin_current_control` or any write.
- It covers `run_verify` (worktree, `--head`, and the manifest-free route that hands off to preview) and `run_preview`. `run_preview` skips it outside Git, where there is no change set to leave anything out of.
- **Which commits it compares depends on the route.** A worktree run compares `HEAD` and the merge base it diffs against, including an auto-detected base and, without a manifest, the local base the host comparison detects. An archived `--head` run compares only the evaluated head: it diffs the merge base against that head in full, so a removal is already in its own change set, and the change that stops committing reports still passes the Action. A ref that does not resolve compares nothing, and the run reports it as before.
- It raises `OutputDirectoryHoldsRepositoryContent`, a `ConfigError` subclass, so the kind stays `config_error` and the exit code stays 2. The message names the directory and what it holds.
- **The remedy** comes from one helper, `output_directory_remedy(default=, kind=)`, shared by the message, `verify`'s `review` next action and `agent control`'s review step. For any directory but the default: omit `--out`, or name a directory that is gitignored or outside the repository. For the default `agents-shipgate-reports` (by physical identity, whether or not `--out` names it), omitting `--out` is never offered: move stray files out, gitignore it, or pass such an `--out`; committed reports need the untracking change, verified with such an `--out` until it merges.
- The existing overlap refusals are unchanged and still come first.

**Readers:**
- `live_workspace` resolves the Git root, then classifies the reports directory against `HEAD` before any other Git read. An unsafe directory returns a `LiveWorkspace` carrying `reports_dir_refusal`.
- Otherwise the workspace carries `reports_dir_refusal_against(commit)`. `_validate_control_currency` calls it with the pointer's recorded `merge_base_sha` for worktree pointers, which catches a legacy pointer that hid a committed deletion. It skips that for `committed_tree` pointers, matching the writer's `--head` rule.
- The check runs first and refuses as `workspace_unverifiable` for every pointer state.
- The classification is never inside the `try` that turns exceptions into `live=None`. `None` withholds completion only and returns every other state unchecked (the #813 path). A classification that cannot run therefore refuses instead of falling into that path.
- This covers `agent control`, `verify --format control` (the run's own in-process read), and the human-review decision reader, which all go through `live_workspace`.
- **Recovery:** for this refusal, `agent control` replays the producing run's own `fix_task.verification_command` with only `--out <v>`/`--out=<v>` removed (`_recovery_away_from`), so it publishes into the default directory with the same `--config`, base, `--head`, policy and baseline options. A rebuilt bare `verify` dropped `--base`, skipped a local base, and read a `human_review_required` change as `complete`. It falls back to `verify` without `--out` only when there is no parseable producing command, and gives a `review` step when the refused directory is the default one.
- **The Stop hook** passes no `--out` (a parity assertion pins that), so it publishes into the default directory and the writer rule applies unchanged. When the default directory is refused, the hook's `verify` exits 2, which the hook, as for every exit other than 0 and 20, surfaces as advisory context without blocking the Stop. STABILITY.md, CHANGELOG, `docs/integrations.md` and `docs/troubleshooting.md` say so.

**Surface discipline:** no new command, flag, schema, schema version, contract version, error kind, refusal code, exit code or discovery surface.
- The change narrows an existing flag contract (`verify --out`) and extends the one currency engine (`read_current_control` / `live_workspace`) rather than adding a parallel check.
- It also consolidates the output-directory rule into one helper that the writer and every reader share.
- The headline metric it moves is the correctness of the local control answer for the product's headline case, a Claude Code shell-permission change. `--out` can no longer turn a widening, or a deny-rule removal, into `complete`/`mergeable`.
- No `docs/distribution-surfaces.md` row makes a claim about output directories, so the registry is unchanged. `tests/test_distribution_surface_parity.py` passes.

**Compatibility:** documented in a STABILITY.md migration note, the Verify Orchestrator section, a CHANGELOG Unreleased entry, `docs/agent-contract-current.md` (with `llms-full.txt` regenerated), `docs/troubleshooting.md`, `docs/integrations.md`, `verify --out` help and the `action.yml` `output_dir` description.
- Pointers published into such a directory by `1.0.0` stop reading as current after upgrade.
- The Action's `output_dir` follows the same rule.
- A default `agents-shipgate-reports` holding committed files, or unrelated unignored files, is refused like any other directory.

## Deviations from the agreed scope, and deferrals

- **"Tracked" now means committed, plus the commits the change is compared against.** The agreed rule ("any tracked path") would have missed a directory that a staged or committed deletion leaves behind. Both reproduce as `complete`/`mergeable` over `SHIP-HOST-BOUNDARY-PERMISSION-DENY-REMOVED`, as measured above. So the rule also counts what `HEAD` and the compared commits held there.
- **Staged artifacts are allowed.** Staged paths that are Shipgate artifacts are judged like untracked ones. The existing advisory contract, "the staged-reports nudge never changes the verdict or the exit code" (`test_verify_warns_when_reports_directory_is_staged`), would otherwise have become a refusal. A staged non-artifact is still refused.
- **Added beyond the spec, each following from the same rule and each tested:**
  - the trust-root rule (a name alone never grants the artifact allowance to a trust-root path);
  - the stored-spelling respelling below the root;
  - `agent control`'s recovery routing away from the refused directory while keeping the producing run's request.
- **Deferred:**
  - **A gitignored output directory is still left out of the static input census as a whole.** The refusal does not reach it, because the agreed scope keeps gitignored directories accepted. So a manifest-named input beneath one (`tool_sources: path: gen/tools.json`, `gen/` gitignored, `--out gen`) is read by the run but not bound: the plan's `inputs.tool_sources` is `[]`, and `agent control` stays `complete` after the file is edited. With the default output directory, the same edit makes it exit 4. `main` behaves the same way. This is documented under "Not changed here" in the STABILITY note. The fix is to narrow the `StaticInputSnapshot` exclusion to the files a run writes, which changes the `input_directories.excluded_paths` identity that replay validates, so it needs its own change;
  - validating the plan-recorded `input_directories.excluded_paths` on replay;
  - `verification prepare`/`assemble`, whose artifacts root is excluded from static inputs but not from the Git change set, and which the writer refusal does not cover.
- **Configured-preview refresh:** the end-to-end assertion stays with #807. That route refuses as `workspace_unverifiable` for an unrelated reason on main. The write-time refusal for preview is tested here.
- **Scan pointers:** `scan` does not exclude its output from any change set and gets no writer refusal. A `scan` pointer in such a directory is refused by the same readers.
- **#813:** the `live=None` fail-open when `ensure_git_workspace` itself fails is unchanged and remains #813's. That includes a `--workspace` that is not a Git checkout with `--reports-dir` inside a repository. This change adds no new route into it.

## Tests

New `tests/test_output_directory_content.py` (53 tests, real Git repositories):
- Every writer route (worktree, `--head`, `--preview`) refuses a tracked `docs/`, writes nothing, and leaves `git status` clean. Manifest-free `verify` refuses `.claude`.
- A manifest-named tool-source directory is refused. The default run binds `tools/tools.json` in the plan.
- A `.claude` widening, committed and uncommitted: the default run is `human_review_required` with `SHIP-HOST-BOUNDARY-PERMISSION-ALLOW-EXPANDED`, and `--out .claude` is refused.
- An untracked new `.claude/settings.json` is refused.
- A deny-rule deletion, staged, committed with `--base main`, and committed with an auto-detected base: the default run reports `SHIP-HOST-BOUNDARY-PERMISSION-DENY-REMOVED`, and `--out .claude` is refused although the directory is gone.
- Trust-root files named like artifacts (`.claude/commands/packet.md`, a staged `.claude/skills/verification-inputs/SKILL.md`, `.agents-shipgate/capabilities.lock.json`): the writer refuses, and a pointer published with the writer disabled refuses on refresh. A gitignored directory inside a trust root still works.
- A legacy pointer that hid a committed deletion refuses on read, through the pointer's merge base.
- Symlink aliases, both to the repository and to `.claude`.
- Case variants of the repository directory and of `.claude`, plus a safe case-variant directory that now works. Skipped where the filesystem is case-sensitive.
- The default directory, a gitignored in-repo directory, an unignored artifacts-only directory (with a `scan` of every packet renderer in between) and a sibling outside the repository: each verifies twice, refreshes `complete`, and still refuses drift.
- `verify --no-base` twice into an unignored default directory for each of the 16 shipped samples with a root manifest: every file left behind is a recognized artifact outside any trust root, and `agent control` exits 0.
- `skill lint`, `skill security` and `skill review` with their default output, between two runs of `verify` into an unignored default directory: `agent control` and the second `verify` both exit 0 with `complete`.
- The `--head` decision is identical for the default and a gitignored directory, and an archived `--head` run may remove committed files from its output directory while the worktree run of the same change is refused.
- A manifest-free run compares the local base its host comparison uses.
- A non-Git `verify --preview`, with the default and with `--out notes`, followed by `agent control`, exits 0.
- Legacy pointers in `complete`, plan-bound `review_publishable`, and plan-less `agent_action_required`/`human_review_required`/`review_publishable` states each read as current with classification disabled, and refuse as `workspace_unverifiable` with it.
- The recovery for a pointer published into `docs` keeps the producing run's `--config` and `--base`; running it as printed gives `human_review_required`.
- A default directory with committed content (worktree and the Action's `--out agents-shipgate-reports --head HEAD`) or stray files is refused, and neither the message nor the recovery offers omitting `--out`.
- A directory that gains content after the run refuses on refresh.
- An inventory failure refuses both the refresh of a non-completion pointer and the next `verify`.
- Classifier unit cases: outside, absent, ignored, artifacts-only, staged artifact, staged non-artifact, committed artifact name, staged deletion, committed deletion against a compared commit, an unresolvable compared ref, trust-root directories and paths, and the root and its ancestors.
- An allowlist test holds the artifact names to every registry: the verifier route names, the pointer bindings, contract `ARTIFACTS`, the scan-removal list, every literal `out_dir / "<name>"` in orchestrator, scan and packet writers, `scenario suggest`'s default, the Action's scripts and `action.yml`, and `skill.runner.REPORT_BASENAME` in every format `_write_reports` renders.

Existing tests changed:
- `tests/test_current_control_closure.py`: the symlink-damage substitute now lives outside the repository instead of as an unignored non-artifact file beside the reports.
- `tests/test_install_hooks.py`: asserts that the Stop hook passes no `--out`.

Mutation checks for each review cycle, reverted afterwards, are listed in the review-cycle comments. CI runs the full suite.

## After

Same scenarios on this branch:

| Scenario | Result |
| --- | --- |
| `--out .claude` with the widening (uncommitted, committed, `--head`, manifest-free, `--preview`) | exit 2 `config_error`: `Verifier --out .claude holds tracked repository files (.claude/settings.json)…` |
| New untracked `.claude/settings.json` | exit 2: `holds untracked files that Git does not ignore and that are not Shipgate artifacts (.claude/settings.json)` |
| Deny-rule deletion, `--out .claude` | staged: exit 2 `held committed repository files that the change being verified removes (.claude/settings.json)`; committed: exit 2, same reason |
| `--out docs`, `--out tools` | exit 2, nothing written, nothing left for a refresh to read |
| `<symlink>/.claude`, `<parent>/REPO/.claude`, `<repo>/.CLAUDE` | exit 2, naming `.claude/settings.json` |
| `--out .claude/commands` over an untracked `packet.md` | exit 2: `lies inside a trust root (agent_instructions) that Git does not ignore…` |
| Default output, gitignored in-repo, unignored artifacts-only in-repo, sibling outside | verify ×2 and refresh `complete`, as before |
| Unignored default directory, `verify` → `skill review` / `lint` / `security` → `agent control` → `verify` | all exit 0 |
| `<repo>/PLAIN-REPORTS` for an existing artifacts-only `plain-reports` | main: exit 3 "path changed identity while it was read"; branch: `complete`, refresh `complete` |
| Non-Git `verify --preview` (default and `--out notes`) → `agent control` | exit 0, as before |
| Pointers written by main into `docs` (`complete`, `review_publishable`) and `.claude` (manifest-free preview, `agent_action_required`), read by this branch | main reads them as current (exit 0); branch: exit 4 `workspace_unverifiable`, naming the directory and what it holds |


pengfei-threemoonslab (commented):
Independent coding-agent review/address cycle 1 at `01b67fe`.

**Not mergeable yet: three P1 and two P2 findings.** The core refusal works for the cases the PR tabulates: `--out .claude` over a widened, new or deleted `.claude/settings.json`, `--out docs`, `--out tools`, and the symlink and case-variant spellings all exit 2 before writing on `01b67fe`, and legacy `1.0.0` pointers in those directories refuse on read. CI is green on `01b67fe` (11 pass, `release-tag-consistency` skipped). Three reviewers checked this independently. Every P1 and P2 finding below was re-run for this comment with `./shipgate` from `01b67fe`, and compared with `9df307ab` where noted, in throwaway repositories built from `samples/clean_read_only_agent` (`.gitignore` = `agents-shipgate-reports/`, branch `feature` off `main`) or `samples/conductor_agent`.

1. **P1: a name-only artifact allowlist still lets `--out` hide a new host-config file under a trust root, and the result is `complete` / `merge: true`.**
   - **Evidence:** `_is_reports_artifact` (`cli/current_workspace.py:300`) accepts any uncommitted path named in `REPORTS_DIRECTORY_ARTIFACT_NAMES`, and anything under a `verification-inputs/` subdirectory. It never asks whether that path is a trust root.
     - An untracked `.claude/commands/packet.md` with `allowed-tools: Bash(curl:*), Bash(rm -rf:*)`. With the default output, `verify --base main` is `review_publishable`, merge=false, `human_review_required`, with `SHIP-VERIFY-TRUST-ROOT-TOUCHED` and `SHIP-AGENT-BOUNDARY-INPUT-INCOMPLETE`. With `--out .claude/commands` it exits 0: `complete`, merge=true, `mergeable`, no findings. The command file is left intact.
     - A staged `.claude/skills/verification-inputs/SKILL.md` (a skill named `verification-inputs`, with `allowed-tools: Bash(curl:*), Bash(rm:*)`). With the default output: `human_review_required`, with `SHIP-AGENT-BOUNDARY-PROTECTED-SURFACE-UNCLASSIFIED` and `SHIP-VERIFY-TRUST-ROOT-TOUCHED`. With `--out .claude/skills`: `complete`, merge=true, `mergeable`, no findings. The staged-reports warning even calls the SKILL.md a "generated Agents Shipgate report file". `agent control --reports-dir .claude/skills` then exits 0 `complete`.
     - An untracked `.agents-shipgate/capabilities.lock.json`. With the default output: `human_review_required`. With `--out .agents-shipgate`: `complete`, merge=true.
     - All three are the same on `9df307ab`, so this is not a regression. But the PR closes #804, whose P1 class is a trust-root change turned into `complete` / `mergeable` through `--out`. The PR body says "`--out` can no longer turn a widening … into `complete`/`mergeable`", and `_artifact_lifecycle.py:29-36` says an artifacts-only directory "can be left out of the change set without hiding anything a change could carry". Neither claim holds for these paths. `packet.md`, `report.md`, `pr-comment.md`, `capabilities.lock.json` and `verification-inputs/**` collide with real inputs under `**/.claude/**`, `**/.agents/skills/**`, `**/.codex/**`, `**/prompts/**`, `**/policies/**`, `**/.agents-shipgate/**` and `**/SKILL.md`.
   - **Fix:**
     - Do not grant the artifact allowance (untracked or staged) to any path that `core/trust_roots.TRUST_ROOT_SURFACES` classifies.
     - The simpler option is to refuse any in-repository, non-ignored output directory that is itself inside a trust-root surface. A gitignored one hides nothing.
     - Add real-Git tests, untracked and staged, at write time and on refresh, for `.claude/commands/packet.md`, `.claude/skills/verification-inputs/SKILL.md` and `.agents-shipgate/capabilities.lock.json`.

2. **P1: `agent control`'s new recovery for this refusal drops the producing run's `--base` and `--config`. Following it exactly as printed gives `complete` / `merge: true` over a `human_review_required` decision.**
   - **Evidence:**
     - On `9df307ab`, I committed a `.claude/settings.json` widening and ran `verify --config shipgate.yaml --base main --out docs` over a tracked `docs/a.txt`. The result was `review_publishable` / `human_review_required` with `SHIP-HOST-BOUNDARY-PERMISSION-ALLOW-EXPANDED`, and `fix_task.verification_command` = `… verify --config shipgate.yaml --base main --workspace <repo> --out docs --json`.
     - On `01b67fe`, `AGENTS_SHIPGATE_AGENT_MODE=1 agent control --reports-dir docs` exits 4 as expected. But `next_actions[0]` is `kind: command`, `verify --workspace <repo> --json` (`_recovery_away_from`, `cli/agent_interface.py:418`).
     - Running that command exactly as printed exits 0: `complete`, merge=true, `mergeable`, no findings. `base_notes` says "Skipped local base 'main' for implicit auto-base…". The following `agent control` is exit 0, `complete`, merge=true.
     - The validated artifacts that hold the exact producing command are in hand at this point (`_validate_control_currency(..., artifacts=validated)`). The new raise at `core/current_control.py:749-762` does not attach them, and `_recovery_away_from` rebuilds a bare command. `_superseded_recovery_command`'s own docstring records dropping `--no-base`, policy and baseline options as the #429 defect. No test runs the recovery command and checks what it yields.
     - (On `9df307ab` this pointer read as current, so there was no recovery to regress from. The defect is in the route this PR adds.)
   - **Fix:**
     - Attach `artifacts=artifacts` to the reports-directory refusal.
     - In `_recovery_away_from`, start from the bound verifier's `fix_task.verification_command` and remove only `--out <value>`, keeping `--config`, `--base`/`--no-base`/`--head`, policy packs and baseline. Fall back to `verify_command_for(workspace, None)` only when there is no producing command.
     - Add a test that runs the printed recovery command and asserts the decision is still not `complete`.

3. **P1: the allowlist misses `suggested-inventory.json`, which `verify` writes into its own output directory. An unignored default `agents-shipgate-reports` then refuses its own next run.**
   - **Evidence:**
     - `cli/scan/writing.py:343` writes `anchor.parent / SUGGESTED_INVENTORY_FILENAME`, which is not in `REPORTS_DIRECTORY_ARTIFACT_NAMES` (`_artifact_lifecycle.py:37`). `test_every_name_a_run_writes_is_a_recognized_artifact` only picks up literal `out_dir / "<name>"` writes, so it misses this one.
     - Setup: `samples/conductor_agent` in a fresh `git init` with no `.gitignore`, so the reports directory is not ignored. `install-hooks` does not add one.
     - On `01b67fe`: `verify --no-base` exits 0 and writes `suggested-inventory.json`. `agent control` then exits 4 `workspace_unverifiable` ("holds untracked files … not Shipgate artifacts (agents-shipgate-reports/suggested-inventory.json)"), and a second `verify --no-base` exits 2 `config_error`.
     - On `9df307ab`, the same sequence gives verify 0, `agent control` 0 and verify 0.
     - This is the configuration the CHANGELOG lists as "Unchanged" (the default directory, and an artifacts-only one).
     - From reading the code: the Stop hook treats any verify exit other than 0 or 20 as advisory context that "did not block completion" (`cli/install_hooks.py:1166-1183`). From the second Stop on, it stops blocking in these repositories.
     - The same gap covers `suggested-scenarios.yaml` (the default of `scenario suggest`, `cli/scenario.py:194`, documented at `docs/agent-recipes.md:360`), and `check-annotations.json` and `check-run-payload.json`, which the Action scripts write into `output_dir`.
   - **Fix:**
     - Add `SUGGESTED_INVENTORY_FILENAME`, `suggested-scenarios.yaml`, `check-annotations.json` and `check-run-payload.json`, or derive the set from the writers.
     - Replace or extend the regex registry test with an empirical one: run `verify` twice over the shipped samples (including a low-confidence one) into an unignored in-repository directory, and assert that every produced name passes `_is_reports_artifact` and that both `verify` runs and `agent control` exit 0.

4. **P2: when the refused directory is the default, `verify`'s next action loops and contradicts the documented remedy.**
   - **Evidence:**
     - In the repository from finding 3, `AGENTS_SHIPGATE_AGENT_MODE=1 verify --no-base` (no `--out`) exits 2 with `next_actions[0].why`: "Choose an output directory other than …/agents-shipgate-reports …: omit --out to use the default agents-shipgate-reports, or name a directory that is gitignored or outside the repository. Do not edit .gitignore or move repository files to get past this refusal."
     - The caller already omitted `--out`. "Do not edit .gitignore" contradicts STABILITY.md:167 ("Move them out, or gitignore the reports directory") and the stderr remedy ("Keep only Shipgate artifacts there").
     - The handler text at `cli/verify/command.py:317-338` is constant. It does not branch on the directory being the default.
     - The Action's shape, `--out agents-shipgate-reports --head HEAD`, takes the explicit branch in `_reject_output_directory_content`, so its message also says "Omit --out to use the default agents-shipgate-reports".
     - The reader's refusal message (`core/current_control.py:749-762`) also suggests "the default agents-shipgate-reports" when the default is the directory refused.
     - `test_a_default_directory_holding_content_is_refused_without_a_looping_route` asserts only `agent control`'s route, not `verify`'s own `next_actions`.
   - **Fix:**
     - Decide "is the default" by physical identity against `default_reports_dir(workspace)`, not by `out is not None`, and carry it on `OutputDirectoryHoldsRepositoryContent`.
     - When the default is refused, never suggest omitting `--out`. Name the removal of the non-artifact files, or an `--out` that is gitignored or outside the repository.
     - Keep "do not edit .gitignore / move repository files" for directories holding repository content such as `.claude` or `docs`, or reword STABILITY so the two agree.
     - Use the same helper for the core message and `_recovery_away_from`; its `realpath` comparison is the P3 below.
     - Assert `verify`'s `next_actions` in the looping-route test.

5. **P2: `--head` runs, which are the Action's shape, now exit 2 for a change that removes committed files from the output directory. `--head` never hid that removal.**
   - **Evidence:**
     - `main` commits `agents-shipgate-reports/report.json`, and `feature` runs `git rm -r agents-shipgate-reports`, adds a `.gitignore` entry and commits.
     - On `9df307ab`, `verify --base main --head HEAD --out agents-shipgate-reports` exits 0 with `changed_files: ['.gitignore', 'agents-shipgate-reports/report.json']`. The removal is visible, because the `archive_head` route does not pathspec-exclude `out_dir` (`orchestrator.py:729`), and the head-tree census cannot see a file the head no longer has.
     - On `01b67fe` it exits 2: "…held committed repository files that the change being verified removes … that content would be hidden from the decision". That is false for `--head`.
     - The same holds for a trust-root deletion. With a committed `git rm .claude/settings.json` that removes deny rules, `9df307ab` `--head HEAD --out .claude` already gives `human_review_required` with `SHIP-HOST-BOUNDARY-PERMISSION-DENY-REMOVED`. `01b67fe` refuses it.
     - So the PR that stops committing reports fails the Action's CI, although the CHANGELOG says "The Action's `--head` decisions were never affected".
     - Content committed at `HEAD` is still correctly refused under `--head`, because the head-tree `StaticInputSnapshot` excludes `out_dir` (`orchestrator.py:1124-1128`).
   - **Fix:**
     - In `_reject_output_directory_content`, when `archive_head` is set, compare `HEAD`/the evaluated head only. Do not add the merge base, whose removals the committed diff already reports.
     - Keep the merge-base comparison for worktree runs.
     - Add a `--head` test for a committed removal beneath the output directory.

Non-blocking (P3), consolidated across reviewers:
- **Case-variant recovery routing.** `_recovery_away_from` decides whether the directory is the default with `os.path.realpath` (`agent_interface.py:402`), which does not fold case. On macOS, `agent control --reports-dir AGENTS-SHIPGATE-REPORTS` over a refused default directory returns a `command` recovery, `verify --workspace <repo> --json`, that publishes back into the refused directory. Compare `_identity`, as the classifier does.
- **Non-Git `--workspace`.** `live_workspace` returns `None` before classifying when `--workspace` is not a Git checkout. So `agent control --workspace <non-git> --reports-dir <repo>/.claude` still reads a legacy non-complete pointer. The PR defers this to #813; record this caller-chosen-workspace case on #813 explicitly.
- **Committed artifacts in the default directory.** The refusal text says "Keep only Shipgate artifacts there" when committed reports are all the directory holds. In worktree mode the untracking PR is itself refused until it merges, so its only route is a different `--out`. Name that case and its working remedy.
- **Help text.** `verify --out` help ("Output directory for verifier and scan artifacts.", `command.py:135`) and `action.yml` `output_dir` ("Output directory for reports.") do not mention the narrowed contract. `docs/troubleshooting.md` quotes only the `Verifier --out <dir> holds …` form, not the default-directory form.
- **"Unchanged" wording.** The CHANGELOG "Unchanged" bullet, the STABILITY Verify Orchestrator paragraph and `docs/agent-contract-current.md`/`llms-full.txt` say a gitignored or artifacts-only directory is accepted, without "and nothing committed beneath it". A committed `out/.gitignore` (`*`, `!.gitignore`) exits 0 on `9df307ab` and 2 on `01b67fe`. The migration note's rule section is already precise.
- **Staged-deletion wording.** For a staged `git rm .claude/settings.json`, where `.claude` no longer exists, the message says the directory "holds tracked repository files". Use the "held … that the change being verified removes" wording when the `HEAD` path is gone from the index and the working tree.
- **Deletion bypass attributed to `main`.** The CHANGELOG and STABILITY say the deny-rule deletion bypass held "on `main`". Per one reviewer's v1.0.0 worktree, it reproduces on published `1.0.0` too (`--out .claude` → `complete`, no findings). Fold it into the `1.0.0` sentence, and drop "on `main`" from release notes.

Verified without findings:
- The writer check runs after `_reject_output_input_overlap` and before `mkdir`/`begin_current_control` in `run_verify` and `run_preview`. The manifest-free route re-checks in `run_preview`, and non-Git preview is skipped.
- Classification in `live_workspace` sits outside the `try` that maps exceptions to `None`, and `output_directory_refusal` never raises. `_validate_control_currency` checks the refusal first for every state, against `HEAD` and the pointer's `merge_base_sha`.
- The classifier pathspec `:(top,literal)` and the change-set exclusion come from the same `worktree_exclusion` path and the same `--exclude-standard` untracked inventory.
- No schema, contract version, error kind, refusal code or exit code moves. `docs/distribution-surfaces.md` makes no output-directory claim, and the parity test passes.
- The Stop hook passes no `--out`.
- Tests run here on `01b67fe`: `tests/test_output_directory_content.py` gave 26 passed. Across the three reviewers, these also passed: `test_current_control_closure`, `test_install_hooks`, `test_agent_control_reports_dir`, `test_current_control`, `test_agent_control_envelope`, `test_current_control_directory_currency`, `test_current_control_input_currency`, `test_adapter_static_only`, `test_distribution_surface_parity`, `test_public_surface_contract`, `test_agent_control_contract`, `test_docs_links` and `test_verify`.
- One reviewer disabled the reader raise, and exactly 5 tests in the new file failed, as the PR states.

pengfei-threemoonslab:
**Addressed review cycle 1.** New head `5e042f6a`: the cycle's fixes are in `5e042f6a` on top of the original commit, and both are rebased onto `origin/main` `acd67c2b`. The only conflict was the CHANGELOG `Unreleased` list, where both entries are kept. All five P1/P2 findings are fixed, and so are all P3s except the one that needs an issue edit. The evidence below was re-run with `./shipgate` from `5e042f6a` in throwaway repositories built from `samples/clean_read_only_agent` and `samples/conductor_agent`, as in the review.

### 1. P1: an artifact name allowed a trust root → fixed

- **Path rule:** the classifier no longer gives the artifact allowance to any path `core/trust_roots.trust_root_class_for` classifies, whatever its name. `_artifact_shaped` now checks only name and place; the trust-root check is separate.
- **Directory rule:** a directory inside a trust root that Git does not ignore is refused before anything is written there, even while empty. `output_directory_inventory` probes the names a run would write with `git check-ignore --stdin -z`, which still sits below `_run_git`, so the static-only allowlist pins don't move. So `--out .claude/reports` and `--out .github/workflows` (where `suggested-declarations.yaml` would be a workflow) are refused, and a gitignored `.agents-shipgate/reports/` still works.
- **Message order:** untracked files that are not artifacts are named first. A directory whose only content is artifact-named trust-root files gets the trust-root message.
- **Evidence (`--base main`):**

| Planted file | Default output | `--out` over its directory |
| --- | --- | --- |
| untracked `.claude/commands/packet.md` | exit 0, `review_publishable`, merge=false | `--out .claude/commands`: **exit 2** `lies inside a trust root (agent_instructions) that Git does not ignore…` |
| staged `.claude/skills/verification-inputs/SKILL.md` | exit 0, merge=false | `--out .claude/skills`: **exit 2**, same reason |
| untracked `.agents-shipgate/capabilities.lock.json` | exit 0, merge=false | `--out .agents-shipgate`: **exit 2** `(shipgate_state)` |

- **Tests:**
  - `test_a_trust_root_named_like_an_artifact_cannot_be_hidden`, for all three: the default run has merge=false, the writer refuses and writes nothing, and a pointer published into the directory with the writer disabled refuses on refresh (`workspace_unverifiable`).
  - `test_a_gitignored_directory_inside_a_trust_root_hides_nothing`.
  - New classifier unit cases: `.claude/reports`, `.github/workflows`, an ignored `.agents-shipgate/reports`, and `rpt2/verification-inputs/.claude/settings.json`.

### 2. P1: the recovery dropped `--base`/`--config` → fixed

- **Change:** `_recovery_away_from` now takes the bound verifier's `fix_task.verification_command` and removes only `--out <v>`/`--out=<v>`, using `split_invocation` and `join_argv`. Every other token stays. It falls back to `verify_command_for(workspace, None)` only when there is no parseable producing command. `_superseded_recovery_command` shares the same `_producing_verification_command` helper.
- **On attaching artifacts to the raise:** no change was needed. The read loop already sets `stale.artifacts = dict(captured)` on every `CurrentControlUnavailable` raised from `_validate_control_currency`, this refusal included (`core/current_control.py`, both observation passes). The bytes were there; the route didn't use them.
- **Evidence** (the review's setup: committed widening, tracked `docs/a.txt`, pointer published into `docs` as `1.0.0` did):
  - `agent control --reports-dir docs` exits 4. `next_actions[0].args` = `['verify', '--config', 'shipgate.yaml', '--base', 'main', '--workspace', <repo>, '--json']`.
  - **Running it as printed:** exit 0, `merge_verdict: human_review_required`, `.claude/settings.json` in `changed_files`.
  - The following `agent control` exits 0 with `review_publishable`, merge=false.
- **Test:** `test_the_recovery_keeps_the_producing_runs_request` runs the printed args and asserts the decision. With the recovery reverted to the bare command and the argv asserts removed, it fails on `assert 'mergeable' == 'human_review_required'`.

### 3. P1: allowlist missed `suggested-inventory.json` → fixed

- **Names added:** `suggested-inventory.json`, `suggested-scenarios.yaml`, `check-annotations.json` and `check-run-payload.json`.
- **Registry test:** it now also reads `SUGGESTED_INVENTORY_FILENAME`/`SUGGESTED_DECLARATIONS_FILENAME`, `scenario suggest`'s `from_path.parent / "…"` default, the three Action scripts (including `PAYLOAD_FILENAME`), and `action.yml`'s `${OUTPUT_DIR}/…` and `path.join(outputDir, …)` writes.
- **Empirical test:** `test_every_file_a_real_run_leaves_is_a_recognized_artifact` covers all 16 shipped samples with a root manifest, in a fresh `git init` with no `.gitignore`. It runs `verify --no-base` twice into the unignored default directory. Every file left behind must be artifact-shaped and outside any trust root, the classifier must return `None`, both runs must exit 0, and `agent control` must exit 0.
- **Evidence (`conductor_agent`):** verify 0 → `agent control` 0 (`review_publishable`) → verify 0, and `suggested-inventory.json` is written.

### 4. P2: the default directory's next action looped → fixed

- **One helper:** `output_directory_remedy(default=, kind=)` in `cli/current_workspace.py` now writes the writer's message, `verify`'s `next_actions`, and `agent control`'s review step.
- **Detecting the default:** `is_default_reports_dir(workspace, directory)` compares physical identity (`_identity`) against `default_reports_dir(workspace)`. `OutputDirectoryHoldsRepositoryContent` carries `default` and the structured `OutputDirectoryRefusal(kind, reason)`.
- **The default directory is never told to omit `--out`.** Its remedy depends on what was found:
  - untracked or staged strays: move them out, gitignore the directory, or pass `--out`;
  - committed or removed files: pass `--out`, and the untracking change (`git rm -r --cached agents-shipgate-reports` plus a gitignore) needs such an `--out` until it merges (your P3);
  - a trust root: gitignore it, or pass `--out`.
- **Other directories:** the advice stays "omit --out or name a gitignored/outside directory", and now says not to gitignore, untrack or move that directory's files. STABILITY now agrees: gitignoring or moving strays out is listed only for the reports directory.
- **Core message:** it no longer recommends the default directory.
- **Evidence:**
  - Stray `notes.md`: exit 2, `…Move the files that are not Shipgate artifacts out of agents-shipgate-reports (unstaging any that are staged), gitignore agents-shipgate-reports, or pass --out naming…`.
  - Committed README, with no `--out` and with the Action's `--out agents-shipgate-reports --head HEAD`: exit 2, and "omit --out" is offered in neither.
- **Tests:** `test_a_default_directory_holding_committed_files_is_refused_without_a_looping_route[worktree|action]` asserts `verify`'s own `next_actions` and the `agent control` route. It also covers the case-variant `--reports-dir AGENTS-SHIPGATE-REPORTS` (your P3; skipped on case-sensitive filesystems) and checks that the named way out works. `test_a_default_directory_holding_stray_files_names_moving_them_out` covers the stray-file case.

### 5. P2: `--head` refused a removal its own diff shows → fixed

- **Writer:** `_reject_output_directory_content(archive_head=…)` compares only the evaluated head for `--head` (and `--preview --head`). Worktree runs keep the merge base.
- **Reader:** `_validate_control_currency` skips the recorded merge base for `snapshot_kind == "committed_tree"` pointers, which otherwise would have refused the Action's pointer on read.
- **Evidence:** `main` commits `agents-shipgate-reports/report.json`, and `feature` runs `git rm -r` and adds a gitignore. Then:
  - `--base main --head HEAD --out agents-shipgate-reports`: exit 0, `changed_files: ['.gitignore', 'agents-shipgate-reports/report.json']`, and `agent control` exits 0.
  - The worktree run of the same change is still refused, with the untracking remedy.
- **Test:** `test_a_head_run_may_remove_committed_files_from_its_output_directory`.
- **Declined in part:** `--head HEAD --out .claude` after a committed `git rm .claude/settings.json` still exits 2. The reason is now the trust-root rule from finding 1, not the removal: `lies inside a trust root (agent_instructions)…`. Every report the Action wrote there would be a `.claude/**` file, and the reader applies the same rule. I don't see a legitimate `output_dir: .claude`.

### Found while fixing: manifest-free base

- **The gap:** a manifest-free worktree run's host comparison auto-detects its base with `allow_local_when_no_remote=True`, and the pointer records that merge base. The writer resolved only the configured verifier's base, which skips a local `main`.
- **Effect:** with no remote and a committed `git rm .vscode/mcp.json`, `verify --out .vscode` was accepted and published human_review_required. `agent control` then refused it (exit 4). That is not a merge bypass, but the writer and reader disagreed.
- **Fix:** without a manifest, the writer now also compares that base, and the run exits 2 with `removes (.vscode/mcp.json)`.
- **Test:** `test_manifest_free_verify_compares_the_base_its_host_comparison_uses`.

### P3s

- **Case-variant recovery:** fixed (finding 4).
- **Committed artifacts in the default directory:** fixed; the case and its remedy are named (finding 4).
- **Help text:** `verify --out` help and the `action.yml` `output_dir` description now state the rule. `docs/troubleshooting.md` quotes the default-directory and trust-root forms too.
- **"Unchanged" wording:** CHANGELOG, the STABILITY Verify Orchestrator paragraph and `docs/agent-contract-current.md` (with `llms-full.txt` regenerated) now say "uncommitted", "outside any trust root" and "nothing committed beneath it". The migration note names a committed `out/.gitignore`.
- **Staged-deletion wording:** a HEAD path gone from both index and disk is now reported as `held committed repository files that the change being verified removes (…)`. The staged deletion test and the classifier unit case assert this.
- **Deletion bypass attributed to `main`:** folded into the `1.0.0` sentence in both CHANGELOG and STABILITY.
- **Non-Git `--workspace`:** this is not fixed here. It is recorded in the STABILITY migration note's "Not changed here" list, which cites #813. I can only reference issues from the PR, so the `--workspace <non-git> --reports-dir <repo>/.claude` case should be added to #813 by hand.

### Mutation checks

Each was reverted afterwards.
- Trust-root class → `None`: 4 tests fail (the 3 trust-root cases and the classifier unit test).
- Recovery without the producing command: `recovery_keeps` fails.
- Dropping `suggested-inventory.json`: the registry test and `real_run_leaves[conductor_agent]` fail.
- `--head` comparing the merge base: `head_run_may_remove` fails.
- Reader not skipping `committed_tree`: `head_run_may_remove` fails.
- Writer deciding the default by `not explicit`: `looping_route[action]` fails.
- Reader deciding the default by string: `looping_route[worktree, action]` fail.
- No manifest-free base: `host_comparison_uses` fails.
- Staged deletion counted as held: `deletion…[staged]` and the classifier test fail.

### Tests run

- **New file:** `tests/test_output_directory_content.py`, 52 passed, before and after the rebase.
- **Related files, run one process per file:** 82 files that reference `verify`, `agent control`, current control, the Stop hook, the Action or the changed modules all pass, except `test_codex_boundary_check.py::test_codex_check_boundary_json_golden_outputs`. That test fails identically without this change: the TOML parse-error evidence differs under this Python 3.14 interpreter.
- **After the rebase:** `test_check_unmodelled_host_config_keys` (from #814), `test_public_surface_contract`, `test_docs_links`, `test_distribution_surface_parity`, `test_agent_control_reports_dir`, `test_current_control_closure`, `test_install_hooks`, `test_verify`, `test_action_metadata`, `test_adapter_static_only` and `test_manifest_free_pr_rows` all pass.
- **Also:** `python -m ruff check .` is clean, and `scripts/generate_schemas.py --check` passes.
- **Surfaces:** no schema, contract version, error kind, refusal code or exit code moved. `docs/distribution-surfaces.md` makes no output-directory claim, so the registry is unchanged; the `action.yml` input description change doesn't touch its `merge_verdict_vocabulary` claim.

pengfei-threemoonslab (commented):
Independent coding-agent review/address cycle 2 at `5e042f6`.

**Not mergeable yet: one P2 finding.** The five cycle-1 findings are fixed at `5e042f6`, and the core refusal holds. CI is green on `5e042f6` (11 pass, `release-tag-consistency` skipped). Everything below was re-run for this comment with `./shipgate` from `5e042f6` and from `origin/main` `acd67c2b`, in throwaway repositories built from `samples/clean_read_only_agent` (`.gitignore` = `agents-shipgate-reports/`, branch `feature` off `main`) unless noted.

1. **P2: the artifact allowlist still misses the reports that `skill lint`, `skill security` and `skill review` write into the default `agents-shipgate-reports`. An unignored default directory then refuses its own next run and every refresh.** This is the same class as cycle-1 finding 3; the fix covered `verify`, `scan`, `scenario suggest` and the Action, but not these commands.
   - **Evidence (code):** `skill/runner.py:86` writes to `workspace / "agents-shipgate-reports"` when `--out` is omitted, and `_write_reports` (`runner.py:160`) names the files `{skill-lint|skill-security|skill-review}.{json,md,sarif}` (`REPORT_BASENAME`, `runner.py:31`). None of these names is in `REPORTS_DIRECTORY_ARTIFACT_NAMES` (`cli/_artifact_lifecycle.py:41`). Neither `test_every_name_a_run_writes_is_a_recognized_artifact` nor `test_every_file_a_real_run_leaves_is_a_recognized_artifact` reads `skill/runner.py` or runs a `skill` command.
   - **Evidence (repro):** same repository with no `.gitignore`: `verify --no-base`, then `agent control`, then `skill review` (default output), then `agent control`, then `verify --no-base`.
     - `acd67c2b`: 0, 0, 0, **0** (`complete`), **0**.
     - `5e042f6`: 0, 0, 0, then **exit 4 `workspace_unverifiable`** ("holds untracked files that Git does not ignore and that are not Shipgate artifacts (agents-shipgate-reports/skill-review.json, agents-shipgate-reports/skill-review.md)"), then **exit 2 `config_error`**, with the same text.
   - **Why it matters:**
     - The message tells the user that Shipgate's own reports are "not Shipgate artifacts".
     - It contradicts three claims made by this PR:
       - the allowlist comment ("Every name a Shipgate run writes directly into a reports directory, whichever command wrote it", `_artifact_lifecycle.py:29`);
       - STABILITY's rule ("…or other artifact a Shipgate command writes into a reports directory", `STABILITY.md:148`);
       - the CHANGELOG `Unchanged` bullet ("the default `agents-shipgate-reports` and any other directory holding only uncommitted Shipgate artifacts").
     - From reading the code: the Stop hook treats a verify exit other than 0 or 20 as advisory context (`cli/install_hooks.py:1175`). In such a repository it stops blocking on `agent_action_required` once a `skill` report is in the directory.
     - It fails closed: no authority is granted, and no pointer reads as current. That is why this is P2, not P1.
   - **Fix:**
     - Add the nine names, preferably derived from `skill.runner.REPORT_BASENAME` × `{json, md, sarif}` (the test can hold the literal set to that product).
     - Extend the registry test to read `skill/runner.py`.
     - Add a `skill review` (or `skill lint --format markdown,json,sarif`) between the two runs of the empirical test, or a dedicated real-Git test: verify, then skill, then `agent control` and `verify` both exit 0.

Non-blocking (P3):
- **Stale PR description.** It still says:
  - `agent control` "does not replay the producing run's command" (it now does, minus `--out`);
  - the writer "resolves the evaluated head and the merge base the way the run does" (`--head` now compares only the head);
  - the new file holds "26 tests" (now 52).
  
  It also never mentions the trust-root rule. The cycle-1 address comment and the commit messages are accurate, so only the description needs refreshing.
- **A gitignored output directory can still drop a manifest-named input from the static input census.** The agreed scope accepts a gitignored directory, so this is not a regression. The setup: manifest `tool_sources: path: gen/tools.json`, with `gen/` gitignored.
  - **Default output:** the plan binds `gen/tools.json`, and after an edit to it, `agent control` exits 4 ("recorded input changed since verification").
  - **`--out gen`:** `inputs.tool_sources: []`, and after the same edit, `agent control` still exits 0 with `complete`, `merge: true`. `acd67c2b` and `5e042f6` behave identically.
  
  So the PR description's deferral rationale, that the refusal makes the census exclusion "unreachable for new runs", does not hold for gitignored directories. Either refuse an `--out` that contains a plan-bound input (extending `_reject_output_input_overlap` to tool sources), or narrow the census exclusion to the files a run writes, in a follow-up issue.

Cycle-1 findings, re-verified on `5e042f6`:
1. **Trust root named like an artifact:** fixed. `--base main` with `--out .claude/commands` (untracked `packet.md`), `--out .claude/skills` (staged `verification-inputs/SKILL.md`) and `--out .agents-shipgate` (untracked `capabilities.lock.json`) each exit 2 ("lies inside a trust root (…) that Git does not ignore"). On `acd67c2b` each is `complete`, merge=true.
   - A gitignored `.claude/reports/` (and `/.claude/reports/*`) verifies twice and refreshes as `complete`.
2. **Recovery dropped `--base`/`--config`:** fixed. Setup: a pointer published by `acd67c2b` with `verify --config shipgate.yaml --base main --out docs`, over a committed widening and a tracked `docs/a.txt`.
   - `agent control --reports-dir docs` exits 4. `next_actions[0].args` = `verify --config shipgate.yaml --base main --workspace <repo> --json`.
   - Running it as printed gives `human_review_required` with `SHIP-HOST-BOUNDARY-PERMISSION-ALLOW-EXPANDED`. `agent control` then exits 0, `review_publishable`, merge=false.
3. **`suggested-inventory.json`:** fixed. `samples/conductor_agent`, with no `.gitignore`: verify 0, `agent control` 0, verify 0, matching `acd67c2b`.
4. **Default-directory looping route:** fixed.
   - A stray `notes.md` gives exit 2, and the `review` next action names moving it out or gitignoring.
   - `agent control` returns a `review` step. Neither offers omitting `--out`.
5. **`--head` removal:** fixed. After `main` commits `agents-shipgate-reports/report.json` and `feature` untracks it:
   - `--base main --head HEAD --out agents-shipgate-reports` exits 0, `changed_files: ['.gitignore', 'agents-shipgate-reports/report.json']`, and `agent control` exits 0.
   - The worktree run of the same change is refused with the untracking remedy.

Also verified without findings:
- **Original defect**, `--out .claude` on `acd67c2b` vs `5e042f6`, uncommitted widening, committed widening, and new untracked `settings.json`: `complete`/merge=true, `complete`/merge=true, `complete`/merge=true on `acd67c2b` → exit 2 on all three at `5e042f6`, with nothing written.
  - `--out docs` gives exit 2, where `acd67c2b` kept a pointer `complete` across a later edit.
- **Aliases** (symlink to the repository, `<parent>/ALIASREPO/.claude`, `<repo>/.CLAUDE`): exit 2, naming `.claude/settings.json`. On `acd67c2b` these were `complete`, `complete` and exit 3.
- **Directories that still work**, each verified twice and refreshed as before: the default directory, a gitignored in-repo `out/`, and a sibling directory outside the repository. For the gitignored `out/`, drift still refuses with `workspace_changed`.
- **Also unchanged:**
  - a `--head` decision (`human_review_required`, with `SHIP-HOST-BOUNDARY-PERMISSION-ALLOW-EXPANDED`);
  - non-Git `verify --preview`, with the default output and with `--out notes`, followed by `agent control`: exit 0;
  - `samples/support_refund_agent` with no `.gitignore`, where verify writes `verification-inputs/baseline/…` and `verification-inputs/policy-packs/…` before stopping at a policy-pack config error, followed by `scan --format markdown,json,sarif`: every refresh reads the same as on `acd67c2b`.
- **`verify --format control`** prints nothing on stdout for this refusal. That matches the existing `--out` overlap refusals on `acd67c2b`.
- **Code placement:**
  - The writer check runs after `_reject_output_input_overlap` and before `mkdir`/`begin_current_control` in `run_verify` and `run_preview`.
  - `live_workspace` classifies outside the `try` that maps exceptions to `None`.
  - `_validate_control_currency` refuses first, for every state.
  - All three readers (`agent control`, `verify --format control`, human-review decision) go through `live_workspace`.
- **Surfaces:** no schema, contract version, error kind, refusal code or exit code moves. `docs/distribution-surfaces.md` makes no output-directory claim.
- **Mutation checks**, run in a separate worktree and reverted:
  - disabling the reader refusal fails 11 tests in `tests/test_output_directory_content.py`;
  - disabling the writer refusal fails 22.
- **Tests run on `5e042f6`**, all passing:
  - `tests/test_output_directory_content.py` (52 tests);
  - `test_current_control_closure`, `test_install_hooks`, `test_agent_control_reports_dir`, `test_current_control`, `test_distribution_surface_parity`, `test_public_surface_contract`, `test_docs_links`, `test_adapter_static_only`, `test_action_metadata`, `test_manifest_free_pr_rows`;
  - `test_current_control_directory_currency`, `test_current_control_input_currency`, `test_agent_control_envelope`, `test_human_review_decision`, `test_verify`.

pengfei-threemoonslab:
**Addressed review cycle 2.** New head `e4f37b14`. The branch is rebased onto `origin/main` `14cdc73b` (#802). The only conflict was STABILITY.md, where the #804 and #802 migration notes both landed at the top; both are kept, #804 first. The cycle's fixes are in `e4f37b14`, on top of the two earlier commits. Evidence was re-run with `./shipgate` from this head in throwaway repositories built from `samples/clean_read_only_agent`.

### 804-c2-1 (P2): the allowlist missed the `skill` reports → fixed

- **Change:** `REPORTS_DIRECTORY_ARTIFACT_NAMES` (`cli/_artifact_lifecycle.py`) now includes `skill-{lint,security,review}.{md,json,sarif}`. That covers `skill.runner.REPORT_BASENAME` in every format `_write_reports` renders.
- **Why the names are written out rather than imported:** that module is a leaf that loads no schema, and importing `skill.runner` would pull in `schemas.report`. Instead, the registry test derives the names from `skill.runner`, so they cannot drift:
  - `test_every_name_a_run_writes_is_a_recognized_artifact` now reads `skill.runner.REPORT_BASENAME` and the suffixes from `_write_reports`. It asserts they are exactly `json`, `md` and `sarif`.
  - It also asserts that `run_skill_review` still defaults to `workspace / "agents-shipgate-reports"`.
- **New real-Git test:** `test_skill_reports_in_the_default_directory_are_recognized_artifacts` uses an unignored default directory:
  1. `verify` completes and `agent control` exits 0.
  2. From the workspace, it runs `skill review` with default formats, then `skill lint`, `skill security` and `skill review` each with `--format markdown,json,sarif`. All nine files must be written, and Git must list them as untracked and not ignored.
  3. `agent control`, a second `verify` and another `agent control` must each exit 0 with `complete`, and the classifier must return `None`.
- **Mutation check:** with the names removed (reverted afterwards), the new test fails on `agent control`: `assert 4 == 0`, `workspace_unverifiable … (agents-shipgate-reports/skill-lint.json, …)`. The registry test fails with the nine names as extra items.
- **Your repro** (no `.gitignore`: `verify --no-base` → `agent control` → `skill review` → `skill lint --format markdown,json,sarif` → `skill security --format markdown,json,sarif` → `agent control` → `verify --no-base` → `agent control`):
  - on `5e042f6a`: `0, 0, 0, 0, 0, 4, 2, 4`;
  - on `e4f37b14`: `0, 0, 0, 0, 0, 0, 0, 0`.
- **Claims now true again:** the allowlist comment, STABILITY.md ("…`skill lint`/`security`/`review` report or other artifact a Shipgate command writes into a reports directory"), and the CHANGELOG "Unchanged" bullet, which now names the skill reports.
- **Other writers checked:** I also went through every other command that writes into a reports directory without `--out`: `verification plan`/`unit-result`/`receipt`, `authorization`, `capability` (generated lock copy), `org` (`attestation.json`), `scenario suggest`, `apply-patches` (`declaration-continuation.json`), `feedback`/`attest`/`audit --host` (explicit `--out` only), `fixture` and `self-check` (their own `reports/`). Each name was already in the allowlist. `skill` was the only gap.

### Nonblocking

- **P3, stale PR description → fixed.** The description is rewritten to match the code:
  - `agent control`'s recovery replays the producing run's `fix_task.verification_command` with only `--out` removed.
  - The writer compares `HEAD` plus the merge base for worktree runs, and only the evaluated head for `--head`.
  - The trust-root rule and the `committed_tree` reader skip are described.
  - The test file now holds 53 tests.
  - The deferral rationale is corrected (next item).
- **P3, a gitignored output directory drops a manifest-named input from the census → documented, fix deferred.**
  - **Reproduced on this head** (`tool_sources: path: gen/tools.json`, `gen/` gitignored, `verify --no-base`, then append a newline to `gen/tools.json`, then `agent control`):
    - default output: the plan binds `gen/tools.json`, and `agent control` exits 0, then 4 after the edit;
    - `--out gen`: `inputs.tool_sources` is `[]`, and `agent control` exits 0, then 0 (`complete`) after the edit.
  - **The old deferral rationale was wrong.** The description said this was unreachable for new runs; that does not hold for gitignored directories, and the description now says so.
  - **Documented:** STABILITY.md's "Not changed here" now states the limit and says to keep inputs out of the output directory.
  - **Not fixed here, for three reasons:**
    - The agreed scope keeps gitignored directories accepted, and `main` behaves identically.
    - A writer refusal limited to manifest-named paths would miss inputs discovered by reading an entrypoint.
    - The general fix, narrowing `StaticInputSnapshot(excluded_paths=[out_dir])` to the files a run writes, changes the `input_directories.excluded_paths` identity that `core/input_directory_identity.py` validates on replay. It also changes the member listings of any input directory that contains the output directory. That deserves its own change and tests. I can't open or edit issues from here, so it needs a follow-up issue filed by hand.
- **P3, the Stop hook is advisory on a refused default directory → documented.**
  - `cli/install_hooks.py` surfaces any `verify` exit other than 0 or 20 as advisory context. That is unchanged: blocking on exit 2 would be a new hook contract.
  - **Documented** in STABILITY.md (a new compatibility bullet), in the CHANGELOG compatibility sub-bullet, in `docs/integrations.md` (hooks section) and in `docs/troubleshooting.md` (default-directory refusal). Each says that when the default directory is refused, the hook's `verify` exits 2 and the hook does not soft-block the Stop, even for a change that would otherwise be `agent_action_required`, so the refusal must be resolved.

### Tests run (on `e4f37b14`'s tree)

- `tests/test_output_directory_content.py`: 53 passed.
- Together with `test_current_control`, `test_scan`, `test_capability_delta_attestation`, `test_cli`, `test_install_hooks` and `test_verify`: 444 passed.
- `test_skill_review`, `test_docs_links`, `test_distribution_surface_parity`, `test_public_surface_contract`, `test_agent_control_reports_dir` and `test_current_control_closure`: 896 passed, 1 skipped.
- After the rebase, `test_workflow_label_redaction` and `test_host_boundary_check` (from #802), plus the output-directory file except the 16-sample empirical test (run above): 144 passed.
- `ruff check .`: clean.
- **Surfaces:** no schema, contract version, error kind, refusal code or exit code moved. No `docs/distribution-surfaces.md` row is touched.

pengfei-threemoonslab (commented):
Independent coding-agent review/address cycle 3 at `e4f37b1`.

**No P0-P2 findings.** The cycle-2 P2 is fixed, the cycle-1 fixes still hold, and the refusal meets the agreed scope. CI is green on `e4f37b1` (11 pass, `release-tag-consistency` skipped). Everything below was re-run for this comment with `./shipgate` from `e4f37b1` and from `origin/main` `14cdc73b`. The repositories are throwaway ones built from `samples/clean_read_only_agent` (`.gitignore` = `agents-shipgate-reports/`, branch `feature` off `main`, no remote) unless noted.

### Cycle-2 finding, re-verified

- **`skill` reports in an unignored default directory: fixed.** Repro, with no `.gitignore`: `verify --no-base` → `agent control` → `skill review` → `skill lint --format markdown,json,sarif` → `skill security --format markdown,json,sarif` → `agent control` → `verify --no-base` → `agent control`.
  - `14cdc73b`: `0 0 0 0 0 0 0 0`.
  - `e4f37b1`: `0 0 0 0 0 0 0 0`, with all eight `skill-*` files in the directory.

### Cycle-1 findings, still fixed

- **What changed:** `git range-diff` shows that the only code difference between `5e042f6` and `e4f37b1` is the nine `skill` names. The rebase onto `14cdc73b` brings only #802. Spot checks:
- **Trust-root file named like an artifact:** untracked `.claude/commands/packet.md` (`allowed-tools: Bash(curl:*), Bash(rm -rf:*)`) with `--out .claude/commands`.
  - main: `complete`, merge=true.
  - branch: exit 2, "lies inside a trust root (agent_instructions) that Git does not ignore…".
- **`--head` removal:** `main` commits `agents-shipgate-reports/report.json`, and `feature` untracks it and gitignores the directory.
  - `--base main --head HEAD --out agents-shipgate-reports` exits 0 (`complete`) on both.
  - The worktree run of the same change exits 2 on the branch, with the untracking remedy.
- **Recovery keeps the request:** setup is a `review_publishable` pointer published by main with `--config shipgate.yaml --base main --out docs`.
  - `agent control --reports-dir docs` exits 4.
  - `next_actions[0].args` = `verify --config shipgate.yaml --base main --workspace <repo> --json`.

### Original defect: main vs branch

| Scenario | `14cdc73b` | `e4f37b1` |
| --- | --- | --- |
| Uncommitted `.claude/settings.json` widening, `--base main --out .claude` | exit 0 `complete`, merge=true | exit 2 `holds tracked repository files (.claude/settings.json)`; `.claude/` still holds only `settings.json` |
| Same widening, committed | `complete`, merge=true | exit 2 |
| New untracked `.claude/settings.json` | `complete`, merge=true | exit 2 `holds untracked files … not Shipgate artifacts` |
| Manifest-free `verify` and `verify --preview` with `--out .claude` | accepted, and the refresh reads it | exit 2; nothing to refresh |
| Tracked `docs/a.txt`: `--no-base --out docs`, edit, refresh | `complete`, then refresh `complete` | exit 2 |
| `<symlink-to-repo>/.claude`, `<PARENT>/REPO/.claude`, `<repo>/.CLAUDE` | `complete`, `complete`, exit 3 | exit 2 each, naming `.claude/settings.json` |
| Legacy pointers published by main into `docs` (`review_publishable`) and `.claude` (`complete`), read by the branch | read as current | exit 4 `workspace_unverifiable` |
| `<repo>/PLAIN-REPORTS` over an artifacts-only `plain-reports` | exit 3 | `complete`; refresh `complete`; an untracked `README.txt` still gives exit 4 `workspace_changed` |

### Unchanged behaviour, checked on both

- **Default directory, gitignored in-repo `out/`, sibling outside the repository:** verify twice, then refresh. All `complete` on the branch.
- **Committed widening with `--head HEAD`:** `review_publishable`/`review_required` on both, with the default output and with a gitignored `out/`.
- **Non-Git `verify --preview`** (default output and `--out notes`), then `agent control`: exit 0 `agent_action_required` on both.
- **`samples/simple_openai_api_agent` into an unignored default directory:** `--base main --suggest-patches` over a new `.claude/settings.json`, `baseline save`, `--baseline`, then `--preview`, each followed by `agent control`. Exit codes and states are identical on both. The last refresh refuses on both for the pre-existing #807 reason.

### Code and claims checked

- **Placement:**
  - The writer check runs after `_reject_output_input_overlap` and before `mkdir`/`begin_current_control` in `run_verify` and `run_preview`.
  - `live_workspace` classifies outside the `try` that maps exceptions to `None`.
  - `_validate_control_currency` refuses first, for every state.
  - All three readers go through `live_workspace`.
  - The Stop hook passes no `--out` and reads no pointer.
- **The classifier matches the change set it guards.** It lists with `:(top,literal)` pathspecs and the same `--exclude-standard` rules as the change-set inventory. Refusing on HEAD, compared commits, the index and unignored untracked paths covers everything `git diff <merge-base>` plus the untracked listing could hide beneath the exclusion.
- **Allowlist against trust roots:**
  - Every artifact name, placed in every directory the boundary-registry globs (and a set of plugin/host directories) name, never makes a boundary-file path that `TRUST_ROOT_SURFACES` does not classify.
  - `verification-inputs/<category>/<digest>-<name>` never matches a trust-root glob.
  - Every `agents-shipgate-reports/<name>` that `docs/`, `action.yml` and the scripts reference is in the allowlist.
- **Mutations** (separate worktree, reverted; the 16-sample empirical test was deselected for speed). Each one is caught:

  | Mutation | Tests that fail |
  | --- | --- |
  | Reader skips the merge-base classification | `hid_a_committed_deletion[docs]` |
  | Trust-root directory rule off | 5 |
  | "Disguised" trust-root rule off | the classifier unit test |
  | Stored-spelling respelling off | the case-variant test and both default-directory route tests |
  | `removed` rule off | 7 |
  | Reader stops skipping `committed_tree` | `head_run_may_remove` |
  | Writer stops comparing the merge base | 4 |
  | Untracked non-artifacts allowed | 4 |

- **Surfaces:** no schema, contract version, error kind, refusal code or exit code moves. `docs/distribution-surfaces.md` makes no output-directory claim.

### Non-blocking (P3)

1. **The docs overstate `agent control`'s recovery for a `complete` pointer.**
   - **The claim:** STABILITY.md (Readers paragraph), the CHANGELOG readers bullet, `docs/agent-contract-current.md`/`llms-full.txt` and `docs/troubleshooting.md` say the recovery keeps the producing run's `--config`, base and policy options.
   - **What happens:** a `complete` verifier has no `fix_task`, so the recovery falls back to a bare `verify --workspace <repo> --json`. `test_a_pointer_already_in_such_a_directory_refuses_in_every_state` pins exactly that.
   - **Repro:** main publishes `verify --config shipgate.yaml --base main --out docs`, which is `complete`. Then a widening is committed.
     - The branch's `agent control --reports-dir docs` exits 4 with `args = ['verify', '--workspace', <repo>, '--json']`.
     - Run as printed, that gives `complete`, merge=true: the local base is skipped because there is no remote.
     - The pre-existing default-directory `workspace_changed` route gives the same bare command (with `--out` naming the default directory) and the same result, so this is not a regression.
   - **Fix:** either qualify the sentence ("when the bound verifier carries a verification command"), or rebuild `--base`/`--head` from the pointer's `workspace_identity` (`base_ref` is recorded) when there is no `fix_task`.
2. **A refused directory with no pointer routes back into itself.**
   - `agent control --reports-dir docs` over a tracked `docs/` with no pointer exits 3 `missing`.
   - Its recovery is `verify --workspace <repo> --out <repo>/docs --json`, which the writer then refuses with the right remedy. That costs one extra hop, but nothing is unsafe.
3. **Refresh latency.** `agent control` on the clean sample went from about 0.86 s to about 1.12 s (three runs each).
   - Each observation classifies twice: against `HEAD` in `live_workspace`, then again with the merge base in `_validate_control_currency`. Each pass re-lists `HEAD`, the index and untracked paths, and the read protocol observes twice.
   - Classifying once with the merge base, or sharing the inventory within a pass, would halve it. Optional.

### Tests run on `e4f37b1`

- `tests/test_output_directory_content.py`: 53 passed.
- These files each pass, run one process per file: `test_action_metadata`, `test_adapter_static_only`, `test_agent_control_contract`, `test_agent_control_envelope`, `test_agent_control_envelope_rows`, `test_agent_control_reports_dir`, `test_agent_controls`, `test_current_control`, `test_current_control_auxiliary_inputs`, `test_current_control_closure`, `test_current_control_directory_currency`, `test_current_control_input_currency`, `test_current_control_input_origins`, `test_distribution_surface_parity`, `test_docs_links`, `test_human_review_decision`, `test_install_hooks`, `test_manifest_free_pr_rows`, `test_public_surface_contract`, `test_skill_review`, `test_verify`.
- `ruff check` on the changed files is clean.

PR #833, opened 2026-09-17
Describe the host-review task and manifest-free diff in discovery copy (#792)

pengfei-threemoonslab:
Refs #792

This is the bounded in-repo slice of #792. The external surfaces listed in the audit below stay open, so this PR does not close the issue.

## Problem

#779/#784 made the README and quickstart lead with the manifest-free `agents-shipgate diff`. The discovery copy that a coding agent or answer engine reads first still described only manifest-first tool-source scanning:

- **`llms.txt`**
  - Mentioned `diff` once, as a footnote under `## Maintainer`.
  - Its continue rule named `is_agent_project`, `suggested_sources` and `codex_plugin_candidates` but not `host_boundary_candidates`. An agent following it stopped on a repository whose only agent configuration is `.claude/settings.json` or `.mcp.json`, even though `detect` routes that repository forward. I reproduced this; see Evidence.
  - Its identity line said `Current source-tree runtime: 1.0.0 (contract 39; published …)`. This tree emits contract 40, and the published `v1.0.0` emits 39. Changing that number to 40 would have told participants' agents that the build they install emits 40.
  - Its `check` instruction named `shipgate.agent_boundary_result/v2`. Both builds emit `v3`.
- **`.well-known/agents-shipgate.json`**: no `when_to_use` entry or keyword for the host-configuration review job.
- **`docs/ai-search-summary.md`**, the `AGENTS.md` intro and "When to use" (and so `llms-full.txt`), and the `pyproject.toml` description and keywords: tool-surface scanning only.

## Design

The task copy is added next to the canonical identity and replaces none of it:

- **`llms.txt`**
  - The first "What it is" and "When to recommend" entries, and a "Quickstart" entry after the three install lines, describe reviewing a change to declared coding-agent configuration: `.claude/settings.json`, `.mcp.json`, hooks, Codex, Cursor or VS Code MCP config, and GitHub workflow permissions. The route is `agents-shipgate diff`. The entries state the limits: no manifest, baseline or skill; the base must already be in the clone's history (it never fetches); static; names recognized sources it could not read, and a no-change answer says nothing about the [surfaces `diff` does not read](https://github.com/ThreeMoonsLab/agents-shipgate/blob/main/docs/host-boundary-support.md#known-unread-surfaces) (hook scripts, composite actions, subagent hooks); implies no verdict; not every change widens authority.
  - The tool-surface bullets stay as the second job.
[...]

PR #835, opened 2026-09-17
Refuse diff cleanly in a partial clone missing base blobs (#817)

### LINKED ISSUE #817: diff exits 1 with an uncaught ConfigError traceback in a partial clone missing base blobs
## Problem

In a partial clone (`--filter=blob:none`) whose base blobs were never fetched, `agents-shipgate diff` exits **1** with an uncaught `ConfigError` traceback. Stdout is empty, and there is no `next_action` even with `AGENTS_SHIPGATE_AGENT_MODE=1`. The error comes from `archive_tree` (`cli/verify/git.py`) and passes through `run_capability_diff` (`cli/diff.py`) uncaught.

Nearby paths already handle the same kind of problem:
- **Shallow clone:** `diff` exits 2 with a structured `config_error` and a `fetch --unshallow` next action.
- **Same partial clone, `verify --preview`:** returns `objects_missing` with a `fetch_base` next action.
- **Same partial clone, `verify` and `check`:** return structured results.

Blobless clones are a common way to scan many repositories cheaply, and `diff` is the promoted entry (#779).

## Reproduction

Observed on `main` `9df307ab`, git 2.54:

```bash
g() { git -c user.name=t -c user.email=t@example.invalid "$@"; }
mkdir -p src/.claude && g -C src init -q -b main && g -C src config uploadpack.allowFilter true
echo '{"permissions":{"allow":["Read"],"deny":["Bash(git push *)"]}}' > src/.claude/settings.json
g -C src add -A && g -C src commit -qm base && g -C src branch -q base
echo '{"permissions":{"allow":["Read","Bash(git push *)"]}}' > src/.claude/settings.json
g -C src commit -qam head
git clone -q --filter=blob:none --no-checkout "file://$PWD/src" partial && git -C partial checkout -q main

AGENTS_SHIPGATE_AGENT_MODE=1 agents-shipgate diff --workspace partial --base origin/base --json; echo "exit=$?"
AGENTS_SHIPGATE_AGENT_MODE=1 agents-shipgate verify --workspace partial --base origin/base --preview --json \
  --out "$PWD/preview-out" | jq -r '.control.next_action.why'
```

Observed:
- **`diff`:** `exit=1`, nothing on stdout. Stderr is a Rich traceback ending in `ConfigError: Could not copy verified Git objects: warning: lazy fetching disabled; some objects may not be available / fatal: could not fetch <blob> from promisor remote`.
- **`verify --preview`:** says the diff body could not be read (`objects_missing`) and that verification runs with `GIT_NO_LAZY_FETCH=1` and will not fetch implicitly. It suggests hydrating "for example `git fetch --refetch origin`, or clone without `--filter`".
- **Suggested hydration command:** `git fetch --refetch origin` does **not** hydrate this clone, because the configured `remote.origin.partialclonefilter=blob:none` still applies and `diff` fails the same way afterwards. `git fetch --refetch --no-filter origin` does hydrate it, and `diff` then returns `comparable` with 2 rows.

## Acceptance

- [ ] `diff` in a partial clone with missing base (or head) objects returns a structured refusal: a documented non-zero exit code, an error kind (for example `objects_missing`, matching preview), a message naming the missing side, and an agent-mode `next_action`. No traceback, in text and `--json`.
- [ ] Shipgate never fetches anything itself: `GIT_NO_LAZY_FETCH` behaviour is unchanged and the refusal only names a command.
- [ ] The named hydration command is exercised by a test and makes the same comparison succeed. `diff` and `verify --preview` use the same wording. Correct preview's `git fetch --refetch origin` example, which keeps the configured filter.
- [ ] Tests cover a `blob:none` clone with head-only hydration (for example a checkout of the head), a hydrated control that stays `comparable`, and a shallow control that keeps its existing `--unshallow` refusal.
- [ ] `check` and `verify` in the same clone keep a structured, fail-closed result. Add a test so they cannot regress to a traceback.

## Sequencing

Phase 1 (#778) reliability reserve, alongside #813 (recoverable refusals). No new surface: this adds an error kind to the existing agent-mode error contract. If the error-kind vocabulary is a stable field, add a STABILITY.md note. Refs #779, #308, #683 (the shallow-clone traceback whose structured refusal this should match).

pengfei-threemoonslab:
Closes #817

## Problem

In a partial clone whose base objects were never fetched, `agents-shipgate diff` exited **1** with an uncaught traceback. Stdout was empty, and agent mode printed no `next_action`.

- **Blobless clone (`--filter=blob:none`):** a `ConfigError` from `archive_tree` ("could not fetch <blob> from promisor remote").
- **Treeless clone (`--filter=tree:0`):** a `CalledProcessError` one step earlier, from the `<commit>^{tree}` lookup.

`verify --preview` in the same clone already reported `objects_missing`, but its hydration example was wrong. `git fetch --refetch origin` applies the clone's configured `remote.origin.partialclonefilter` again and fetches no blob, so `diff`, `verify` and `check` kept failing the same way.

## Design

**1. Detect the case from the repository.** `promised_objects_missing` (`cli/verify/git.py`) walks the commit's objects with `rev-list --objects --no-walk`. It reports missing objects only when that walk fails and the same walk succeeds with `--missing=allow-promisor`.
- It checks the repository itself rather than matching Git's error text.
- `--no-walk` keeps it from crossing a shallow boundary.
- A corrupt repository, whose missing objects have no promisor remote behind them, does not count.
- Both walks go through the existing `_git_object_environment` (`GIT_NO_LAZY_FETCH=1`), so the check never fetches anything.

**2. Leave `verify` alone.** `archive_fetched_tree` wraps `archive_tree` and only `diff` calls it. When the check above holds, it raises `PromisedObjectsMissingError` (a `ConfigError`). Every other failure is re-raised unchanged.

I first tried a typed subclass raised inside `archive_tree`. It changed `verify`'s published `host_comparison.incomparable_reasons` from `comparison_input_unavailable:ConfigError` to `…:GitObjectsMissingError`, because that reason is built from the exception's class name. So `archive_tree` and its `verify` callers stay exactly as they were.

**3. Refuse the way the shallow case does (#683).** `diff` exits `2` and prints one stderr line naming the side it could not read. This works the same in text and `--json`. In agent mode it emits:
```json
{"error": "objects_missing", "exit_code": 2, "next_actions": [{"kind": "command", "command": "git -C <workspace> fetch --refetch --no-filter <remote>", ...}], ...}
```
- `next_actions` has one such command for each promisor remote in the repository config (`remote.<name>.promisor`, parsed by Git as a boolean, then `extensions.partialClone`), in configuration order.
- A remote is named only if `git check-ref-format` accepts `refs/remotes/<name>/test` (the check `git remote add` makes, so `up/stream` is named), it has a `remote.<name>.url`, and its name does not start with `-`.
- `origin` is never used as a fallback. If no promisor remote qualifies, the one next action is a `review`, and the example ends in a `<remote>` placeholder.
- The error kind is the same word `verify` reports as `diff_status.reason`.

**4. One wording.** `objects_missing_remediation(hydrate)` builds the sentence. `verify`'s `_DIFF_REASON_REMEDIATION` uses it with the corrected example `git fetch --refetch --no-filter origin`. `diff` passes the same command bound to its workspace. "Verification runs with GIT_NO_LAZY_FETCH=1" now reads "Shipgate runs Git with GIT_NO_LAZY_FETCH=1", because `diff` says it too.

**5. Docs.**
- `docs/errors.json`: new `objects_missing` entry, exit `2`. `schema_version` stays `0.1`.
- `AGENTS.md`: the kind is listed; `llms-full.txt` is rebuilt.
- `STABILITY.md`: migration note.
- `CHANGELOG.md`: Unreleased entry.
- `docs/quickstart.md`: the recovery sits next to the shallow one and is held to real output by `tests/test_host_diff_entry_docs.py`.
- `docs/distribution-surfaces.md`: one sentence on the `capability_diff` row. Its claims are unchanged (still none), so `tests/test_distribution_surface_parity.py` needs no change.
- `tests/test_adapter_static_only.py`: the two `cli/verify/git.py` line pins moved.

### Surface discipline

- **Metric:** time-to-first-verdict and activation. Blobless clones are a common cheap way to scan many repositories, and `diff` is the promoted entry (#779). Before this change they got a traceback. Now they get one runnable command, after which the same comparison succeeds.
- **Existing surface:** this adds no command, schema, schema version, report or `verifier.json` block, or JSON field. It extends the existing agent-mode error catalog by one kind, as the issue proposes, reusing `verify`'s `objects_missing` vocabulary and wording rather than adding a parallel one.
- **Non-goals:** there is no network access in the scanner. Shipgate names the fetch and never runs it.

### Not changed / deferred

- A shallow checkout is still refused first with `git fetch --unshallow`. After that fetch, a partial clone can still report `objects_missing` (tested as one sequence).
- `verify` without `--preview` in a **treeless** clone still exits `4` with `internal_error`, as before. The same uncaught `CalledProcessError` comes through its own `archive_tree` call. Fixing that changes `verify`'s answer and is outside #817's `diff` scope; it needs its own issue.
- In the same blobless clone, `check` still says "restore a readable git worktree" rather than naming hydration. It is structured and fail-closed, and is now pinned by tests.

## Tests

New `tests/test_capability_diff_partial_clone.py` (28 tests after review cycle 1) uses real `file://` partial clones hydrated only by a checkout of the head:
- **Refusal:** `blob:none` and `tree:0` × text/`--json` × agent mode on/off. Each case checks exit 2, no traceback, empty stdout, the message naming the base side, and the exact `objects_missing` line with `executable`/`args`. It then checks that the object store and the set of missing objects are unchanged. Finally it runs the published command from outside the checkout and repeats the same comparison (`comparable`, `Bash(git push *)` row).
- **No implicit fetch:** the promisor remote is replaced with `ext::touch <marker>`, and `GIT_NO_LAZY_FETCH` is removed from the caller's environment. `diff` leaves the marker absent. The test runs four ways: with Shipgate's Git environment as shipped, with only `GIT_NO_LAZY_FETCH` (skipped before Git 2.44), and with only `GIT_ALLOW_PROTOCOL=""`, the marker stays absent. With both protections removed, the marker is created. As a positive control, an ordinary `git cat-file` of the same object also creates it.
- **Remote naming:** `--origin upstream` and `--origin up/stream` clones each get a command naming that remote, which hydrates when run. A clone with `alpha` and `beta` as promisor remotes (and `gamma` with `promisor = false`) gets commands for `alpha` and then `beta`, and none for `origin`. The second command, run as written, hydrates. A promisor remote whose name starts with `-`, one with no URL, and one with a space in its name each get a single `review` action and no command.
- **Corruption control:** a deleted loose blob in a normal repository is not reported as `objects_missing`.
- **Wording:** `verify --preview`'s `diff_status.remediation` equals `objects_missing_remediation()`, and `diff`'s message ends with the same sentence. `git fetch --refetch origin` is shown **not** to hydrate (diff still exit 2, preview still `objects_missing`). The new example, run as written, gives `completeness: complete` and a comparable diff.
- **Controls:**
  - `--base HEAD` compares in both `blob:none` and `tree:0` clones.
  - A `--depth 1 --filter=blob:none` clone goes: shallow `config_error` → `--unshallow` → `objects_missing` → hydrate → comparable.
- **Same clone, other commands:**
  - `check` (`agent-boundary-json`, `agent-control-json`, `text`) exits 0 with `human_review_required` and every permission false.
  - `verify --json` exits 0 with `objects_missing`, the shared remediation, `can_merge_without_human: false`, and every permission false.
- **Catalog:** the `docs/errors.json` entry's exit code and fields match what is emitted.

`tests/test_host_diff_entry_docs.py::test_the_documented_partial_clone_recovery_holds` holds the quickstart sentence, including the `--refetch`-alone trap, to this tree's output.

Run locally (macOS, git 2.54, Python 3.14). Every run passed; 1 test was skipped.
- **Run 1, about 2,300 tests:** the new module, `test_capability_diff`, `test_diff_input_status`, `test_public_surface_contract`, `test_distribution_surface_parity`, `test_agent_mode`, `test_host_diff_entry_docs`, `test_live_workspace_cause`, `test_manifest_free_pr_rows`, `test_workflow_capability_diff`, `test_adapter_static_only`, `test_cli_on_demand_loading`, `test_cli`, `test_install_hooks`, and the other host and workflow suites that invoke `diff`.
- **Run 2, 1,191 tests:** `test_verify`, the new module, `test_host_diff_entry_docs`, `test_adapter_static_only` and `test_public_surface_contract` again.
- **Checks:** `ruff check .` and `compileall` are clean. CI runs the full suite.

## Before / after

Measured on the issue's synthetic repository. `main` is `origin/main` 34d8b38d run from an extracted source tree; `branch` is this PR. Clones are made with `--filter=<f> --no-checkout` followed by `checkout main`.

| clone | engine | `diff --base origin/base` | `--json`, `AGENTS_SHIPGATE_AGENT_MODE=1` |
| --- | --- | --- | --- |
| `blob:none` | main | exit 1, traceback, stdout empty | exit 1, traceback, no agent line |
| `blob:none` | branch | exit 2, one line, stdout empty | exit 2, `objects_missing`, `next_action: git -C <clone> fetch --refetch --no-filter origin` |
| `tree:0` | main | exit 1, `CalledProcessError` traceback | exit 1, traceback, no agent line |
| `tree:0` | branch | exit 2, one line | exit 2, `objects_missing`, same command |

After running the emitted command, `diff` exits 0: `comparable`, rows `Bash(git push *)` added and the deny rule removed. This held for both filters.

Branch refusal line:
```
The base side of this diff, origin/base (9c3515cb), could not be read (objects_missing). This checkout is a partial clone and the objects the diff needs were never fetched. Shipgate runs Git with GIT_NO_LAZY_FETCH=1 and will not fetch them implicitly. Hydrate them (for example `git -C <clone> fetch --refetch --no-filter origin`, or clone without `--filter`), then rerun.
```

`verify --preview` example, run as printed in the clone:
- **main:** `git fetch --refetch origin` → `diff_status` still `partial` / `objects_missing`.
- **branch:** `git fetch --refetch --no-filter origin` → `diff_status` `complete`.

In the same blobless clone, `check --format agent-control-json` is `human_review_required` and `verify --json` is `agent_action_required` with `objects_missing`, no permission, and `comparison_input_unavailable:ConfigError` on both main and branch.


pengfei-threemoonslab (commented):
Independent coding-agent review/address cycle 1 at `2857e52a`.

**Not mergeable yet: one P2 finding.** The #817 fix works, and CI is green on this head.

1. **P2: a valid single promisor remote whose name is not a "plain token" gets a command that names `origin`, and that command fails.**
   - **What the docs say:** `CHANGELOG.md`, the `STABILITY.md` migration note and the `docs/errors.json` `objects_missing` `recovery_hint` all say the command names the clone's promisor remote, with `origin` only "when the configuration does not name exactly one". Agents read the catalog to route, so that sentence is part of the contract.
   - **What the code does:** `promisor_remote()` also returns `None` when the one name does not match `_REMOTE_NAME_RE` (`[A-Za-z0-9][A-Za-z0-9._-]{0,99}`), and `diff` then falls back to `origin`. Git allows `/` in remote names.
   - **Reproduction** (the issue's `src` repository, git 2.54):
     - `git clone -q --filter=blob:none --no-checkout --origin up/stream file://$PWD/src slash && git -C slash checkout -q main`. The configuration has exactly one promisor, `remote.up/stream.promisor=true`.
     - `AGENTS_SHIPGATE_AGENT_MODE=1 ./shipgate diff --workspace slash --base up/stream/base --json` gives `objects_missing` with `next_action` = `git -C <abs>/slash fetch --refetch --no-filter origin`.
     - Running that command fails: `fatal: 'origin' does not appear to be a git repository`, exit 128.
     - `git -C slash fetch --refetch --no-filter up/stream` hydrates, and the same `diff` is then `comparable` with 2 rows.
   - **Fix, either one:**
     - Accept any remote name Git accepts that does not start with `-`. `join_argv`/`shlex.join` already quotes the token. For example, check it with `git check-ref-format "refs/remotes/<name>/x"`, or add `/` to the pattern. Add an `--origin up/stream` case next to `test_the_command_names_the_remote_the_clone_was_promised_by`.
     - Or keep the token restriction and state it in the three published claims, e.g. "`origin` when the configuration does not name exactly one remote, or names one that is not a plain token".

**What I verified**

- **Defect on `origin/main` `34d8b38d`:** reproduced with the issue's synthetic repository via `./shipgate` from a detached `origin/main` worktree.
  - `blob:none` clone: exit 1 with a `ConfigError` traceback, empty stdout, no agent line.
  - `tree:0` clone: exit 1 with a `CalledProcessError` traceback.
- **Fix at `2857e52a`:**
  - Both clones, text and `--json`: exit 2, empty stdout, one stderr line naming the base side, and in agent mode one `objects_missing` line. Its `next_actions[0]` is a `command` whose `executable`/`args` match.
  - Ran each emitted command from `/`, then reran `diff`: exit 0, `comparable`, `Bash(git push *)` added and the deny rule removed.
  - `git fetch --refetch origin` alone leaves `diff` at exit 2 in both clones, so the corrected example is right.
  - Also refused correctly: a `blob:limit=10` clone and a linked `git worktree` of the blobless clone. The worktree's command is bound to the worktree path.
  - With no `--base`, a partial clone whose default base it already holds still compares.
- **`verify --preview` in both clones:** `diff_status.remediation` and `control.next_action.why` now carry the `--no-filter` example and the "Shipgate runs Git with GIT_NO_LAZY_FETCH=1" wording, which is the sentence `diff` ends with.
- **Same clones, other commands:**
  - `check --format agent-control-json` exits 0 with `human_review_required` and no permission, in both clones.
  - `verify --json` in the blobless clone exits 0 and is structured.
  - The treeless `verify` still exits 4 with `internal_error` on both main and this head, as disclosed.
- **Detection logic:**
  - Mutating `promised_objects_missing` to return `True` fails `test_an_object_missing_without_a_promise_is_not_called_a_partial_clone`, so the corruption control is live.
  - `promisor_remote` handles `extensions.partialClone` alone, `promisor=false`, `TRUE`, and a missing key.
  - Every non-promise failure is re-raised unchanged, and the refusal never yields rows, so I found no fail-open.
- **No fetch:** every probe goes through `_git_object_environment`, which sets `GIT_NO_LAZY_FETCH=1` and `GIT_ALLOW_PROTOCOL=""`.
- **Tests run locally** (Python 3.14, git 2.54), all passing:
  - `test_capability_diff_partial_clone`, `test_host_diff_entry_docs`, `test_public_surface_contract`, `test_distribution_surface_parity`, `test_adapter_static_only`, `test_agent_mode`
  - `test_diff_input_status`, `test_host_file_links`, `test_host_link_read_through`, `test_scoped_base_tree`, `test_source_head_identity`, `test_verification_git_snapshot`, `test_capability_diff`
  - `ruff check` on the changed files is clean, and `scripts/build-llms-full.py` reproduces `llms-full.txt` byte for byte.
- **CI:** every check on `2857e52a` passes, including `suite (1-3)`, `coverage` and `windows-launcher`.

**Non-blocking (P3)**

- **The no-fetch test does not isolate `GIT_NO_LAZY_FETCH`.** With only that line removed from `_git_object_environment`, `test_diff_never_asks_the_promisor_remote_for_what_is_missing` and the eight refusal cases' object-store assertions still pass, because `GIT_ALLOW_PROTOCOL=""` also blocks the `ext::` probe. They fail only when both lines are removed. The guarantee (no contact) is proven, but the docstring's "Shipgate sets `GIT_NO_LAZY_FETCH` itself" is not what the test pins.
- **Wording in the `STABILITY.md` note.** Its first sentence says no "exit code" moves, while its Compatibility bullet says a consumer "now sees `2`". "No new exit code" would say it exactly.
- **PR description.** "With two promisor remotes, no name is guessed" is true of `promisor_remote()`, but the emitted command still names `origin`. The CHANGELOG says so; a clone with no `origin` remote gets a failing command.
- **The first probe walk in `promised_objects_missing` passes no `--missing=`.** On Git older than 2.44, which ignores `GIT_NO_LAZY_FETCH`, it would try a lazy fetch. `GIT_ALLOW_PROTOCOL=""` still blocks the transport, and `archive_tree` already relied on the same thing, so this is optional hardening (for example `--missing=print`).
- **Pre-existing and disclosed, out of #817's scope; worth their own issues:**
  - `verify` without `--preview` in a `tree:0` clone exits 4 with `internal_error`.
  - `check` says "restore a readable git worktree" instead of naming hydration.
  - `verify --preview`'s `fetch_base` next action (`command: null`, `expects: <base>...HEAD`) does not hydrate a partial clone.
- **`skills/agents-shipgate/SKILL.md`** (and its plugin copy) do not list `objects_missing` in their error-kind table. The table was already missing `malformed_patch`, `environment_error` and `unknown_fingerprint`, so this is not a regression.

pengfei-threemoonslab (commented):
Independent coding-agent review/address cycle 1 at `2857e52a`.

**Not mergeable yet: one P2 finding.** The #817 fix works and CI is green on this head. The head has not changed since the earlier cycle-1 review on this PR. I re-derived its P2 independently, and it is still open.

1. **P2: a clone with exactly one promisor remote gets a command naming `origin` when that remote's name is not a "plain token". The command fails.**
   - **What the docs promise:** three published places say the command names the clone's promisor remote, and uses `origin` only when the configuration "does not name exactly one". They are the `CHANGELOG.md` entry, the `STABILITY.md` note (`#partial-clone-diff-objects-missing-817`) and the `recovery_hint` of the `objects_missing` entry in `docs/errors.json`. Agents route on that catalog.
   - **What the code does:** `promisor_remote()` (`cli/verify/git.py:2534`) also returns `None` when the one name does not match `_REMOTE_NAME_RE` (`[A-Za-z0-9][A-Za-z0-9._-]{0,99}`, line 2496). `diff.py:147` then falls back to `origin`. Git accepts `/` in remote names: `git check-ref-format refs/remotes/up/stream/x` exits 0.
   - **Reproduction** (the issue's `src` repository, git 2.54, `./shipgate` from this head):
     - Clone: `git clone -q --filter=blob:none --no-checkout --origin up/stream file://$PWD/src slash && git -C slash checkout -q main`. The only promisor key is `remote.up/stream.promisor true`.
     - Run `AGENTS_SHIPGATE_AGENT_MODE=1 ./shipgate diff --workspace slash --base up/stream/base --json`. It gives `objects_missing`, exit 2, and `next_actions[0].args` ends in `fetch --refetch --no-filter origin`.
     - Running that command fails with `fatal: 'origin' does not appear to be a git repository`, exit 128.
     - `git -C slash fetch --refetch --no-filter up/stream` exits 0, and the same `diff` then returns `comparable` with 2 rows.
   - **Fix, either one:**
     - Accept any name Git accepts that does not start with `-`, for example with `git check-ref-format "refs/remotes/<name>/x"`. `join_argv` already quotes the token. Add an `--origin up/stream` case next to `test_the_command_names_the_remote_the_clone_was_promised_by`.
     - Or keep the restriction and say so in all three published claims.
   - Either way, consider not emitting `origin` when the repository has no remote by that name (see the first P3).

**What I verified**

- **Defect on `origin/main` `34d8b38d`** (detached worktree, `./shipgate`, the issue's synthetic repository):
  - `blob:none` clone, agent mode, `--json`: exit 1, `ConfigError` traceback (`could not fetch … from promisor remote`), empty stdout, no agent line.
  - `tree:0` clone: exit 1, `CalledProcessError` traceback from `rev-parse <commit>^{tree}`.
- **The fix at `2857e52a`:**
  - Both clones exit 2 with empty stdout and one stderr line naming `origin/base (0baf3f94)`. In agent mode there is one `objects_missing` line whose `executable`/`args` match `command`, and with `AGENTS_SHIPGATE_AGENT_MODE=0` there is none.
  - I ran each emitted `args` from `/`. `diff` then exits 0 as `comparable`, with `Bash(git push *)` added and the deny rule removed.
- **No implicit fetch:** I pointed `remote.origin.url` at `ext::touch <marker>` and unset `GIT_NO_LAZY_FETCH` in the caller.
  - These left the marker absent: the `diff` refusal, `diff` with the default base, `verify --preview` and `check --format agent-control-json`.
  - A plain `git cat-file` of the same object then created it.
- **Detection:**
  - A loose blob deleted from a local, non-promised base commit inside a blobless clone still fails as before (exit 1, `unable to read`). It is not reported as `objects_missing`, which matches the "fails as before" claim.
  - After a failed archive, the rev-list walk checks the same object set that `pack-objects --revs <tree>` needed, so a non-object failure is never relabelled.
- **Wording:** `verify --preview` on main still printed `git fetch --refetch origin` and "Verification runs with". This head prints `--no-filter` and "Shipgate runs Git with" in both `diff_status.remediation` and `control.next_action.why`. Reason, `fetch_repairable: true` and the `fetch_base` route are unchanged, as claimed.
- **Same clones, other commands:**
  - `check --format agent-control-json` is `human_review_required` with no permission, in the blobless and treeless clones.
  - In the treeless clone, `verify --json` is exit 4 `internal_error` with no traceback on both main and this head, as disclosed.
- **Tests fail without the fix:** with `diff.py` reverted to a plain `archive_tree` call, 14 tests fail: 13 in `test_capability_diff_partial_clone.py` plus `test_the_documented_partial_clone_recovery_holds`.
- **CI ran the new tests:** the `suite (1-3)` totals are 11,354 passed / 22 skipped. On main `34d8b38d` they are 11,332 / 22. So all 22 new tests ran and passed on git 2.55; none were skipped.
- **Tests run locally** (Python 3.14, git 2.54), all passing:
  - 1,520 tests: `test_capability_diff_partial_clone`, `test_host_diff_entry_docs`, `test_public_surface_contract`, `test_distribution_surface_parity`, `test_adapter_static_only`, `test_agent_mode`, `test_diff_input_status`, `test_capability_diff`, `test_scoped_base_tree`, `test_verification_git_snapshot`, `test_cli_on_demand_loading`.
  - 224 tests: `test_verify`, `test_host_file_links`, `test_host_link_read_through`, `test_source_head_identity`, `test_workflow_capability_diff`, `test_manifest_free_pr_rows`, `test_host_path_privacy`, `test_unchanged_host_limits`.
  - `ruff check` on the changed Python files is clean, and `scripts/build-llms-full.py` rebuilds `llms-full.txt` byte for byte.
- **CI:** every check on `2857e52a` passes.

**Non-blocking (P3)**

- **Fallback when no remote is named `origin`.** Git 2.54 confirms that with two promisor remotes and none named `origin` (`alpha`, `beta`), the emitted command names `origin` and fails with exit 128. The CHANGELOG documents that fallback. But the PR description's "With two promisor remotes, no name is guessed" reads as though no command is guessed either, and an agent following `next_actions` retries a command that cannot work.
- **The no-fetch test does not pin `GIT_NO_LAZY_FETCH`.** I deleted only the `"GIT_NO_LAZY_FETCH": "1"` line from `_git_object_environment`, and all 21 tests in `test_capability_diff_partial_clone.py` still pass, because `GIT_ALLOW_PROTOCOL=""` also blocks the `ext::` probe. No contact is proven, but the docstring's "Shipgate sets `GIT_NO_LAZY_FETCH` itself" is not what the test pins.
- **`STABILITY.md` note wording.** The first sentence says no "exit code" moves, while the Compatibility bullet says a consumer "now sees `2`" where it saw `1`. "No new exit code", the phrasing of the 1.0.0 "an absent input is refused" note, is exact.
- **Optional hardening.** The first walk in `promised_objects_missing` passes no `--missing=`. On Git older than 2.44, which ignores `GIT_NO_LAZY_FETCH`, it would attempt a lazy fetch that only `GIT_ALLOW_PROTOCOL=""` stops. `archive_tree` already relies on the same thing.
- **Pre-existing and out of #817's scope (each worth its own issue):**
  - A treeless `verify` exits 4 with `internal_error`.
  - `check` says "restore a readable git worktree" rather than naming hydration.
  - `verify --preview`'s `fetch_base` next action (`command: null`) does not hydrate a partial clone.
  - The error-kind tables in `skills/agents-shipgate/SKILL.md` and its copies were already missing several kinds, and do not list `objects_missing`.

pengfei-threemoonslab:
**Addressed review cycle 1.** New head `ec1ddbee91592912daffb0d8327860d0372cd9bd`

I rebased onto `origin/main` `a4dbd139` (#833), so this push is a force push. The only conflict was the CHANGELOG Unreleased list, and I kept both entries. The fixes are in one commit on top of the rebased original commit.

## Blocking

### P2-1: a single promisor remote named `up/stream` got a command naming `origin`

**Fixed.** `promisor_remote` is now `promisor_remotes` (`src/agents_shipgate/cli/verify/git.py`). It no longer checks names against `[A-Za-z0-9][A-Za-z0-9._-]{0,99}`. Instead it keeps any promisor remote that meets three conditions:
- Git accepts the name as a remote name: `git check-ref-format refs/remotes/<name>/test` exits 0. `git remote add` makes the same check, so `up/stream` is kept.
- It has a `remote.<name>.url`. Without one, `git fetch <name>` would treat the name as a path.
- The name is UTF-8 and does not start with `-`, which `git fetch` would read as an option.

`join_argv` already quotes the name.

I also took the optional part: `diff.py` never uses `origin` as a fallback. When no promisor remote qualifies, the one next action is `kind: "review"` with no command. The example in the message ends in a `<remote>` placeholder, after the clause "no remote this partial clone was promised objects by can be named in a command".

The three published claims (the CHANGELOG entry, the STABILITY.md #817 note and the `docs/errors.json` `objects_missing` `recovery_hint`) now describe this rule. So does AGENTS.md, and `llms-full.txt` is rebuilt.

Evidence, git 2.54:
- `test_the_command_names_the_remote_the_clone_was_promised_by[up/stream]` clones with `--origin up/stream`, as in your reproduction. It asserts `promisor_remotes(clone) == ("up/stream",)` and that the only next action's argv is exactly `git -C <clone> fetch --refetch --no-filter up/stream`. It runs that argv from outside the checkout, then asserts `diff --base up/stream/base --json` is `comparable` with the `Bash(git push *)` row. The `[upstream]` case is unchanged.
- `test_a_promisor_no_command_can_name_is_left_to_the_operator` covers `[leading-dash]` (`remote.-origin`), `[no-url]` and `[not-a-remote-name]` (`remote.or igin`). Each asserts exit 2, one `review` action with `command: null` and no `executable`, and `next_action == "Review: <message>"`. It also asserts the message ends with `objects_missing_remediation("git -C <clone> fetch --refetch --no-filter <remote>")`.

## Nonblocking

**P3, two promisor remotes and neither named `origin`: fixed.** `next_actions` now has one ranked command per qualifying promisor remote, in configuration order.
- Git parses the flag (`git config --type=bool`), so `yes` counts and `false` does not.
- `extensions.partialClone` is read after the flags.
- `next_actions[0]`'s `why` is the message. Each later action's `why` says it is an alternative if the first remote cannot supply the objects.

`test_every_promisor_remote_is_named_and_none_is_guessed` sets up `alpha` (the clone's source), `beta` (`promisor = yes`) and `gamma` (`promisor = false`). It asserts the commands are exactly `[... alpha]` and then `[... beta]`, and that `origin` appears nowhere in stderr. It then runs the second command as written, and `diff` compares. The config reads use the existing bounded reader (`_run_git_bounded_output`, with `--includes` and a 1 MiB bound), like the filter and diff-driver config reads.

**P3, the no-contact test did not pin `GIT_NO_LAZY_FETCH`: fixed.** `test_diff_never_asks_the_promisor_remote_for_what_is_missing` now monkeypatches `_git_object_environment` and runs four ways:

| case | Git environment | marker |
| --- | --- | --- |
| `as-shipped` | both protections | absent |
| `no-lazy-fetch-alone` | `GIT_ALLOW_PROTOCOL` removed | absent (skipped on Git before 2.44) |
| `protocol-allow-list-alone` | `GIT_NO_LAZY_FETCH` removed | absent |
| `neither` | both removed | created (control) |

Mutation check: I deleted only the `"GIT_NO_LAZY_FETCH": "1"` line from `_git_object_environment` and reran the module. `test_diff_never_asks_the_promisor_remote_for_what_is_missing[no-lazy-fetch-alone]` failed at `assert marker.exists() is contacted`, and nothing else did. I restored the line before committing.

**P3, the STABILITY.md wording: fixed.** The note now reads "No schema, contract or `minimum_control_contract_version` moves, no new exit code is added, and no JSON field is added", which agrees with its Compatibility bullet (1 becomes 2 for this case).

**P3, the first `rev-list` walk passes no `--missing=`: declined.**
- `promised_objects_missing` runs only after `archive_tree` has failed. `archive_tree`'s `pack-objects --revs` has already read the same missing objects under the same environment, so on Git before 2.44 the lazy fetch would already have been attempted there and stopped only by `GIT_ALLOW_PROTOCOL=""`. Hardening the walk alone would not change whether Shipgate attempts a fetch on old Git.
- The protection old Git relies on is now pinned by the `protocol-allow-list-alone` case above.
- `print`, `allow-any` and `allow-promisor` are the `--missing=` values documented as tolerating missing objects. Each one either prints every missing object id or makes the walk succeed. The probe answers "does the walk fail without promisor allowance?", so it would have to buffer and count those ids instead.

**P3, pre-existing and out of scope: not changed.** On `main`, `verify` without `--preview` in a `tree:0` clone exits 4 with `internal_error`. Also pre-existing: `check`'s "restore a readable git worktree" wording, and `verify --preview`'s `fetch_base` action that has no command. Changing any of these changes `verify`/`check` answers outside #817's `diff` scope, and they need their own issue. The SKILL.md error-kind tables were already missing several kinds. Completing them is a separate docs change, so I did not add a single row here.

## Tests run (macOS, git 2.54, Python 3.14)

- `tests/test_capability_diff_partial_clone.py`: 28 passed, up from 21.
- Together with `test_host_diff_entry_docs`, `test_adapter_static_only` (line pins moved), `test_public_surface_contract`, `test_distribution_surface_parity`, `test_agent_mode`, `test_docs_links`, `test_diff_input_status`, `test_capability_diff`, `test_invocation_policy`, `test_cli_on_demand_loading`, `test_scoped_base_tree`, `test_verification_git_snapshot` and `test_live_workspace_cause`: 1654 passed, 1 skipped.
- `test_verify`, `test_cli`, `test_workflow_capability_diff`, `test_manifest_free_pr_rows`, `test_unchanged_host_limits` and `test_host_change_route_parity`: 263 passed.
- `ruff check .` is clean.

I updated the PR description's Design and Tests sections to match. It had said that with no single remote the command uses `origin`, and that no name is guessed.

pengfei-threemoonslab (commented):
Independent coding-agent review/address cycle 2 at `ec1ddbee`.

**No P0-P2 findings.** The cycle-1 P2 is fixed, CI is green on this head, and every #817 acceptance item holds when run against real partial clones.

**The cycle-1 P2 is fixed**

The finding was that a single promisor remote named `up/stream` got a command naming `origin`, and that command failed.

- **Reproduction** (the issue's `src` repository, git 2.54, `./shipgate` from this head):
  - `git clone -q --filter=blob:none --no-checkout --origin up/stream file://$PWD/src slash && git -C slash checkout -q main`
  - `AGENTS_SHIPGATE_AGENT_MODE=1 ./shipgate diff --workspace slash --base up/stream/base --json` exits 2 with `objects_missing`. The one next action's `args` are `-C <abs>/slash fetch --refetch --no-filter up/stream`.
  - I ran that command from `/`: exit 0. The same `diff` is then `comparable` with 2 rows.
- **Other names** get the same result: `Upstream` (the subsection keeps its case), `up.stream` and `café`. For each one I ran the emitted command, and `diff` then compared.
- **Two promisor remotes** (`zeta` first in the configuration, then `origin`) give two ranked commands. The second one's `why` names `zeta` as the one to try first.

**Defect on `origin/main` `a4dbd139`** (detached worktree, `./shipgate`)

- **`blob:none` clone, `--json`, agent mode:** exit 1 with a `ConfigError` traceback (`could not fetch <blob> from promisor remote`). Stdout is empty and there is no agent line.
- **`tree:0` clone:** exit 1 with a `CalledProcessError` traceback from `rev-parse <commit>^{tree}`.

**The fix at `ec1ddbee`**

- **Both filters, text and `--json`, agent mode on and off:** exit 2 and empty stdout. Stderr has one line naming `origin/base (68c9ae32)`. In agent mode there is exactly one `objects_missing` line: its `message` equals `next_actions[0].why`, and its `executable`/`args` match `command`.
- **Emitted command:** I ran it from `/`, then reran `diff`. It exits 0 as `comparable`, with `Bash(git push *)` added and the deny rule removed.
- **The trap:** `git fetch --refetch origin` alone leaves `diff` at exit 2 in both clones.
- **Also refused correctly:**
  - a `blob:limit=10` clone
  - a linked `git worktree`, whose command is bound to the worktree path
  - `--workspace .` from `.claude/` in a directory whose path has a space. The command is bound to the repository root and quoted.
  - `extensions.partialClone` with no `remote.origin.promisor`
  - a promisor flag that comes from an `include.path` file
  - a bare `promisor` key. With `--type=bool` Git prints `true` for it.
- **Not a partial clone:** with the promisor configuration removed and the `.promisor` packs kept, `diff` still exits 1 with the old traceback. It is not called `objects_missing`, which is the fail-closed direction.
- **No-base control:** with no `--base`, a clone that already holds its default base compares with exit 0.

**No implicit fetch**

In both clones I pointed `remote.origin.url` at `ext::touch <marker>` and unset `GIT_NO_LAZY_FETCH` in the caller. The marker stayed absent through each of these:
- the `diff` refusal
- `diff` with the default base
- `verify --preview`
- `check --format agent-control-json`

A plain `git cat-file` of the same object then created it.

**Tests fail without the fix** (mutations on a copy of this tree, running the new module and `test_host_diff_entry_docs.py`)

| Mutation | Failing tests |
| --- | --- |
| Catch nothing in `diff.py` (the refusal never runs) | 22 |
| Drop `--no-filter` | 17 |
| Skip the `check-ref-format` test | 1 (`[not-a-remote-name]`) |
| Skip the leading-`-` test | 1 (`[leading-dash]`) |
| Skip the URL test | 1 (`[no-url]`) |
| Make `promised_objects_missing` always return true | 1 (the corruption control) |
| Fall back to `origin` | 3 (every `review` case) |

**Contracts and claims**

- **Wording:** `verify --preview`'s `diff_status.remediation` and `control.next_action.why` carry the `--no-filter` example and "Shipgate runs Git with GIT_NO_LAZY_FETCH=1". The old example now appears only where it is named as the trap (CHANGELOG, STABILITY, quickstart, the tests), and "Verification runs with" only in the STABILITY note that describes the change.
- **Other commands, same clones:**
  - `verify --json` in a blobless clone is `agent_action_required`, with `objects_missing`, no permission and `comparison_input_unavailable:ConfigError`, as the description says.
  - `check` in the treeless clone is `human_review_required` with no permission.
- **Published claims:** the CHANGELOG, the STABILITY note, the `docs/errors.json` entry and AGENTS.md all describe the naming rule the code applies. That rule is: Git accepts the name as a remote name, the remote has a URL, the name does not start with `-`, and no fallback name is used. The `review` fallback is described too.
- **Precedent:** "no new exit code" matches the `environment_error` note.
- **No sensitive text:** the refusal reads remote URLs only to check that one exists, and never prints them.
- **Surfaces:** the new catalog entry, `llms-full.txt` (`scripts/build-llms-full.py` rebuilds it byte for byte) and the `capability_diff` row pass `test_public_surface_contract` and `test_distribution_surface_parity`.
- **Scale:** on a synthetic history of 30,000 commits and 50,000 files, the `--missing=allow-promisor` walk took 1.2 s. The refusal took 15 s end to end.

**Tests run locally** (Python 3.14, git 2.54), all passing:

- **1,516 tests:** `test_capability_diff_partial_clone`, `test_host_diff_entry_docs`, `test_adapter_static_only`, `test_public_surface_contract`, `test_distribution_surface_parity`, `test_agent_mode`, `test_diff_input_status`, `test_capability_diff`, `test_docs_links`.
- **659 tests:** `test_verify`, `test_cli`, `test_scoped_base_tree`, `test_verification_git_snapshot`, `test_workflow_capability_diff`, `test_manifest_free_pr_rows`, `test_host_change_route_parity`, `test_unchanged_host_limits`, `test_host_file_links`, `test_host_link_read_through`, `test_source_head_identity`, `test_host_path_privacy`, `test_cli_on_demand_loading`, `test_workflow_step_action_references`, `test_reusable_workflow_secret_mappings`, `test_workflow_label_redaction`, `test_host_boundary_unread_surfaces`, `test_host_only_advisory_recipe`, `test_live_workspace_cause`.
- **Lint:** `ruff check` on the changed Python files is clean.

**CI:** every check on `ec1ddbee` passes. The `suite (1-3)` totals are 11,365 passed and 22 skipped. That is the same skip count as before, so the new `no-lazy-fetch-alone` case ran rather than skipped.

**Non-blocking (P3)**

- **"The same sentence" is slightly loose.** The `docs/errors.json` description ("the message ends with the same remediation sentence"), the `capability_diff` row and the CHANGELOG's "One wording" bullet all say `diff` ends with the sentence `verify` reports. It is built by the same function, but its example is the bound `git -C <workspace> ... <remote>` command, not `git fetch --refetch --no-filter origin`. The STABILITY note says this exactly ("with the first command (or the placeholder) as its example"); the other three could say the same.
- **The PR description's before/after table** cites `origin/main` `34d8b38d`. That was the base before the rebase. The same defect reproduces on the current base `a4dbd139`, so only the SHA is out of date.
- **`docs/design-partner-verifier-pilot.md`, Route H,** says to report and stop when "a shallow clone is refused" but does not mention the partial-clone refusal. It is no-stdout exit 2, so an agent is unlikely to read it as a result, but naming it would close the list.
- **Still open from cycle 1, pre-existing and outside #817's `diff` scope** (each worth its own issue):
  - `verify` without `--preview` in a `tree:0` clone exits 4 with `internal_error`.
  - `check` says "restore a readable git worktree" rather than naming hydration.
  - `verify --preview`'s `fetch_base` next action has no command.
  - The error-kind tables in the SKILL.md copies do not list `objects_missing`, and were already missing other kinds.

